PR #1, opened 2015-08-27
Fix NPE if an embeddable query info provider is not set on the snapsh…

ilooner:
…ot server.


tweise:
JIRA + smart commit please.

PR #5, opened 2015-08-31
Fixed NPE

pramodin:
Not sure what happened but couple of lines were missing in NetworkManager. Added them back.


tweise:
Pramod, this should go to devel-3 branch.

PR #6, opened 2015-09-01
MLHR-1825 #comment handling case when no open part is updated but str…

otterc:
…eam is not created

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
@@ -484,34 +484,40 @@ public abstract class AbstractFileOutputOperator<INPUT> extends BaseOperator imp
         //of this operator.
         for(String seenFileName: endOffsets.keySet()) {
           try {
-            Integer part = openPart.get(seenFileName).getValue() + 1;
+            Integer fileOpenPart = this.openPart.get(seenFileName).getValue();
+            int nextPart = fileOpenPart + 1;
             String seenPartFileName;
             while (true) {
-              seenPartFileName = getPartFileName(seenFileName, part);
-              Path activePath;
+              seenPartFileName = getPartFileName(seenFileName, nextPart);
+              Path activePath = null;
               if (alwaysWriteToTmp) {
                 String tmpFileName = fileNameToTmpName.get(seenPartFileName);
-                activePath = new Path(filePath + Path.SEPARATOR + tmpFileName);
+                if (tmpFileName != null) {
+                  activePath = new Path(filePath + Path.SEPARATOR + tmpFileName);
+                }
               } else {
                 activePath = new Path(filePath + Path.SEPARATOR + seenPartFileName);
               }
-              if (!fs.exists(activePath)) {
+              if (activePath == null || !fs.exists(activePath)) {
                 break;
               }
 
               fs.delete(activePath, true);
-              part = part + 1;
+              nextPart++;
             }
 
-            seenPartFileName = getPartFileName(seenFileName, openPart.get(seenFileName).intValue());
-            Path activeFilePath;
+            seenPartFileName = getPartFileName(seenFileName, fileOpenPart);
+            Path activePath = null;
             if (alwaysWriteToTmp) {
-              activeFilePath = new Path(filePath + Path.SEPARATOR + fileNameToTmpName.get(seenPartFileName));
+              String tmpFileName = fileNameToTmpName.get(seenPartFileName);
+              if (tmpFileName != null) {
+                activePath = new Path(filePath + Path.SEPARATOR + fileNameToTmpName.get(seenPartFileName));
+              }
             } else {
-              activeFilePath = new Path(filePath + Path.SEPARATOR + seenPartFileName);
+              activePath = new Path(filePath + Path.SEPARATOR + seenPartFileName);
             }
 
-            if (fs.getFileStatus(activeFilePath).getLen() > maxLength) {
+            if (activePath != null && fs.getFileStatus(activePath).getLen() > maxLength) {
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
```diff
@@ -484,34 +484,40 @@ public FSFilterStreamContext load(@Nonnull String filename)
         //of this operator.
         for(String seenFileName: endOffsets.keySet()) {
           try {
-            Integer part = openPart.get(seenFileName).getValue() + 1;
+            Integer fileOpenPart = this.openPart.get(seenFileName).getValue();
+            int nextPart = fileOpenPart + 1;
```
PramodSSImmaneni: Looks like future part tmp files are being deleted but not the current. In the preceding section where the tmp file is recovered to a new tmp file the previous tmp file is left as is. Consider deleting the old tmp file in that section.

chandnisingh: We can't delete old tmp files here because when the operator is restored to  a previous state after recovery those tmp files will be needed. It is mentioned in the comment. The old tmp files are deleted in committed call back.

PramodSSImmaneni: One more thing on line 446 there is an optimization to not create the new tmp file if the file hasn't changed from the recovery offset (i.e., file.length == recovery offset). This may be a problem as some other process holds the write lease to this file and we may not be able to open it in append mode later on when data is going to be written

chandnisingh: The files written by the operator belong to this operator and if some one is using this operator to write to files which are also being written by another process then that is an invalid use-case.

PramodSSImmaneni: Regd 1. good point. However, if there was a fault recovery the tmp filename for a part will change and later when it becomes eligible for finalization as you mentioned in committed callback the new tmp filename will be moved, the old tmp file would still remain isn't it.

Regd 2. if there is a failure the lease would be held by a prior instance of this operator and the new instance would try to write to the same file as the tmp file isn't being recreated.

chandnisingh: Regd 1: 
    //when writing to tmp files there can be vagrant tmp files which we have to clean
    FileStatus[] statuses = fs.listStatus(new Path(filePath));
    for (FileStatus status : statuses) {
      String statusName = status.getPath().getName();
      if (statusName.endsWith(TMP_EXTENSION) && statusName.startsWith(fileName)) {
        LOG.debug("deleting vagrant file {}", statusName);
        fs.delete(status.getPath(), true);
      }
    }

chandnisingh: Regd2:
If there is a failure then the file will be corrupted so it will go into that block and recover it. Are we facing any lease issue?

PramodSSImmaneni: 1. I didn't notice that code. Thats great. However there is some ambiguity here. For example startsWith(fileName) condition could match for tmp files of filename.10 when looking for tmp files of filename.1. Also since the user can determine how to form the part filename looking for something like startsWith(fileName + ".") is not always guaranteed either as I could choose my part 10 filename to be filename.1.0, will have to look for pattern filename.<long>.tmp specifically.

Regd 2 it could take some time for namenode to realize client is dead. Not related to earlier issue but since there was no proper close and since there cant be multiple writes to same file I am guessing there could be an issue. Will be good to try it out.

chandnisingh: Regd1-> why will there be future tmp files for a file  at this point of time?

Regd2 -> will create another ticket to address this as a potential bug. however we are not facing this now so will the existing code as is.

PramodSSImmaneni: 1. Since commit can lag by the time part 1 is being materialized in finalize part 10 could already be created and wont this code potentially delete tmp files of part10 because startsWith would match.

chandnisingh: Yes you are right. Made the fix. Existing tests validate the implementation.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
```diff
@@ -1143,8 +1149,12 @@ protected void finalizeFile(String fileName) throws IOException
     for (FileStatus status : statuses) {
       String statusName = status.getPath().getName();
       if (statusName.endsWith(TMP_EXTENSION) && statusName.startsWith(fileName)) {
-        LOG.debug("deleting vagrant file {}", statusName);
-        fs.delete(status.getPath(), true);
+        //a tmp file has tmp extension always preceded by timestamp
+        String actualFileName = statusName.substring(0, statusName.lastIndexOf('.', statusName.lastIndexOf('.') - 1));
```
PramodSSImmaneni: There could be other random files in the folder not related to the ones the operator is writing where lastIndexOf returns -1 and the above compound expression could result in IndexOutOfBoundsException for those.

chandnisingh: The probability of a random file which ends with .tmp and starts with an actual file name used by this is just too low. This method can be overridden though, it someone using it face such issues

PR #7, opened 2015-09-02
MLHR-1824 Add App Data support as a separate app

amberarrow:
Undo previous merge and add App Data support as a new application.


amberarrow:
Doesn't restore the old Application.java, closing; will re-open another pull request

PR #11, opened 2015-09-02
Schema integration with JDBC operators

otterc:


library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInputOperator.java
```diff
@@ -0,0 +1,582 @@
...
+ * @category Input
+ * @tags database, sql, pojo, jdbc
+ * @since 2.1.0
+ */
+public class JdbcPOJOInputOperator extends AbstractJdbcInputOperator<Object> implements Operator.ActivationListener<Context.OperatorContext>
+{
+  private static int DEF_BATCH_SIZE = 100;
+
+  //table and query definition
+  protected String tableName;
```
ilooner: Why no \@NotNull annotation? Also if there are getters and setters the field should be private. This is because any modification to a non private field is potentially not backwards compatible (like adding an annotation or removing it).

chandnisingh: It is nullable because either a query is provided or individual constructs (like table name etc. need to be provided)

chandnisingh: backwards compatible is a good point but I think for any extensions it is better to access the fields directly not through getters. the code is compact and readable. I think this is subjective and should be left up to the discretion of author.

ilooner: I had a semantic versioning issue caused by making fields protected. Please check with @tweise about this for the correct convention to follow.

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInputOperator.java
```diff
@@ -0,0 +1,582 @@
...
+      }
+      if (orderByExpr != null) {
+        builder.append(" ORDER BY ").append(orderByExpr);
+      }
+    }
+    if (limitSupported) {
+      builder.append(" LIMIT ? , ? ");
+    } else {
+      //Starting from Oracle 12c R1 there is a row limiting clause.
+      builder.append(" OFFSET ? ROWS FETCH NEXT ? ROWS FETCH NEXT ? ROWS ONLY ");
```
ilooner: is this right? On line 138 you only set two query parameters

ilooner: This is also only supported for Oracle, what if we want to make this work with another flavor of SQL? Instead of having limitSupported can we set the limit statement as a string like you are doing for everything else?

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInputOperator.java
```diff
@@ -0,0 +1,582 @@
...
+ * A user can provide there own query by setting the {@link #query} property which takes precedence.<br/>
+ *
+ * For eg. user can set the query property to a complex one : "select x1, x2 from t1, t2 where t1.x3 = t2.x3 ;"<br/>
+ *
+ * @displayName Jdbc Input Operator
+ * @category Input
+ * @tags database, sql, pojo, jdbc
+ * @since 2.1.0
+ */
+public class JdbcPOJOInputOperator extends AbstractJdbcInputOperator<Object> implements Operator.ActivationListener<Context.OperatorContext>
```
ilooner: Document performance issue with using row id based batching

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInputOperator.java
```diff
@@ -0,0 +1,582 @@
+/**
```
ilooner: I don't see a unit test for this operator

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInputOperator.java
chandnisingh: The reason for not having limit as another clause is because then we need to infer the batch size from the clause. Also this method can be overridden in case of another flavor of SQL

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInputOperator.java
chandnisingh: Had added a test case. There are tests for all jdbc operators here library/src/test/java/com/datatorrent/lib/db/jdbc/JdbcOperatorTest.java

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInputOperator.java
ilooner: Also please document that this operator is only idempotent if the table is not modified.

ilooner: please add checkpointableWithinWindow = false

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInputOperator.java
```diff
@@ -0,0 +1,582 @@
...
+  @Override
+  public void emitTuples()
+  {
+    if (!windowDone) {
+      try {
+        preparedStatement.setLong(1, offset);
+        preparedStatement.setLong(2, batchSize);
+
+        ResultSet resultSet = preparedStatement.executeQuery();
+        if (resultSet.next()) {
```
ilooner: I was mistaken on this. You were right.

ilooner:
Rest looks good to me.

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInputOperator.java
chandnisingh: We can't make this implementation checkpointableWithinWindow=false. The table can be modified

chandnisingh:
@ilooner Please review and merge

chandnisingh:
@PramodSSImmaneni please review

chandnisingh:
I have addressed all the comments

PR #20, opened 2015-09-10
Refactoring of file splitter

otterc:
Resolves MLHR-1804, MLHR-1805, MLHR-1806 and a kryo exception due to race condition


chandnisingh:
@PramodSSImmaneni please review. this is blocking another project

PR #21, opened 2015-09-11
MLHR-1804 MLHR-1805 MLHR-1806 #resolve #comment Refactoring of file s…

otterc:
…plitter and fix for kryo exception

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/FileSplitter.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/FileSplitter.java
@@ -59,14 +59,15 @@ import com.datatorrent.lib.io.block.BlockMetadata.FileBlockMetadata;
  * The operator emits block metadata and file metadata.<br/>
  *
  * The file system/directory space should be different for different partitions of file splitter.
- * The scanning of
  *
+ * @deprecated  use {@link FileSplitterInput}. This splitter has issues with recovery and fixing that breaks backward compatibility.
  * @displayName File Splitter
  * @category Input
  * @tags file
  * @since 2.0.0
  */
 @OperatorAnnotation(checkpointableWithinAppWindow = false)
+@Deprecated
 public class FileSplitter implements InputOperator, Operator.CheckpointListener
 {
   protected Long blockSize;
@@ -454,6 +455,7 @@ public class FileSplitter implements InputOperator, Operator.CheckpointListener
   /**
    * Represents the file metadata - file path, name, no. of blocks, etc.
    */
+  @Deprecated
   public static class FileMetadata
   {
     @NotNull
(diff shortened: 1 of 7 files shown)


chandnisingh:
@PramodSSImmaneni  please review

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitter.java
```diff
@@ -654,6 +673,7 @@ public TimeBasedDirectoryScanner()
       discoveredFiles = new LinkedBlockingDeque<FileInfo>();
       atomicThrowable = new AtomicReference<Throwable>();
       ignoredFiles = Sets.newHashSet();
+      lock = new Lock();
```
pramodin: Why not create this setup.. transient anyway

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitter.java
```diff
@@ -99,6 +105,8 @@
 
   protected final BasicCounters<MutableLong> fileCounters;
 
+  private final static Lock LOCK = new Lock();
```
pramodin: Why not use Object

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitter.java
otterc: If we initialize it here and don't expect the object to change, can declare it final. Don't see why it should be moved to setup()?

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitter.java
otterc: Later if we want to enhance the lock then we will be able to do so. With Object we can't

chandnisingh:
Please let me know if this can be merged?

chandnisingh:
@PramodSSImmaneni please review and merge. This is blocking other changes.

chandnisingh:
@PramodSSImmaneni updated the code as per review  comments. Please merge

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -0,0 +1,594 @@
...
+      }
+    }
+
+    /**
+     * Operations that need to be done once a scan is complete.
+     */
+    protected void scanComplete()
+    {
+      LOG.debug("scan complete {}", lastScanMillis);
+      ScannedFileInfo fileInfo = discoveredFiles.peekLast();
```
pramodin: discoveredFiles could become empty by the time this executes

otterc: Yes. This is tricky to fix.

otterc:
@PramodSSImmaneni added the fix. Please review and merge

otterc:
As per review, removed delimiter which informs the operator when a scan iteration is complete. If any specific implementation requires it, it will need to add it.

@PramodSSImmaneni  please merge the pull request

PR #24, opened 2015-09-13
Fixed eol to LF

vrozov:


tweise:
Can you commit this as MalharJenkins to distinguish from real changes.

vrozov:
Changed author to MalharJenkins.

PR #26, opened 2015-09-16
exclude couchbase mock snapshot dependency from release enforce rule

vrozov:
@davidyan74 please review


tweise:
Not needed in release-3.1 - see #29

PR #27, opened 2015-09-16
exclude couchbase mock snapshot dependency from release enforce rule

vrozov:
@davidyan74 please review


vrozov:
Obsoleted by #29

PR #32, opened 2015-09-22
Fix removal of vagrant tmp files when output file is in subdirectory

darinnee:
This pull request adds removal of left-over tmp files for output files that may be located in a subdirectory of the writer's filePath. The use case is a single writer that outputs to multiple files using a file naming scheme that includes directory components. With this change, the writer removes the matching tmp files that have the same parent directory as the output file being finalized.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
@@ -1168,10 +1168,10 @@ public abstract class AbstractFileOutputOperator<INPUT> extends BaseOperator imp
     fileNameToTmpName.remove(fileName);
 
     //when writing to tmp files there can be vagrant tmp files which we have to clean
-    FileStatus[] statuses = fs.listStatus(new Path(filePath));
+    FileStatus[] statuses = fs.listStatus(destPath.getParent());
     for (FileStatus status : statuses) {
       String statusName = status.getPath().getName();
-      if (statusName.endsWith(TMP_EXTENSION) && statusName.startsWith(fileName)) {
+      if (statusName.endsWith(TMP_EXTENSION) && statusName.startsWith(destPath.getName())) {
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
```diff
@@ -1168,10 +1168,10 @@ protected void finalizeFile(String fileName) throws IOException
     fileNameToTmpName.remove(fileName);
 
     //when writing to tmp files there can be vagrant tmp files which we have to clean
-    FileStatus[] statuses = fs.listStatus(new Path(filePath));
+    FileStatus[] statuses = fs.listStatus(destPath.getParent());
```
chandnisingh: Will this work when the output file is not in a sub-directory but the main directory?

darinnee: Yes, it should. Only assumption is that the tmp file is located in the same directory as the output file.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
```diff
@@ -1168,10 +1168,10 @@ protected void finalizeFile(String fileName) throws IOException
...
     for (FileStatus status : statuses) {
       String statusName = status.getPath().getName();
-      if (statusName.endsWith(TMP_EXTENSION) && statusName.startsWith(fileName)) {
+      if (statusName.endsWith(TMP_EXTENSION) && statusName.startsWith(destPath.getName())) {
```
chandnisingh: daestPath.getName() will be same as fileName so do we need this change?

darinnee: I think so.  For example, fileName could be "a/b/c/part-00000", so getName() returns "part-00000", needed to match statusName.

PR #35, opened 2015-09-24
SPOI-5728: Update Malhar with changes made by Ingestion app to Malhar operators

DT-Priyanka:
FileReaderUtils is going to be used by FileSplitter (The pull request yet to be raised).
The use case is:
We want to handle upper/lower case scheme input for FileSplitter input.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/block/AbstractFSBlockReader.java
+++ b/library/src/main/java/com/datatorrent/lib/io/block/AbstractFSBlockReader.java
@@ -56,7 +56,9 @@ public abstract class AbstractFSBlockReader<R> extends AbstractBlockReader<R, Bl
   {
     super.teardown();
     try {
-      fs.close();
+      if (fs != null) {
+        fs.close();
+      }
     }
     catch (IOException e) {
       throw new RuntimeException(e);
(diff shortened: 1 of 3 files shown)


otterc:
Hi Priyanka,

FileReaderUtils is specific to your use-case. The implementations in Malhar library are general and none of them work with a notion of scheme.  

We have discussed earlier that even FileSplitterInput in Malhar will not have any schemes. 
If you are facing issues in extending FileSplitterInput and then plugging schemes in then please let me. 

Chandni

DT-Priyanka:
Hi Chandni,
Looks like there is some confusion. There was one more code change related to FTP input scheme, to address FTP parallel read issue and that one was ingestion specific.
This change is about accepting case insensitive input scheme, i.e. we should be able to accept input as HDFS://localhost:8020/user/priyanka/input as well as hdfs://localhost:8020/user/priyanka/input.
Here is the jira reference for this change: https://datatorrent.atlassian.net/browse/SPOI-4830

If we still agree on not having this fix for Malhar, I will remove the classes.

otterc:
Malhar basic implementations (Not just FileSplitter) work with URL and the notion that the user specified URL is intact. If not, the operator throws error.
So if in your use-case, you want to let the user specify scheme in any case and then automatically correct it then I think that should be part of your specific implementation.

DT-Priyanka:
Okay. I will close this pull request then.
So now if user specifies capital case scheme as part of input application can fail and we won't do any auto correction.

PR #38, opened 2015-09-27
File input operator that emits the lines of the file

pramodin:
Ready to use implementation, refer MLHR-1858. @chandnisingh see

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
@@ -1158,4 +1158,50 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
       return result;
     }
   }
+
+  /**
+   * This is an implementation of the {@link AbstractFileInputOperator} that outputs the lines in a file.&nbsp;
+   * Each line is emitted as a separate tuple.&nbsp; It is emitted as a String.
+   * <p>
+   * The directory path where to scan and read files from should be specified using the {@link #directory} property.
+   * </p>
+   * @displayName File Line Input
+   * @category Input
+   * @tags fs, file, line, lines, input operator
+   *
+   */
+  public static class FileLineInputOperator extends AbstractFileInputOperator<String>
+  {
+    public transient final DefaultOutputPort<String> output = new DefaultOutputPort<String>();
+
+    protected transient BufferedReader br;
+
+    @Override
+    protected InputStream openFile(Path path) throws IOException
+    {
+      InputStream is = super.openFile(path);
+      br = new BufferedReader(new InputStreamReader(is));
+      return is;
+    }
+
+    @Override
+    protected void closeFile(InputStream is) throws IOException
+    {
+      super.closeFile(is);
+      br.close();
+      br = null;
+    }
+
+    @Override
+    protected String readEntity() throws IOException
+    {
+      return br.readLine();
+    }
+
+    @Override
+    protected void emit(String tuple)
+    {
+      output.emit(tuple);
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


pramodin:
This is the order I have been using for a while

all java/javax
com
net
org
org.apache
gnu
com.datatorrent.lib
com.datatorrent.contrib
com.datatorrent.api
static imports

Why do you have org before com. com is more common than org.

otterc:
Till now everyone has been using different order. This is a common one  which is  default Eclipse formatter configuration. The style for Malhar and Apex needs to be same.

pramodin:
What is the . imports in your list. Also is there a rationale for the eclipse order. For example the java/javax classes are more fundamental and are at beginning, why are the static imports before that, if there is a static inner class within a static import it could have name collision with java.\* classes.

otterc:
Some of us have been following this order from the start. I believe Netbeans is similar except the placement of static imports. Static imports are for members (static fields and methods)  of a class. I don't understand your point of name collision as those will result in compilation error and resolved at the time of writing code.

Also I am not sure if  the import order that you are proposing follows the order of most common package to least common package.  
And isn't this discussion a little late? We have added the checks to Apex already. APEX-16 mentioned all the checkstyle rules being applied

pramodin:
You can also import inner classes with a static wildcard import. These would come before more fundamental classes which is contrary to the goal. The order is from more fundamental/lower level to more higher level.

The discussion is a little late and I think I missed the original discussion. Its too late to change it. Are these not enabled on malhar yet as I didn't see checkstyle errors.

pramodin:
Changed order to what you suggested but we should seriously take another look/discuss the order if there was no specific analysis done the first time around.

otterc:
I don't see much value in doing that analysis. Static import documentation itself says to use it very sparingly.
https://docs.oracle.com/javase/1.5.0/docs/guide/language/static-import.html

So we should actually try avoiding static wildcard imports and moreover each of the popular IDE has different default placement for it. Eclibse - top, Intellij - bottom, Netbeans - doesn't differentiate between static and non-static.

So IMHO we just need to adopt one and follow that which we did.

243826:
Chandni - just a remark about "discussion is a little late". It's never too late. You have gotten the pass to discuss an issue a few months after the discussion had closed and it's only fair that others get the consistent treatment.

Pramod has made all the valid points. "Eclipse does it in some way" is not a valid argument. If it were so, we would not need other editors or other plugins. That said this is why I have always had a recommended order which most of the files I wrote have and I believe you also followed at some point:

The order is defined by our ability to influence the change in the package/class/code. The more we are in control of changing code, the lower the position of the package is in the import order. The idea being if there is ever a conflict we change the lower ordered code.

java,javax are global, backward compatible so go at the top. We never can modify these classes.
datatorrent packages are completely adaptable so go at the bottom.

In between, argument is to be made between org, gnu, com, net and we use our best judgement. Still I recommend that we follow alphabetical order except for com:

gnu
net
org
com - here because com is in most cases is likely to be less proliferated.

A quick glance at java source code will reveal that static import is reasonably common even for "java developers" and there is not a set pattern. Some of them use at the top, some at the bottom, some intermingled before the group or after the group. So static imports are here regardless of recommendation and people use it. Even though I use Netbeans, I do not follow Netbeans order (not sure if it has recommendation for one). There is another plugin called "organize imports" which is what I use to define custom ordering and it does allow various options and the one which looks properly grouped and aesthetically pleasing is putting static imports in alphabetical order at the bottom of the group to which the import would belong if it was not static.

Of course this is just one way of doing things but use the one which suits the most of you and keep the feedback channels open. There is no time expiry on a good idea.

otterc:
@243826  About "You have gotten the pass to discuss an issue a few months after the discussion had closed and it's only fair that others get the consistent treatment."
I think you are just drawing parallels  between completely different discussions.
Moreover the main reason to re-open the discussion  you are referring to was that it was very hard to debug Mocha tests that takes developers time.

Here we just adopt a rule and enforce consistency. Good coding practices dictate using static imports sparingly so why should we spend  more time and effort in finding which is the best suitable place for it and debate about it.

pramodin:
Chandni if things are fine with the rest of it can you review it

PR #40, opened 2015-09-28
MLHR-1861 Port PI demo to Scala

tushargosavi:
Review only
Investigation work to see if we can use Scala language to write streaming application. I have ported Pi demo in Scala along with operators.


tushargosavi:
Closing the pull request, will open a new PR.

PR #45, opened 2015-10-02
 - MLHR-1864 #resolve #comment Made adding a query to a queue atomic

ilooner:
- Fixed race condition where a query could be accessed from the queue and then have its countdown decremented.


library/src/main/java/com/datatorrent/lib/appdata/query/QueryManagerAsynchronous.java
```diff
@@ -123,7 +124,7 @@ public void endWindow()
   {
     queueManager.haltEnqueue();
 
-    while(queueManager.getNumLeft() > 0) {
+    while(!isProcessingDone()) {
```
davidyan74: space after while

ilooner:
@davidyan74 fixed formatting

PR #46, opened 2015-10-02
 - MLHR-1864 #resolve #comment Made adding a query to a queue atomic

ilooner:
- Fixed race condition where a query could be accessed from the queue and then have its countdown decremented.


ilooner:
@davidyan74 fixed formatting

PR #54, opened 2015-10-04
MLHR-1857 License header format

tweise:
This PR formats all files with the ASF license header (exclusions listed in pom.xml).

@sashadt @andyperlitch There are a large number of demo UI files in this repository, some of them may be dead? Can you please check that the header changes on those files are correct and don't break anything.

@sandeepdeshmukh Please look at the .R files. 

The goal is to get these changes in soon so before branching 3.2.0


tweise:
@vrozov please review. If there is no other feedback, it would be nice to get this merged within the next days.

vrozov:
I'll wait till Friday for the feedback.

sashadt:
I're reviewed, and we can skip all the web demos.  I've opened PR to delete web/ folder from this request against your branch, @tweise

PR #57, opened 2015-10-12
 - MLHR-1868 #resolve Improved hash function

ilooner:


ilooner:
@amberarrow please review

PR #58, opened 2015-10-14
MLHR-1869 Update Maven coordinates for ASF release. 

tweise:
@vrozov  @davidyan74 First cut replacement. We need to discuss the .apa artifact distribution - they should not go into the ASF repo.

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -23,45 +23,26 @@
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <artifactId>malhar</artifactId>
-    <groupId>com.datatorrent</groupId>
-    <version>3.2.0-SNAPSHOT</version>
+    <groupId>org.apache.apex</groupId>
+    <version>3.2.0-incubating-SNAPSHOT</version>
   </parent>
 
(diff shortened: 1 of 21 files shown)


benchmark/pom.xml
```diff
@@ -519,13 +499,13 @@
...
-      <artifactId>dt-engine</artifactId>
+      <groupId>org.apache.apex</groupId>
+      <artifactId>apex-engine</artifactId>
       <version>3.0.0</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
-      <groupId>com.datatorrent</groupId>
+      <groupId>org.apache.apex</groupId>
       <artifactId>dt-api</artifactId>
```
davidyan74: apex-api?

contrib/pom.xml
```diff
@@ -601,9 +601,9 @@
       <optional>true</optional>
     </dependency>
     <dependency>
-      <groupId>com.datatorrent</groupId>
-      <artifactId>dt-common</artifactId>
-      <version>${dt.framework.version}</version>
```
davidyan74, line 606: I think it's better to use apex.core.version instead of just apex.version since apex technically includes malhar.

PR #60, opened 2015-10-14
Improved hash function

ilooner:
- MLHR-1868 improved the hashcode function to produce better hashes.


ilooner:
@amberarrow implemented changes

PR #69, opened 2015-10-22
MLHR-1877 #resolve #comment moved DTFile implementation and Cache man…

otterc:
…ager to library


otterc:
@tweise @siyuanh please review and merge

otterc:
I suspect that the DTFileTest which fails was even failing before my change. Since this code was in contrib before and the tests in contrib are only manually run, this wasn't seen before.

tweise:
@siyuanh please have a look

otterc:
@siyuanh I have fixed the test. Can you please merge this pull request. The change just moves these files from contrib to lib. Made a few changes to CacheManager as it was using a higher version of guava which wasn't what we were using in library. Also took liberty to make few style fixes to CacheManager.

library/src/main/java/org/apache/hadoop/io/file/tfile/CacheManager.java
```diff
@@ -110,41 +116,45 @@
...
+    return SINGLE_CACHE.getIfPresent(key);
   }
 
-  public static final void invalidateKeys(Collection<String> keys)
+  public static void invalidateKeys(Collection<String> keys)
   {
-    if (singleCache != null)
-      singleCache.invalidateAll(keys);
+    if (SINGLE_CACHE != null)
+      SINGLE_CACHE.invalidateAll(keys);
```
siyuanh, line 146: Fix the cs violation

PR #71, opened 2015-10-23
- MLHR-1880 #resolve fixed default value documentation for setMaxLeng…

ilooner:
…th method.


ilooner:
@chandnisingh if the maxLength is Long.MAX_VALUE then the operator is NOT in rolling mode.

line 173 has the following documentation stating this

  /**
- True if {@link #maxLength} < {@link Long#MAX_VALUE} or {@link #rotationWindows} > 0
  */
  protected transient boolean rollingFile = false;

PR #73, opened 2015-10-30
- MLHR-1887 #resolve made shutdown field volatile since it is written…

ilooner:
… and read by separate threads.


ilooner:
@davidyan74 please review

PR #77, opened 2015-11-02
MLHR-1885: Adding getters to the variables of KafkaMessage nested class

chaithu14:


chaithu14:
@siyuanh: Please review

PR #79, opened 2015-11-03
MLHR-1886: moving restoring of a file in its own method and added a t…

otterc:
…est that covers re-writing the file when it isn't closed before failure


otterc:

PR #80, opened 2015-11-04
MLHR-1889 #resolve #comment moved atomic renaming to its own method

otterc:
@ilooner can you please review and merge

Thanks,
Chandni

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
@@ -21,7 +21,11 @@ package com.datatorrent.lib.io.fs;
 import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
```diff
@@ -642,6 +658,22 @@ protected void closeStream(FSFilterStreamContext streamContext) throws IOExcepti
...
 
   /**
+   * Renames source path to destination atomically. This relies on the FileContext api. If
+   * the underlying filesystem doesn't have an {@link AbstractFileSystem} then this should be overridden.
+   *
+   * @param source      source path
+   * @param destination destination path
+   * @throws IOException
+   */
+  protected void rename(Path source, Path destination) throws IOException
```
ilooner, line 668: Could this method be used to rename files in finalizeFile?

otterc, line 668: yeah good catch

otterc, line 668: Done

PR #81, opened 2015-11-05
NOTICE original developer

tweise:


tweise:
Adding fix for MLHR-1891

PR #82, opened 2015-11-05
MLHR-1874 #resolve #comment Implementation of file output operator in…

otterc:
… hive doesn't move tmp files to actual files

### DIFF
--- a/contrib/src/main/java/com/datatorrent/contrib/hive/AbstractFSRollingOutputOperator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/hive/AbstractFSRollingOutputOperator.java
@@ -148,6 +148,7 @@ public abstract class AbstractFSRollingOutputOperator<T> extends AbstractFileOut
   @Override
   public void committed(long windowId)
   {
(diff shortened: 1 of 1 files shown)


otterc:
@brightchen please review and merge

contrib/src/main/java/com/datatorrent/contrib/hive/AbstractFSRollingOutputOperator.java
```diff
@@ -183,12 +184,7 @@ protected void rotateCall(String lastFile)
     try {
       this.rotate(lastFile);
     }
-    catch (IOException ex) {
-      logger.debug(ex.getMessage());
-      DTThrowable.rethrow(ex);
-    }
-    catch (ExecutionException ex) {
-      logger.debug(ex.getMessage());
+    catch (IOException | ExecutionException ex) {
```
brightchen, line 187: The HiveMockTest also has a issue. the  fsRolling.setRotationWindows should be called to set the window. 
And the HiveMockTest is using the embed Hive, you in fact can test even if Hive not installed.

PR #83, opened 2015-11-06
- MLHR-1880 #resolve fixed default value documentation for setMaxLeng…

otterc:
…th method.


otterc:
@ilooner Please merge to 3.2

PR #84, opened 2015-11-06
MLHR-1886 #resolve #comment optimizing recovery and updating rotation…

otterc:
… states only when new part is open


otterc:
@ilooner Tim please merge this change to 3.2

PR #88, opened 2015-11-06
- MLHR-1894 #resolve #comment Added ConnectableInputPort which has is…

ilooner:
…Connected method.


ilooner:
@chandnisingh please review

PR #92, opened 2015-11-10
Removed DataTorrent & Malhar copyrights. Changed DataTorrent platform wording to Apex platform.

pramodin:
It doesn't yet remove the copyrighted rss and text files


pramodin:
@tweise see

PR #99, opened 2015-11-15
MLHR-1852 #comment temporary ignoring the test case

otterc:


otterc:
@ilooner please merge. Ignoring the test case so that other pull requests are not affected. the ticket is still open. Made formatting changes but that does't affect any attribution because it was just me who worked on these files.

PR #100, opened 2015-11-17
MLHR-1906 #resolve #comment Snapshot Server support tags

brightchen:


brightchen:
@ilooner Hi Tim, Could you please review my commitment?

library/src/test/java/com/datatorrent/lib/appdata/snapshot/AppDataSnapshotServerTagsSupportTest.java
```diff
@@ -0,0 +1,67 @@
+package com.datatorrent.lib.appdata.snapshot;
```
ilooner, line 19: Travis build is failing because correct license is not added at the top of the file. Please see other files in Malhar and copy the license header over. Also make sure that the year stated in the license is this year.

library/src/test/resources/satisfactionRatingSnapshotSchema_test.json
```diff
@@ -0,0 +1,8 @@
+{
+  "values": [
+    {"name": "current", "type": "long", "tags": ["current"]}, 
+    {"name": "min", "type": "long", "tags": ["min"]}, 
+    {"name": "max", "type": "long", "tags": ["max"]}, 
+    {"name": "threshold", "type": "long", "tags": ["threshold"]}
+  ]
+}
```
ilooner: Add a newline at the end of the file.

library/src/main/java/com/datatorrent/lib/appdata/schemas/SnapshotSchema.java
```diff
@@ -246,6 +247,26 @@ private void initialize() throws JSONException
...
+    JSONArray tagArray = new JSONArray();
+
+    try {
+      int index = 0;
+      for (String tag : tags)
+        tagArray.put(index++, tag);
+
+      schema.put(FIELD_SCHEMA_TAGS, tagArray);
+    } catch (JSONException e) {
+      Preconditions.checkState(false, e.getMessage());
```
ilooner, line 254: Rethrow the exception so that the stack trace will be printed

library/src/main/java/com/datatorrent/lib/appdata/schemas/SnapshotSchema.java
```diff
@@ -246,6 +247,26 @@ private void initialize() throws JSONException
     schemaJSON = this.schema.toString();
   }
 
+  public void setTags(Set<String> tags)
+  {
+    if (tags == null || tags.isEmpty())
+      return;
+
+    JSONArray tagArray = new JSONArray();
```
ilooner: move this logic into initialize so that the schema is built in one place

library/src/main/java/com/datatorrent/lib/appdata/schemas/SnapshotSchema.java
```diff
@@ -246,6 +247,26 @@ private void initialize() throws JSONException
     schemaJSON = this.schema.toString();
   }
 
+  public void setTags(Set<String> tags)
+  {
+    if (tags == null || tags.isEmpty())
```
ilooner, line 246: throw IllegalArgumentException

library/src/test/java/com/datatorrent/lib/appdata/snapshot/AppDataSnapshotServerTagsSupportTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+      queryProcessor.endWindow();
+    }
+  }
+
+  private static final String schemaLocation = "satisfactionRatingSnapshotSchema_test.json";
+  private static final String TAG = "bulletin";
+
+  @Test
+  public void testSchema() throws Exception
```
ilooner, line 64: Please add a test for a predifined schema string satisfactionRatingSnapshotSchema_test_2.json

{
  "tags":["something"],
  "values": [
    {"name": "current", "type": "long", "tags": ["current"]}, 
    {"name": "min", "type": "long", "tags": ["min"]}, 
    {"name": "max", "type": "long", "tags": ["max"]}, 
    {"name": "threshold", "type": "long", "tags": ["threshold"]}
  ]
}

ilooner:
looks good, Thanks :)

library/src/main/java/com/datatorrent/lib/appdata/schemas/SnapshotSchema.java
brightchen, line 254: The method initialize() was called by the constructor. and the tags haven't set at that time.

library/src/test/java/com/datatorrent/lib/appdata/snapshot/AppDataSnapshotServerTagsSupportTest.java
brightchen, line 64: The Preconditions.checkState() in the initialize() method forbid keys in the first level. I am not sure why this precondition was checked, but maybe have some trap here if break this logic.

library/src/main/java/com/datatorrent/lib/appdata/schemas/SnapshotSchema.java
ilooner, line 254: I see, Instead of a setTags method could you create a new constructor that takes the tags, and then do this work in initialize?

library/src/test/java/com/datatorrent/lib/appdata/snapshot/AppDataSnapshotServerTagsSupportTest.java
ilooner, line 64: I'm not sure I understand, maybe you can explain it to me tomorrow.

library/src/main/java/com/datatorrent/lib/appdata/schemas/SnapshotSchema.java
brightchen, line 254: There are several constructors, add this parameter to the constructors will add more constructors( one for no tags, one for tags ). So I think maybe use setTags more clear.

PR #102, opened 2015-11-17
MLHR-1910 #resolve #comment fixed style violations and warnings in Bl…

otterc:
…ock Reader classes


otterc:
@ilooner Please merge. The attribution is consistent as only I have worked on these classes.

PR #106, opened 2015-11-19
MLHR-1859 #comment checkstyle execution phase is inherited from paren…

otterc:
…t with an older version


otterc:

PR #108, opened 2015-11-20
MLHR-1913: Need access to modifiedTime of ScannedFileInfo class

DT-Priyanka:


DT-Priyanka:
@chandnisingh can you please review and pull this change?

PR #111, opened 2015-11-24
Implement semi-join operator with two input tables

d9liang:
Please review.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/join/SemiJoinOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/join/SemiJoinOperator.java
@@ -1,0 +1,206 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


otterc:
Is this a left-join implementation? This is a simple implementation over an application window. @chaithu14 has worked on an implementation which does join over much larger time windows. We are working to pull that change in. 
Meanwhile can we change the name here to call it 'SimpleLeftJoin"?

d9liang:
@chandnisingh No this is not a left-join or left-outer-join. In semi-join only the joined rows from the left table are returned. However, in a left-outer-join, all the rows from left table will be returned (also padding with nulls for columns from the right table when not joined).

otterc:
I see. Can you please explain a little bit about what semi-join is in the operator. Maybe a link will be helpful.

d9liang:
@chandnisingh I've added sample cases in the Jira page

https://malhar.atlassian.net/projects/MLHR/issues/MLHR-1813

You can also find more info here:
http://docs.oracle.com/cd/B28359_01/server.111/b28286/statements_10002.htm#i2166436

otterc:
@d9liang Can you please add the link to the info in the javadoc for this operator. Thanks.

d9liang:
@chandnisingh I've added a link to the info in operator's Javadoc.

otterc:
@d9liang the license check is failing. Can you please check?

d9liang:
@chandnisingh I couldn't figure out what's wrong with the build, as it still fails even after removing the recent code with link info at javadoc. Please advise. I will add back the link info within javadoc section.

otterc:
@d9liang I see this is not a problem with this pull request then. Please squash your changes and I will pull them in

d9liang:
@chandnisingh PR squashed.

PR #112, opened 2015-11-25
MLHR-1918: [FileSplitter] Need stopScanning method in Scanner

DT-Priyanka:
@chandnisingh Please review and merge the changes.


DT-Priyanka:
Somehow I don't see previous comments on this PR, copying @chandnisingh comment from mail thread:
[Chandni]: Well what you described is something specific to your use case.
The method isn't being invoked from anywhere in the code. Don't see the need of it. If you are trying to add a util method which will be helpful for different implementations then call it 'pauseScanning'

[Priyanka]: Yes this is specific to my use case that I want to stop scan on some trigger. But if parent class doesn't allow me to override and call some method to stop scan, how do I achieve that? I might need to replicate entire logic of scan in my subclass to achieve this.
"pauseScanning": There is no way to resume it so I named it as stopScanning, added call to method from current class code.

PR #118, opened 2015-12-04
Avoid busy wait when maximum number of tuples emitted in a window are reached

pramodin:


pramodin:
@siyuanh please review

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
```diff
@@ -363,18 +361,26 @@ public void emitTuples()
...
+        MutablePair<Long, Integer> offsetAndCount = currentWindowRecoveryState.get(message.kafkaPart);
+        if (offsetAndCount == null) {
+          currentWindowRecoveryState.put(message.kafkaPart, new MutablePair<Long, Integer>(message.offSet, 1));
+        } else {
+          offsetAndCount.setRight(offsetAndCount.right + 1);
+        }
+      }
+    } else {
+      try {
+        Thread.sleep(spinMillis);
```
siyuanh, line 381: @PramodSSImmaneni  my understanding is in the InputNode, it already sleep spinmillis if counter of the emitTuples() is 0, do we need to double sleep here?

PR #121, opened 2015-12-09
MLHR-1901 #resolve

tushargosavi:
Changed location of the directory where test files were generated.

### DIFF
--- a/library/src/test/java/org/apache/hadoop/io/file/tfile/DTFileTest.java
+++ b/library/src/test/java/org/apache/hadoop/io/file/tfile/DTFileTest.java
@@ -36,6 +36,9 @@ import java.util.Random;
 
 public class DTFileTest
 {
+   private static String ROOT =
+      System.getProperty("test.build.data", "target/tfile-test");
+
   private Configuration conf;
   private Path path;
   private FileSystem fs;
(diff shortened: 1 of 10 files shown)


library/src/test/java/org/apache/hadoop/io/file/tfile/DTFileTest.java
```diff
@@ -70,7 +73,7 @@ public void setUp() throws IOException
 
     conf.setInt("tfile.fs.input.buffer.size", tconf.fsInputBufferSize);
     conf.setInt("tfile.fs.output.buffer.size", tconf.fsOutputBufferSize);
-    path = new Path("tmp/dtfile");
+    path = new Path(ROOT, "/tmp/dtfile");
```
otterc: @tushargosavi we usually create any test data files under target directory

tushargosavi: @chandnisingh other tests in tfile using same pattern. I used the same path for consistency. This code was copied from apache tfile implementation. Do you want all tests to use target directory for test data?

otterc:
@tushargosavi yes, tests should create files under target directory.

tushargosavi:
@chandnisingh I have updated patch to use target directory for DTFile tests.

PR #125, opened 2015-12-11
MLHR-1938 - Operator checkpointing in distributed in-memory store

ashishtadose:
Depends on https://issues.apache.org/jira/browse/APEXCORE-283

@tushargosavi kindly review.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
+++ b/library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
@@ -1,0 +1,233 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.lib.util;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.datatorrent.api.Attribute.AttributeMap;
+import com.datatorrent.api.Context.DAGContext;
+import com.datatorrent.lib.util.StorageAgent.ApplicationAwareStorageAgent;
+
+/**
+ * Abstract implementation of {@link ApplicationAwareStorageAgent} which can be
+ * configured be KeyValue store witch implementation of {@link StorageAgentKeyValueStore}
+ * 
+ * NOTE - this should be picked from APEX-CORE once below feature is release
+ * https://issues.apache.org/jira/browse/APEXCORE-283
+ * 
+ * @param <S>
+ *          Store implementation
+ */
+public abstract class AbstractKeyValueStorageAgent<S extends StorageAgentKeyValueStore>
+    implements ApplicationAwareStorageAgent, Serializable
+{
+
+  protected S store;
+  protected String applicationId;
+  public static final String CHECKPOINT_KEY_SEPARATOR = "-";
+
+  /**
+   * Gets the store
+   *
+   * @return the store
+   */
+  public S getStore()
+  {
+    return store;
+  }
+
+  /**
+   * Sets the store
+   *
+   * @param store
+   */
+  public void setStore(S store)
+  {
+    this.store = store;
+  }
+
+  /**
+   * Return yarn application id of running application
+   * 
+   * @return
+   */
+  public String getApplicationId()
+  {
+    return applicationId;
+  }
+
+  /**
+   * Set yarn application id
+   * 
+   * @param applicationId
+   */
+  public void setApplicationId(String applicationId)
+  {
+    this.applicationId = applicationId;
+  }
+
+  /**
+   * Generates key from operator id and window id to store unique operator
+   * checkpoints
+   * 
+   * @param operatorId
+   * @param windowId
+   * @return unique key for store
+   */
+  public static String generateKey(int operatorId, long windowId)
+  {
+    return String.valueOf(operatorId) + CHECKPOINT_KEY_SEPARATOR + String.valueOf(windowId);
+  }
+
+  /**
+   * Stores the given operator object in configured store
+   * 
+   * @param object
+   *          Operator object to store
+   * @param operatorId
+   *          of operator
+   * @param windowId
+   *          window id of operator to checkpoint
+   * 
+   */
+  @Override
+  public void save(Object object, int operatorId, long windowId) throws IOException
+  {
+
+    try {
+      store(generateKey(operatorId, windowId), object);
+      logger.debug("saved check point object key {} region {}", generateKey(operatorId, windowId), applicationId);
+    } catch (Exception ex) {
+      throw new RuntimeException(ex);
+    }
+  }
+
+  private synchronized void store(String checkpointKey, Object operator) throws IOException
+  {
+    if (!getStore().isConnected()) {
+      getStore().connect();
+    }
+    getStore().put(checkpointKey, operator);
+  }
+
+  /**
+   * Retrieves the operator object for given operator & window from configured
+   * store
+   * 
+   * @param operatorId
+   *          of operator
+   * @param windowId
+   *          window id of operator to checkpoint
+   */
+  @Override
+  public Object load(int operatorId, long windowId)
+  {
+    Object obj = null;
+    try {
+      obj = retrieve(generateKey(operatorId, windowId));
+      logger.debug("retrieved object from store  key {} region {} ", generateKey(operatorId, windowId), applicationId);
+    } catch (Exception ex) {
+      throw new RuntimeException(ex);
+    }
+
+    return obj;
+  }
+
+  private synchronized Object retrieve(String checkpointKey) throws IOException
+  {
+    if (!getStore().isConnected()) {
+      getStore().connect();
+    }
+
+    return getStore().get(checkpointKey);
+  }
+
+  /**
+   * Removes stored operator object for given operatorId & windowId from store
+   * 
+   */
+  @Override
+  public void delete(int operatorId, long windowId) throws IOException
+  {
+
+    if (!getStore().isConnected()) {
+      getStore().connect();
+    }
+
+    try {
+      getStore().remove(generateKey(operatorId, windowId));
+      logger.debug("deleted object from store key {} region {}", generateKey(operatorId, windowId));
+    } catch (Exception ex) {
+      throw new RuntimeException(ex);
+    }
+
+  }
+
+  /**
+   * Returns list window id for given operator id for which operator objects are
+   * stored but not removed
+   * 
+   */
+  @Override
+  public long[] getWindowIds(int operatorId) throws IOException
+  {
+    if (!getStore().isConnected()) {
+      getStore().connect();
+    }
+
+    List<String> keys = getStore().getKeys(operatorId);
+    if (keys.size() > 0) {
+      long[] windowsIds = new long[keys.size()];
+      int count = 0;
+      for (String key : keys) {
+        windowsIds[count] = extractwindowId(key);
+        count++;
+      }
+      return windowsIds;
+    } else {
+      return new long[0];
+    }
+  }
+
+  public static long extractwindowId(String checkpointKey)
+  {
+    String[] parts = checkpointKey.split(CHECKPOINT_KEY_SEPARATOR);
+    return Long.parseLong(parts[1]);
+  }
+
+  /**
+   * Saves the yarn application id which can be used by create application
+   * specific table/region in KeyValue sore.
+   */
+  @Override
+  public void setApplicationAttributes(AttributeMap map)
+  {
+    this.applicationId = map.get(DAGContext.APPLICATION_ID);
+    getStore().setTableName(applicationId);
+  }
+
+  private static final long serialVersionUID = 7065320156997171116L;
+  private static final Logger logger = LoggerFactory.getLogger(AbstractKeyValueStorageAgent.class);
+
+}
--- a/library/src/main/java/com/datatorrent/lib/util/StorageAgentKeyValueStore.java
+++ b/library/src/main/java/com/datatorrent/lib/util/StorageAgentKeyValueStore.java
@@ -1,0 +1,48 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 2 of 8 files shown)


library/src/main/java/com/datatorrent/lib/util/AppIdAwareStorageAgent.java
```diff
@@ -0,0 +1,37 @@
...
+package com.datatorrent.lib.util;
+
+import com.datatorrent.api.StorageAgent;
+
+/**
+ * Interface to pass application id to storage agent
+ * 
+ *
+ */
+public interface AppIdAwareStorageAgent extends StorageAgent
```
gauravgopi123: Is this required here? Will ApplicationAwareStorageAgent defined in Apex not take care of this?

library/src/main/java/com/datatorrent/lib/util/StorageAgentKeyValueStore.java
```diff
@@ -0,0 +1,88 @@
...
+package com.datatorrent.lib.util;
+
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Interface for KeyValue store
+ * 
+ */
+public interface StorageAgentKeyValueStore
```
gauravgopi123: There is com.datatorrent.lib.db.KeyValueStore in malhar. Can this not be used?

library/src/main/java/com/datatorrent/lib/util/StorageAgentKeyValueStore.java
```diff
@@ -0,0 +1,88 @@
...
+   * @return the list of all associated keys
+   */
+  public List<String> getKeys(Object key);
+
+  /**
+   * Set table/region name of store
+   * 
+   * @param tableName
+   */
+  public void setTableName(String tableName);
```
gauravgopi123, line 45: This is very specific to an implementation.. I don't think this is required in API

library/src/main/java/com/datatorrent/lib/util/StorageAgentKeyValueStore.java
```diff
@@ -0,0 +1,88 @@
...
+   */
+  public void remove(Object key);
+
+  /**
+   * Get all the keys associated with key
+   * 
+   * @param key
+   * @return the list of all associated keys
+   */
+  public List<String> getKeys(Object key);
```
gauravgopi123, line 38: What is the usecase for this? Can you please give example how will it work?

library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
```diff
@@ -0,0 +1,234 @@
...
+   * Generates key from operator id and window id to store unique operator
+   * checkpoints
+   * 
+   * @param operatorId
+   * @param windowId
+   * @return unique key for store
+   */
+  public static String generateKey(int operatorId, long windowId)
+  {
+    return String.valueOf(operatorId) + CHECKPOINT_KEY_SEPARATOR + String.valueOf(windowId);
```
gauravgopi123, line 100: Should application Id not be used to generate this key? What happens if there are multiple applications running and they have operators with same operatorId?

library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
```diff
@@ -0,0 +1,234 @@
...
+   * 
+   */
+  @Override
+  public void save(Object object, int operatorId, long windowId) throws IOException
+  {
+
+    try {
+      store(generateKey(operatorId, windowId), object);
+      logger.info("saved check point object key {} region {}", generateKey(operatorId, windowId), applicationId);
+    } catch (Throwable t) {
```
gauravgopi123: why catch Throwable?

library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
```diff
@@ -0,0 +1,234 @@
...
+  {
+
+    if (!getStore().isConnected()) {
+      getStore().connect();
+    }
+
+    try {
+      store.remove(generateKey(operatorId, windowId));
+      logger.info("deleted object from store key {} region {}", generateKey(operatorId, windowId));
+    } catch (Throwable t) {
```
gauravgopi123: Why catch Throwable?

library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
```diff
@@ -0,0 +1,234 @@
...
+ * 
+ * @param <S>
+ *          Store implementation
+ */
+public abstract class AbstractKeyValueStorageAgent<S extends StorageAgentKeyValueStore>
+    implements AppIdAwareStorageAgent, Serializable
+{
+
+  protected S store;
+  protected String applicationId;
```
gauravgopi123, line 47: should applicationId be final?

gauravgopi123, line 47: This should reflect the changes for ApplicationAwareStorageAgent in Apex, which it is not

library/src/main/java/com/datatorrent/lib/util/AppIdAwareStorageAgent.java
ashishtadose: This is required as ApplicationAwareStorageAgent is not released in Apex core and Malhar build takes core dependency of release candidates but not develop branches.

library/src/main/java/com/datatorrent/lib/util/StorageAgentKeyValueStore.java
ashishtadose: com.datatorrent.lib.db.KeyValueStore it contains additional methods such as getAll & putAll which are not relevant here also this requires one additional method which is not present there.

library/src/main/java/com/datatorrent/lib/util/StorageAgentKeyValueStore.java
ashishtadose, line 45: I thought it would be common for any key value store service as most of them have notion  of table, region to denote to particular storage instance.

library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
ashishtadose, line 100: This would not happen as application ids are being used to create application specific tables or regions. 

For checkpointing of each application's operators a new table/region is created once the application is started.

library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
ashishtadose: Initial though was, since this is abstract implementation and to ensure that any errors or exceptions thrown by underlying store should not cause disruption to application.

I can change this to catch to run time exception instead.

library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
ashishtadose, line 47: Yes

library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
gauravgopi123: You should not catch any exception.

library/src/main/java/com/datatorrent/lib/util/StorageAgentKeyValueStore.java
ashishtadose, line 38: Rationale for getKeys() method.

StorageAgent needs to provide below fours operations for operator checkpointing 
1 save(opertorId, windowId)
2 load(opertorId, windowId)
3 delete(opertorId, windowId)
4 getWindowIds(opertorId)

All Key value based systems perform operations based on key so to accommodate these operations need to form a key with combination of operatorId & windowId to uniquely identify checkpoints.
Note that each checkpoints will be written to different table/region by store implementation by using name provided by setTableName method.

To support 4th operation of getting windows ids related given operator id store implementation will have to scan keys using some form of query and return related windowids.

Please check provided concrete implementation of GeodeStore

ashishtadose:
@tushargosavi  @gauravgopi123 
Please review.

contrib/pom.xml
```diff
@@ -623,5 +628,12 @@
       <artifactId>super-csv-joda</artifactId>
       <version>2.3.1</version>
     </dependency>
+    <dependency>
+      <groupId>com.gemstone.gemfire</groupId>
+      <artifactId>gemfire</artifactId>
+      <version>8.2.0</version>
+      <type>jar</type>
+      <scope>compile</scope>
```
gauravgopi123: Default scope is compile so not required. I think the scope should be optional

gauravgopi123: I meant that this dependency should be set optional

ashishtadose: Yes dependency compile scope is default and should be optional. Fixed it.

contrib/pom.xml
```diff
@@ -623,5 +628,11 @@
       <artifactId>super-csv-joda</artifactId>
       <version>2.3.1</version>
     </dependency>
+    <dependency>
+      <groupId>com.gemstone.gemfire</groupId>
+      <artifactId>gemfire</artifactId>
+      <version>8.2.0</version>
+      <type>jar</type>
```
gauravgopi123: please add `<optional>true</optional>`
`type` is by default jar so not required

ashishtadose:
@gauravgopi123  Added optional tag to dependency.

pom.xml
```diff
@@ -184,7 +184,6 @@
       <groupId>org.apache.apex</groupId>
       <artifactId>apex-engine</artifactId>
       <version>${apex.core.version}</version>
-      <scope>test</scope>
```
tweise, line 195: Why this change?

contrib/pom.xml
```diff
@@ -53,6 +53,11 @@
         <updatePolicy>daily</updatePolicy>
       </releases>
     </repository>
+     <repository>
```
tweise: Is there no Geode release in Maven central?

pom.xml
ashishtadose, line 195: This is required as AbstractKeyValueStorageAgent.java needs access to com.datatorrent.stram.plan.logical and test scope would cause compilation failure.

Please check - incubator-apex-malhar/library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java:[229,34]

contrib/pom.xml
ashishtadose: No. Geode has not released a major release version yet.

library/src/main/java/com/datatorrent/lib/util/StorageAgentKeyValueStore.java
gauravgopi123: Additional method can be added to extended Storage Agent. For the ones which are not required you can leave their implementation empty or throw exception

pom.xml
tweise, line 195: Malhar code (except tests) should only depend on apex-api and apex-common, not engine internals.

ashishtadose, line 195: Hi @tweise 
Can you please take a look at incubator-apex-malhar/library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java:[229,34]

and suggest what can be done to accommodate this change without changing test score of apex-engine.

library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  /**
+   * Saves the yarn application id which can be used by create application
+   * specific table/region in KeyValue sore.
+   */
+  @Override
+  public void setApplicationAttributes(AttributeMap map)
+  {
+    this.applicationId = map.get(LogicalPlan.APPLICATION_ID);
```
tweise: Change this to DAGContext.APPLICATION_ID

library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
```diff
@@ -0,0 +1,236 @@
...
+   * @param windowId
+   *          window id of operator to checkpoint
+   */
+  @Override
+  public Object load(int operatorId, long windowId)
+  {
+    Object obj = null;
+    try {
+      obj = retrieve(generateKey(operatorId, windowId));
+      logger.info("retrieved object from store  key {} region {} ", generateKey(operatorId, windowId), applicationId);
```
tweise: change this to debug

library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
```diff
@@ -0,0 +1,236 @@
...
+  public void delete(int operatorId, long windowId) throws IOException
+  {
+
+    if (!getStore().isConnected()) {
+      getStore().connect();
+    }
+
+    try {
+      getStore().remove(generateKey(operatorId, windowId));
+      logger.info("deleted object from store key {} region {}", generateKey(operatorId, windowId));
```
tweise: debug

library/src/main/java/com/datatorrent/lib/util/AbstractKeyValueStorageAgent.java
ashishtadose: Ok. will make these changes and push.

ashishtadose:
@gauravgopi123 updated StorageAgentKeyValueStore to extend com.datatorrent.lib.db.KeyValueStore.
Can't completely remove it as to support Abstract implementation of storage agent which needs to integrate methods not supported in KeyValueStore. 
Also can't make these additional methods to throw exception as it is not supported in base interface.

@tweise @gauravgopi123  review comments have been incorporated.

contrib/pom.xml
```diff
@@ -623,5 +628,11 @@
       <artifactId>super-csv-joda</artifactId>
       <version>2.3.1</version>
     </dependency>
+    <dependency>
+      <groupId>com.gemstone.gemfire</groupId>
```
markito: Would you guys like to point it to Geode instead of GemFire ?  For now it would need to be pulled from incubator repository and from nightlies but a release is just about to start. 

``` <dependency>
      <groupId>org.apache.geode</groupId>
      <artifactId>gemfire-core</artifactId>
      <version>${geode.version}</version>
</dependency>
```

tweise: That would be preferred. We would get the dependency from Apache snapshots,
right? In that case, no extra repository declaration should be required.

ashishtadose:
During development Geode incubation branch wasn't stable so this was pointed other repo.
Recent Apache Geode alpha release is very stable.

Updated Geode dependency with Incubator Geode.

ashishtadose:
@tweise can anyone look into branch conflicts.

tweise:
@ashishtadose You need to rebase your branch and resolve all conflicts. You should also squash to commits. 

http://apex.incubator.apache.org/contributing.html#opening-pull-requests-contributors-

Doesn't this work depend on the upcoming Apex 3.3.0 release?

ashishtadose:
Yes it depends on Apex 3.3.0 release. 
I will work on branch rebase & resolving conflicts.

tweise:
@ashishtadose the 3.3.0 core dependency is in place now. Please squash and rebase.

ashishtadose:
@tweise  Added test setup instructions, fixed logging & exceptions.

PR #126, opened 2015-12-13
- MLHR-1940 #resolve #comment utility class which has methods to conv…

ilooner:
…ert from time to windows.


ilooner:
@chandnisingh please review

PR #127, opened 2015-12-13
MLHR-1941 #resolve #comment added slice comparator

otterc:


otterc:
This is taken out from an earlier release and added to util.

PR #129, opened 2015-12-14
MLHR-1943  #resolve #comment copied from Megh

brightchen:
@ilooner Could you please review and merge?

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/dimensions/aggregator/AggregateEvent.java
+++ b/library/src/main/java/com/datatorrent/lib/dimensions/aggregator/AggregateEvent.java
@@ -1,0 +1,35 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.lib.dimensions.aggregator;
+
+import it.unimi.dsi.fastutil.Hash;
+
+public interface AggregateEvent
+{
+  int getAggregatorIndex();
+
+  public static interface Aggregator<EVENT, AGGREGATE extends AggregateEvent> extends Hash.Strategy<EVENT>
+  {
+    AGGREGATE getGroup(EVENT src, int aggregatorIndex);
+
+    void aggregate(AGGREGATE dest, EVENT src);
+
+    void aggregate(AGGREGATE dest, AGGREGATE src);
+  }
+}


otterc:
licenses are missing

library/src/main/java/com/datatorrent/lib/dimensions/aggregator/Aggregator.java
```diff
@@ -0,0 +1,14 @@
+package com.datatorrent.lib.dimensions.aggregator;
+
+import com.datatorrent.lib.dimensions.AggregateEvent;
+
+import gnu.trove.strategy.HashingStrategy;
+
+public interface Aggregator<EVENT, AGGREGATE extends AggregateEvent> extends HashingStrategy<EVENT>
```
otterc: can this not be a nested interface of AggregateEvent since it depends on it.

otterc:
Please change the commit message "Added from previous release"

library/src/main/java/com/datatorrent/lib/dimensions/aggregator/Aggregator.java
ilooner: +1 This is a rare case where I agree with nesting the interface. An Aggregator will never be used without an AggregateEvent.

brightchen: what about vice versa ?

ilooner: There needs to be a consumer of an AggregateEvent, while it is possible to have an AggregateEvent by itself, the interface is meaningless without an Aggregator to operate on it. So I would say that the dependence goes both ways.

library/pom.xml
```diff
@@ -310,6 +310,11 @@
       <artifactId>commons-beanutils</artifactId>
       <version>1.8.3</version>
     </dependency>
+    <dependency>
+      <groupId>net.sf.trove4j</groupId>
+      <artifactId>trove4j</artifactId>
+      <version>3.0.3</version>
+    </dependency>
   </dependencies>
```
davidyan74, line 313: http://trove4j.sourceforge.net/html/license.html
We can't use trove4j because of GPL

ilooner:
Thanks Bright can you squash the commits, then I'll merge.

library/pom.xml
ilooner, line 313: @davidyan74 We know. This is part of a bigger change that Bright is making to move some classes to Malhar. The changes are being done incrementally and the removal of trove4j will be done as a separate change shortly.

otterc:
@ilooner Even if this is part of a big change, it should not contain gpl license. So for now @brightchen will remove this trove4j dependency and its usage. Then we can merge it.

ilooner:
@brightchen I would prefer to hold off removing trove4j till later, but since two people have raised this issue the trov4j replacement should be done now. The classes used from trov4j can be replaced by these classes from fastutil. The API is almost identical for the two libraries so it should be a small change.

http://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/Hash.Strategy.html

http://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/objects/Object2ObjectOpenCustomHashMap.html

brightchen:
@ilooner I think maybe better delay this task into the 3rd sub-task. It should do some test such as performance etc( I don't know if you had performance test or not ). And only the implementation need Hash Strategy interface.

ilooner:
@brightchen I think people's concern may come from the idea that trove4j may be included in a release. Can you find out when Malhar will be released from @tweise? If you are confident the 3rd subtask will be completed before release, then I don't see a problem with temporarily including it. If people insist more on not allowing it in at all, then you can reorder your sub-tasks to remove trove4j first and then make these changes.

brightchen:
@ilooner I don't think its a big deal even Malhar released with this interface. As we are going to do is only expand interface rather than change or limit the interface. Anyway, It's better to keep it as stable as possible. So I'll add the hash strategy interface.

brightchen:
@ilooner could you review and merge

tweise:
@brightchen trove4j cannot appear in Malhar. 
@ilooner We are getting close to 3.3.0 release: https://malhar.atlassian.net/issues/?jql=project%20%3D%20Malhar%20and%20fixVersion%20%3D%203.3.0%20and%20resolution%20%3D%20Unresolved

brightchen:
trove4j had already removed

PR #130, opened 2015-12-15
MLHR-1946 #resolve #comment Marked WindowUtils as unstable.

ilooner:


ilooner:
@chandnisingh Please review

otterc:
Don't think we need to mark this util as unstable. Why will you want to mark it unstable?

ilooner:
@chandnisingh The location of the class and the signature of the methods may change if it is used in @siyuanh api changes.

otterc:
I think then we need to go with performing semver only for @Stable classes. Don't think it is a good idea to make it Unstable. There is some time for high level api

PR #133, opened 2015-12-15
MLHR-1936 Adding NiFi operators to contrib

bbende:
This pull request adds NiFi operators to the contrib module. The integration is based off the NiFi site-to-site client which is the mechanism for pulling/pushing data to/from a NiFi instance or cluster. It is used internally when a NiFi instance communicates with another instance, and also between other third party frameworks. In a cluster, the site-to-site client knows about the nodes in the cluster and will pull/push to/from all nodes.

To test the operators against a running NiFi instance you can do the following:
- Download the latest NiFi release and extract it: https://nifi.apache.org/download.html
- Edit conf/nifi.properties and set nifi.remote.input.socket.port to any available port, and nifi.remote.input.secure=false
- Start NiFi with bin/nifi.sh start
- Go the NiFi UI in your browser http://localhost:8080/nifi/
- Import the NiFi_Apex template that will be attached to the JIRA, template button in top-right toolbar
- Drag the template on to the canvas, button near top-left of toolbar
- Start all of the processors and ports 
- Run either TestNiFiInputApplication or TestNiFiOutputApplication in contrib/src/test/java/com.datatorrent.contrib.nifi.demo

nifi-apex-flow

The left side of the flow generates fake data to a NiFi output port, the NiFi Site-To-Site client will pull data from this output port in the case of the Malhar NiFi input operator. The right side has an a NiFi input port waiting for data to be pushed from the Malhar NiFi output operator.

Some questions/comments....
- Is using activate/deactivate preferred over setup/teardown? 
- I wasn't sure if calling Thread.sleep in emitTuples was appropriate here. We did this in other integrations such as storm and spark, but in those cases we were with in a while loop so we didn't want to make a ton of requests when no data was available. I'm not familiar with how Malhar will call emitTuples, so let me know if this doesn't sense here.
- Wasn't sure about the preferred approach for exception handling, seems wrapping a checked exception with RuntimeException is appropriate?
- I noticed some other operators used a WindowDataManager which I think was to be able to replay data,  didn't really know enough about that to know if that was needed here.

### DIFF
--- a/contrib/pom.xml
+++ b/contrib/pom.xml
@@ -220,6 +220,12 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.apache.nifi</groupId>
+      <artifactId>nifi-site-to-site-client</artifactId>
+      <version>0.4.1</version>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
       <groupId>org.codehaus.janino</groupId>
       <artifactId>janino</artifactId>
       <version>2.7.8</version>
--- a/contrib/src/main/java/com/datatorrent/contrib/nifi/AbstractNiFiInputOperator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/nifi/AbstractNiFiInputOperator.java
@@ -1,0 +1,212 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.contrib.nifi;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.nifi.remote.Transaction;
+import org.apache.nifi.remote.TransferDirection;
+import org.apache.nifi.remote.client.SiteToSiteClient;
+import org.apache.nifi.remote.protocol.DataPacket;
+
+import com.datatorrent.api.Context;
+import com.datatorrent.api.InputOperator;
+import com.datatorrent.lib.util.WindowDataManager;
+import com.datatorrent.netlet.util.DTThrowable;
+
+/**
+ * This is the base implementation of a NiFi input operator.&nbsp;
+ * Subclasses should implement the methods which convert NiFi DataPackets to tuples and emit them.
+ * <p>
+ * Ports:<br>
+ * <b>Input</b>: No input port<br>
+ * <b>Output</b>: Can have any number of output ports<br>
+ * <br>
+ * Properties:<br>
+ * None<br>
+ * <br>
+ * Compile time checks:<br>
+ * Classes derived from this have to implement the abstract methods emitTuples(List<T> tuples)&nbsp;
+ * and createTuple(DataPacket dp)<br>
+ * <br>
+ * Run time checks:<br>
+ * None<br>
+ * <br>
+ * Benchmarks:<br>
+ * TBD<br>
+ * </p>
+ *
+ * @displayName Abstract NiFi Input
+ * @category Messaging
+ * @tags input operator
+ */
+
+public abstract class AbstractNiFiInputOperator<T> implements InputOperator
+{
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(AbstractNiFiInputOperator.class);
+
+  private transient SiteToSiteClient client;
+  private final SiteToSiteClient.Builder siteToSiteBuilder;
+
+  private transient int operatorContextId;
+  private transient long currentWindowId;
+  private transient List<T> currentWindowTuples;
+  private transient List<T> recoveredTuples;
+  private final WindowDataManager windowDataManager;
+
+  /**
+   * @param siteToSiteBuilder the builder for a NiFi SiteToSiteClient
+   * @param windowDataManager a WindowDataManager to save and load state for windows of tuples
+   */
+  public AbstractNiFiInputOperator(final SiteToSiteClient.Builder siteToSiteBuilder,
+      final WindowDataManager windowDataManager)
+  {
+    this.siteToSiteBuilder = siteToSiteBuilder;
+    this.windowDataManager = windowDataManager;
+  }
+
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    this.client = siteToSiteBuilder.build();
+    this.operatorContextId = context.getId();
+    this.currentWindowTuples = new ArrayList<>();
+    this.recoveredTuples = new ArrayList<>();
+    this.windowDataManager.setup(context);
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
+    currentWindowId = windowId;
+
+    // if the current window is now less than the largest window, then we need to replay data
+    if (currentWindowId <= windowDataManager.getLargestRecoveryWindow()) {
+      try {
+        List<T> recoveredData =  (List<T>)this.windowDataManager.load(operatorContextId, windowId);
+        if (recoveredData == null) {
+          return;
+        }
+
+        // if we recovered tuples then load them to be processed by next call to emitTuples()
+        recoveredTuples.addAll(recoveredData);
+      } catch (IOException e) {
+        DTThrowable.rethrow(e);
+      }
+    }
+  }
+
+  @Override
+  public void emitTuples()
+  {
+    // if we have recovered tuples we must be replaying a previous window so emit them,
+    // clear the recovered list, and return until we have no more recovered data
+    if (recoveredTuples.size() > 0) {
+      emitTuples(recoveredTuples);
+      recoveredTuples.clear();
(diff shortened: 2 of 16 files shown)


otterc:
@bbende Thanks for the contribution. We follow a certain style and there are checkstyle violations in the pull request. If you use IntelliJ, the code style settings can be found here:
https://github.com/apache/incubator-apex-core/blob/devel-3/.idea/codeStyleSettings.xml

The settings for other IDEs are outdated but let me know if you use any of those.

otterc:
Wasn't sure about the preferred approach for exception handling, seems wrapping a checked exception with RuntimeException is appropriate?

Yes, that is appropriate. Swallowing exceptions has proven to be troublesome.

bbende:
@chandnisingh thanks for pointing out the styling, i should have looked at that before submitting. I use IntelliJ and think I have it all reformatted now, ran mvn checkstyle:check and there are no more violations for nifi related classes.

Since this modified most of the files, do you prefer I squash the commit and force push? or do you want to see it as a separate commit?

otterc:
Thanks @bbende We do prefer a single commit in the end but you don't have to do this right away. Once the pull request is ready to be merged, you can squash all your commits then. I will be responding to your other questions soon

otterc:
I wasn't sure if calling Thread.sleep in emitTuples was appropriate here. We did this in other integrations such as storm and spark, but in those cases we were with in a while loop so we didn't want to make a ton of requests when no data was available. I'm not familiar with how Malhar will call emitTuples, so let me know if this doesn't sense here.

@bbende 
If emitTuples doesn't generate tuples, then the main operator thread will sleep before emitTuples is called again. So there isn't a need to sleep here.

otterc:
Is using activate/deactivate preferred over setup/teardown?

We use activate/deactivate in the following scenario:
While integration with certain systems (for example message queues), if setting up session initiates receiving messages immediately then setting of this session belongs in activate rather than setup because  in setup the operator might be doing some intensive tasks like IO operations, etc. and it will not help to buffer the messages.

otterc:
I noticed some other operators used a WindowDataManager which I think was to be able to replay data, didn't really know enough about that to know if that was needed here.

WindowDataManager is used for making an operator idempotent. I think we don't have to address this now. In the first cut we can focus on making the adapter fault-tolerant.

bbende:
@chandnisingh thanks for answering those questions, pushed some changes up based on that feedback, and got the travis build passing again.

contrib/pom.xml
```diff
@@ -220,6 +220,11 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.apache.nifi</groupId>
+      <artifactId>nifi-site-to-site-client</artifactId>
+      <version>0.4.0</version>
```
gauravgopi123: This dependency should be marked optional

contrib/src/main/java/com/datatorrent/contrib/nifi/AbstractNiFiInputOperator.java
```diff
@@ -0,0 +1,162 @@
...
+ * None<br>
+ * <br>
+ * Benchmarks:<br>
+ * TBD<br>
+ * </p>
+ *
+ * @displayName Abstract NiFi Input
+ * @category Messaging
+ * @tags input operator
+ * @since 3.3.0
```
gauravgopi123: since tag not required

contrib/src/main/java/com/datatorrent/contrib/nifi/AbstractNiFiInputOperator.java
```diff
@@ -0,0 +1,162 @@
...
+ */
+
+public abstract class AbstractNiFiInputOperator<T> implements InputOperator
+{
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(AbstractNiFiInputOperator.class);
+
+  private SiteToSiteClient client;
```
gauravgopi123: This should be set transient as it is initialized in setup

tweise:
Bryan, thanks for contributing the NiFi support! 

There are a few code style related things to be addressed, we also realized that the pointers were missing and added them now: http://apex.incubator.apache.org/contributing.html#code-style

contrib/src/main/java/com/datatorrent/contrib/nifi/AbstractNiFiInputOperator.java
```diff
@@ -0,0 +1,162 @@
...
+
+      // confirm all of the expected data was received
+      transaction.confirm();
+
+      // delegate to sub-classes to emit the tuples
+      emitTuples(tuples);
+
+      // everything was successful so complete the transaction
+      transaction.complete();
+
```
tweise, line 164: Doing the confirm here will lead to data loss when the operator fails. We would need to retrieve the data into a fault tolerant storage before confirming the transaction. This looks similar to JMS, where we cannot acknowledge the message prior to having it stored. The place to store data would be endWindow. Looks like the pattern from JSM applies here. Kafka and files are different because they retain the data and we can simply go back to where the operator left off.

contrib/src/main/java/com/datatorrent/contrib/nifi/AbstractNiFiInputOperator.java
```diff
@@ -0,0 +1,162 @@
...
+    this.client = siteToSiteBuilder.build();
+  }
+
+  @Override
+  public void teardown()
+  {
+    try {
+      client.close();
+    } catch (IOException e) {
+      LOGGER.error(e.getMessage(), e);
```
gauravgopi123: don't swallow exception

contrib/src/main/java/com/datatorrent/contrib/nifi/AbstractNiFiInputOperator.java
tweise, line 164: The JMS code is here:  https://github.com/apache/incubator-apex-malhar/blob/master/library/src/main/java/com/datatorrent/lib/io/jms/AbstractJMSInputOperator.java

It also shows how data is managed by window for replay (you already asked about the window data manager). It's a write-ahead-log that on replay let's us retain the relationship between individual events and windows. That's to achieve idempotency. Downstream processing that is window based (vs. event by event) will require this.

contrib/src/main/java/com/datatorrent/contrib/nifi/AbstractNiFiInputOperator.java
```diff
@@ -0,0 +1,162 @@
...
+        LOGGER.debug("No data available to pull, returning and will try again...");
+        return;
+      }
+
+      // read all of the available data packets and convert to the given type
+      final List<T> tuples = new ArrayList<>();
+      do {
+        tuples.add(createTuple(dataPacket));
+        dataPacket = transaction.receive();
+      } while (dataPacket != null);
```
gauravgopi123, line 153: can this lead to infinite loop? May be you want to add number of DataPacket condition as well?

contrib/src/main/java/com/datatorrent/contrib/nifi/AbstractNiFiOutputOperator.java
```diff
@@ -0,0 +1,118 @@
...
+    this.client = siteToSiteBuilder.build();
+  }
+
+  @Override
+  public void teardown()
+  {
+    try {
+      client.close();
+    } catch (IOException e) {
+      LOGGER.error(e.getMessage(), e);
```
gauravgopi123: throw exception

contrib/src/main/java/com/datatorrent/contrib/nifi/AbstractNiFiOutputOperator.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  protected void processTuple(T tuple)
+  {
+    final NiFiDataPacket niFiDataPacket = dataPacketBuilder.createNiFiDataPacket(tuple);
+
+    try {
+      final Transaction transaction = client.createTransaction(TransferDirection.SEND);
```
DT-Priyanka, line 168: If creating a transaction is costly operation we might create batch of tuples and send them in one transaction during "endWindow". Creating a transaction to send only one tuple is optimal?

bbende:
Just wanted to let everyone know I haven't forgot about this. Hoping to get back to it soon and make the recommended changes, thanks for all of the feedback.

contrib/src/main/java/com/datatorrent/contrib/nifi/AbstractNiFiOutputOperator.java
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * @since 3.3.0
+ *
+ */
+public abstract class AbstractNiFiOutputOperator<T> extends BaseOperator
+{
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(AbstractNiFiOutputOperator.class);
+
+  protected SiteToSiteClient client;
```
chinmaykolhatkar: Can you please make this transient as this gets initialized in setup.

bbende:
I updated this pull request based on everyone's feedback. The biggest change here is adding the WindowDataManager to ensure data is saved before we complete a NiFi transaction, at which point it would be gone if we lost it. I based this mostly off looking at the RabbitMQ operators, but let me know if anything doesn't seem right.

contrib/src/main/java/com/datatorrent/contrib/nifi/AbstractNiFiInputOperator.java
bbende, line 153: shouldn't be able to get an infinite loop... transaction.receive() will return null when there is no matter data to pull, or when it is has pulled the maximum number of data packets for a transaction (configured on the site-to-site client)

tweise:
@bbende thanks for addressing the comments, now Apex officially integrates with NiFi!

I see potential for improvement around the efficiency of flushing the WAL for recovery. It is done in emitTuples and endWindow, it would be sufficient to do it in endWindow only if you can defer the txn confirm to endWindow as well?

Also note that the development branch and target for new PRs is now master and your changes are already there.

bbende:
@tweise thanks for merging, excited to have this integration!

I agree about the possible performance of flushing the WAL in every call to emitTuples. I was a little bit hesitant about leaving the transactions open and trying to commit them later, although it may work fine. I figured we could see how this first pass goes and then definitely consider addressing that if needed.

PR #135, opened 2015-12-15
MLHR-1944 #comment Added from previous release

brightchen:
@ilooner @chandnisingh please review and merge

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/dimensions/CustomTimeBucketRegistry.java
+++ b/library/src/main/java/com/datatorrent/lib/dimensions/CustomTimeBucketRegistry.java
@@ -1,0 +1,138 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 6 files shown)


otterc:
are there tests for it too?

otterc:
Just want to make sure if we missed moving the tests. If there weren't any tests it is still file.

brightchen:
DimensionsConversionContext is a class, but in fact don't have any logic, only properties.
I don't find any test case related to it.

ilooner:
@brightchen There are some for CustomTimeBucketRegistry and DimensionsDescriptor

brightchen:
@ilooner why the tester in different package? which package make more sense, dimensions or appdata/dimensions?
bright-mac:Megh bright$ find . -iname "CustomTimeBucketRegistry*.java"
./library/src/main/java/com/datatorrent/lib/dimensions/CustomTimeBucketRegistry.java
./library/src/test/java/com/datatorrent/lib/appdata/dimensions/CustomTimeBucketRegistryTest.java

ilooner:
@brightchen Good point, I agree they should be in the same package. I made the mistake of putting them in different packages.

ilooner:
@brightchen please remove @since tags in the java doc at the top of the class in CustomTimeBucketRegistry and DimensionsDescriptor. Those will be inserted by the release process. Otherwise looks good and I will merge after that.

brightchen:
@ilooner the @since tags had removed, please merge

PR #137, opened 2015-12-16
Added XSD validation to XML parsers and moved XML and JSON parsers to Malhar Library

ishark:


library/src/main/java/com/datatorrent/lib/formatter/XmlFormatter.java
```diff
@@ -57,45 +55,32 @@ public XmlFormatter()
...
+      StringWriter writer = new StringWriter();
+      if (getAlias() != null) {
+        marshaller.marshal(new JAXBElement(new QName(getAlias()), tuple.getClass(), tuple), writer);
+      } else {
+        marshaller.marshal(new JAXBElement(new QName(getClazz().getSimpleName()), tuple.getClass(), tuple), writer);
+      }
+      return writer.toString();
+    } catch (Exception e) {
+      logger.debug("Error while converting tuple {} {} ", tuple, e.getMessage());
       return null;
```
otterc, line 84: Swallowing exception?

library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
```diff
@@ -0,0 +1,139 @@
...
+      if (parsedOutput.isConnected()) {
+        DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+        DocumentBuilder builder;
+        try {
+          builder = factory.newDocumentBuilder();
+          Document doc = builder.parse(inputTuple);
+          parsedOutput.emit(doc);
+
+        } catch (Exception e) {
+          e.printStackTrace();
```
otterc: swallowing exception?

library/src/main/java/com/datatorrent/lib/formatter/XmlFormatter.java
ishark, line 84: This is because we do not want the operator to stop if one of the input strings is incorrect. The super class emits the tuple to error port if parsing fails.

library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
ishark: Removed print statement. The input is already validated before it parses. So, it should not reach in catch block.

library/src/main/java/com/datatorrent/lib/parser/Parser.java
```diff
@@ -47,10 +47,19 @@
...
+  @AutoMetric
+  protected long emittedObjectCount;
+  @AutoMetric
+  protected long rollingPercentValidTuples;
+  @AutoMetric
+  protected long rollingPercentErrorTuples;
+  @AutoMetric
+  protected long incomingTuplesCount;
 
   @OutputPortFieldAnnotation(schemaRequired = true)
```
shubham-pathak22, line 60: schemaRequired=true is very restrictive. Suppose XMLParser  operator is used  and i don't want pojos and so i will not connect to this port. But because of schemaRequired=true  application will fail during verification.

library/src/main/java/com/datatorrent/lib/parser/Parser.java
```diff
@@ -61,25 +70,48 @@ public void setup(PortContext context)
     }
   };
 
-  @OutputPortFieldAnnotation(optional = true)
   public transient DefaultOutputPort<INPUT> err = new DefaultOutputPort<INPUT>();
-
+  public transient DefaultOutputPort<INPUT> validatedOutput = new DefaultOutputPort<INPUT>();
```
shubham-pathak22: Do we need this ? We had a discussion on eng regarding whether this port would be useful and the opinion was this wont be used.

library/src/main/java/com/datatorrent/lib/parser/Parser.java
```diff
@@ -61,25 +70,48 @@ public void setup(PortContext context)
     }
   };
 
-  @OutputPortFieldAnnotation(optional = true)
   public transient DefaultOutputPort<INPUT> err = new DefaultOutputPort<INPUT>();
```
shubham-pathak22: Instead of err<INPUT>, I feel error port with KeyValPair<STRING,STRING>  would be better as we could give the error reason along with that.

shubham-pathak22:
I feel we should move Csv parser also to malhar/lib. Having parsers in contrib as well as in lib doesn't seem correct. All of them must belong to same place either lib/contrib. This would also help users to locate operators easily.

library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
```diff
@@ -0,0 +1,139 @@
...
+    // This method is not invoked for XML parser
+    return null;
+  }
+
+  @Override
+  public void processTuple(String inputTuple)
+  {
+    try {
+      if (out.isConnected()) {
+        StringReader reader = new StringReader(inputTuple);
```
shubham-pathak22: Can we consider ReusableStringReader here ?

library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
```diff
@@ -0,0 +1,139 @@
...
+
+        } catch (Exception e) {
+          // It should not reach here, since inputTuple string is already validaed.
+        }
+      }
+    } catch (Exception e) {
+      LOG.debug("Failed to parse xml tuple {}, Exception = {} ", inputTuple, e);
+      errorTupleCount++;
+      if (err.isConnected()) {
+        err.emit(inputTuple);
```
shubham-pathak22, line 112: Can we have error port emit reason as well ? User would want to know reason for validation failure.

library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
```diff
@@ -0,0 +1,139 @@
...
+  {
+    try {
+      if (out.isConnected()) {
+        StringReader reader = new StringReader(inputTuple);
+        JAXBElement<?> output = unmarshaller.unmarshal(new StreamSource(reader), getClazz());
+        LOG.debug(output.getValue().toString());
+        emittedObjectCount++;
+        out.emit(output.getValue());
+      } else {
+        validator.validate(new StreamSource(inputTuple));
```
shubham-pathak22, line 90: why is validator.validate in else part ? Shouldn't it be the first step?

library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  @Override
+  public void setup(com.datatorrent.api.Context.OperatorContext context)
+  {
+    try {
+      JAXBContext ctx = JAXBContext.newInstance(getClazz());
+      unmarshaller = ctx.createUnmarshaller();
+      if (schemaXSDFile != null) {
+        Path filePath = new Path(schemaXSDFile);
```
shubham-pathak22, line 138: Can we have reading the file part be put into a utility class that has method to take uri and return contents of the file as stream / string . so that even others can reuse the code ?

library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
```diff
@@ -0,0 +1,139 @@
...
+        Configuration configuration = new Configuration();
+        FileSystem fs = FileSystem.newInstance(filePath.toUri(), configuration);
+        FSDataInputStream inputStream = fs.open(filePath);
```
shubham-pathak22, line 141: Closing fs ?

library/src/main/java/com/datatorrent/lib/parser/Parser.java
```diff
@@ -61,25 +70,48 @@ public void setup(PortContext context)
...
+  {
+    errorTupleCount = 0;
+    emittedObjectCount = 0;
+    rollingPercentValidTuples = 0;
+    rollingPercentErrorTuples = 0;
+  }
+
+  @Override
+  public void endWindow()
+  {
```
shubham-pathak22, line 106: Why are we emitting percentage ? AppDataTracker aggregates the counters per window basis. On ADT side we have sum/max/avg etc which will not hold true/have no meaning w.r.t percentages.

library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
shubham-pathak22, line 90: Can validator be null if schemaXSDFile is not specified ?

library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
```diff
@@ -0,0 +1,139 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+
+import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.netlet.util.DTThrowable;
+
+public class XmlParser extends Parser<String>
```
shubham-pathak22: java docs

library/src/main/java/com/datatorrent/lib/parser/Parser.java
ishark, line 60: This only applies to POJO output port, This is another way to set clazz attribute in operator, by setting Context.PortContext.TUPLE_CLASS property to expected pojo output class type. If pojo output port is not connected, it wouldn't fail in this check.

library/src/main/java/com/datatorrent/lib/parser/Parser.java
ishark: Removed it.

library/src/main/java/com/datatorrent/lib/parser/Parser.java
ishark, line 106: Valid point. I added these, since it was mentioned to capture percent metrics as well. Will check what is the best way to add these. I think max and avg still will hold for percentages. Sum probably is not relevant.

library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
ishark, line 90: Unmarshal does validation as well, so no need to explicitly validate in POJO case.
Yes, validate should only be applied if Schema file is specified. Will fix that.

library/src/main/java/com/datatorrent/lib/parser/Parser.java
ishark: I think logs is better place to capture reason for error tuples as it can print stack trace as well.

shubham-pathak22: Certainly, logs can contain stack trace. But a short error message on error port along with error tuple would be useful for users to quickly find out validation errors by connecting to the error port instead of digging into the logs.
What i suggest, the port should at least have provision to send error message.

library/src/main/java/com/datatorrent/lib/parser/Parser.java
shubham-pathak22, line 60: Well, what i meant is if user doesn't want pojos he/she wont connect to this  port and would not set the class property as well. Basically user wont provide class if pojo's aren't required. But because of schemaRequired = true validation would fail

library/src/main/java/com/datatorrent/lib/parser/Parser.java
otterc: Why don't we remove the error port and add process error tuples method which by default logs error messages. I think logging is the obvious default behavior.

Users can add error port and  override this processErrorTuples().  The reason is that we don't know what kind of tuples are error tuples - same type as Input tuples or different.  
From the customer use cases seen so far they are either input tuples which are logged to files or the type is completely different.
Currently we don't have a reason to assume that error tuples are KeyValPair so that shouldn't be picked as default.

library/src/main/java/com/datatorrent/lib/parser/Parser.java
otterc, line 106: AppDataTracker does first, last as well. Application developer can provide the dimension scheme for the %metrics to be just FIRST/LAST

library/src/main/java/com/datatorrent/lib/parser/Parser.java
otterc, line 60: @ishark is correct. If the schemaRequired port is not connected, then validation will not fail even though the TUPLE_CLASS attribute is missing.

@shubham-pathak22 Schema required annotation is an expressive way to let the users know that processing of tuples which are received on this port requires the tuple schema. I will suggest that if you face any limitations or problems with this feature, then you should take it up on apex dev.

library/src/main/java/com/datatorrent/lib/parser/Parser.java
ishark: Would it be good to have default as  emitting same input to error port and logging error.
If transformation is needed, like you suggested, users can add/override additional port for processedError or connect another operator for error processing to the default error port, whichever applies to specific scenario.

otterc: The problem with that is they will have to add another port because the type here is tied to INPUT. I know its optional here but still there will be unused error ports.

otterc: @ishark please consider my comment as just a suggestion. I don't have a strong opinion here. These things evolve as we see more use cases.

ishark: yes, thanks Chandni. I will make the type of error port as generics. But I will leave the behavior of emitting of error tuples by default as is for now. We can revisit it later if needed.

otterc:
Right now we don't have a decision of how to address CSVFormatter in contrib which adds another dependency. That can be done later when we decide.

I think this change is good to go. I will merge it as all the other concerns are addressed.

library/src/main/java/com/datatorrent/lib/parser/Parser.java
shubham-pathak22, line 60: @ishark @chandnisingh Yes, you are correct. I tested it again. No need to remove the annotation

library/src/main/java/com/datatorrent/lib/parser/Parser.java
shubham-pathak22: As part of requirement, Csv parser operator , json and xml parser  would have to be enhanced to emit error tuples with error message . I get @chandnisingh suggestion of not keeping error port as KeyValuePair by default, but also its type shouldn't be INPUT as well which is the case now.   Better approach would be to have generic ERROR and leave the implementation to child operators

shubham-pathak22:
@chandnisingh i suggest we decide on moving CsvParser and CSVFormatter to lib in this PR only . Its just one dependency and also having single place for all parsers would be convinient for users to locate them.

library/src/main/java/com/datatorrent/lib/parser/Parser.java
shubham-pathak22, line 106: Thanks for the suggestion. Did not know that we can provide dimension scheme within the operator for the metrics. Do we have any examples ?

library/src/main/java/com/datatorrent/lib/parser/Parser.java
ishark: hi @shubham-pathak22, I have already changed the type of error port to be ERROROUT in the latest diff. However, the concrete implementations still emit the same tuple on error.

otterc:
@shubham-pathak22 it has been discussed on apex dev that we cannot add another dependency to library. There are couple of things that we can do with SuperCsv usages. We can combine them in their own small module like what we are doing with kafka, hbase, etc. So until that decision is made, we can't move it. This pull request moves other parsers to library which do not have additional dependency.  IMO it is not efficient to block this change because we have not concluded how to address SuperCsv.

library/src/test/java/com/datatorrent/lib/parser/XmlParserTest.java
```diff
@@ -65,6 +73,19 @@ protected void finished(Description description)
   }
 
   @Test
+  public void testOperatorSerialization()
```
shubham-pathak22, line 76: Curious to know purpose of this test. Isn't it something that applies to all operators ?

shubham-pathak22:
@chandnisingh i agree its not efficient to block the change. I see there's discussion going on creating small modules, but seems like it will take some time to materialize. Lets go ahead and merge this for now.

library/src/test/java/com/datatorrent/lib/parser/XmlParserTest.java
ishark, line 76: @shubham-pathak22  this test is only to check if the operator is kryo serializable. If operator's test run a dag then serialization automatically gets validated in unit tests. However, all the tests for this operator directly call process on operator input port, so serialization never got checked. Hence added the test. 
But yes, @chandnisingh had mentioned there could be a utility for this test going forward.

shubham-pathak22:
Thanks @ishark for addressing all comments.
@chandnisingh seems like all comments have been addressed. If there are no further concerns/review comments, lets merge this. Enhancing json and csv parser depends on this.

PR #139, opened 2015-12-16
MLHR-1931 #resolve

otterc:
@tushargosavi Please review and merge

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/fileaccess/DTFileReader.java
+++ b/library/src/main/java/com/datatorrent/lib/fileaccess/DTFileReader.java
@@ -68,13 +68,13 @@ public class DTFileReader implements FileAccess.FileReader
   }
 
   @Override
-  public void readFully(TreeMap<Slice, byte[]> data) throws IOException
+  public void readFully(TreeMap<Slice, Slice> data) throws IOException
   {
     scanner.rewind();
     for (; !scanner.atEnd(); scanner.advance()) {
       Entry en = scanner.entry();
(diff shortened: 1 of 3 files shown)


tweise:
Why is this not done using the reader instead of adding code to each implementation? Also please use human readable commit messages and pull request descriptions.

otterc:
Updated the commit message.
The change adds retrieving values as Slice with readFully. 
There is already a comment on existing readFully(...) -

```
    //Move to
    // void readFully(TreeMap<Slice, Slice> data) throws IOException;
       void readFully(TreeMap<Slice, byte[]> data) throws IOException;
```

Also when next reads value as Slice, then readFully should be consistent with it.

tweise:
Would like to understand what this is needed for.

otterc:
I need values as Slice so I don't have to convert it to byte[]. 
When reading value of a particular key,  FileReader.next(Slice key, Slice value) let's me do that.
However the existing readFully expects the value as byte[]. There was already a comment there that it should move to `(TreeMap<Slice, Slice>)` so since I needed it now I added that.

tweise:
The comment was move, not duplicate :-)

otterc:
I can move it though if later HDHT uses this API in Malhar, then it will have to change its implementation and so I didn't remove the existing method.

otterc:
Also because of type erasure I can't add another method 
`readFully(TreeMap<Slice, Slice> data)`

tweise:
The class is "new" in Malhar and we should make the changes to the old code unless you foresee a real problem doing so.

otterc:
All right. Moved the method instead of adding another one.

@tushargosavi @tweise please merge

PR #141, opened 2015-12-16
MLHR-1950 #comment marked classes that are frequently used as stable

otterc:


otterc:
@PramodSSImmaneni @ilooner Please review

otterc:
@PramodSSImmaneni  @ilooner Please let me know if you are going to review this pull request

ilooner:
Sorry @chandnisingh  somehow missed the first mail. Will look now.

contrib/src/main/java/com/datatorrent/contrib/kafka/KafkaSinglePortByteArrayInputOperator.java
```diff
@@ -19,6 +19,9 @@
 package com.datatorrent.contrib.kafka;
 
 import java.nio.ByteBuffer;
+
+import org.apache.hadoop.classification.InterfaceStability;
+
 import kafka.message.Message;
```
ilooner, line 22: why only the import statement added here?

siyuanh:
@chandnisingh Please mark the whole kafka package to stable

otterc:
@ilooner @siyuanh made the suggested changes

pramodin:
Will do @chandnisingh

pramodin:
This is introducing a dependency from operators directly to Hadoop. How about defining our own annotations or some other independent standard if it exists.

All the operators you picked seemed fine. How about applying it to other components such as codecs and partitioners. The other operators should also be marked as evolving or unstable so that there is no ambiguity.

We also need to document the definitions, for example stable may still mean the API could change in the next major version.

otterc:
@PramodSSImmaneni 
1. Agreed on the point of defining our own annotations so that the operators do not directly depend on Hadoop.
2. I picked components which were being used by demos and our POC applications. If partitioners and codecs are being used as well then I will mark them as stable.
3. We discussed on dev mailing list that the benefit of this approach is that we don't have to mark the majority of the class as unstable. Individual developers can mark operators/components evolving when they think it is not unstable. That effort is not part of this ticket.
4. In this change we were using hadoop's annotation and they do describe it.

pramodin:
Regd 3, one of the intention of this exercise is I think is to easily tell the user whether they need to be plan ahead for possible changes in future for the operator when they make a choice of using a particular operator. 

In my opinion it is much more useful when there is an evolving annotation than when there is a stable annotation as it serves as a caution notice. If they don't see anything a user may just proceed as everything is normal.

otterc:
@PramodSSImmaneni This exercise is specifically meant for japicmp plugin to perform semantic versioning of Stable classes.
Identifying @Evolving classes is part of another exercise. If we don't mark classes as Evolving or Unstable that anyways serves as a caution notice to the users because the Stable classes are marked stable.
There needs to be a different task which differentiates @Evolving from @Unstable but that is not part of this.

pramodin:
Unless that exercise follows immediately the intended goal is not fully served. If somebody picks an operator and they don't see anything mentioned there I don't think they will always reach a conclusion that there are other operators that are marked stable so this one is not. They may not even look at other packages or operators.

otterc:
The intended goal for this exercise is that japicmp plugin only verifies semantic versioning of Stable operators. We can add that to Malhar README or elsewhere.  But marking everything Evolving or Unstable is a different task.

ilooner:
I agree with Chandni that marking things Evolving or Unstable is a separate exercise. Yes it is more explicit to the user and should be done, but the annotation of operators should be done incrementally. Annotating everything in one pull request will be difficult to review. Furthermore there will likely be some debate on how to annotate some of the classes. Debate about those straggling classes should not prevent a needed change from being pushed, especially when doing things incrementally will not have a downside.

tweise:
Any further comments or can this be merged?

tweise:
@chandnisingh can you please look at the checkstyle failure and rebase?

otterc:
@tweise rebased. please merge

PR #145, opened 2015-12-22
APEXMALHAR-1897 added managed state 

otterc:
Still working on tests and will add documentation. 

@amberarrow @ilooner @PramodSSImmaneni 
Please review

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/fileaccess/FileAccess.java
+++ b/library/src/main/java/com/datatorrent/lib/fileaccess/FileAccess.java
@@ -52,7 +52,10 @@ public interface FileAccess extends Closeable
    * @throws IOException
    */
   void rename(long bucketKey, String oldName, String newName) throws IOException;
+
   void delete(long bucketKey, String fileName) throws IOException;
+
+  void deleteBucket(long bucketKey) throws IOException;
 
   long getFileSize(long bucketKey, String s) throws IOException;
 
--- a/library/src/main/java/com/datatorrent/lib/fileaccess/FileAccessFSImpl.java
+++ b/library/src/main/java/com/datatorrent/lib/fileaccess/FileAccessFSImpl.java
@@ -135,10 +135,13 @@ public abstract class FileAccessFSImpl implements FileAccess
   public RemoteIterator<LocatedFileStatus> listFiles(long bucketKey) throws IOException
   {
     Path bucketPath = getBucketPath(bucketKey);
-    if (!fs.exists(bucketPath)) {
-      return null;
-    }
-    return fs.listFiles(bucketPath, true);
+    return fs.exists(bucketPath) ? fs.listFiles(bucketPath, true) : null;
+  }
+
+  @Override
+  public void deleteBucket(long bucketKey) throws IOException
+  {
+    fs.delete(getBucketPath(bucketKey), true);
   }
 
   @Override
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/BucketedState.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/BucketedState.java
@@ -1,0 +1,72 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.apex.malhar.lib.state;
+
+import java.util.concurrent.Future;
+
+import javax.validation.constraints.NotNull;
+
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * A state where keys are grouped in buckets.
+ */
+public interface BucketedState
+{
+  /**
+   * An expired value. In some implementations where bucketId is time related then the event can be old and
+   * the get methods- getSync & getAsync return this fixed slice instance.<br/>
+   * In the usages, comparisons with EXPIRED should be made using <code>==</code> instead of <code>equals</code>.
+   */
+  Slice EXPIRED = new Slice(null, -1, -1);
+
+  /**
+   * Sets the value of the key in bucket identified by bucketId.
+   *
+   * @param bucketId identifier of the bucket.
+   * @param key      key (not null)
+   * @param value    value (not null)
+   */
+  void put(long bucketId, @NotNull Slice key, @NotNull Slice value);
+
+  /**
+   * Returns the value of the key in a bucket identified by bucketId. Fetching a key can be expensive if the key
+   * is not in memory and is present on disk. This fetches the key synchronously. <br/>
+   * {@link #getAsync(long, Slice)} is recommended for efficiently reading the value of a key.
+   *
+   * @param bucketId identifier of the bucket.
+   * @param key     key (not null)
+   *
+   * @return value of the key if found; null if the key is not found;
+   * {@link #EXPIRED} if the bucketId is time based and very old.
+   */
+  Slice getSync(long bucketId, @NotNull Slice key);
+
+  /**
+   * Returns the future using which the value is obtained.<br/>
+   *
+   * @param key key (not null)
+   *
+   * @return value of the key if found; null if the key is not found;
+   * {@link #EXPIRED} if the bucketId is time based and very old.
+   */
+  Future<Slice> getAsync(long bucketId, @NotNull Slice key);
+
+}
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/TimeSlicedBucketedState.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/TimeSlicedBucketedState.java
@@ -1,0 +1,104 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.lib.state;
+
+import java.util.concurrent.Future;
+
+import javax.validation.constraints.NotNull;
+
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * A type of bucketed state where a bucket's data is further divided into time buckets. This requires
+ * time per key to figure out which time bucket a particular key belongs to.
+ * <p/>
+ * The time here is mainly used for purging of aged key/value pair.
+ */
+public interface TimeSlicedBucketedState
+{
+  /**
+   * Sets the value of a key in the bucket identified by bucketId. Time is used to derive which time bucket (within
+   * the main bucket) a key belongs to.
+   *
+   * @param bucketId identifier of the bucket.
+   * @param time    time associated with the key.
+   * @param key     key   (not null)
+   * @param value   value (not null)
+   */
+  void put(long bucketId, long time, @NotNull Slice key, @NotNull Slice value);
+
+  /**
+   * Returns the value of the key in the bucket identified by bucketId.<br/>
+   * If the value of the key is not present in the bucket cache then this scans all the time bucket files on disk from
+   * the latest to the oldest.
+   *
+   * It retrieves the value synchronously that can be expensive.<br/>
+   * {@link #getAsync(long, Slice)} is recommended for efficient reading the value of a key.
+   *
+   *
+   * @param bucketId identifier of the bucket
+   * @param key key (not null)
+   *
+   * @return value of the key if found; null if the key is not found;
+   */
+  Slice getSync(long bucketId, @NotNull Slice key);
+
+
+  /**
+   * Returns the value of key in the bucket identified by bucketId.<br/>
+   * If the value of the key is not present in the bucket cache then this will use the time to derive the time
+   * bucket and just search for the key in a particular time bucket file.<br/>
+   *
+   * It retrieves the value synchronously which can be expensive.<br/>
+   * {@link #getAsync(long, long, Slice)} is recommended for efficiently reading the value of a key.
+   *
+   * @param bucketId identifier of the bucket.
+   * @param time  time for deriving the time bucket.
+   * @param key   key (not null)
+   *
+   * @return value of the key if found; null if the key is not found; {@link BucketedState#EXPIRED} if the time is old.
+   */
+  Slice getSync(long bucketId, long time, @NotNull Slice key);
+
+  /**
+   * Returns the future using which the value is obtained.<br/>
+   * If the value of the key is not present in the bucket cache then this searches for it in all the time buckets on
+   * disk.<br/>
+   * Time-buckets are looked-up in order from newest to oldest.
+   *
+   * @param bucketId identifier of the bucket.
+   * @param key      key (not null)
+   *
+   * @return value of the key if found; null if the key is not found;
+   */
+  Future<Slice> getAsync(long bucketId, @NotNull Slice key);
+
+  /**
+   * Returns the future using which the value is obtained.<br/>
+   * If the value of the key is not present in the bucket cache then this will use the time to derive the time
+   * bucket and just search for the key in a particular time bucket file.<br/>
+   *
+   * @param bucketId  identifier of the bucket.
+   * @param time     time associated with the key.
+   * @param key      key  (not null)
+   *
+   * @return value of the key if found; null if the key is not found; {@link BucketedState#EXPIRED} if time is very old.
+   */
+  Future<Slice> getAsync(long bucketId, long time, @NotNull Slice key);
+}
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/managed/AbstractManagedStateImpl.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/managed/AbstractManagedStateImpl.java
@@ -1,0 +1,583 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.lib.state.managed;
+
+import java.io.IOException;
+import java.util.Comparator;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicReference;
+
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+
+import org.joda.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.esotericsoftware.kryo.serializers.FieldSerializer;
+import com.esotericsoftware.kryo.serializers.JavaSerializer;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Throwables;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.ListMultimap;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Multimaps;
+import com.google.common.util.concurrent.Futures;
+
+import com.datatorrent.api.Component;
+import com.datatorrent.api.Context.DAGContext;
+import com.datatorrent.api.Context.OperatorContext;
+import com.datatorrent.api.Operator;
+import com.datatorrent.api.annotation.Stateless;
+import com.datatorrent.common.util.NameableThreadFactory;
+import com.datatorrent.lib.fileaccess.FileAccess;
+import com.datatorrent.lib.fileaccess.TFileImpl;
+import com.datatorrent.lib.util.comparator.SliceComparator;
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * An abstract implementation of managed state.<br/>
+ *
+ * The important sub-components here are:
+ * <ol>
+ *   <li>
+ *     {@link #checkpointManager}: writes incremental checkpoints in window files and transfers data from window
+ *     files to bucket files.
+ *   </li>
+ *   <li>
+ *     {@link #bucketsFileSystem}: manages writing/reading from all the buckets. A bucket on disk is further sub-divided
+ *     into time-buckets. This abstracts out updating time-buckets and meta files and reading from them.
+ *   </li>
+ *   <li>
+ *     {@link #timeBucketAssigner}: assigns time-buckets to keys and manages the time boundaries.
+ *   </li>
+ *   <li>
+ *     {@link #stateTracker}: tracks the size of data in memory and requests buckets to free memory when enough memory
+ *     is not available.
+ *   </li>
+ *   <li>
+ *     {@link #fileAccess}: plug-able file system abstraction.
+ *   </li>
+ * </ol>
+ * <p/>
+ * <b>Differences between different concrete implementations of {@link AbstractManagedStateImpl}</b>
+ * <table>
+ *   <tr>
+ *     <td></td>
+ *     <td>{@link ManagedStateImpl}</td>
+ *     <td>{@link ManagedTimeStateImpl}</td>
+ *     <td>{@link ManagedTimeUnifiedStateImpl}</td>
+ *   </tr>
+ *   <tr>
+ *     <td>Main buckets</td>
+ *     <td>identified by unique adhoc long ids that the user provides with the key.</td>
+ *     <td>same as ManagedStateImpl.</td>
+ *     <td>user doesn't provide bucket ids and instead just provides time. Time is used to derive the time buckets
+ *     and these are the main buckets.</td>
+ *   </tr>
+ *   <tr>
+ *     <td>Data on disk: data in buckets is persisted on disk with each bucket data further divided into
+ *     time-buckets, i.e., {base_path}/{bucketId}/{time-bucket id}</td>
+ *     <td>time-bucket is computed using the system time corresponding to the application window.</td>
+ *     <td>time-bucket is derived from the user provided time.</td>
+ *     <td>time-bucket is derived from the user provided time.
+ *     In this implementation operator id is used to isolate data of different partitions on disk, i.e.,
+ *     {base_path}/{operatorId}/{time-bucket id}</td>
+ *   </tr>
+ *   <tr>
+ *     <td>Bucket partitioning</td>
+ *     <td>bucket belongs to just one partition. Multiple partitions cannot write to the same bucket.</td>
+ *     <td>same as ManagedStateImpl.</td>
+ *     <td>multiple partitions can be working with the same time-bucket since time-bucket is derived from time.
+ *     This works because on disk each partition's data is segregated by the operator id.</td>
+ *   </tr>
+ *   <tr>
+ *     <td>Dynamic partitioning</td>
+ *     <td>can support dynamic partitioning by pre-allocating buckets.</td>
+ *     <td>same as ManagedStateImpl.</td>
+ *     <td>will not be able to support dynamic partitioning efficiently.</td>
+ *   </tr>
+ * </table>
+ *
+ */
+public abstract class AbstractManagedStateImpl
+    implements ManagedState, Component<OperatorContext>, Operator.CheckpointNotificationListener, ManagedStateContext,
+    TimeBucketAssigner.PurgeListener
+{
+  private long maxMemorySize;
+
+  protected int numBuckets;
+
+  @NotNull
+  private FileAccess fileAccess = new TFileImpl.DTFileImpl();
+  @NotNull
+  protected TimeBucketAssigner timeBucketAssigner = new TimeBucketAssigner();
+
+  protected Bucket[] buckets;
+
+  @Min(1)
+  private int numReaders = 1;
+  @NotNull
+  protected transient ExecutorService readerService;
+
+  @NotNull
+  protected IncrementalCheckpointManager checkpointManager = new IncrementalCheckpointManager();
+
+  @NotNull
+  protected BucketsFileSystem bucketsFileSystem = new BucketsFileSystem();
+
+  protected transient OperatorContext operatorContext;
+
+  @NotNull
+  protected Comparator<Slice> keyComparator = new SliceComparator();
+
+  protected final transient AtomicReference<Throwable> throwable = new AtomicReference<>();
+
+  @NotNull
+  @FieldSerializer.Bind(JavaSerializer.class)
+  private Duration checkStateSizeInterval = Duration.millis(
+      DAGContext.STREAMING_WINDOW_SIZE_MILLIS.defaultValue * OperatorContext.APPLICATION_WINDOW_COUNT.defaultValue);
+
+  @FieldSerializer.Bind(JavaSerializer.class)
+  private Duration durationPreventingFreeingSpace;
+
+  private transient StateTracker stateTracker = new StateTracker();
+
+  //accessible to StateTracker
+  final transient Object commitLock = new Object();
+
+  protected final transient ListMultimap<Long, ValueFetchTask> tasksPerBucketId =
+      Multimaps.synchronizedListMultimap(ArrayListMultimap.<Long, ValueFetchTask>create());
+
+  @Override
+  public void setup(OperatorContext context)
+  {
+    operatorContext = context;
+    fileAccess.init();
+
+    timeBucketAssigner.setPurgeListener(this);
+
+    //setup all the managed state components
+    timeBucketAssigner.setup(this);
+    checkpointManager.setup(this);
+    bucketsFileSystem.setup(this);
+
+    if (buckets == null) {
+      //create buckets array only once at start when it is not created.
+      numBuckets = getNumBuckets();
+      buckets = new Bucket[numBuckets];
+    }
+    for (Bucket bucket : buckets) {
+      if (bucket != null) {
+        bucket.setup(this);
+      }
+    }
+
+    stateTracker.setup(this);
+    long activationWindow = context.getValue(OperatorContext.ACTIVATION_WINDOW_ID);
+
+    if (activationWindow != Stateless.WINDOW_ID) {
+      //delete all the wal files with windows > activationWindow.
+      //All the wal files with windows <= activationWindow are loaded and kept separately as recovered data.
+      try {
+        for (long recoveredWindow : checkpointManager.getWindowIds(operatorContext.getId())) {
+          if (recoveredWindow <= activationWindow) {
+            @SuppressWarnings("unchecked")
+            Map<Long, Map<Slice, Bucket.BucketedValue>> recoveredData = (Map<Long, Map<Slice, Bucket.BucketedValue>>)
+                checkpointManager.load(operatorContext.getId(), recoveredWindow);
+            if (recoveredData != null && !recoveredData.isEmpty()) {
+              for (Map.Entry<Long, Map<Slice, Bucket.BucketedValue>> entry : recoveredData.entrySet()) {
+                int bucketIdx = prepareBucket(entry.getKey());
+                buckets[bucketIdx].recoveredData(recoveredWindow, entry.getValue());
+              }
+            }
+            checkpointManager.save(recoveredData, operatorContext.getId(), recoveredWindow,
+                true /*skipWritingToWindowFile*/);
+
+          } else {
+            checkpointManager.delete(operatorContext.getId(), recoveredWindow);
+          }
+        }
+      } catch (IOException e) {
+        throw new RuntimeException("recovering", e);
+      }
+    }
+
+    readerService = Executors.newFixedThreadPool(numReaders, new NameableThreadFactory("managedStateReaders"));
+  }
+
+  /**
+   * Gets the number of buckets which is required during setup to create the array of buckets.<br/>
+   * {@link ManagedTimeStateImpl} provides num of buckets which is injected using a property.<br/>
+   * {@link ManagedTimeUnifiedStateImpl} provides num of buckets which are calculated based on time settings.
+   *
+   * @return number of buckets.
+   */
+  public abstract int getNumBuckets();
+
+  public void beginWindow(long windowId)
+  {
+    if (throwable.get() != null) {
+      Throwables.propagate(throwable.get());
+    }
+    timeBucketAssigner.beginWindow(windowId);
+  }
+
+
+  /**
+   * Prepares the bucket and returns its index.
+   * @param bucketId bucket key
+   * @return bucket index
+   */
+  protected int prepareBucket(long bucketId)
+  {
+    stateTracker.bucketAccessed(bucketId);
+    int bucketIdx = getBucketIdx(bucketId);
+
+    Bucket bucket = buckets[bucketIdx];
+    if (bucket == null) {
+      //bucket is not in memory
+      bucket = newBucket(bucketId);
+      bucket.setup(this);
+      buckets[bucketIdx] = bucket;
+    } else if (bucket.getBucketId() != bucketId) {
+      handleBucketConflict(bucketIdx, bucketId);
+    }
+    return bucketIdx;
+  }
+
+  protected void putInBucket(long bucketId, long timeBucket, @NotNull Slice key, @NotNull Slice value)
+  {
+    Preconditions.checkNotNull(key, "key");
+    Preconditions.checkNotNull(value, "value");
+    if (timeBucket != -1) {
+      //time bucket is invalid data is not stored
+      int bucketIdx = prepareBucket(bucketId);
+      buckets[bucketIdx].put(key, timeBucket, value);
+    }
+  }
+
+  @SuppressWarnings("SynchronizationOnLocalVariableOrMethodParameter")
+  protected Slice getValueFromBucketSync(long bucketId, long timeBucket, @NotNull Slice key)
+  {
+    Preconditions.checkNotNull(key, "key");
+    int bucketIdx = prepareBucket(bucketId);
+    Bucket bucket = buckets[bucketIdx];
+    synchronized (bucket) {
+      return bucket.get(key, timeBucket, Bucket.ReadSource.ALL);
+    }
+  }
+
+  @SuppressWarnings("SynchronizationOnLocalVariableOrMethodParameter")
+  protected Future<Slice> getValueFromBucketAsync(long bucketId, long timeBucket, @NotNull Slice key)
+  {
+    Preconditions.checkNotNull(key, "key");
+    int bucketIdx = prepareBucket(bucketId);
+    Bucket bucket = buckets[bucketIdx];
+    synchronized (bucket) {
+      Slice cachedVal = bucket.get(key, timeBucket, Bucket.ReadSource.MEMORY);
+      if (cachedVal != null) {
+        return Futures.immediateFuture(cachedVal);
+      }
+      ValueFetchTask valueFetchTask = new ValueFetchTask(bucket, key, timeBucket, this);
+      tasksPerBucketId.put(bucket.getBucketId(), valueFetchTask);
+      return readerService.submit(valueFetchTask);
+    }
+  }
+
+  protected void handleBucketConflict(int bucketIdx, long newBucketId)
+  {
+    throw new IllegalArgumentException("bucket conflict " + buckets[bucketIdx].getBucketId() + " " + newBucketId);
+  }
+
+  protected int getBucketIdx(long bucketId)
+  {
+    return (int)(bucketId % numBuckets);
+  }
+
+  Bucket getBucket(long bucketId)
+  {
+    return buckets[getBucketIdx(bucketId)];
+  }
+
+  protected Bucket newBucket(long bucketId)
+  {
+    return new Bucket.DefaultBucket(bucketId);
+  }
+
+  public void endWindow()
+  {
+    timeBucketAssigner.endWindow();
+  }
+
+  @SuppressWarnings("SynchronizationOnLocalVariableOrMethodParameter")
+  @Override
+  public void beforeCheckpoint(long windowId)
+  {
+    Map<Long, Map<Slice, Bucket.BucketedValue>> flashData = Maps.newHashMap();
+
+    for (Bucket bucket : buckets) {
+      if (bucket != null) {
+        synchronized (bucket) {
+          Map<Slice, Bucket.BucketedValue> flashDataForBucket = bucket.checkpoint(windowId);
+          if (!flashDataForBucket.isEmpty()) {
+            flashData.put(bucket.getBucketId(), flashDataForBucket);
+          }
+        }
+      }
+    }
+    if (!flashData.isEmpty()) {
+      try {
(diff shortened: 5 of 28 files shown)


library/src/main/java/com/datatorrent/lib/state/BucketedState.java
```diff
@@ -0,0 +1,49 @@
...
+ * under the License.
+ */
+
+package com.datatorrent.lib.state;
+
+import java.util.concurrent.Future;
+
+import com.datatorrent.netlet.util.Slice;
+
+public interface BucketedState
```
tweise: Javadoc?

library/src/main/java/com/datatorrent/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -0,0 +1,487 @@
...
+import com.datatorrent.lib.fileaccess.FileAccess;
+import com.datatorrent.lib.fileaccess.TFileImpl;
+import com.datatorrent.lib.util.comparator.SliceComparator;
+import com.datatorrent.netlet.util.DTThrowable;
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * Not thread safe.
+ */
+public abstract class AbstractManagedStateImpl
```
tweise: Javadoc?

library/src/main/java/com/datatorrent/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -0,0 +1,487 @@
...
+  @NotNull
+  @FieldSerializer.Bind(JavaSerializer.class)
+  private Duration checkStateSizeInterval = Duration.millis(
+      DAGContext.STREAMING_WINDOW_SIZE_MILLIS.defaultValue * OperatorContext.APPLICATION_WINDOW_COUNT.defaultValue);
+
+  private transient StateTracker stateTracker;
+
+  private final transient Object commitLock = new Object();
+
+  @SuppressWarnings("unchecked")
```
tweise: Can the suppress be localized?

library/src/main/java/com/datatorrent/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -0,0 +1,487 @@
...
+    int bucketIdx = prepareBucket(groupId);
+    return buckets[bucketIdx].get(key, -1, Bucket.ReadSource.ALL);
+  }
+
+  @SuppressWarnings("SynchronizationOnLocalVariableOrMethodParameter")
+  @Override
+  public Future<Slice> getAsync(long groupId, Slice key)
+  {
+    int bucketIdx = prepareBucket(groupId);
+    Bucket bucket = buckets[bucketIdx];
```
tweise: Why the local variable?

library/src/main/java/com/datatorrent/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -0,0 +1,487 @@
...
+    return keyComparator;
+  }
+
+  @Override
+  public BucketsMetaDataManager getBucketsMetaDataManager()
+  {
+    return bucketsMetaDataManager;
+  }
+
+  public int getNumReaders()
```
tweise: Purpose of numReaders?

otterc: Haven't completed the documentation yet. There are multiple threads that can load different buckets from disk. This configures the number of threads in the pool. This is why synchronizing on a bucket in the async call so that fetching of a single bucket is done by a single thread.

library/src/main/java/com/datatorrent/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -0,0 +1,487 @@
...
+      replay = true;
+    }
+
+    readerService = Executors.newFixedThreadPool(numReaders, new NameableThreadFactory("managedStateReaders"));
+
+    stateTracker = new StateTracker(this, throwable);
+    stateTracker.setup(context);
+  }
+
+  public abstract int getNumBuckets();
```
tweise: Javadoc?

otterc:
I have addressed some of the the review comments by @tweise and @amberarrow. 
Added more test cases but still working on adding more of it. 
Added javadocs.

otterc:
@ilooner made TimeBucketAssigner use WindowBoundedService. please review

otterc:
Still need to address the review comments by @amberarrow

pramodin:
looking at the latest

library/src/main/java/com/datatorrent/lib/state/managed/TimeBucketAssigner.java
```diff
@@ -0,0 +1,223 @@
...
+  private long startTime;
+  private long endTime;
+  private int numBuckets;
+
+  private boolean initialized;
+
+  private transient WindowBoundedService windowBoundedService;
+
+  @NotNull
+  private final transient Set<Listener> listeners = Sets.newHashSet();
```
otterc: From @amberarrow 
Do we need a set here, i.e. do we expect more than one listener ? With a set, several questions come up:
1. We should document the fact that the listener should ensure that set membership works correctly by implementing hashCode()/equals().
2. The purgeTimeBucketsBefore() method should not take a long time since it is run with a lock held in the expiry task.
3. Do we need an unregister() method to remove a listener ?

All these go away if we limit it to just 1 listener.

---

Will change

library/src/main/java/com/datatorrent/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -0,0 +1,529 @@
...
+  /**
+   * Gets the number of buckets which is required during setup to create the array of buckets.<br/>
+   * {@link ManagedStateImpl} provides num of buckets which is injected using a property.<br/>
+   * {@link TimeManagedStateImpl} provides num of buckets which are calculated based on time settings.
+   *
+   * @return number of buckets.
+   */
+  public abstract int getNumBuckets();
+
+  public void beginWindow(long l)
```
otterc: From @amberarrow 
 descriptive parameter name like 'windowId' ? (also in TimeBucketAssigner)

---

will change

library/src/main/java/com/datatorrent/lib/state/managed/TimeBucketAssigner.java
```diff
@@ -0,0 +1,223 @@
...
+  {
+    @Override
+    public void run()
+    {
+      synchronized (lock) {
+        startTime += bucketSpanMillis;
+        endTime += bucketSpanMillis;
+        for (Listener listener : listeners) {
+          listener.purgeTimeBucketsBefore(startTime);
+        }
```
otterc: From @amberarrow 
Does this loop need to be inside the synchronized block ?

otterc: There will be only one listener. Notification is within the synchronized block because `startTime` is modified by the operator thread as well.

library/src/main/java/com/datatorrent/lib/state/managed/TimeBucketAssigner.java
```diff
@@ -0,0 +1,223 @@
...
+ * <p/>
+ *
+ * The configuration {@link #expireBefore} and {@link #bucketSpan}  are used to calculate number of time-buckets.
+ * For eg. if <code>expireBefore = 1 hour</code> and <code>bucketSpan = 30 minutes</code>, then <code>
+ *   numBuckets = 60 minutes/ 30 minutes = 2 </code>.
+ * <p/>
+ *
+ * The time boundaries- start and end, periodically move by span of a single time-bucket. Any event with time < start
+ * is expired. These boundaries slide between application window by another thread and not the operator thread.
+ */
```
otterc: From @amberarrow 

Seems like the operator thread could also move them in getTimeBucketFor(). Are startTime and endTime intended to bound the current time bucket ? If so, it's worth stating it here.

---

ok

library/src/main/java/com/datatorrent/lib/state/managed/TimeBucketAssigner.java
```diff
@@ -0,0 +1,223 @@
...
+      return -1;
+    }
+    long diffFromStart = value - fixedStartTime;
+    long key = diffFromStart / bucketSpanMillis;
+    if (value > lend) {
+      long move = ((value - lend) / bucketSpanMillis + 1) * bucketSpanMillis;
+      synchronized (lock) {
+        startTime = lstart + move;
+        endTime = lend + move;
+      }
```
otterc: From @amberarrow 

Not clear what we're trying to achieve with the code above; could you explain ? Can (value - lend) ever exceed bucketSpanMillis ?
Second, it seems like this assignment could overwrite the assignment made to startTime and endTime by the expiryTask thread in some scenarios, for example: (a) We copy the values in the first synchronized block above (b) The expiryTask thread increments values by bucketSpanMillis (c) We overwrite those values by the values computed here. Is this OK ?

library/src/main/java/com/datatorrent/lib/state/managed/TimeBucketAssigner.java
```diff
@@ -0,0 +1,223 @@
...
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    if (!initialized) {
+      if (bucketSpan == null) {
+        bucketSpan = Duration.millis(context.getValue(Context.OperatorContext.APPLICATION_WINDOW_COUNT) *
+            context.getValue(Context.DAGContext.STREAMING_WINDOW_SIZE_MILLIS));
+      }
+      Calendar calendar = Calendar.getInstance();
+      long now = calendar.getTimeInMillis();
```
otterc: From @amberarrow 
System.currentTimeMillis() ?

otterc: Done

library/src/main/java/com/datatorrent/lib/state/managed/TimeBucketAssigner.java
otterc: > > Can (value - lend) ever exceed bucketSpanMillis ?
> > Yes. `startTime` is initialized during setup() and `endTime = startTime + numBucket*bucketSpan`. 

However `value`  can come from the event. This time in the event can be ahead of `endTime` (we have seen this in a real application).  This triggers the whole boundary to move ahead.

> > Second, it seems like this assignment could overwrite the assignment made to startTime and endTime by the expiryTask thread in some scenarios, for example: (a) We copy the values in the first synchronized block above (b) The expiryTask thread increments values by bucketSpanMillis (c) We overwrite those values by the values computed here. Is this OK ?

We haven't seen a problem so far with this in BucketManager because the time in the event is given a priority over expiry task. However here there maybe an issue because the listener (`BucketsDataManager`) is notified to purge data before start time.

Will club (a) and (c) in one block

library/src/main/java/com/datatorrent/lib/state/managed/Bucket.java
```diff
@@ -0,0 +1,505 @@
...
+        sizeInBytes.getAndAdd(Long.SIZE);
+      }
+      if (timeBucket > bucketedValue.getTimeBucket()) {
+
+        if (bucketedValue.getValue() != null) {
+          //overriding the value so need to subtract the size of previous value
+          sizeInBytes.getAndAdd(-bucketedValue.getValue().length);
+        }
+
+        sizeInBytes.getAndAdd(value.length);
```
otterc: From @amberarrow 
Can avoid 2 calls to getAndAdd() with something like:

``` java
int inc = null == bucketedValue.getValue() ? value.length : value.length - bucketedValue.getValue().length;
sizeInBytes.getAndAdd(inc);
```

otterc: Ok

library/src/main/java/com/datatorrent/lib/state/managed/StateTracker.java
```diff
@@ -0,0 +1,153 @@
...
+
+import com.google.common.base.Preconditions;
+
+import com.datatorrent.api.Component;
+import com.datatorrent.api.Context;
+
+/**
+ * Tracks the size of state in memory and evicts buckets.
+ */
+class StateTracker extends TimerTask implements Component<Context.OperatorContext>
```
otterc: From @amberarrow 
Why does this class need to implement Component ?

otterc: Most of the classes that I have added implement Component. The reason for making them implement Component is that the `setup(...)` and `teardown(...)` also becomes part of their definition.

library/src/main/java/com/datatorrent/lib/state/managed/StateTracker.java
```diff
@@ -0,0 +1,153 @@
...
+{
+  //bucket id -> last time the bucket was accessed
+  private final transient ConcurrentHashMap<Long, Long> bucketAccessTimes = new ConcurrentHashMap<>();
+
+  private transient PriorityQueue<Map.Entry<Long, Long>> bucketHeap;
```
otterc: From @amberarrow 
One last suggestion to think about: ConcurrentSkipListSet may work better and also avoid re-populating and clearing the heap each time since it is thread-safe and maintains sorted order. Feel free to stay with the current version if you prefer it.

The synchronization cost is sometimes unavoidable when multiple threads are involved. For example, TimeBucketAssigner.getTimeBucketFor() is called by all the put, getSync and getAsync methods and yet has 2 synchronized blocks.

otterc:
@amberarrow addressed the review comments so far.

amberarrow:
Ok, I'll take a look in a day or two.

library/src/main/java/com/datatorrent/lib/state/BucketedState.java
```diff
@@ -0,0 +1,101 @@
...
+{
+  /**
+   * Sets the value of the key in bucket identified by bucketId.
+   *
+   * @param bucketId identifier of the bucket.
+   * @param key     key
+   * @param value   value
+   */
+  void put(long bucketId, Slice key, Slice value);
```
pramodin: Is bucketId an artifact of the implementation? The key is fine as it has a manifestation for the user within the application. This way of specification puts an implicit 2 level hierarchy. Questions arise why not 3 levels?

otterc: @PramodSSImmaneni  
This is a comment/question about design and large state management design document was shared couple of months back with dev@apex. 

IMO the pull request is for code review. 

The bucketId is for partitioning. The user of managed state knows best how to partition the data.

tweise:
@chandnisingh please update the pull request title so info gets linked to ASF JIRA:  https://issues.apache.org/jira/browse/APEXMALHAR-1897

pramodin:
As discussed do check that state is consistent for reads when WAL is materialized. Also please check that when purge happens it will still lead to idempotent results in fault conditions.

otterc:
Please look at the revised API. I have added more documentation so hopefully this makes it easier. 
I will update the document that Ram, Chaitanya and I are working on with these changes.

otterc:
Except adding  a version number to meta file, all the other changes have been made. Please take a look.

Would like to squash the commit but the comments here then may disappear.

tweise:
You can go ahead and squash, comments should still be associated with outdated diff.

tweise:
Just merged in head of master, please rebase.

library/src/main/java/com/datatorrent/lib/state/BucketedState.java
```diff
@@ -0,0 +1,60 @@
...
+   *
+   * @param bucketId identifier of the bucket.
+   * @param key     key
+   * @return        value
+   */
+  Slice getSync(long bucketId, Slice key);
+
+  /**
+   * Returns the future using which the value is obtained.<br/>
+   * If the key is present in memory, then the future has its value set when constructed.
```
amberarrow: Suggest omitting "in memory" or replacing with "in the bucket" (multiple places)

library/src/main/java/com/datatorrent/lib/state/managed/TimeBucketAssigner.java
```diff
@@ -0,0 +1,236 @@
...
+ * Keeps track of time buckets.<br/>
+ *
+ * The data of a bucket is further divided into time-buckets. This component controls the length of time buckets,
+ * which time-bucket an event falls into and sliding the time boundaries.
+ * <p/>
+ *
+ * The configuration {@link #expireBefore} and {@link #bucketSpan}  are used to calculate number of time-buckets.
+ * For eg. if <code>expireBefore = 1 hour</code> and <code>bucketSpan = 30 minutes</code>, then <code>
+ *   numBuckets = 60 minutes/ 30 minutes = 2 </code>.
+ * <p/>
```
amberarrow: Are the values of `expireBefore` and `bucketSpan` intended to be fixed after initialization or are they modifiable ?

library/src/main/java/com/datatorrent/lib/state/managed/TimeBucketAssigner.java
```diff
@@ -0,0 +1,236 @@
...
+   * Sets the length of a time bucket.
+   * @param bucketSpan length of time bucket
+   */
+  public void setBucketSpan(Duration bucketSpan)
+  {
+    this.bucketSpan = bucketSpan;
+  }
+
+  /**
+   * The listeners are informed when the time slides and time buckets which are older than the smallest time bucket
```
amberarrow: listeners are => listener is

library/src/main/java/com/datatorrent/lib/state/managed/TimeBucketAssigner.java
otterc: These are configurable properties so they need to be just set once. Shouldn't be change after application is deployed

amberarrow: It's worth adding a comment to that effect; otherwise, users may try to do something clever like increasing the span when data is sparse and decreasing it when data is dense.

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeUnifiedStateImpl.java
```diff
@@ -0,0 +1,238 @@
...
+  {
+    purgedTimeBuckets.add(timeBucket);
+    super.purgeTimeBucketsLessThanEqualTo(timeBucket);
+  }
+
+  /**
+   * This uses operator id instead of bucket id as the name of parent folder of time-buckets. This is because
+   * multiple partitions may work on same time-buckets.
+   */
+  public static class TimeUnifiedBucketsFileSystem extends BucketsFileSystem
```
amberarrow: Can this be private ?

library/src/main/java/com/datatorrent/lib/state/BucketedState.java
```diff
@@ -29,11 +31,17 @@
 public interface BucketedState
 {
   /**
+   * An expired value. In some implementations where bucketId is time related then the event can be old and
+   * the get methods- getSync & getAsync return this special slice instance.
+   */
+  Slice EXPIRED = new Slice(Ints.toByteArray(-1));
```
amberarrow: Couldn't -1 be a valid user value ?
Might be better to use invalid slices, e.g. Slice(null, -1, -1). Or, if you want to be more user friendly, extend Slice and add an isExpired() method which checks offset and length against -1.

otterc: I can use Slice(null, -1, -1). In the usages, comparisons should be made using `==` and not `equals`.

For every expired event I didn't want to instantiate a Slice object. So even if I extend Slice, I will re-use a special instance of slice so thought why not use just a special instance.

There is a problem with all the approaches I can think of which is that Slice cannot have an immutable extension because all the members are publicly exposed without getters and setters.

amberarrow: Using a fixed instance is fine. Worth documenting the `==` part.

ilooner:
@chandnisingh Taking a look out of curiosity, please ignore my comments if you feel they are not important.

library/src/main/java/com/datatorrent/lib/state/TimeSlicedBucketedState.java
```diff
@@ -0,0 +1,98 @@
...
+   * @param time    time associated with the key.
+   * @param key     key
+   * @param value   value
+   */
+  void put(long bucketId, long time, Slice key, Slice value);
+
+  /**
+   * Returns the value of the key in the bucket identified by bucketId. This will search for the key in all the
+   * time buckets.</br>
+   *
```
ilooner: If a key is present in multiple time buckets, which value is returned?

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeUnifiedStateImpl.java
otterc: I think it can be private. I don't foresee any need of extending this.

library/src/main/java/com/datatorrent/lib/state/TimeSlicedBucketedState.java
otterc: The latest value will be returned. Will update the java docs.

library/src/main/java/com/datatorrent/lib/state/TimeSlicedBucketedState.java
```diff
@@ -0,0 +1,98 @@
...
+   * @param bucketId identifier of the bucket
+   * @param key key
+   * @return value of the key if found; null if the key is not found;
+   */
+  Slice getSync(long bucketId, Slice key);
+
+
+  /**
+   * Returns the value of key in the bucket identified by bucketId. This expects the time value.
+   * Time is be used to derive the time bucket (within the main bucket) to which the key belonged and only that
```
ilooner: I think there is a type. "Time is be used to derive the time bucket" should be "Time is used to derive the time bucket"

library/src/main/java/com/datatorrent/lib/state/managed/Bucket.java
```diff
@@ -0,0 +1,534 @@
...
+    private Map<Slice, BucketedValue> flash = Maps.newHashMap();
+
+    //Data persisted in write ahead logs. window -> bucket
+    private final transient TreeMap<Long, Map<Slice, BucketedValue>> checkpointedData = Maps.newTreeMap();
+
+    //Data persisted in bucket data files
+    private final transient Map<Slice, BucketedValue> committedData = Maps.newHashMap();
+
+    //Data recovered
+    private final transient TreeMap<Long, Map<Slice, BucketedValue>> recoveredData = Maps.newTreeMap();
```
ilooner: can recovered data be treated the same as checkpointed data?

otterc: Yes. I think it can be. I will make the change.

library/src/main/java/com/datatorrent/lib/state/managed/Bucket.java
```diff
@@ -0,0 +1,534 @@
...
+      return null;
+    }
+
+    private Slice getFromReaders(Slice key, long timeBucket)
+    {
+      if (timeBucket != -1) {
+        Slice valSlice = getKeyFromTimeBucketReader(key, timeBucket);
+        if (valSlice != null) {
+          BucketedValue bucketedValue = new BucketedValue(timeBucket, valSlice);
+          readCache.put(key, bucketedValue);
```
ilooner: If the user does a get for key "A" time bucket 1 first. Then that value will be put in the read cache. If the user then does a get for key "A" time bucket 2, then the time bucket 1 value is retrieved from the read cache and returned. So the value for an older timebucket is returned.

otterc: Will fix this. However the expectation here is that the latest value is retrieved always. If the latest time-bucket is known then the process to read it is faster otherwise reading all time-buckets is slower. This is again missing from the documentation  and I will add to it.

library/src/main/java/com/datatorrent/lib/state/BucketedState.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+package com.datatorrent.lib.state;
+
+import java.util.concurrent.Future;
+
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * A state where keys are grouped in buckets.
```
amberarrow: Add: A key must not be null.

library/src/main/java/com/datatorrent/lib/state/BucketedState.java
```diff
@@ -0,0 +1,69 @@
...
+   */
+  Slice EXPIRED = new Slice(null, -1, -1);
+
+  /**
+   * Sets the value of the key in bucket identified by bucketId.
+   *
+   * @param bucketId identifier of the bucket.
+   * @param key      key
+   * @param value    value
+   */
```
amberarrow: value => value (must not be null)

library/src/main/java/com/datatorrent/lib/state/TimeSlicedBucketedState.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+package com.datatorrent.lib.state;
+
+import java.util.concurrent.Future;
+
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * A type of bucketed state where a bucket's data is further divided into time buckets. This requires
+ * time per key to figure out which time bucket a particular key belongs to.
```
amberarrow: Add: The key must not be null.

library/src/main/java/com/datatorrent/lib/state/TimeSlicedBucketedState.java
```diff
@@ -0,0 +1,98 @@
...
+public interface TimeSlicedBucketedState
+{
+  /**
+   * Sets the value of a key in the bucket identified by bucketId. Time is used to derive which time bucket (within
+   * the main bucket) a key belongs to.
+   *
+   * @param bucketId identifier of the bucket.
+   * @param time    time associated with the key.
+   * @param key     key
+   * @param value   value
```
amberarrow: value => value (must not be null)

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeUnifiedStateImpl.java
```diff
@@ -0,0 +1,238 @@
...
+import com.datatorrent.lib.state.BucketedState;
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * In this implementation of {@link ManagedState} the buckets in memory are time-buckets.
+ * <p/>
+ *
+ * <b>Difference from {@link ManagedTimeStateImpl}</b>: <br/>
+ * <ol>
+ * <li>The main buckets in {@link ManagedTimeStateImpl} are unique adhoc long ids which the user provides with the
```
amberarrow: `ManagedTimeStateImpl` has a time parameter so seems like this should be `ManagedStateImpl`.

This documentation would be easier to read with a 3-column table -- one column for each of the implementation classes.

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeUnifiedStateImpl.java
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * <li>In {@link ManagedTimeStateImpl} a bucket belongs to just one partition. Multiple partitions cannot write to
+ * the same bucket. <br/>
+ * In this implementation multiple partitions can be working with the same time-bucket (since time-bucket is derived
+ * from time). This works because on the disk the time-bucket data is segregated under each operator id.
+ * </li>
+ * <br/>
+ *
+ * <li>While {@link ManagedTimeStateImpl} can support dynamic partitioning by pre-allocating buckets this will not
+ * be able to support dynamic partitioning efficiently.
```
amberarrow: These contrasting statements should also include the third implementation class.

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeUnifiedStateImpl.java
```diff
@@ -0,0 +1,238 @@
...
+  {
+    return timeBucketAssigner.getNumBuckets();
+  }
+
+  @Override
+  public void put(long time, Slice key, Slice value)
+  {
+    long timeBucket = timeBucketAssigner.getTimeBucketFor(time);
+    if (timeBucket == -1) {
+      //time is expired so return null.
```
amberarrow: return null => ignore key/value

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeUnifiedStateImpl.java
```diff
@@ -0,0 +1,238 @@
...
+    buckets[bucketIdx].put(key, timeBucket, value);
+
+  }
+
+  @Override
+  public Slice getSync(long time, Slice key)
+  {
+    long timeBucket = timeBucketAssigner.getTimeBucketFor(time);
+    if (timeBucket == -1) {
+      //time is expired so return null.
```
amberarrow: return null. => return expired token.

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeUnifiedStateImpl.java
```diff
@@ -0,0 +1,238 @@
...
+    return buckets[bucketIdx].get(key, timeBucket, Bucket.ReadSource.ALL);
+  }
+
+  @SuppressWarnings("SynchronizationOnLocalVariableOrMethodParameter")
+  @Override
+  public Future<Slice> getAsync(long time, Slice key)
+  {
+    long timeBucket = timeBucketAssigner.getTimeBucketFor(time);
+    if (timeBucket == -1) {
+      //time is expired so return null.
```
amberarrow: return null. => return expired token.

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeUnifiedStateImpl.java
```diff
@@ -0,0 +1,238 @@
...
+      return Futures.immediateFuture(BucketedState.EXPIRED);
+    }
+    int bucketIdx = prepareBucket(timeBucket);
+    Bucket bucket = buckets[bucketIdx];
+    synchronized (bucket) {
+      Slice cachedVal = buckets[bucketIdx].get(key, timeBucket, Bucket.ReadSource.MEMORY);
```
amberarrow: buckets[bucketIdx] => bucket

library/src/main/java/com/datatorrent/lib/state/managed/ManagedStateImpl.java
```diff
@@ -0,0 +1,93 @@
...
+  @SuppressWarnings("SynchronizationOnLocalVariableOrMethodParameter")
+  @Override
+  public Future<Slice> getAsync(long bucketId, Slice key)
+  {
+    int bucketIdx = prepareBucket(bucketId);
+    Bucket bucket = buckets[bucketIdx];
+
+    synchronized (bucket) {
+
+      Slice cachedVal = buckets[bucketIdx].get(key, -1, Bucket.ReadSource.MEMORY);
```
amberarrow: buckets[bucketIdx] => bucket

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeStateImpl.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  @SuppressWarnings("SynchronizationOnLocalVariableOrMethodParameter")
+  @Override
+  public Future<Slice> getAsync(long bucketId, Slice key)
+  {
+    int bucketIdx = prepareBucket(bucketId);
+    Bucket bucket = buckets[bucketIdx];
+    synchronized (bucket) {
+      Slice cachedVal = buckets[bucketIdx].get(key, -1, Bucket.ReadSource.MEMORY);
```
amberarrow: buckets[bucketIdx] => bucket

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeStateImpl.java
```diff
@@ -0,0 +1,125 @@
...
+  {
+    int bucketIdx = prepareBucket(bucketId);
+    Bucket bucket = buckets[bucketIdx];
+    long timeBucket = timeBucketAssigner.getTimeBucketFor(time);
+    if (timeBucket == -1) {
+      //time is expired so no point in looking further.
+      return Futures.immediateFuture(BucketedState.EXPIRED);
+    }
+    synchronized (bucket) {
+      Slice cachedVal = buckets[bucketIdx].get(key, timeBucket, Bucket.ReadSource.MEMORY);
```
amberarrow: buckets[bucketIdx] => bucket

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeStateImpl.java
```diff
@@ -0,0 +1,125 @@
...
+      if (cachedVal != null) {
+        return Futures.immediateFuture(cachedVal);
+      }
+      return readerService.submit(new KeyFetchTask(bucket, key, timeBucket, throwable));
+    }
```
amberarrow: This synchronized block and the assignment to `bucket` are duplicated in multiple places (3 or 4); could it be factored out to a helper method in the base class ?

library/src/main/java/com/datatorrent/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -0,0 +1,499 @@
...
+    checkpointManager.setup(this);
+    bucketsFileSystem.setup(this);
+
+    if (buckets == null) {
+      //create buckets array only once at start when it is not created.
+      numBuckets = getNumBuckets();
+      buckets = new Bucket[numBuckets];
+    }
+    for (Bucket bucket : buckets) {
+      if (bucket != null) {
```
amberarrow: Is it possible for bucket to be null here ?

library/src/main/java/com/datatorrent/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -0,0 +1,499 @@
...
+ *   <li>
+ *     {@link #stateTracker}: tracks the size of data in memory and requests buckets to free memory when enough memory
+ *     is not available.
+ *   </li>
+ *   <li>
+ *     {@link #fileAccess}: plug-able file system abstraction.
+ *   </li>
+ * </ol>
+ *
+ * The implementations of put, getSync and getAsync here use windowId as the time field to derive timeBucket of a key.
```
amberarrow: This comment seems to belong in ManagedStateImpl

library/src/main/java/com/datatorrent/lib/state/managed/BucketsFileSystem.java
```diff
@@ -0,0 +1,573 @@
...
+    }
+
+  }
+
+  /**
+   * This serves the readers - {@link com.datatorrent.lib.state.managed.Bucket.DefaultBucket}.
+   * It is accessible outside the package unlike {@link MutableTimeBucketMeta} that can be only modified by
+   * BucketPersistor.
+   */
+  public static final class ImmutableTimeBucketMeta extends TimeBucketMeta
```
amberarrow: Do we need this class ?
`TimeBucketMeta` itself seems immutable since it has no setters and private constructors, so could we not just use that instead ?

library/src/main/java/com/datatorrent/lib/state/managed/BucketsFileSystem.java
```diff
@@ -0,0 +1,573 @@
...
+      }
+      return immutableTimeBucketMeta;
+    }
+
+  }
+
+  /**
+   * This serves the readers - {@link com.datatorrent.lib.state.managed.Bucket.DefaultBucket}.
+   * It is accessible outside the package unlike {@link MutableTimeBucketMeta} that can be only modified by
+   * BucketPersistor.
```
amberarrow: BucketPersistor ?

library/src/main/java/com/datatorrent/lib/state/managed/BucketsFileSystem.java
```diff
@@ -0,0 +1,573 @@
...
+   * @param bucketId     bucket id
+   * @param timeBucketId time bucket id
+   * @return immutable time bucket meta
+   * @throws IOException
+   */
+  @Nullable
+  public ImmutableTimeBucketMeta getTimeBucketMeta(long bucketId, long timeBucketId) throws IOException
+  {
+    synchronized (timeBucketsMeta) {
+      MutableTimeBucketMeta tbm = timeBucketMetaHelper(bucketId, timeBucketId);
```
amberarrow: synchronized block could end here.

library/src/main/java/com/datatorrent/lib/state/managed/BucketsFileSystem.java
```diff
@@ -0,0 +1,573 @@
...
+   *
+   * @param bucketId bucket id
+   * @throws IOException
+   */
+  void updateBucketMetaFile(long bucketId) throws IOException
+  {
+    Map<Long, MutableTimeBucketMeta> timeBuckets;
+    synchronized (timeBucketsMeta) {
+      timeBuckets = timeBucketsMeta.row(bucketId);
+    }
```
amberarrow: synchronized block may have to cover most of this function because the map returned by the row() call is not a separate copy, so if another thread modifies the row while the loop below is iterating, there could be problems. The docs for the row() call say: "Changes to the returned map will update the underlying table, and vice versa."

library/src/main/java/com/datatorrent/lib/state/managed/ManagedStateImpl.java
```diff
@@ -0,0 +1,93 @@
...
+      int bucketIdx = prepareBucket(bucketId);
+      buckets[bucketIdx].put(key, timeBucket, value);
+    }
+  }
+
+
+  @Override
+  public Slice getSync(long bucketId, Slice key)
+  {
+    int bucketIdx = prepareBucket(bucketId);
```
ilooner: This method should also acquire the bucket lock incase the user mixes Synchronous and Asynchronous get calls.

library/src/main/java/com/datatorrent/lib/state/managed/Bucket.java
```diff
@@ -0,0 +1,534 @@
...
+      return false;
+    }
+
+    @Override
+    public void put(Slice key, long timeBucket, Slice value)
+    {
+      BucketedValue bucketedValue = flash.get(key);
+      if (bucketedValue == null) {
+        bucketedValue = new BucketedValue();
+        flash.put(key, bucketedValue);
```
ilooner: A put operation can modify flash in one thread while an asynchronous get operation can do a get from flash in another thread. Could this cause a problem?

otterc: Should be a concurrent hash map

library/src/main/java/com/datatorrent/lib/state/managed/BucketsFileSystem.java
otterc: Checking if this could be removed.

library/src/main/java/com/datatorrent/lib/state/managed/BucketsFileSystem.java
otterc: BucketFileSystem. Left over from refactoring.

library/src/main/java/com/datatorrent/lib/state/managed/BucketsFileSystem.java
otterc: Modification to a particular row is just being done by one thread which is the thread that transfers data from a window file to data files. I will double check and add a comment

otterc: There is another thread reading from the row so I think I need to synchronize on the MutableTimeBucket in the row.

otterc: Will make the suggested change, i.e. synchronize block cover most of the function

ilooner:
Can purging be disabled? Some of the customer use cases I've encountered require the data to be kept indefinitely

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeUnifiedStateImpl.java
```diff
@@ -0,0 +1,243 @@
...
+  public void endWindow()
+  {
+    super.endWindow();
+    Long purgedTimeBucket;
+
+    //tear down all the purged time buckets
+    while (null != (purgedTimeBucket = purgedTimeBuckets.poll())) {
+      int purgedTimeBucketIdx = getBucketIdx(purgedTimeBucket);
+      if (buckets[purgedTimeBucketIdx] != null && buckets[purgedTimeBucketIdx].getBucketId() == purgedTimeBucket) {
+        buckets[purgedTimeBucketIdx].teardown();
```
ilooner: If there are asynchronous get calls happening when a bucket is purged they may attempt to retrieve data from a closed reader. I would suggest keeping track of the asynchronous get calls and wait for them to finish before purging a time bucket.

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeUnifiedStateImpl.java
```diff
@@ -0,0 +1,243 @@
...
+    purgedTimeBuckets.add(timeBucket);
+    super.purgeTimeBucketsLessThanEqualTo(timeBucket);
+  }
+
+  /**
+   * This uses operator id instead of bucket id as the name of parent folder of time-buckets. This is because
+   * multiple partitions may work on same time-buckets.
+   */
+  public static class TimeUnifiedBucketsFileSystem extends BucketsFileSystem
+  {
```
ilooner: Missing Override annotations on methods for this class.

library/src/main/java/com/datatorrent/lib/state/managed/BucketsFileSystem.java
```diff
@@ -0,0 +1,580 @@
...
+        if (timeBucketStr.equals(BucketsFileSystem.META_FILE_NAME) || timeBucketStr.endsWith(".tmp")) {
+          //ignoring meta and tmp files
+          continue;
+        }
+        long timeBucket = Long.parseLong(timeBucketStr);
+
+        if (timeBucket <= latestExpiredTimeBucket) {
+          LOG.debug("deleting bucket {} time-bucket {}", timeBucket);
+          delete(bucketName, timeBucketStatus.getPath().getName());
+
```
ilooner: The time bucket data file is deleted then the meta file is updated. If the operator fails after deleting the time bucket but before the meta file is updated, the operator could load the meta data for a deleted time bucket on recovery. Could that cause a problem? Could the update of the Meta file be done before the delete of the time bucket?

library/src/main/java/com/datatorrent/lib/state/managed/ManagedStateImpl.java
```diff
@@ -0,0 +1,98 @@
...
+
+  public ManagedStateImpl()
+  {
+    this.numBuckets = 1;
+  }
+
+  @Override
+  public void put(long bucketId, Slice key, Slice value)
+  {
+    long timeBucket = timeBucketAssigner.getTimeBucketFor(windowId);
```
amberarrow: `getTimeBucketFor` expects its argument to be a timestamp somewhere near the current time; the `windowId`, at least when run in local mode, seems to be a much larger value, e.g.
windowId = 6264531193618759682, current = 1458574832005, diff = 6264529735043927677

So, how is this expected to work ? Do we have any platform guarantees about the windowId ? If the windowId happens be less than `start` would we not always get a -1 back ?

otterc: window id is derived from time and its increasing with time. TimeBucketAssigner reference instant can be configured. If TimeBucketAssigner is configured correctly windowId will not be less than start.

otterc: I think I need to make have better default configuration for TimeBucketAssigner in this case

otterc: @ilooner suggested to keep initial time and add millis corresponding to app window to it. So made that change.

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeStateImpl.java
otterc: getSync and getAsync both return different types of object - Slice and Future<Slice>.
I can create 2 helper methods for each and reduce the code a little.

amberarrow: Yes, I think it's worth doing for 2 reasons: (a) Reduces code duplication thus helping with maintainability and (b) Helps readability since the reader wouldn't have to visually compare code fragments to confirm that all locations are doing the same thing.

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeStateImpl.java
```diff
@@ -51,63 +51,57 @@ public void put(long bucketId, long time, Slice key, Slice value)
...
     long timeBucket = timeBucketAssigner.getTimeBucketFor(time);
     if (timeBucket == -1) {
       //time is expired so no point in looking further.
       return Futures.immediateFuture(BucketedState.EXPIRED);
     }
+    return getValueFromBucketAsync(bucketId, timeBucket, key);
+  }
+
+  @SuppressWarnings("SynchronizationOnLocalVariableOrMethodParameter")
+  private Future<Slice> getValueFromBucketAsync(long bucketId, long timeBucket, Slice key)
```
amberarrow: If these new functions getValueFromBucketAsync() and getValueFromBucketSync() are moved to the base class and made static, could they also be used in the other 2 implementation classes ?

otterc: I moved it to the AbstractManagedStateImpl however can't make it static because they need to access some non-static members and methods - `buckets` array and `prepareBucket()`

otterc:
@ilooner currently purging cannot be disabled. However that support can be added later. I can create a ticket for that as a pending item

library/src/main/java/com/datatorrent/lib/state/managed/AbstractManagedStateImpl.java
otterc: Yes it is possible. Let's say user specified numBuckets = 4. This will create `buckets` array of size 4. However if we don't see a key for bucket 3 then `buckets[2]` will never be initialized to a a bucket. We are following a lazy initialization strategy.

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeUnifiedStateImpl.java
otterc: How about I create a markdown file in this package with a tabular structure highlighting the differences?

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeUnifiedStateImpl.java
otterc: How about I create a markdown file in this package with a tabular structure highlighting the differences?

library/src/main/java/com/datatorrent/lib/state/managed/ManagedTimeUnifiedStateImpl.java
amberarrow: I think having the table in javadoc is better since it will be visible on our API reference page.
Here is one example (though not a real fancy table):
https://docs.oracle.com/javase/7/docs/jre/api/net/httpserver/spec/com/sun/net/httpserver/HttpServer.html

You can put it in one implementation class and just have links in the other 2.

library/src/main/java/com/datatorrent/lib/state/BucketedState.java
otterc: A bucket includes all the  key/values -  both in-memory and on-disk belonging to that bucket.  So if I say key is present in the bucket that could mean that the key could be in bucket cache or in the data files belonging to the bucket.

amberarrow: Oh, I misunderstood the comment.

If the intent is to say "If the key is present in memory, then the future has its value set when constructed; if not the value is set some time after construction"
(where I've added the part after the semicolon), then isn't this an implementation detail that is better placed in the implementation file ?

otterc: Yeah should be placed in the implementation file. Will move it there.

library/src/main/java/com/datatorrent/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -226,6 +232,15 @@ protected int prepareBucket(long bucketId)
     return bucketIdx;
   }
 
+  protected void putInBucket(long bucketId, long timeBucket, Slice key, Slice value)
+  {
+    if (timeBucket != -1) {
+      //time bucket is invalid data is not stored
```
amberarrow: Suggest moving comment above the if and rephrasing:
// if time bucket is invalid, data is not stored

library/src/main/java/com/datatorrent/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -0,0 +1,499 @@
...
+
+    if (activationWindow != Stateless.WINDOW_ID) {
+      //delete all the wal files with windows > activationWindow.
+      //All the wal files with windows <= activationWindow are loaded and kept separately as recovered data.
+      try {
+        for (long recoveredWindow : checkpointManager.getWindowIds(operatorContext.getId())) {
+          if (recoveredWindow <= activationWindow) {
+            @SuppressWarnings("unchecked")
+            Map<Long, Map<Slice, Bucket.BucketedValue>> recoveredData =
+                (Map<Long, Map<Slice, Bucket.BucketedValue>>)checkpointManager.load(operatorContext.getId(), windowId);
```
amberarrow: windowId => recoveredWindow ?
(seems like windowId is never used, other than here, and can be removed ?)

tweise:
@chandnisingh please let me know after package rename and rebase/squash are done.

otterc:
@tweise I have renamed package and squash

PR #147, opened 2015-12-28
MLHR-1953 - Added JdbcPOJOGenericUpdateOperator. Added Unit tests.

bhupeshchawda:


otterc:
Why is this called Generic? Are there any other update operators which are not Generic?

bhupeshchawda:
The "generic" keyword was actually because it is not specific to the query. It can do insert as well as update, delete, and merge.

otterc:
There is already a JdbcPojoOutputOperator so why add a new one 
Why update and  merge support was not added to JdbcPojoOutputOperator?

Adding extensions likes this decreases the usability. It was pointed out on Apex dev list that update and merge support should be added to JdbcPOJOOutputOperator but I don't understand why that approach was not followed.

bhupeshchawda:
Hi Chandni,
The approach followed by JDBCPOJOOutputOperator is to formulate the insert query in setup() using the list of fieldInfo taken from the user. This kind of approach would not be suitable for update / merge statements. In other words, it would be very difficult to formulate the update / merge statement in the operator logic, especially in case the where clause expressions are even moderately complex.

The reason for not adding the update / merge support in JdbcPojoOutputOperator was that including the update/merge support in the same class would not have been clean. A new type of input format is needed as explained in the generic implementation. A by product of the new implementation is that it is applicable to inserts as well. So, if we had added the update support to the existing class, then we would have two approaches of handling inserts within the same class which does not seem to be a good implementation approach.

otterc:
@bhupeshchawda This should have been discussed on dev list when it was recommended to modify JdbcPojoOutputOperator.  It is marked as @Evovling and so backward compatibility is not a concern. 

You can reply to same email thread by asking suggestions of how to cleanly add update/merge support to the existing operator, how FieldInfo is limiting, etc.

IMO the approach of adding another operator instead of augmenting an existing one is a bad approach. 
Also having "generic" in the name trivializes all the operators hinting that they aren't.

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOOutputOperator.java
```diff
@@ -69,18 +69,18 @@
     implements Operator.ActivationListener<OperatorContext>
 {
   @NotNull
-  private List<FieldInfo> fieldInfos;
+  protected List<FieldInfo> fieldInfos;
```
DT-Priyanka, line 72: As this has getters and setters we need not change the access level.

otterc:
-1 for this change as explained

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOOutputOperator.java
```diff
@@ -69,18 +69,18 @@
...
 
-  private List<Integer> columnDataTypes;
+  protected List<Integer> columnDataTypes;
 
   @NotNull
   private String tablename;
 
-  private final transient List<JdbcPOJOInputOperator.ActiveFieldInfo> columnFieldGetters;
+  protected final transient List<JdbcPOJOInputOperator.ActiveFieldInfo> columnFieldGetters;
```
DT-Priyanka, line 79: why have you changed this access level?

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOOutputOperator.java
```diff
@@ -69,18 +69,18 @@
...
 
   private String insertStatement;
 
-  private transient Class<?> pojoClass;
+  protected transient Class<?> pojoClass;
```
DT-Priyanka, line 83: Same, is there need to change this access level?

bhupeshchawda:
Closing this pull request. Will open another one with the updated design as discussed on dev

PR #148, opened 2015-12-28
MLHR-1956 Added POJO Kafka Output operator with auto-metrics and batchprocessing

chaithu14:
@siyuanh : Please review.

### DIFF
--- a/contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
@@ -1,0 +1,261 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.contrib.kafka;
+
+import java.lang.reflect.Field;
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+import org.apache.commons.lang3.ClassUtils;
+import com.datatorrent.api.AutoMetric;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.lib.util.PojoUtils;
+
+import kafka.producer.KeyedMessage;
+import kafka.producer.ProducerConfig;
+
+/**
+ * POJOKafkaOutputOperator extends from AbstractKafkaOutputOperator receives the POJO
+ * from upstream and converts to Kafka Messages and writes to kafka topic.&nbsp;
+ * <p>
+ * <br>
+ * Ports:<br>
+ * <b>Input</b>: Have only one input port<br>
+ * <b>Output</b>: No Output Port <br>
+ * <br>
+ * Properties:<br>
+ * <b>brokerList</b>: List of brokers in the form of Host1:Port1,Host2:Port2,Host3:port3,...<br>
+ * <b>keyField</b>: Specifies the field creates distribution of tuples to kafka partition. <br>
+ * <b>isBatchProcessing</b>: Specifies whether to write messages in batch or not. By default,
+ *                           the value is true <br>
+ * <b>batchSize</b>: Specifies the batch size.<br>
+ * <br>
+ * <br>
+ * </p>
+ *
+ * @displayName POJO Kafka Output
+ * @category Messaging
+ * @tags Output operator
+ *
+ */
+public class POJOKafkaOutputOperator extends AbstractKafkaOutputOperator<Object,Object>
+{
+  @AutoMetric
+  private long outputMessagesPerSec;
+  @AutoMetric
+  private long outputBytesPerSec;
+  protected final String BROKER_KEY = "metadata.broker.list";
+  protected final String BATCH_NUM_KEY = "batch.num.messages";
+  protected final String PRODUCER_KEY = "producer.type";
+  protected final String QUEUE_BUFFER_KEY = "queue.buffering.max.ms";
+  protected final String ASYNC_PRODUCER_TYPE = "async";
+  private long messageCount;
+  private long byteCount;
+  private String brokerList;
+  private double windowTimeSec;
+  private String keyField = "";
+  protected boolean isBatchProcessing = true;
+  @Min(2)
+  protected int batchSize;
+  protected transient PojoUtils.Getter keyMethod;
+  protected transient Class<?> pojoClass;
+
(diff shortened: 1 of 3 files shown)


contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
```diff
@@ -0,0 +1,272 @@
...
+  /**
+   * setup producer configuration.
+   * @return ProducerConfig
+   */
+  @Override
+  protected ProducerConfig createKafkaProducerConfig(){
+    if(brokerList != null) {
+      getConfigProperties().setProperty(BROKER_KEY, brokerList);
+    }
+    getConfigProperties().setProperty(BATCH_NUM_KEY, String.valueOf(batchSize));
```
siyuanh: Enforce this only if it is in batch mode, right?  And also you should set producer.type as true to make the batch mode really work

contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
```diff
@@ -0,0 +1,272 @@
...
+    // Convert the given tuple to KeyedMessage
+    KeyedMessage msg;
+    if(keyMethod != null) {
+      msg = new KeyedMessage(getTopic(), keyMethod.get(tuple), tuple);
+    } else {
+      msg = new KeyedMessage(getTopic(), tuple, tuple);
+    }
+
+    if (isBatchProcessing) {
+      if (messageList.size() == batchSize) {
```
siyuanh: I think you don't have to do batch manually. you can set the  batch.num.messages to batch size and also set the queue.buffering.max.ms to streaming window size. It's not strictly same as what you are doing here. But the logic and performance are similar

siyuanh: Unless you want to clear the buffer exactly at the endwindow

contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
```diff
@@ -0,0 +1,272 @@
...
+ * @category Messaging
+ * @tags Output operator
+ *
+ */
+public class POJOKafkaOutputOperator extends AbstractKafkaOutputOperator<Object,Object>
+{
+  @AutoMetric
+  private long outputMessagesPerSec;
+  @AutoMetric
+  private long outputBytesPerSec;
```
siyuanh: If I remember correctly AutoMetric only works when you have setter/getter method for this. @chandnisingh  am I correct?

siyuanh: Sorry, I was wrong, ignore this

contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
chaithu14: Yes. If it is a batch mode, added the producer.type as async to config properties.

contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
chaithu14: Not changing this because want to send the remaining batch messages to Kafka at end window.

siyuanh: @chaithu14  What is the purpose of this batch producer? If you want to enforce the batch processing, you have to set the batch property anyways, and if you set it, you can't make sure the message will be send out at the end of window.   You can't guarantee both.

chaithu14: Purpose is to process batch of messages instead of message by message. But, producer is already taking care this. Will update it as per the previous review comment.

chaithu14: Incorporated the changes.

contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
```diff
@@ -0,0 +1,266 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.contrib.kafka;
+
+import java.lang.reflect.Field;
+import java.util.ArrayList;
```
siyuanh: Remove unused imports

contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
```diff
@@ -0,0 +1,266 @@
...
+    }
+  };
+
+  /**
+   * setup producer configuration.
+   * @return ProducerConfig
+   */
+  @Override
+  protected ProducerConfig createKafkaProducerConfig(){
+    if(brokerList != null) {
```
siyuanh: Space after "if". Please follow the checkstyle rules

contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
```diff
@@ -0,0 +1,266 @@
...
+      processTuple(tuple);
+    }
+  };
+
+  /**
+   * setup producer configuration.
+   * @return ProducerConfig
+   */
+  @Override
+  protected ProducerConfig createKafkaProducerConfig(){
```
siyuanh: { on new line

contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
```diff
@@ -0,0 +1,266 @@
...
+ * @tags Output operator
+ *
+ */
+public class POJOKafkaOutputOperator extends AbstractKafkaOutputOperator<Object,Object>
+{
+  @AutoMetric
+  private long outputMessagesPerSec;
+  @AutoMetric
+  private long outputBytesPerSec;
+  protected final Integer DEFAULT_BATCH_SIZE = 200;
```
siyuanh: We don't need extra static value here, right?

contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
```diff
@@ -0,0 +1,266 @@
...
+   * setup producer configuration.
+   * @return ProducerConfig
+   */
+  @Override
+  protected ProducerConfig createKafkaProducerConfig(){
+    if(brokerList != null) {
+      getConfigProperties().setProperty(BROKER_KEY, brokerList);
+    }
+    if(isBatchProcessing) {
+      getConfigProperties().setProperty(BATCH_NUM_KEY, String.valueOf(batchSize));
```
siyuanh: Also the producer type should be async

contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
chaithu14: Yes. DEFAULT_BATCH_SIZE static value is not needed.

contrib/src/main/java/com/datatorrent/contrib/kafka/POJOKafkaOutputOperator.java
chaithu14: Done

chaithu14:
Incorporated the review comments.

PR #149, opened 2015-12-29
APEXMALHAR-1958 Provide access to doneTuple field in AbstractReconciler for derived classes

yogidevendra:
1. Making queues as protected for access in derived classes.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/AbstractReconciler.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/AbstractReconciler.java
@@ -73,8 +73,8 @@ public abstract class AbstractReconciler<INPUT, QUEUETUPLE> extends BaseOperator
   // this stores the mapping from the window to the list of enqueued tuples
   private Map<Long, List<QUEUETUPLE>> currentWindowTuples = Maps.newConcurrentMap();
   private Queue<Long> currentWindows = Queues.newLinkedBlockingQueue();
-  private Queue<QUEUETUPLE> committedTuples = Queues.newLinkedBlockingQueue();
-  private transient Queue<QUEUETUPLE> doneTuples = Queues.newLinkedBlockingQueue();
+  protected Queue<QUEUETUPLE> committedTuples = Queues.newLinkedBlockingQueue();
+  protected transient Queue<QUEUETUPLE> doneTuples = Queues.newLinkedBlockingQueue();
   private transient Queue<QUEUETUPLE> waitingTuples = Queues.newLinkedBlockingQueue();
   private transient volatile boolean execute;
   private transient AtomicReference<Throwable> cause;
@@ -185,7 +185,6 @@ public abstract class AbstractReconciler<INPUT, QUEUETUPLE> extends BaseOperator
       }
     };
   }
-
   /**
    * The implementation class should call this method to enqueue output once input is converted to queue input.
    *


yogidevendra:
@tushargosavi Can you please review this?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractReconciler.java
```diff
@@ -73,9 +73,9 @@ public void process(INPUT input)
   // this stores the mapping from the window to the list of enqueued tuples
   private Map<Long, List<QUEUETUPLE>> currentWindowTuples = Maps.newConcurrentMap();
   private Queue<Long> currentWindows = Queues.newLinkedBlockingQueue();
-  private Queue<QUEUETUPLE> committedTuples = Queues.newLinkedBlockingQueue();
```
chinmaykolhatkar, line 76: Any reason why these are required to be protected?

I believe these fields are very important ones for correct functioning of Reconciler along with correct checkpointing.
By making these protected, we're exposing the overall functionality of of this class to get compromised.

In my opinion, we need a strong case for making this protected.

yogidevendra:
@chinmaykolhatkar 
Are you suggesting to use protected getter methods instead?

Reason why derived class need access to these fields:
In the endWindow() [of the derived class]  needs some custom post processing of doneTuples, committedTuples. 

How should we achieve this if we do not expose these fields or getters?

chinmaykolhatkar:
Yes.. Maybe getter method is a better idea. Also, getter method should return unmodifiable collection.

chinmaykolhatkar:
Please also add the javadoc for the getter method and how it should be used and other gotchas.
getter method for doneTuples should be called and processed before super.endWindow is called.

yogidevendra:
Collections.unmodifiableCollection in getters should solve the purpose.
I will incorporate these changes.

yogidevendra:
Regarding Javadoc:
For this specific usecase, derived class is overriding endWindow() method. 
But, it could potentially be any other method. Thus, explicit mention of super.endWindow() in the javadoc for getDoneTuples() will lead to confusion.

chinmaykolhatkar:
Agreed.
But the javadoc should specify the when this method is safe to be called.
When its unsafe to be called and what happens then.

~ Chinmay.

chinmaykolhatkar:
Looks fine to me. I can merge this in a day if there are no comments.

sandeepdeshmukh:
Run these changes through Gaurav as he is author of Reconciler code.
@gauravgopi123 : Please review.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractReconciler.java
```diff
@@ -187,6 +189,18 @@ public void run()
...
+   * Returns the unmodifiable collection of the tuples on which
+   * processCommittedData has been called in this window. Underlying collection
+   * is cleared at every endWindow() of AbstractReconciler.
+   * 
+   * @return done tuples
+   */
+  protected Collection<QUEUETUPLE> getDoneTuples()
+  {
+    return Collections.unmodifiableCollection(doneTuples);
+  }
```
gauravgopi123: Can you give details about the usecase why you need only read view and not write view? Why even read view is required as I think this is internal to operator?

yogidevendra:
Actually, after carefully looking at the concrete derived class; I noticed following:
1. In the derived class; usecase needs some custom handling in endWindow().
2. endWindow() is the place where derived class can look at the queue processed by the worker thread.

Other option:
1. Let derived class maintain its own copy of doneTuples (say doneTuples2). Derived class will maintain consistency of doneTuples2.
2. doneTuples2.add(tuple) will be called from processCommitedData() of the derived class.
3. But, this leads to unnecessary code duplication in the derived class. Instead, making these fields protected in the Abstract class would be useful.

gauravgopi123:
I think we should make the variables protected in Abstract Base Class

yogidevendra:
Thanks Gaurav for the reply. I will make the relevant changes and update the PR.

yogidevendra:
@gauravgopi123 Incorporated review comments. 
Changed the field access in question to protected.

yogidevendra:
Can you please review this and merge?

PR #150, opened 2015-12-30
MLHR-1959 Set offset manager properties from property files

pramodin:
Added getter for offset manager so that its properties can be set from property files


pramodin:
@siyuanh Please take a look

PR #153, opened 2015-12-31
Committing files for additional twitter operators and the RSS parser

NithinPerumal:
In this commit I have included the necessary JSON files and XML files to make my code work.  All applications were tested on the Sandbox through the Eclipse IDE and they compile + run successfully.  The XML and JSON files are contained in the resource folders, the testing files are in the testing folder and the main folder has the Twitter Demos + RSS parsing.  

I have modified and improved the code as per the feedback provided from my last pull request (Some heavy editing to the RSS reading).  The extraneous operators have also been located and removed.

### DIFF
--- a/demos/twitter/src/main/java/com/datatorrent/demos/twitter/RSSFeedExtractor.java
+++ b/demos/twitter/src/main/java/com/datatorrent/demos/twitter/RSSFeedExtractor.java
@@ -1,0 +1,133 @@
+/**
+ * Copyright (C) 2015 DataTorrent, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datatorrent.demos.twitter;
+
+import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.api.Context.OperatorContext;
+import com.datatorrent.api.InputOperator;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+
+import org.w3c.dom.Document;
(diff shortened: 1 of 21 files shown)


NithinPerumal:
I could use some help here - I'm not sure how to proceed here as it seems to be breaking due to a modified class (That needed to be modified).  How should I proceed to get past this build issue?

tweise:
@NithinPerumal Navigate to "Details" and look at the build log:

```
***! MODIFIED CLASS: PUBLIC com.datatorrent.contrib.twitter.TwitterSampleInput  (not serializable)
    ---! REMOVED FIELD: PUBLIC(-) FINAL(-) com.datatorrent.api.DefaultOutputPort media
    +++  NEW FIELD: PUBLIC(+) FINAL(+) com.datatorrent.api.DefaultOutputPort mediaType
```

Looks like you renamed or removed the media port property. This is not allowed for a minor version as it breaks backward compatibility.

NithinPerumal:
I was unaware that would happen as I replaced the existing code with working code.  I will go back and simply modify the already existing code in that case.

tweise:
Pull requests are tied to the branches. As soon as you update the source branch, the changes will reflect in the pull request, no need to open another one. See http://apex.incubator.apache.org/contributing.html#opening-pull-requests-contributors-

NithinPerumal:
I recommitted with necessary changes - still seeing a failure to build.  I don't quite understand the issue as the error I am seeing from the failed build is:

[ERROR] Failed to execute goal org.apache.rat:apache-rat-plugin:0.11:check (default-cli) on project twitter-demo: Too many files with unapproved license: 7 See RAT report in: /home/travis/build/apache/incubator-apex-malhar/demos/twitter/target/rat.txt -> [Help 1]

Not sure what to make of this.

tweise:
That's a problem with the license headers (either missing or not following the expected format). You can run these checks locally using: 

```
mvn apache-rat:check verify -Dlicense.skip=false -Pall-modules
```

chinmaykolhatkar:
The license headers seems to have more than required spaces before LICENSE link (http://www.apache.org/licenses/LICENSE-2.0)

Please have a look at below license header:
https://github.com/apache/incubator-apex-core/blob/devel-3/header.txt

demos/twitter/src/main/java/com/datatorrent/demos/twitter/RSSFeedExtractor.java
```diff
@@ -0,0 +1,133 @@
...
+			  if (j % 2 != 0) {
+				String tagName = tagList.item(j).getNodeName().toUpperCase();
+				String tagValue = tagList.item(j).getTextContent();
+			    System.out.println(tagName + " : " + tagValue);
+			  }
+		  }
+		  System.out.println();
+		  System.out.println("-----------------------------------");
+		  System.out.println();
+	    }
```
gauravgopi123, line 128: Instead of System.out.println please use logger

tweise:
@NithinPerumal please submit a new PR against master once review comments are addressed. We are removing the devel-3 branch.

PR #154, opened 2015-12-31
MLHR-1961 enhanced existing csv parser to take in schema for validations

shubham-pathak22:
enhanced existing csv parser to take in schema for validations
@gauravgopi123 could you please review.

### DIFF
--- a/contrib/pom.xml
+++ b/contrib/pom.xml
@@ -477,7 +477,7 @@
     <dependency>
       <groupId>net.sf.supercsv</groupId>
       <artifactId>super-csv</artifactId>
-      <version>2.2.0</version>
+      <version>2.4.0</version>
       <optional>true</optional>
     </dependency>
     <dependency>
@@ -610,12 +610,6 @@
       <artifactId>apex-common</artifactId>
       <version>${apex.core.version}</version>
       <type>jar</type>
-    </dependency>
-    <dependency>
-      <!-- required by Csv parser and formatter -->
-      <groupId>net.sf.supercsv</groupId>
-      <artifactId>super-csv-joda</artifactId>
-      <version>2.3.1</version>
-    </dependency>
+    </dependency>  
   </dependencies>
 </project>
--- a/contrib/src/main/java/com/datatorrent/contrib/parser/CellProcessorBuilder.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/parser/CellProcessorBuilder.java
@@ -1,0 +1,456 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.contrib.parser;
+
+import java.util.Map;
+
+import org.supercsv.cellprocessor.Optional;
+import org.supercsv.cellprocessor.ParseBool;
+import org.supercsv.cellprocessor.ParseChar;
+import org.supercsv.cellprocessor.ParseDate;
+import org.supercsv.cellprocessor.ParseDouble;
+import org.supercsv.cellprocessor.ParseInt;
+import org.supercsv.cellprocessor.ParseLong;
+import org.supercsv.cellprocessor.constraint.DMinMax;
+import org.supercsv.cellprocessor.constraint.Equals;
+import org.supercsv.cellprocessor.constraint.LMinMax;
+import org.supercsv.cellprocessor.constraint.StrMinMax;
+import org.supercsv.cellprocessor.constraint.StrRegEx;
+import org.supercsv.cellprocessor.constraint.Strlen;
+import org.supercsv.cellprocessor.ift.CellProcessor;
+import org.supercsv.cellprocessor.ift.DoubleCellProcessor;
+import org.supercsv.cellprocessor.ift.LongCellProcessor;
+import org.supercsv.util.CsvContext;
+
+import org.apache.commons.lang3.StringUtils;
+
+import com.datatorrent.contrib.parser.DelimitedSchema.FieldType;
+
+/**
+ * Helper class with methods to generate CellProcessor objects. Cell processors
+ * are an integral part of reading and writing with Super CSV - they automate
+ * the data type conversions, and enforce constraints. They implement the chain
+ * of responsibility design pattern - each processor has a single, well-defined
+ * purpose and can be chained together with other processors to fully automate
+ * all of the required conversions and constraint validation for a single
+ * delimited record.
+ * 
+ */
+public class CellProcessorBuilder
+{
+
+  /**
+   * Method to get cell processors for given field type and constraints
+   * 
+   * @param fieldType
+   *          data type of the field
+   * @param constraints
+   *          a map of constraints
+   * @return
+   */
+  public static CellProcessor getCellProcessor(FieldType fieldType, Map<String, Object> constraints)
+  {
+    switch (fieldType) {
+      case STRING:
+        return getStringCellProcessor(constraints);
+      case INTEGER:
+        return getIntegerCellProcessor(constraints);
+      case LONG:
+        return getLongCellProcessor(constraints);
+      case FLOAT:
+      case DOUBLE:
+        return getDoubleCellProcessor(constraints);
+      case CHARACTER:
+        return getCharCellProcessor(constraints);
+      case BOOLEAN:
+        return getBooleanCellProcessor(constraints);
+      case DATE:
+        return getDateCellProcessor(constraints);
+      default:
+        return null;
+    }
+  }
+
+  /**
+   * Method to get cellprocessor for String with constraints. These constraints
+   * are evaluated against the String field for which this cellprocessor is
+   * defined.
+   * 
+   * @param constraints
+   *          map of constraints applicable to String
+   * @return CellProcessor
+   */
+  private static CellProcessor getStringCellProcessor(Map<String, Object> constraints)
+  {
+    Boolean required = constraints.get(DelimitedSchema.REQUIRED) == null ? null : Boolean
+        .parseBoolean((String)constraints.get(DelimitedSchema.REQUIRED));
+    Integer strLen = constraints.get(DelimitedSchema.LENGTH) == null ? null : Integer.parseInt((String)constraints
+        .get(DelimitedSchema.LENGTH));
+    Integer minLength = constraints.get(DelimitedSchema.MIN_LENGTH) == null ? null : Integer
+        .parseInt((String)constraints.get(DelimitedSchema.MIN_LENGTH));
+    Integer maxLength = constraints.get(DelimitedSchema.MAX_LENGTH) == null ? null : Integer
+        .parseInt((String)constraints.get(DelimitedSchema.MAX_LENGTH));
+    String equals = constraints.get(DelimitedSchema.EQUALS) == null ? null : (String)constraints
+        .get(DelimitedSchema.EQUALS);
+    String pattern = constraints.get(DelimitedSchema.REGEX_PATTERN) == null ? null : (String)constraints
+        .get(DelimitedSchema.REGEX_PATTERN);
+
+    CellProcessor cellProcessor = null;
+    if (StringUtils.isNotBlank(equals)) {
+      cellProcessor = new Equals(equals);
+    } else if (StringUtils.isNotBlank(pattern)) {
+      cellProcessor = new StrRegEx(pattern);
+    } else if (strLen != null) {
+      cellProcessor = new Strlen(strLen);
+    } else if (maxLength != null || minLength != null) {
+      Long min = minLength == null ? 0L : minLength;
+      Long max = maxLength == null ? LMinMax.MAX_LONG : maxLength;
+      cellProcessor = new StrMinMax(min, max);
+    }
+    if (required == null || !required) {
+      cellProcessor = addOptional(cellProcessor);
+    }
+    return cellProcessor;
+  }
+
+  /**
+   * Method to get cellprocessor for Integer with constraints. These constraints
+   * are evaluated against the Integer field for which this cellprocessor is
+   * defined.
+   * 
+   * @param constraints
+   *          map of constraints applicable to Integer
+   * @return CellProcessor
+   */
+  private static CellProcessor getIntegerCellProcessor(Map<String, Object> constraints)
+  {
+    Boolean required = constraints.get(DelimitedSchema.REQUIRED) == null ? null : Boolean
+        .parseBoolean((String)constraints.get(DelimitedSchema.REQUIRED));
+    Integer equals = constraints.get(DelimitedSchema.EQUALS) == null ? null : Integer.parseInt((String)constraints
+        .get(DelimitedSchema.EQUALS));
+    Integer minValue = constraints.get(DelimitedSchema.MIN_VALUE) == null ? null : Integer.parseInt((String)constraints
+        .get(DelimitedSchema.MIN_VALUE));
+    Integer maxValue = constraints.get(DelimitedSchema.MAX_VALUE) == null ? null : Integer.parseInt((String)constraints
+        .get(DelimitedSchema.MAX_VALUE));
+
+    CellProcessor cellProcessor = null;
+    if (equals != null) {
+      cellProcessor = new Equals(equals);
+      cellProcessor = addParseInt(cellProcessor);
+    } else if (minValue != null || maxValue != null) {
+      cellProcessor = addIntMinMax(minValue, maxValue);
+    } else {
+      cellProcessor = addParseInt(null);
+    }
+    if (required == null || !required) {
+      cellProcessor = addOptional(cellProcessor);
+    }
+    return cellProcessor;
+  }
+
+  /**
+   * Method to get cellprocessor for Long with constraints. These constraints
+   * are evaluated against the Long field for which this cellprocessor is
+   * defined.
+   * 
+   * @param constraints
+   *          map of constraints applicable to Long
+   * @return CellProcessor
+   */
+  private static CellProcessor getLongCellProcessor(Map<String, Object> constraints)
+  {
+    Boolean required = constraints.get(DelimitedSchema.REQUIRED) == null ? null : Boolean
+        .parseBoolean((String)constraints.get(DelimitedSchema.REQUIRED));
+    Long equals = constraints.get(DelimitedSchema.EQUALS) == null ? null : Long.parseLong((String)constraints
+        .get(DelimitedSchema.EQUALS));
+    Long minValue = constraints.get(DelimitedSchema.MIN_VALUE) == null ? null : Long.parseLong((String)constraints
+        .get(DelimitedSchema.MIN_VALUE));
+    Long maxValue = constraints.get(DelimitedSchema.MAX_VALUE) == null ? null : Long.parseLong((String)constraints
+        .get(DelimitedSchema.MAX_VALUE));
+    CellProcessor cellProcessor = null;
+    if (equals != null) {
+      cellProcessor = new Equals(equals);
+      cellProcessor = addParseLong(cellProcessor);
+    } else if (minValue != null || maxValue != null) {
+      cellProcessor = addLongMinMax(minValue, maxValue);
+    } else {
+      cellProcessor = addParseLong(null);
+    }
+    if (required == null || !required) {
+      cellProcessor = addOptional(cellProcessor);
+    }
+    return cellProcessor;
+  }
+
+  /**
+   * Method to get cellprocessor for Float/Double with constraints. These
+   * constraints are evaluated against the Float/Double field for which this
+   * cellprocessor is defined.
+   * 
+   * @param constraints
+   *          map of constraints applicable to Float/Double
+   * @return CellProcessor
+   */
+  private static CellProcessor getDoubleCellProcessor(Map<String, Object> constraints)
+  {
+    Boolean required = constraints.get(DelimitedSchema.REQUIRED) == null ? null : Boolean
+        .parseBoolean((String)constraints.get(DelimitedSchema.REQUIRED));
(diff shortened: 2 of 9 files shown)


contrib/src/main/java/com/datatorrent/contrib/parser/CellProcessorBuilder.java
```diff
@@ -0,0 +1,456 @@
...
+   * defined.
+   * 
+   * @param constraints
+   *          map of constraints applicable to String
+   * @return CellProcessor
+   */
+  private static CellProcessor getStringCellProcessor(Map<String, Object> constraints)
+  {
+    Boolean required = constraints.get(DelimitedSchema.REQUIRED) == null ? null : Boolean
+        .parseBoolean((String)constraints.get(DelimitedSchema.REQUIRED));
```
chinmaykolhatkar, line 102: Why should this class care about the conversion to from strings to respective types?

Also if everything is going to be string, why not have a Map<String, String>

contrib/src/main/java/com/datatorrent/contrib/parser/CellProcessorBuilder.java
```diff
@@ -0,0 +1,456 @@
...
+        .parseInt((String)constraints.get(DelimitedSchema.MIN_LENGTH));
+    Integer maxLength = constraints.get(DelimitedSchema.MAX_LENGTH) == null ? null : Integer
+        .parseInt((String)constraints.get(DelimitedSchema.MAX_LENGTH));
+    String equals = constraints.get(DelimitedSchema.EQUALS) == null ? null : (String)constraints
+        .get(DelimitedSchema.EQUALS);
+    String pattern = constraints.get(DelimitedSchema.REGEX_PATTERN) == null ? null : (String)constraints
+        .get(DelimitedSchema.REGEX_PATTERN);
+
+    CellProcessor cellProcessor = null;
+    if (StringUtils.isNotBlank(equals)) {
```
chinmaykolhatkar, line 115: For the if/else block:
Does this mean there is its not allowed to have multiple of them as constraints?
If yes, why?

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +71,193 @@
  * @since 3.2.0
  */
 @InterfaceStability.Evolving
-public class CsvParser extends Parser<String, String>
+public class CsvParser extends Parser<byte[], KeyValPair<String, String>>
```
chinmaykolhatkar, line 75: Should this be renamed to DelimitedParser rather than CsvParser.
I believe it can work on delimited strings, correct?

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +71,193 @@
...
   {
-    try {
-      csvStringReader.open(tuple);
-      return csvReader.read(clazz, nameMapping, processors);
-    } catch (IOException e) {
-      logger.debug("Error while converting tuple {} {}",tuple,e.getMessage());
-      return null;
-    }
+    //This method is not invoked for CSV parser
+    return null;
```
chinmaykolhatkar: If this is not invoked, please throw UnsupportedOperationException rather then returning any value.

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +71,193 @@
...
 
   @Override
-  public void teardown()
+  public void processTuple(byte[] tuple)
   {
-    try {
-      if (csvReader != null) {
-        csvReader.close();
+    if (tuple == null) {
+      if (err.isConnected()) {
```
chinmaykolhatkar, line 156: If condition of err.isConnected is not required when action is just to emit.

By default its connected to Sink.BLACKHOLE on which put object does not do anything.

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +71,193 @@
...
-    {
-      return type;
-    }
-
-    public void setType(String type)
-    {
-      this.type = FIELD_TYPE.valueOf(type);
+    String incomingString = new String(tuple);
+    if (StringUtils.isBlank(incomingString) || StringUtils.equals(incomingString, header)) {
+      if (err.isConnected()) {
```
chinmaykolhatkar, line 164: err.isConnected not required if action is just emit.

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +71,193 @@
...
-    {
-      this.format = format;
+    } catch (SuperCsvException e) {
+      if (err.isConnected()) {
+        err.emit(new KeyValPair<String, String>(incomingString, e.getMessage()));
+      }
+      errorTupleCount++;
+      logger.error("Tuple could not be parsed. Reason {}", e.getMessage());
+    } catch (IOException e) {
+      logger.error("Exception in process method {}", e.getMessage());
```
chinmaykolhatkar: Why is IOException case distinguished from SuprtCsvException?

In case of SuperCsvException, we know for sure that it failed to parse, hence sent to err port.
In case of IOException do we know for sure that its not the case for sending to err port?

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +71,193 @@
...
-      logger.debug("Error while converting tuple {} {}",tuple,e.getMessage());
-      return null;
-    }
+    //This method is not invoked for CSV parser
+    return null;
   }
 
   @Override
-  public void teardown()
+  public void processTuple(byte[] tuple)
```
chinmaykolhatkar, line 153: Lot of code duplicate here.

Can you please create 2 methods, emitError and emitSuccess... Let these 2 methods emit on respective ports and also do other activities like increment the counter. This way the tasks are associated and never missed in any future code changes.

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +71,193 @@
...
   }
 
-  /**
-   * Gets the delimiter which separates fields in incoming data.
-   * 
-   * @return fieldDelimiter
-   */
-  public int getFieldDelimiter()
+  @Override
+  public KeyValPair<String, String> processErorrTuple(byte[] input)
```
chinmaykolhatkar: Spelling of "Error"...

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +71,193 @@
...
   {
-    return fieldDelimiter;
+    //This method is not invoked for CSV parser
+    return null;
```
chinmaykolhatkar: Please throw UnsupportedOperationException if this is not meant to be called.

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +71,193 @@
...
+   * output port to emit validate records as map
    */
-  public void setFieldInfo(String fieldInfo)
-  {
-    this.fieldInfo = fieldInfo;
-  }
-
-  private static final Logger logger = LoggerFactory.getLogger(CsvParser.class);
+  public final transient DefaultOutputPort<Map<String, Object>> parsedOutput = new DefaultOutputPort<Map<String, Object>>();
+  private static final Logger logger = LoggerFactory.getLogger(Parser.class);
```
chinmaykolhatkar: Not sure if there is any convention here.. But I have seen most of the files declaring the variables are the top of the file.

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +71,193 @@
...
-   * separated string in same order as incoming data. e.g
-   * name:string,dept:string,eid:integer,dateOfJoining:date|dd/mm/yyyy
+   * Set path of schema
    * 
-   * @return fieldInfo
+   * @param schemaPath
+   *          path of the schema file in hdfs
    */
-  public String getFieldInfo()
+  public void setSchemaPath(String schemaPath)
```
chinmaykolhatkar: I it a requirement that this needs to be retrieved from HDFS. User might have a file in local file system.

I would suggest, reading config files from FileSystems in not responsibility of any operator. The file should be read by application and the content in string/byte[] format should be passed to the operator.

contrib/src/main/java/com/datatorrent/contrib/parser/DelimitedSchema.java
```diff
@@ -0,0 +1,359 @@
...
+  private List<String> fieldNames = new LinkedList<String>();
+  /**
+   * This holds list of {@link Field}
+   */
+  private List<Field> fields = new LinkedList<Field>();
+
+  /**
+   * Supported data types
+   */
+  public enum FIELD_TYPE
```
chinmaykolhatkar: Can you please rename this to "FieldType"
All caps is usually used for parameterized classes OR static final constants etc.

I actually got confused in CellProcessorBuilder class when I saw FIELD_TYPE as input param type to be method.

contrib/src/main/java/com/datatorrent/contrib/parser/DelimitedSchema.java
```diff
@@ -0,0 +1,359 @@
...
+
+    /**
+     * Sets the map of constraints associated with the field
+     * 
+     * @param constraints
+     */
+    public void setConstraints(Map<String, Object> constraints)
+    {
+      this.constraints = constraints;
+    }
```
chinmaykolhatkar, line 348: Do you want to provide addConstraint method here as well?

contrib/src/test/resources/schema.json
```diff
@@ -0,0 +1,96 @@
...
+                "required": "true",
+                "equals": "Y"
+            }
+        },
+        {
+            "name": "valid",
+            "type": "String"            
+        }
+    ]
+}
```
chinmaykolhatkar, line 96: New line missing at the end of the file.

library/src/main/java/com/datatorrent/lib/util/FileUtils.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.commons.io.IOUtils;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+/**
+ * This class holds utility methods related to files
+ *
+ */
+public class FileUtils
```
chinmaykolhatkar: Can you please check if there is already a utility like this existing in malhar?

contrib/src/main/java/com/datatorrent/contrib/parser/CellProcessorBuilder.java
shubham-pathak22, line 115: We can surely have multiple constraints. However for String datatype, the constraints we are supporting are  equals, regex, string length, min and max length. For these, specifying multiple wouldn't make much sense. So, even if user provide multiple constraints, e.g regex and equals OR equals and length, the if/else block adds appropriate processors.

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
shubham-pathak22, line 75: This is enhancement to existing operator in the library. Not sure whether i can rename the operator itself.

contrib/src/main/java/com/datatorrent/contrib/parser/CellProcessorBuilder.java
shubham-pathak22, line 102: Why not have a map ? It is  a Map<String, Object> 
And methods will have to convert the values to the required types.

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
shubham-pathak22: Good catch. Will have to fix this in parent class

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
shubham-pathak22: Good suggestion. Will make necessary changes.

library/src/main/java/com/datatorrent/lib/util/FileUtils.java
shubham-pathak22: would not need this anymore.

contrib/src/main/java/com/datatorrent/contrib/parser/DelimitedSchema.java
shubham-pathak22, line 348: Currently don't see the need . :)

shubham-pathak22:
incorporated review comments

contrib/src/main/java/com/datatorrent/contrib/parser/CellProcessorBuilder.java
chinmaykolhatkar, line 102: I think the comment did not take Map < String, String >

contrib/src/test/java/com/datatorrent/contrib/parser/CsvPOJOParserTest.java
```diff
@@ -20,169 +20,425 @@
...
-  public Watcher watcher = new Watcher();
+  /*
+  * adId,campaignId,adName,bidPrice,startDate,endDate,securityCode,isActive,isOptimized,parentCampaign,weatherTargeted,valid
+  * 1234,98233,adxyz,0.2,2015-03-08 03:37:12,11/12/2012,12,y,OPTIMIZE,CAMP_AD,Y,yes
+  * Constraints are defined in schema.json
+  */
 
-  public class Watcher extends TestWatcher
+  @Before
+  public void setup()
```
gauravgopi123: Why TestWatcher can't be used for same thing?

contrib/src/test/java/com/datatorrent/contrib/parser/CsvPOJOParserTest.java
```diff
@@ -20,169 +20,425 @@
...
+    parser.in.process("1234,98233,adxyz,0.2,2015-03-08 03:37:12,11/12/2012,12,y,OPTIMIZATION,CAMP_AD,Y,yes,ExtraField"
+        .getBytes());
+    parser.teardown();
+    Assert.assertEquals(0, objectPort.collectedTuples.size());
+    Assert.assertEquals(0, pojoPort.collectedTuples.size());
+    Assert.assertEquals(1, error.collectedTuples.size());
   }
 
-  public static class EmployeeBean
+  public static class Ad
```
gauravgopi123, line 383: Could we not enhance EmployeeBean?

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +69,177 @@
...
+  long parsedOutputCount;
 
-  public CsvParser()
+  @Override
+  public void beginWindow(long windowId)
   {
-    fields = new ArrayList<Field>();
-    fieldDelimiter = ',';
-    lineDelimiter = "\r\n";
+    parsedOutputCount = 0;
```
gauravgopi123, line 128: It should call super.beginWindow() to reset other autometric every window. Please add test case to cover this case

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +69,177 @@
...
 
-    public void setFormat(String format)
-    {
-      this.format = format;
+    } catch (SuperCsvException | IOException e) {
+      if (err.isConnected()) {
+        err.emit(new KeyValPair<String, String>(incomingString, e.getMessage()));
+      }
+      errorTupleCount++;
+      logger.error("Tuple could not be parsed. Reason {} Stacktrace {}", e.getMessage(), e.getStackTrace());
```
gauravgopi123: simply do `logger.error("Tuple could not be parsed. Exception: ", e);`

contrib/pom.xml
```diff
@@ -614,8 +614,8 @@
     <dependency>
       <!-- required by Csv parser and formatter -->
       <groupId>net.sf.supercsv</groupId>
-      <artifactId>super-csv-joda</artifactId>
-      <version>2.3.1</version>
+      <artifactId>super-csv</artifactId>
+      <version>2.4.0</version>
```
gauravgopi123: Make this dependency as optional

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -19,42 +19,49 @@
...
  * <b>Properties</b> <br>
- * <b>fieldInfo</b>:User need to specify fields and their types as a comma
- * separated string having format &lt;NAME&gt;:&lt;TYPE&gt;|&lt;FORMAT&gt; in
- * the same order as incoming data. FORMAT refers to dates with dd/mm/yyyy as
- * default e.g name:string,dept:string,eid:integer,dateOfJoining:date|dd/mm/yyyy <br>
- * <b>fieldDelimiter</b>: Default is comma <br>
- * <b>lineDelimiter</b>: Default is '\r\n'
+ * <b>schemaPath</b>:Complete path of schema file in HDFS <br>
+ * <b>clazz</b>:Pojo class <br>
+ * <b>Ports</b> <br>
```
gauravgopi123, line 59: Port names mentioned below don't match actual variable names. Port names are `in` for input, `out` for pojo, `err` for error and `parsedOutput` for  map

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
gauravgopi123: LoggerFactory.getLogger(CsvParser.class)

contrib/src/test/java/com/datatorrent/contrib/parser/CsvPOJOParserTest.java
shubham-pathak22, line 383: I could have , however the schema  and unit tests already existed for Ad class, hence used those.

contrib/src/test/java/com/datatorrent/contrib/parser/CsvPOJOParserTest.java
shubham-pathak22: It can be. I used the same code that i had already written for enhanced parser.

gauravgopi123: If we can use then I think we should use the existing code.

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +69,178 @@
...
+  long parsedOutputCount;
 
-  public CsvParser()
+  @Override
+  public void beginWindow(long windowId)
   {
-    fields = new ArrayList<Field>();
-    fieldDelimiter = ',';
-    lineDelimiter = "\r\n";
+    super.beginWindow(windowId);
```
gauravgopi123, line 127: Please add test case for auto-metric verification

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +69,178 @@
...
-   */
-  public String getLineDelimiter()
+  @Override
+  public void teardown()
   {
-    return lineDelimiter;
+    try {
+      csvMapReader.close();
+    } catch (IOException e) {
+      logger.error("Error while closing csv map reader {}", e.getMessage());
```
gauravgopi123, line 219: It should throw exception

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +69,178 @@
...
+    }
+    try {
+      csvBeanReader.close();
+    } catch (IOException e) {
+      logger.error("Error while closing csv bean reader {}", e.getMessage());
```
gauravgopi123, line 225: It should throw exception

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +69,178 @@
...
+  /**
+   * Bean Reader to read delimited records
+   */
+  private transient CsvBeanReader csvBeanReader;
+  /**
+   * Reader used by csvMapReader and csvBeanReader
+   */
+  private transient ReusableStringReader csvStringReader;
+  /**
+   * Contents of the schema
```
gauravgopi123: can you add example for schema?

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -62,247 +69,178 @@
...
+  private transient String header;
+  /**
+   * Reading preferences that are passed through schema
+   */
+  private transient CsvPreference preference;
 
-  @NotNull
-  private transient ReusableStringReader csvStringReader = new ReusableStringReader();
+  @AutoMetric
+  long parsedOutputCount;
```
gauravgopi123, line 122: Javadoc for this?

shubham-pathak22:
Incorporated review comments

shubham-pathak22:
Hi, Can we merge this if there aren't any more review comments ?

shubham-pathak22:
Hello, Can we merge this if there aren't any more review comments ?

contrib/src/test/java/com/datatorrent/contrib/parser/CsvPOJOParserTest.java
```diff
@@ -20,169 +20,453 @@
...
+    Assert.assertEquals(0, pojoPort.collectedTuples.size());
+    Assert.assertEquals(1, error.collectedTuples.size());
+  }
+
+  @Test
+  public void TestParserInvalidBidPriceInput()
+  {
+    String input = "1234,98233,adxyz,3.3,2015-03-08 03:37:12,11/12/2012,12,y,,CAMP_AD,Y,yes";
+    parser.in.process(input.getBytes());
+    parser.teardown();
```
gauravgopi123: parser.teardown() call should be part of finished() call of TestWatcher

gauravgopi123:
I don't see the test case for testing autometric counts being reset every begin window

shubham-pathak22:
Earlier had added test for just auto-metric verification. Added another test case for testing autometric counts being reset every begin window

shubham-pathak22:
Comments related to test case have been incorporated.

shubham-pathak22:
Updated tests for autometric verication and squashed commits

gauravgopi123:
Please squash commits and rebase it..

shubham-pathak22:
Done.

shubham-pathak22:
Closing the PR as it is merged.

PR #157, opened 2016-01-04
APEXMALHAR-1963 AbstractAsyncProcessor processes tuples asynchronously

chinmaykolhatkar:
Processes tuples async/parallel threads.
Also takes care of order and timeout of processing.
Also take care of checkpointing.

@gauravgopi123 Please review.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
+++ b/library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
@@ -36,28 +36,27 @@ import com.datatorrent.common.util.BaseOperator;
 
 
 /**
- * This base operator can be used to enqueue tuples for processing asynchronously by seperate Worker Threads.
- * This operator can be implemented to do async and parallel operations.
+ * This base operator can be used to enqueue tuples for asynchronous and parallel processing by separate worker threads.
  *
- * User can configure a number of threads that are required for async processing. Defaulting to no limit, meaning the
- * threads will be created for as per need basis. These worker executor threads are seperate than operator thread.
- * The user can also configure the timeout for the operations to happen.
+ * User can configure number of threads required for async processing. By default
+ * threads will be created on need basis with no upper limit. These worker threads, spawned by operator thread,
+ * can also be configured to processing timeout.
  *
  * <b>Workflow is as follows:</b>
  * <ol>
  *   <li>Implementing class can enqueue the tuples of type QUEUETUPLE using <i>enqueueTupleForProcessing</i>
  *       method.</li>
- *   <li>When a tuple gets its turn for execution, <i>processTupleAsync</i> gets called from processing thread pool
- *       other than operator thread.</li>
+ *   <li>When a tuple gets its turn for execution, <i>processTupleAsync</i> gets called
+ *       by worker thread. This is asynchronous call</li>
  *   <li>Based on maintainTupleOrder parameter, the results are consolidated at the end of every window and respective
- *       call to <i>handleProcessedTuple</i> will be made. This call will happen in operator thread.</li>
+ *       call to <i>handleProcessedTuple</i> will be made. This callback is in operator thread.</li>
  * </ol>
  *
  *
  * <b>Use case examples:</b>
  * <ul>
  *   <li>Parallel reads from external datastore/database system.</li>
- *   <li>Doing any operations which are long running tasks per tuple but DAG io should not be blocked.</li>
+ *   <li>Operations which are long running tasks per tuple but DAG i/o should not be blocked.</li>
  * </ul>
  *
  * @param <QUEUETUPLE> input type of tuple
@@ -68,7 +67,7 @@ public abstract class AbstractAsyncProcessor<QUEUETUPLE, RESULTTUPLE> extends Ba
   private static final Logger logger = LoggerFactory.getLogger(AbstractAsyncProcessor.class);
 
   /**
-   * This is a executor service which hold thread pool and enqueues the task for processing.
+   * This is an executor service which holds thread pool and enqueues task for processing.
    */
   private transient ExecutorService executor;
 
@@ -87,18 +86,22 @@ public abstract class AbstractAsyncProcessor<QUEUETUPLE, RESULTTUPLE> extends Ba
   private boolean maintainTupleOrder = false;
 
   /**
-   * Timeout is milliseconds for how long a tuple can go through processing.
+   * Timeout in milliseconds after which processing of tuple may be terminated.
    */
   private long processTimeoutMs = 0;
 
   /**
-   * This Queue will hold all the tuples that are in process, has completed processing but yet to be notified for
-   * processing completion.
+   * The queue holds following types of tuples:
+   * <ul>
+   *   <li>Yet to be processed</li>
+   *   <li>In Process</li>
+   *   <li>Completed processing but yet to be notified for process completion</li>
+   * </ul>
    */
-  private Queue<ProcessTuple> waitingTuples = Lists.newLinkedList();
+  private Queue<WorkItem> waitingTuples = Lists.newLinkedList();
 
   /**
-   * Setup method of this operator will initialize the various types of executors possible.
+   * Setup method of this operator will initialize various types of executors possible.
    *
    * @param context OperatorContext
    */
@@ -114,18 +117,26 @@ public abstract class AbstractAsyncProcessor<QUEUETUPLE, RESULTTUPLE> extends Ba
   }
 
   /**
-   * This method will be called from setup of this operator. The method will enqueue all the tuples for prcessing to
-   * Executors which are pre-existing in processing queue i.e. <i>waitingTuples</i> variable.
-   * The <i>waitingTuple</i> is expected to contain anything only if the operators gets restored back to previous
-   * checkpointed state.
+   * This method is called from setup of the operator. If <i>waitingTuples</i> contains anything, the method enqueues
+   * all those tuples for processing to worker threads.
+   *
+   * Dependening on condition the method will behave as follows:
+   * <ul>
+   *   <li>During fresh start of operator, this method will not have any effect as there are no tuples in
+   *       <i>waitingTuples</i>.</li>
+   *   <li>During restore of operator to previous checkpointed state, if there are any tuples already present
+   *       in <i>waitingTuples</i> that means they're yet to be notified, hence enqueued again.</li>
+   * </ul>
    */
   private void replayTuplesIfRequired()
   {
     // Enqueue everything that is still left.
     // Its left because its either not completed processing OR its not emitted because of ordering restrictions.
-    for (ProcessTuple processTuple : waitingTuples) {
-      processTuple.taskHandle = executor.submit(new Processor(processTuple));
-      processTuple.processState = State.SUBMITTED;
+    for (WorkItem item : waitingTuples) {
+      item.taskHandle = executor.submit(new Processor(item));
+      item.processState = State.SUBMITTED;
+      item.processStartTime = 0;
+      item.outTuple = null;
     }
   }
 
@@ -138,25 +149,28 @@ public abstract class AbstractAsyncProcessor<QUEUETUPLE, RESULTTUPLE> extends Ba
   }
 
   /**
-   * Enqueues tuple of type QUEUETUPLE for processing to Worker threads.
+   * Enqueues tuple of type QUEUETUPLE for processing to worker threads.
    * This method should be called only from operator thread.
    *
    * @param QUEUETUPLE Input tuple of type QUEUETUPLE
    */
   protected void enqueueTupleForProcessing(QUEUETUPLE QUEUETUPLE)
   {
-    ProcessTuple processTuple = new ProcessTuple(QUEUETUPLE);
-    waitingTuples.add(processTuple);
-    processTuple.taskHandle = executor.submit(new Processor(processTuple));
-    processTuple.processState = State.SUBMITTED;
+    WorkItem item = new WorkItem(QUEUETUPLE);
+    waitingTuples.add(item);
+    item.taskHandle = executor.submit(new Processor(item));
+    item.processState = State.SUBMITTED;
   }
 
   /**
-   * Abstract method which will get called from this AsyncProcessor when the results for certain input tuple
+   * Abstract method which gets called from AsyncProcessor when results for certain input tuple
    * are ready to be notified.
-   * The method will be called from operator thread hence this is a sync method of this operator.
+   * The method is called from operator thread hence this is synchronous method of this operator.
    *
-   * The implementer of this method should keep this method to minimal overhead.
+   * <b>
+   *   The implementer of this method should keep this method to minimal overhead because this being part of operator
+   *   thread, the long running tasks may block DAG i/o.
+   * </b>
    *
    * @param inpTuple     Input tuple which was processed of type QUEUETUPLE
    * @param resultTuple  Result tuple of type RESULTTUPLE
@@ -165,7 +179,7 @@ public abstract class AbstractAsyncProcessor<QUEUETUPLE, RESULTTUPLE> extends Ba
   protected abstract void handleProcessedTuple(QUEUETUPLE inpTuple, RESULTTUPLE resultTuple, State processState);
 
   /**
-   * Abstract method which will get called from Worker threads for processing given input tuple of type QUEUETUPLE.
+   * Abstract method which gets called from Worker threads for processing given input tuple of type QUEUETUPLE.
    * The method will be called in Async mode.
    * The implementer of this method can chose to have long running tasks here.
    *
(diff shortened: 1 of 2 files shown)


chinmaykolhatkar:
@tushargosavi @sandeepdeshmukh Can you please review and merge?

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
```diff
@@ -0,0 +1,357 @@
...
+ *
+ * <b>Workflow is as follows:</b>
+ * <ol>
+ *   <li>Tuple of type INPUT is received on input port.</li>
+ *   <li>Tuple gets enqueued into Executors by calling <i>enqueueTupleForProcessing</i>.
+ *       Once can override <i>processTuple</i> method in which case, its users responsibility to enqueue tuple for
+ *       processing by calling <i>enqueueTupleForProcessing</i>.</li>
+ *   <li>When tuple gets its turn of execution, <i>processTupleAsync</i> gets called from thread other than
+ *       operator thread.</li>
+ *   <li>Based on maintainTupleOrder parameter, the results are consolidated at end of every window and respective call
```
sandeepdeshmukh: at _the_ end of

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
```diff
@@ -0,0 +1,357 @@
...
+ *
+ * User can configure number of of threads that are required for processing. Defaulting to single thread which is
+ * different than operator thread.
+ * The user will also configure the timeout for the operations to happen.
+ *
+ * <b>Workflow is as follows:</b>
+ * <ol>
+ *   <li>Tuple of type INPUT is received on input port.</li>
+ *   <li>Tuple gets enqueued into Executors by calling <i>enqueueTupleForProcessing</i>.
+ *       Once can override <i>processTuple</i> method in which case, its users responsibility to enqueue tuple for
```
sandeepdeshmukh: Once -> One

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
```diff
@@ -0,0 +1,357 @@
...
+ *       processing by calling <i>enqueueTupleForProcessing</i>.</li>
+ *   <li>When tuple gets its turn of execution, <i>processTupleAsync</i> gets called from thread other than
```
sandeepdeshmukh: -> When a tuple
Give some name to the other thread.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
```diff
@@ -0,0 +1,357 @@
...
+ * @param <INPUT> input type of tuple
+ * @param <RESULT> Result of async processing of tuple.
+ * @since 3.3.0
+ */
+public abstract class AbstractAsyncProcessor<INPUT, RESULT> extends BaseOperator implements IdleTimeHandler
+{
+  private static final Logger logger = LoggerFactory.getLogger(AbstractAsyncProcessor.class);
+
+  private transient ExecutorService executor;
+  private int numProcessors = 1;
```
sandeepdeshmukh: Default of 1 for this purpose is not recommended. That will almost make it serial for each tuple. Suggest to increase it to a higher number. There is no perfect number for this. It will certainly depend on the system you are interacting with. 
Suggested number is 16 at least.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
```diff
@@ -0,0 +1,357 @@
...
+   */
+  @Override public void setup(Context.OperatorContext context)
+  {
+    if (numProcessors == 1) {
+      this.executor = Executors.newSingleThreadExecutor();
+    } else if (numProcessors == 0) {
+      this.executor = Executors.newCachedThreadPool();
+    } else if (numProcessors > 0) {
+      this.executor = Executors.newFixedThreadPool(numProcessors);
+    } else {
```
sandeepdeshmukh: @Min = 0  for numProcessors? Will save this else case.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
```diff
@@ -0,0 +1,357 @@
...
+  private boolean maintainTupleOrder = false;
+  private long processTimeoutMs = 0;
+
+  /**
+   * This Queue will hold all the tuples that are in process, has completed processing but yet to be notified for
+   * processing completion.
+   */
+  private Queue<ProcessTuple> waitingTuples = Lists.newLinkedList();
+
+  @InputPortFieldAnnotation(optional = true)
```
shubham-pathak22: why optional ?

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
```diff
@@ -0,0 +1,357 @@
...
+   * @param context OperatorContext
+   */
+  @Override public void setup(Context.OperatorContext context)
+  {
+    if (numProcessors == 1) {
+      this.executor = Executors.newSingleThreadExecutor();
+    } else if (numProcessors == 0) {
+      this.executor = Executors.newCachedThreadPool();
+    } else if (numProcessors > 0) {
+      this.executor = Executors.newFixedThreadPool(numProcessors);
```
shubham-pathak22: Do we need to distinguish between Executors.newSingleThreadExecutor() and Executors.newFixedThreadPool(numProcessors) with numProcessors = 1 ? Would it be functionally different ?

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
shubham-pathak22: Do we need to have max limit ?

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
chinmaykolhatkar: Done

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
chinmaykolhatkar: Done

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
chinmaykolhatkar: Done.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
chinmaykolhatkar: Made it 16.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
chinmaykolhatkar: Done.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
chinmaykolhatkar: Done.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
chinmaykolhatkar: I'm actually considering to remove the input port form this.

I just need to ensure that this abstract operator is bound to operator lifecycle, so that the checkpointing can be handled properly.

The abstract has provided enqueue method which should take care of queing the tuples for processing.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
chinmaykolhatkar: Actually let me keep this to 0. This means, the threads will be created as per need.

sandeepdeshmukh:
@gauravgopi123 : Could you please review this further. I have given my review comments and @chinmaykolhatkar  has incorporated them.

chinmaykolhatkar:
@gauravgopi123 Can you please review this pull request?

@sandeepdeshmukh @tushargosavi @shubham-pathak22 @gauravgopi123 , Once you're happy can you please merge this pull request?

Thanks.

sandeepdeshmukh:
Looks good to me. If there are no further comments, I will merge this in a day.

gauravgopi123:
@chinmaykolhatkar : Can you please add the use cases where this is required and AbstractReconciler won't solve the issue?

chinmaykolhatkar:
@gauravgopi123 The usecase is as follows:
1. Parallel reads from external datastore/database system.
2. Operations which are long running tasks per tuple but DAG i/o should not be blocked.

I've already added thses use cases as a part of javadoc for AbstractAsyncProcessor.

Difference between AbstractReconciler & AbstractAsyncProcessor are following:
1. AbstractReconciler waits till committed call and then start processing. AbstractAsyncProcessor won't, it'll start processing as soon as the tuple gets enqueued. Waiting till committed call will delay the processing of tuple and hence add to the latency. A seperate datastructure takes care queued data and hence checkpoint is taken care off that way.
2. AbstractReconciler does async processing but on a single thread. Whereas AbstractAsyncProcessor can work off a configurable number of thread.

Please correct me if I'm wrong about understanding of AbstractReconciler.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
```diff
@@ -0,0 +1,344 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(AbstractAsyncProcessor.class);
+
+  /**
+   * This is an executor service which holds thread pool and enqueues task for processing.
+   */
+  private transient ExecutorService executor;
+
+  /**
+   * Number of worker threads for processing
+   * 0   - Threads will be created as per need.
```
gauravgopi123: I don't think creating threads on need basis is good idea.. You need to have control on the number of threads

library/src/test/java/com/datatorrent/lib/async/AbstractAsyncProcessorTest.java
```diff
@@ -0,0 +1,145 @@
...
+      for (int i = 0; i < 100; i++) {
+        async.input.put(Integer.toString(windowId * 100 + i));
+      }
+      async.endWindow();
+    }
+
+    for (int i = 0; i < 10; i++) {
+      Thread.sleep(waitInterval / 10);
+      async.endWindow();
+    }
```
gauravgopi123, line 104: what is this loop doing? endWindow() is called for every beginWindow and that has been done in previous loop (line 91-97)

library/src/test/java/com/datatorrent/lib/async/AbstractAsyncProcessorTest.java
```diff
@@ -0,0 +1,145 @@
...
+    Assert.assertEquals(666, outTuples.size());
+    Assert.assertEquals(334, errTuples.size());
+
+    int previous = -1;
+    for (Object outTuple : outTuples) {
+      Assert.assertTrue(outTuple instanceof String);
+      String outTuple1 = (String)outTuple;
+      String[] split = outTuple1.split(";");
+      Assert.assertEquals(2, split.length);
+      int current = Integer.parseInt(split[0]);
```
gauravgopi123, line 118: why not use swtich case?

library/src/test/java/com/datatorrent/lib/async/AbstractAsyncProcessorTest.java
chinmaykolhatkar, line 104: Got it. I should have added the beginWindow also here. Will update.
This loop is just to make sure to wait for all the items to finish processing.

library/src/test/java/com/datatorrent/lib/async/AbstractAsyncProcessorTest.java
chinmaykolhatkar, line 118: Done.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
chinmaykolhatkar: I agree.What would you suggest a good ma number will be. I'll remove the cached pool all together.
And put a limit on that.

chinmaykolhatkar: So here is what I've done:
Min = 1 & Max = 32.
32 is the cap on number of threads enforced by javax Max annotation.

Default is set to 32.

Let me know if some other cap on number of threads need to be there.

vrozov:
IMO, the functionality must be part of the platform (core) and not part of operators library (malhar). It applies to various types of operators like JDBC, S3, HDFS. It should be possible to have any operator configured for synchronous and asynchronious processing.

chinmaykolhatkar:
@vrozov I think that's a great idea. Though the question is whether the operator should have control over checkpointing.
Having an abstract operator async processing gives operator control over checkpointing and how to replay the tuples.

sandeepdeshmukh:
@vrozov : Good idea to absorb this functionality in Apex core.
However, as suggested by Chinmay, this would need a lot broader discussion on the topic and needs to consider many aspects beyond the operator life cycle. Hence, I suggest that we go ahead with this pull request in Malhar and once we have a full fledged solution in Apex core, this can be deprecated.

chinmaykolhatkar:
If there are no further comments, can this pull request be merged?

vrozov:
1. What is the immediate need for this functionality? Are there other malhar operators that depend on the AbstractAsyncProcessor?
2. Why is it necessary to introduce the new operator instead of extending the existing one that provides related functionality?
3. Checkpointing is a joint functionality of the Apex platform and operators. In general, operator should define what state to checkpoint and platform defines how checkpoints are saved and restored.

chinmaykolhatkar:
1. I'm writing a database point query operator which needs async operations. Hence this abstract async processor. There are no operators in malhar as of now which depend on AbstractAsyncProcessor.
2. Can you please point out which operator provides such functionality? I see there is AsyncReconciler. But that waits till commit window and then operate on tuple. This one will start processing right away in async mode. Hence not affecting latency.

vrozov:
1. Is it possible to extend AsyncReconciler to support what is missing by adding more properties?
2. I don't understand how application designers will use AbstractAsyncProcessor operator. Suppose application designer wants to read from HBase instead of RDBMS. Does she need to re-implement all HBase operator functionality if support for asynchronous processing is also a requirement?

chinmaykolhatkar:
1. I would prefer to keep this seperate from AsyncReconciler. Instead of making a single abstract more complicated by adding properties, it looks to me, its better to the 2 types of functionality same.
2. Operator developer will have to extend from this class to have async processing of tuples. Please refer to unit test for the example.

vrozov:
1. If every time new functionality is introduced, a completely new operator is also introduced, we will end up with dozens of operators that expose related functionality. Such approach complicates life for an application developer as it becomes confusing which operator out of those dozens to use or extend. IMO, there should be an attempt to reuse and extend existing functionality first and only in case there are technical reasons not to do so, a brand new operator should be introduced.
2. The value of such abstract operator is limited as it does not allow to reuse existing wide range of input and output connectors and functional operators and requires reimplementation of every connector if async processing is necessary (please correct me if I am wrong).
3. For every tuple it processes it creates at minimum 2 additional objects not counting input and output tuples. For high throughput operators, this will lead to extensive GC.
4. While Apex was designed to adjust to the input rate and either propagate back pressure or elastically scale if configured, AbstractAsyncProcessor is configured with the fixed number of threads and if those threads are not capable of keeping up with the input rate, the queue of tuples to process will grow indefinitely, leading to out of memory errors and huge checkpoint snapshots with no way to recover out of this condition.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
```diff
@@ -0,0 +1,341 @@
...
+   *       in <i>waitingTuples</i> that means they're yet to be notified, hence enqueued again.</li>
+   * </ul>
+   */
+  private void replayTuplesIfRequired()
+  {
+    // Enqueue everything that is still left.
+    // Its left because its either not completed processing OR its not emitted because of ordering restrictions.
+    for (WorkItem item : waitingTuples) {
+      item.taskHandle = executor.submit(new Processor(item));
+      item.processState = State.SUBMITTED;
```
vrozov, line 133: This possibly overwrites state of processed item.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
```diff
@@ -0,0 +1,341 @@
...
+   * This method should be called only from operator thread.
+   *
+   * @param QUEUETUPLE Input tuple of type QUEUETUPLE
+   */
+  protected void enqueueTupleForProcessing(QUEUETUPLE QUEUETUPLE)
+  {
+    WorkItem item = new WorkItem(QUEUETUPLE);
+    waitingTuples.add(item);
+    item.taskHandle = executor.submit(new Processor(item));
+    item.processState = State.SUBMITTED;
```
vrozov: This possibly overwrites state of processed item.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
chinmaykolhatkar, line 133: calling executor.submit after variables are set. This should take care of overrides.

library/src/main/java/com/datatorrent/lib/async/AbstractAsyncProcessor.java
chinmaykolhatkar: calling executor.submit after variables are set. This should take care of overrides.

chinmaykolhatkar:
Hi @vrozov , 

I'm trying to answer your questions as per points. But please let me me know if anything is not correct.
1. The closest/similar functionality that I found in malhar to what AbstractAsyncProcessor provides is AbstractReconciler. But here are the pain points for see with extending AbstractReconciler:
   a. Simplicity - The purpose of reconciler is to process the tuples after they're are committed. The     purpose of async processor is to process the tuples in async and parallel mode. I think they have seperate intent of existence. And off course we can merge the functionality and create something common out of these 2, but simplicity will be lost, as they separately complicated enough.
   b. Configuration - If we merge these 2, then I think the configuration to overall entity would be fairly complicated for one to use. This will beat the purpose of keeping application developer's life simple.
   c. Backward compatibility - AbstractReconciler, I think, is used in quite some places, if we tent to extend reconciler or add functionality of reconciler in async processor, this might beat the backward compatibility.
2. Yes you're correct, this can be used only by new operators. The older operators will have to modify the code to do async processor. But I don't think that much change. One just have to move processing of tuples to processAsync method and where processing is done, enqueue it.
   If you're pointing to moving the async processing of tuples via apex core itself, I agree with you. I'll be happy to be part of that work. Please let me know whether I should log a Jira for that.
   But, I have a usecase where I need to process tuples async. And as a part of that I thought this would be a good addition to library. Hence added here. 
   Do you think we can consume this in malhar-library for now? Once apex supports it, we can deal with what to be done with this operator.
3. & 4. Good catch. I want to propose following changes to take care of 3. & 4.:
   a. The Executors can be configured with the blocking queue of fixed size. Because of blocking nature, the concrete operator trying to enqueue a tuple for processing will be blocked for enqueuing any further tuples unless there is a place available in queue. This way, the buffer server will get the chance to take care of backpressure and also the memory consumption can be controlled.
   b. The 2 objects created are small but we can reuse the objects if need so.
   c. About amount of data that gets checkpointed, we can have a configuration which limits the amount of items that will get checkpointed. With new api beforeCheckpoint, we can control over what will get checkpointed. The beforeCheckpoint will block till the size is under given limit.

Please share your thoughts.

Thanks.

PR #159, opened 2016-01-06
*Review only* N-DelayOperator

sandeshh:
@davidyan74 please review.

This is the basic implementation of N-DelayOperator using WindowDataManager for iteration.


library/src/main/java/com/datatorrent/lib/iteration/nDelayOperator.java
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * This operator is used in iteration and can do n window delay.
+ * Required delay needs to be set in the constructor, default delay is 1.
+  *
+ * @displayName nDelayOperator
+ * @category iteration
+
+ */
+public class nDelayOperator<T> implements Operator.DelayOperator, Operator.CheckpointListener
```
otterc: Sorry for nitpicking but the name is not according to Java Convention of naming classes

otterc: maybe DelayByNOperator?

library/src/main/java/com/datatorrent/lib/iteration/nDelayOperator.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
+    currentWindowId = windowId;
+
+    if ( delay > 1 ) {
+      replay(windowId - delay + 1);
```
pramodin: I don't think operator should be doing this. We should enhance the platform to do the n-delay by incrementing the window id by n for data transmitted in a window. This should not have to be managed within the operator where operator is doing (n-1) and platform is doing +1. Platform should also provide necessary hints for the operator to know about first n windows in the case of a replay. My suggestion is to implement n delay in platform and take this operator after. Also when that happens the operator needs to only save data from (checkpoint-n, checkpoint] window range and not for all windows.

tweise:
I don't see this in core. It is better done in Malhar where you can use other dependencies to implement the state management. It also makes it easier for the user to swap out or modify.

library/src/main/java/com/datatorrent/lib/iteration/nDelayOperator.java
sandeshh: will use that name.

pramodin:
Core will not save the tuples. You won't need to manage n window worth of tuples if core attributes emitted tuples to the right window (+n) when data is emitted. You would still do whatever data saving you need to do in the operator.

library/src/main/java/com/datatorrent/lib/iteration/nDelayOperator.java
```diff
@@ -0,0 +1,186 @@
...
+  @Override
+  public void teardown()
+  {
+    this.windowDataManager.teardown();
+  }
+
+  @Override
+  public void firstWindow()
+  {
+    replay(currentWindowId - delay);
```
davidyan74: We cannot simply subtract from windowId since there is a baseSecond rollover mechanism.  Refer to WindowGenerator.getAheadWindowId()

library/src/main/java/com/datatorrent/lib/iteration/nDelayOperator.java
```diff
@@ -0,0 +1,186 @@
...
+    try {
+      recoveredData = (ArrayList<T>)this.windowDataManager.load(operatorContextId, windowId);
+      if (recoveredData == null) {
+        return;
+      }
+      for ( T tuple : recoveredData) {
+        output.emit(tuple);
+      }
+    } catch (IOException e) {
+      DTThrowable.rethrow(e);
```
davidyan74: rethrow() is deprecated.

library/src/main/java/com/datatorrent/lib/iteration/nDelayOperator.java
```diff
@@ -0,0 +1,186 @@
...
+    windowData.clear();
+  }
+
+  protected void processTuple(T tuple)
+  {
+    windowData.add(tuple);
+
+    if ( delay == 1 ) {
+      output.emit(tuple);
+    }
```
davidyan74: What happens if delay > 1?

library/src/main/java/com/datatorrent/lib/iteration/nDelayOperator.java
davidyan74: We cannot do simple arithmetics on window id because of baseSecond rollover.  See my comment above.

library/src/main/java/com/datatorrent/lib/iteration/nDelayOperator.java
sandeshh: Tuples are just stored in WindowDataManager and replayed during BeginWindow. So it is important not to change delay after the object is constructed, so it can only be set through constructor.

davidyan74: I see.

library/src/main/java/com/datatorrent/lib/iteration/nDelayOperator.java
sandeshh: Making the necessary changes. It also means cleanup logic needs to be handled carefully.

library/src/main/java/com/datatorrent/lib/iteration/nDelayOperator.java
```diff
@@ -0,0 +1,186 @@
...
+  public void checkpointed(long l)
+  {
+
+  }
+
+  @Override
+  public void committed(long windowId)
+  {
+    try {
+      windowDataManager.deleteUpTo(operatorContextId, windowId - delay - 1);
```
davidyan74: windowId arithmetic needs to be changed.

library/src/test/java/com/datatorrent/lib/iteration/nDelayOperatorTest.java
```diff
@@ -0,0 +1,186 @@
...
+      OutputOperator outputOperator = dag.addOperator("output", new OutputOperator());
+
+      nDelayOperator.setWindowDataManager(new WindowDataManager.NoopWindowDataManager());
+
+      dag.addStream("rand_inputToDelay", rand.integer_data, outputOperator.inputDummy);
+      dag.addStream("inputToDelay_toDelay", outputOperator.output, nDelayOperator.input);
+      dag.addStream("DelayOperator_output", nDelayOperator.output, outputOperator.input);
+    }
+  }
+}
```
davidyan74: Can you add a couple of test cases that test the window id baseSecond rollover?

library/src/main/java/com/datatorrent/lib/iteration/DelayByNOperator.java
```diff
@@ -0,0 +1,225 @@
...
+  {
+    init();
+  }
+
+  /*
+  * @param delay set the delay in number of windows for the tuples.
+   */
+  public DelayByNOperator(int delay)
+  {
+    if ( delay < 1 ) {
```
davidyan74, line 74: When the logic is done, please correct the code style so it can be merged after https://github.com/apache/incubator-apex-core/pull/185 is merged.
- no spaces after open parentheses or before close parentheses.
- classes (including unit test classes) must be capitalized.
- (someBoolean == false) should be changed to (!someBoolean)
- all open braces that denote a definition must be in its own line

library/src/main/java/com/datatorrent/lib/iteration/DelayByNOperator.java
```diff
@@ -0,0 +1,225 @@
...
+    try {
+      recoveredData = (ArrayList<T>)this.windowDataManager.load(operatorContextId, windowId);
+      if (recoveredData == null) {
+        return;
+      }
+      for ( T tuple : recoveredData) {
+        output.emit(tuple);
+      }
+    } catch (IOException e) {
+      DTThrowable.rethrow(e);
```
davidyan74, line 144: rethrow is deprecated

davidyan74:
@sandeshh can you issue another PR once your changes are ready?

PR #163, opened 2016-01-06
MLHR-1962 json parser enhancements

shubham-pathak22:
- Enhanced existing JSON Parser.
- Now user can provide json schema against which the incoming tuple would be validated.
- JSON Schema is as per http://json-schema.org/
- Operator has 3 output ports. 
  - **ParsedOutputPort :**  emits incoming JSON as JSONObject. 
  - **Out port:** emits incoming JSON as POJO
  - **Err port:** emits tuples that could not be parsed as a KeyValPair where key is the tuple and value is the reason.
- POJO class needs to be provided if downstream operator connects to port that emits POJO.
- Metrics are provided to give count of incomingTuples, errorTuples, pojos , and parsedObjects (JSONObject)


shubham-pathak22:
@gauravgopi123  Please review.

library/pom.xml
```diff
@@ -315,7 +315,17 @@
       <artifactId>joda-time</artifactId>
       <version>2.9.1</version>
     </dependency>
-
+    <dependency>
```
tweise, line 317: As discussed on dev@, we don't want to add dependencies to malhar-library that aren't broadly applicable.

shubham-pathak22, line 317: I agree, but this case requires it. This is an enhancement to existing operator in malhar-library

chinmaykolhatkar:
There is a test that is failing: DTFileTest.seekDTFile as per travis build.

I checked locally on devel-3 and its passing for me.

Can you please if this is happening because of the dependency added? Seems unlikely, but please do a quick check if possible.

shubham-pathak22:
Thanks Chinmay for pointing it out. Yes, the build fail was happening because of jar conflicts after adding new dependency. I have moved json parser to contrib so that malhar-lib is unaffected.

shubham-pathak22:
Hello, @chinmaykolhatkar , @gauravgopi123, @sandeepdeshmukh , @tushargosavi   please review

library/.gitignore
```diff
@@ -0,0 +1 @@
+/target
```
chinmaykolhatkar: Can you please remove this change?

contrib/src/test/java/com/datatorrent/contrib/parser/JsonParserApplicationTest.java
```diff
@@ -0,0 +1,78 @@
+package com.datatorrent.contrib.parser;
```
chinmaykolhatkar, line 19: License header is missing.

chinmaykolhatkar, line 19: Can you please run mvn apache-rat:check verify -Dlicense.skip=false before uploading the changes?

contrib/src/test/java/com/datatorrent/contrib/parser/JsonParserApplicationTest.java
```diff
@@ -0,0 +1,78 @@
...
+        + "\"tags\": [\"cold\", \"ice\"]," + "\"dimensions\": {" + "\"length\": 7.0," + "\"width\" : 8.0,"
+        + "\"height\": 9.5" + "}," + "\"warehouseLocation\": {" + "\"latitude\": -78.75," + "\"longitude\": 20.4"
+        + "}," + "\"dateOfManufacture\": \"2013/09/29\"," + "\"dateOfExpiry\": \"2013\"" + "}";
+
+    public final transient DefaultOutputPort<String> output = new DefaultOutputPort<String>();
+
+    @Override
+    public void emitTuples()
+    {
+
```
chinmaykolhatkar: Remove space.

contrib/src/test/java/com/datatorrent/contrib/parser/JsonParserTest.java
```diff
@@ -0,0 +1,340 @@
+/**
```
chinmaykolhatkar, line 1: Just a thought..
If I understand correctly, this test won't be run as all the contrib tests are skipped by default.
Can parser tests be allowed to be executed as they do not require external setup?

contrib/src/test/resources/json-parser-schema.json
```diff
@@ -0,0 +1,51 @@
...
+                
+            }, 
+            "dateOfExpiry": {
+                "description": "expiry date",
+                 "type": "string"
+                
+            }            
+        },
+        "required": ["id", "name", "price"]
+}
```
chinmaykolhatkar: New line at the end of the file is missing.

contrib/src/test/java/com/datatorrent/contrib/parser/JsonParserApplicationTest.java
shubham-pathak22, line 19: Sure.

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  @Override
+  public KeyValPair<String, String> processErrorTuple(String input)
+  {
+    throw new UnsupportedOperationException("Not supported");
+  }
+
+  /**
+   *
```
chinmaykolhatkar, line 183: Description missing for getJsonSchema function.

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
```diff
@@ -0,0 +1,187 @@
...
+    } catch (ProcessingException | IOException e) {
+      DTThrowable.wrapIfChecked(e);
+    }
+  }
+
+  @Override
+  public void processTuple(String tuple)
+  {
+    try {
+      if (schema != null) {
```
chinmaykolhatkar, line 132: What happens when schema is null? Is this condition required? 
I see the variable gets assigned in setup method.

Still, if this condition is required, please take care of null case.

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
```diff
@@ -0,0 +1,187 @@
...
+ * @since 3.2.0
+ */
+@InterfaceStability.Evolving
+public class JsonParser extends Parser<String, KeyValPair<String, String>>
+{
+
+  /**
+   * Contents of the schema.Schema is specified as per http://json-schema.org/
+   */
+  private String jsonSchema;
```
chinmaykolhatkar, line 83: You can put @NotNull attribute to jsonSchema variable. After that jsonSchema != null check won't be required in setup.

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
```diff
@@ -0,0 +1,187 @@
...
+      errorTupleCount++;
+      err.emit(new KeyValPair<String, String>(tuple, e.getMessage()));
+      logger.error("Failed to parse json tuple {}, Exception = {} ", tuple, e);
+    }
+  }
+
+  @Override
+  public Object convert(String tuple)
+  {
+    throw new UnsupportedOperationException("Not supported");
```
chinmaykolhatkar, line 172: Can you please return null here instead of throwing UnsupportedOperationException.
I see similar is done in XMLParser.

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  @Override
+  public KeyValPair<String, String> processErrorTuple(String input)
+  {
+    throw new UnsupportedOperationException("Not supported");
```
chinmaykolhatkar, line 178: Again, can you please return null here and log it, instead of throwing exception. This call (if I understand) correctly is coming from super class. If tomorrow super class changes, this should not get impacted.

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
shubham-pathak22, line 132: User may chose to not provide schema if he doesn't wish to perform any validations. In such cases, "jsonSchema" field can be null and hence the check.

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
shubham-pathak22, line 83: jsonSchema can be null.

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
shubham-pathak22, line 172: throwing UnsupportedOperationException is better approach

contrib/src/test/java/com/datatorrent/contrib/parser/JsonParserTest.java
```diff
@@ -0,0 +1,356 @@
...
+  }
+
+  @Test
+  public void TestParserInvalidInputMetricVerification()
+  {
+    Assert.assertEquals(0, parser.parsedOutputCount);
+    String tuple = "{" + "\"id\": 2" + "}";
+    parser.in.process(tuple);
+    Assert.assertEquals(0, parser.parsedOutputCount);
+  }
```
gauravgopi123, line 309: I don't see beginWindow() and endWindow() call on parser in any of the above tests? 
parser.teardown() has been removed all together. I suggested to move it to finished() function call of TestWatcher...

contrib/src/test/java/com/datatorrent/contrib/parser/JsonParserTest.java
```diff
@@ -0,0 +1,356 @@
...
+  
+  @Test
+  public void TestParserMetricResetVerification()
+  {
+    Assert.assertEquals(0, parser.parsedOutputCount);
+    String tuple = "{" + "\"id\": 2," + "\"name\": \"An ice sculpture\"," + "\"price\": 1,"
+        + "\"tags\": [\"cold\", \"ice\"]," + "\"dimensions\": {" + "\"length\": 7.0," + "\"width\" : 8.0,"
+        + "\"height\": 9.5" + "}," + "\"warehouseLocation\": {" + "\"latitude\": -78.75," + "\"longitude\": 20.4"
+        + "}," + "\"dateOfManufacture\": \"2013/09/29\"," + "\"dateOfExpiry\": \"2013\"" + "}";
+    parser.in.process(tuple);
```
gauravgopi123: how is this process happening before beginWindow() call???

contrib/src/test/java/com/datatorrent/contrib/parser/JsonParserTest.java
shubham-pathak22, line 309: Sure @gauravgopi123 . Because beingWindow and endWindow did not have any functionality except resetting autometric which actually defaults to 0, i did not explicitly call them. But i agree with you, the test must reflect function calls in the same order as they are called in application. I have added it now.

shubham-pathak22:
Comments related to test case have been incorporated.

shubham-pathak22:
Hello @gauravgopi123 i have incorporated review comments. Could you take a look ?

shubham-pathak22:
Hi @tushargosavi @chinmaykolhatkar @sandeepdeshmukh , can anyone of you review and merge if there aren't anymore review comments ?

chinmaykolhatkar:
@shubham-pathak22 Please squash the commits.
Also please put Jira ID in the final commit.

contrib/src/test/java/com/datatorrent/contrib/parser/JsonParserTest.java
```diff
@@ -0,0 +1,414 @@
...
+    Assert.assertEquals(Product.class, obj.getClass());
+    Product pojo = (Product)obj;
+    Assert.assertEquals(2, pojo.getId());
+    Assert.assertEquals(1, pojo.getPrice());
+    Assert.assertEquals("An ice sculpture", pojo.getName());
+    Assert.assertEquals(7.0, (double)pojo.getDimensions().get("length"), 0);
+  }
+
+  @Test
+  public void TestParserValidInputMetricVerification()
```
chinmaykolhatkar: Please keep the same format for method name. All the unit tests should start from test\* and not Test*

contrib/src/test/java/com/datatorrent/contrib/parser/JsonParserTest.java
```diff
@@ -0,0 +1,414 @@
...
+    parser.in.process(tuple.getBytes());
+    parser.endWindow();
+    Assert.assertEquals(1, parser.parsedOutputCount);
+    Assert.assertEquals(1, parser.getIncomingTuplesCount());
+    Assert.assertEquals(0, parser.getErrorTupleCount());
+    Assert.assertEquals(1, parser.getEmittedObjectCount());
+  }
+
+  @Test
+  public void TestParserInvalidInputMetricVerification()
```
chinmaykolhatkar: Test case name should start with "test". That's what is followed everywhere.

contrib/src/test/java/com/datatorrent/contrib/parser/JsonParserTest.java
```diff
@@ -0,0 +1,414 @@
...
+    parser.in.process(tuple.getBytes());
+    parser.endWindow();
+    Assert.assertEquals(0, parser.parsedOutputCount);
+    Assert.assertEquals(1, parser.getIncomingTuplesCount());
+    Assert.assertEquals(1, parser.getErrorTupleCount());
+    Assert.assertEquals(0, parser.getEmittedObjectCount());
+  }
+
+  @Test
+  public void TestParserMetricResetVerification()
```
chinmaykolhatkar: Keep consistency in name of the test case.

shubham-pathak22:
Fixed unit test related comments and Squashed commits

chinmaykolhatkar:
Changes looks fine to me.
Will wait for another day for any comments and then merge this to devel-3 and master.

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
```diff
@@ -0,0 +1,241 @@
...
+    if (tuple == null) {
+      if (err.isConnected()) {
+        err.emit(new KeyValPair<String, String>(null, "null tuple"));
+      }
+      errorTupleCount++;
+      return;
+    }
+    String incomingString = new String(tuple);
+    try {
+      if (schema != null) {
```
tushargosavi, line 132: validation is only required if error port is connected, the condition can be schema != null && err.isConnected()

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
```diff
@@ -0,0 +1,241 @@
...
+        parsedOutput.emit(new JSONObject(incomingString));
+        parsedOutputCount++;
+      }
+      if (out.isConnected()) {
+        out.emit(objMapper.readValue(tuple, clazz));
+        emittedObjectCount++;
+      }
+    } catch (JSONException | ProcessingException | IOException e) {
+      errorTupleCount++;
+      err.emit(new KeyValPair<String, String>(incomingString, e.getMessage()));
```
tushargosavi: emit only when error port is connected

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
```diff
@@ -0,0 +1,241 @@
...
+  @Override
+  public void setup(OperatorContext context)
+  {
+    try {
+      if (jsonSchema != null) {
+        JsonSchemaFactory factory = JsonSchemaFactory.byDefault();
+        JsonNode schemaNode = JsonLoader.fromString(jsonSchema);
+        schema = factory.getJsonSchema(schemaNode);
+      }
+      objMapper = new ObjectMapper();
```
tushargosavi, line 113: By default object mapper will check if all fields specified in json are available in the class, if it does not find such mapping it will fail the parsing. you need to relax this condition by setting

```
mapper.configure(DeserializationConfig.Feature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```

This will allow users to specify pojo with less number of fields than what are present in json. This way user can only extract interested fields from json.

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
shubham-pathak22, line 132: Agreed. But errorTupleCount must be incremented either ways. I will add the condition before err.emit is called.

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
shubham-pathak22, line 113: Good catch !. Will do the same and add a test case too.

shubham-pathak22:
Incorporated reivew comments

contrib/src/main/java/com/datatorrent/contrib/parser/JsonParser.java
tushargosavi, line 113: let me know once you have fixed it, I will merge after that.

shubham-pathak22:
I have made the required changes.

shubham-pathak22:
Hi @tushargosavi . I have made the required changes. Can we merge this if there aren't any more review comments ?

chinmaykolhatkar:
@shubham-pathak22 While trying to merge these changes, I got failures in running travis build command (same as the failure shown here).
I tried to run travis build on devel-3 which is compiling successfully.
The failure is happening in logstream application.

Can you please check what is the issue?

shubham-pathak22:
Hi @chinmaykolhatkar i have fixed the issue.

PR #164, opened 2016-01-06
APEXMALHAR-1950 #resolve #comment added stable annotation to semantic…

otterc:
… version inclusion list


otterc:
@tweise please review and merge

PR #165, opened 2016-01-07
APEXMALHAR-1950 #resolve #comment includes :

otterc:
1. removed @Stable annotation from classes
2. add @Evolving to classes and packages
3. change japi configuration to just exclude classes based on the annotation

### DIFF
--- a/contrib/src/main/java/com/datatorrent/contrib/accumulo/package-info.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/accumulo/package-info.java
@@ -1,0 +1,20 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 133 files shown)


otterc:
@PramodSSImmaneni @tweise please review
Currently exclusion of classes/packages based on annotation in japicmp plugin is broken. I have already informed japicmp developer about it and will be creating a detailed issue.

I went ahead and reverted the change that marks classes as @Stable. I have annotated rest of the classes and packages as @Evolving which sets the intention that they can be modified with minor version.

Meanwhile I will work to get in the japicmp annotation bug fix.

tweise:
If the annotation based exclusion isn't working, then let's hold this until it is fixed. Perhaps for the 3.3.0 release, make a separate change that restores the name based exclusion?

tweise:
Actually I'm OK merging it if we agree that's what we want once annotations are working. Let's wait for @PramodSSImmaneni to comment.

tweise:
@PramodSSImmaneni waiting for your comment so that we can proceed and cut release candidate.

otterc:
@tweise @PramodSSImmaneni please see this comment.
https://github.com/siom79/japicmp/issues/88#issuecomment-165711360

In this version we need to mark the classes as @Evolving so that in the next version we can add the @Evolving annotation as excluded.

pramodin:
@chandnisingh @tweise I am fine with going ahead with this change now and waiting for the fix from japicmp

PR #168, opened 2016-01-13
Skip all the modules that should not be deployed to central (separate…

tweise:
… from -Pall-modules).

### DIFF
--- a/samples/pom.xml
+++ b/samples/pom.xml
@@ -34,6 +34,10 @@
 
   <name>Apache Apex Malhar (incubating) Samples</name>
 
+  <properties>
+    <maven.deploy.skip>true</maven.deploy.skip>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/apps/logstream/pom.xml
+++ b/apps/logstream/pom.xml
@@ -32,7 +32,7 @@
   <packaging>jar</packaging>
 
   <properties>
-    <maven.deploy.skip>false</maven.deploy.skip>
+    <maven.deploy.skip>true</maven.deploy.skip>
     <skipTests>false</skipTests>
     <semver.plugin.skip>true</semver.plugin.skip>
   </properties>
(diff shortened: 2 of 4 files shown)


tweise:
@davidyan74 @vrozov please merge

samples/pom.xml
```diff
@@ -34,6 +34,11 @@
 
   <name>Apache Apex Malhar (incubating) Samples</name>
 
+  <properties>
+    <semver.plugin.skip>true</semver.plugin.skip>
```
vrozov: Is semver plugin enabled for samples? japicmp is part of plugin management and should not be invoked unless explicitly enabled.

tweise: I guess not, but this is how I found it in another module. However, why would be require a per module plugin configuration for something that by default should apply to all modules?

vrozov: The common to all modules configuration should be part of plugin management and each module may modify it. It is not clear to me why it is necessary to provide semver plugin configuration settings here as semver plugin does not seem to be enabled for the samples module.

tweise: No issue with the plugin management part. But I do not see the point of repeating the plugin like this:

```
      <plugin>
        <groupId>com.github.siom79.japicmp</groupId>
        <artifactId>japicmp-maven-plugin</artifactId>
      </plugin>
```

in every module. Instead, I would prefer to see this in the malhar parent pom only and use the property that turns it off for those modules that should be exempt. That's how the deploy plugin is setup already.

vrozov: It depends on the percentage of modules that semver should be enabled and disabled for. In Malhar, there are more modules where it needs to be disabled, it seems that repeating 

```
      <plugin>
        <groupId>com.github.siom79.japicmp</groupId>
        <artifactId>japicmp-maven-plugin</artifactId>
      </plugin>
```

only for modules where semver needs to be enabled will be more applicable in this case.

tweise:
I removed the property as it is not related to this PR. I also think we should revisit how we manage semver execution, making it property driven is less verbose and gives more flexibility through command line and profile properties.

PR #170, opened 2016-01-13
APEXMALHAR-1972 Added support for Expression Evaluator

chinmaykolhatkar:
Added support for Expression Evaluator for quasi-Java Expression Language.

Design details are all given in Jira.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/expressions/ConversionUtils.java
+++ b/library/src/main/java/com/datatorrent/lib/expressions/ConversionUtils.java
@@ -1,0 +1,204 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.lib.expressions;
+
+import org.apache.commons.lang3.math.NumberUtils;
+
+public class ConversionUtils
+{
+  public static int toInt(double a)
+  {
+    return new Double(a).intValue();
+  }
+
+  public static int toInt(float a)
+  {
+    return new Float(a).intValue();
+  }
+
+  public static int toInt(long a)
+  {
+    return new Long(a).intValue();
+  }
+
+  public static int toInt(byte a)
+  {
+    return new Byte(a).intValue();
+  }
+
+  public static int toInt(short a)
+  {
+    return new Short(a).intValue();
+  }
+
+  public static int toInt(String a)
+  {
+    return NumberUtils.toInt(a);
+  }
+
+  public static long toLong(float a)
+  {
+    return new Float(a).longValue();
+  }
+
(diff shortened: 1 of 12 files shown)


chinmaykolhatkar:
@tushargosavi @sandeepdeshmukh @gauravgopi123 Please review.

library/src/main/java/com/datatorrent/lib/expressions/ConversionUtils.java
```diff
@@ -0,0 +1,204 @@
...
+  {
+    return new Byte(a).intValue();
+  }
+
+  public static int toInt(short a)
+  {
+    return new Short(a).intValue();
+  }
+
+  public static int toInt(String a)
```
shubham-pathak22, line 79: can we add java docs ? NumberUtils returns 0 for null , blank or incorrect string.

library/src/main/java/com/datatorrent/lib/expressions/ExpressionEvaluator.java
```diff
@@ -0,0 +1,440 @@
...
+ *
+ * ExpressionEvaluator also allows you to set extra imports that needs to be added over default is java.lang.*
+ *
+ * ExpressionEvaluator needs to be configured with following configurations as minimal configuration:
+ * 1. Mapping of input object place holders to it corresponding types.
+ * This can be done with setInputObjectPlaceholders method.
+ * 2. Return type of of expression evaluation.
+ * 3. Expression to be evaluated. This is a standard java expression except for referencing the variable inside object JEL syntax needs to be used i.e. ${objectPlaceHolder.varName}
+ *
+ * Expression cab be defined as per JSL 7 syntax:
```
shubham-pathak22: can

library/src/main/java/com/datatorrent/lib/expressions/ExpressionEvaluator.java
```diff
@@ -0,0 +1,440 @@
...
+  @SuppressWarnings({"unchecked"})
+  private <O> Expression<O> createExecutableCode(String expression, Class<?> returnType,
+      boolean containsReturnStatement)
+  {
+    if (this.placeholderClassMapping.size() <= 0) {
+      throw new RuntimeException("Mapping needs to be provided.");
+    }
+
+    Pattern entry = Pattern.compile(variablePlaceholderPattern);
+    Matcher matcher = entry.matcher(expression);
```
shubham-pathak22: Should we consider throwing runtime exception if expression is null / blank ?

chinmaykolhatkar: Considering this will be run in some operator, when the expression variable is encountered, it'll anyway throw exception. 
Do you see any advantage in adding this check before that?

vrozov:
The language for expressions is an important decision that has a significant impact on malhar library and other operators/modules. Prior to opening pull request please initiate discussion on dev@apex to gather community opinion on language to support and use cases.

chinmaykolhatkar:
@vrozov The details of expression language are there on Jira ticket.
Can we discuss there? The Jira comments are anyway forwarded to dev@apex.

vrozov:
@chinmaykolhatkar IMO discussion on the pull request is not sufficient for the decision on the language to support to make as the discussion on the pull request mostly focuses on the existing implementation. There is a bigger decision for the community to make.

chinmaykolhatkar:
Hi Vlad,

I was asking whether discussion can be done on jira.

~ Chinmay

vrozov:
I would prefer broader discussion on the expression language to support. It can be done on JIRA as well, but please have a separate from automatically generated email sent to dev@apex. The goal is to gather as much response/opinions as possible.

chinmaykolhatkar:
Hi Vlad,

I agree that we can and should have a discussion but I had to provide a
starting point. So here is what I suggest:
1. Jira has all the details about this expression language and its
implementation.
2. One can take that as a starting point and express their opinion for
current  language.

I'm all open to suggestion and will be happy to modify the language
semantics as per discussion. But because lot of content is already there on
jira, I would prefer the discussion to be there. Plus this way the
discussion can also be contextualized.

~ Chinmay

PR #171, opened 2016-01-14
APEXMALHAR-1975 Added group id to all malhar app packages

davidyan74:


davidyan74:
@vrozov please merge when you get a chance

PR #172, opened 2016-01-14
APEXMALHAR-1977: Add metrics to jdbc input operator

DT-Priyanka:


DT-Priyanka:
@bhupeshchawda can you please review?

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInputOperator.java
```diff
@@ -97,6 +97,9 @@
 
   protected int pageNumber;
 
+  @AutoMetric
+  protected long tuplesRead;
```
bhupeshchawda, line 101: can we make it private?

bhupeshchawda:
Change looks okay to me. Will merge this tomorrow if there are no other comments.

davidyan74:
Remember to correct the spelling error (i.e. matrics) of the JIRA and the commit message before merging.

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInputOperator.java
DT-Priyanka, line 101: As suggested by Gaurav  keeping it protected so that the subclass can update the metrics if required.

PR #175, opened 2016-01-22
APEXMALHAR-1982: CassandraInputOperator query setter fails with null …

DT-Priyanka:
…pointer exception

### DIFF
--- a/contrib/src/test/java/com/datatorrent/contrib/cassandra/CassandraOperatorTest.java
+++ b/contrib/src/test/java/com/datatorrent/contrib/cassandra/CassandraOperatorTest.java
@@ -293,8 +293,8 @@ public class CassandraOperatorTest
 
     TestInputOperator inputOperator = new TestInputOperator();
     inputOperator.setStore(store);
-    inputOperator.setTablename(TABLE_NAME_INPUT);
     inputOperator.setQuery(query1);
+    inputOperator.setTablename(TABLE_NAME_INPUT);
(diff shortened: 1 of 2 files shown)


DT-Priyanka:
@chandnisingh  please review.

contrib/src/test/java/com/datatorrent/contrib/cassandra/CassandraOperatorTest.java
```diff
@@ -293,8 +293,8 @@ public void testCassandraInputOperator()
 
     TestInputOperator inputOperator = new TestInputOperator();
     inputOperator.setStore(store);
-    inputOperator.setTablename(TABLE_NAME_INPUT);
     inputOperator.setQuery(query1);
+    inputOperator.setTablename(TABLE_NAME_INPUT);
```
bhupeshchawda, line 297: Please comment on why this change was needed, or undo this change as it is just reordering of setters on the operator

DT-Priyanka, line 297: Previously, query was dependent on "tableName" variable. As the order of setting variables was tableName followed by query here the test case always passed. To simulate situation where platform could set tablename after query (there is no guarantee in which order variables are set) I did this change.

bhupeshchawda:
Change looks good. Will merge this if there are no other comments.

PR #180, opened 2016-01-27
MLHR-1919 #resolve #comment add dimension schema to malhar

brightchen:


brightchen:

ilooner:
@brightchen Thanks for doing this move. My only comment is that the package for the classes should start with org.apache.apex.malhar instead of com.datatorrent.lib

That's it. Thanks!

tweise:
@ilooner Agreed. Can you please pull in the changes and then we do the rename in a separate PR.

tweise:
@ilooner can this be merged?

PR #181, opened 2016-01-28
APEXMALHAR-1979 Generate compiled class on the fly

chinmaykolhatkar:
@tushargosavi @chandnisingh Please review.

### DIFF
--- a/library/pom.xml
+++ b/library/pom.xml
@@ -314,7 +314,12 @@
       <artifactId>joda-time</artifactId>
       <version>2.9.1</version>
     </dependency>
-
+    <!-- use shaded asm library to avoid conflict -->
+    <dependency>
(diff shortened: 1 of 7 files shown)


pramodin:
A general question, do you see this being used instead of generating the code for the schema classes. What situations it would be better to do it this way. Having schema classes also helps somebody who is mixing custom operators with generic available operators to use the schema classes without having to learn this methodology.

chinmaykolhatkar:
So here are couple of usecases:
1. This utility allows the user of application/operator to provide schema in a non-java fashion
2. Allows user to specify new fields on existing schema.
3. Allows user to provide joining of 2 schema via simple JSON type representation of object.

Basically, a non-java developer can leverage features in operators or application or platform which lets him/her specify the schema using simple representation.

PR #185, opened 2016-02-10
APEXMALHAR-1993 & 1994: Fix for offset storage and reporting in kafka input operator

pramodin:


pramodin:
@tweise @siyuanh please see, these issues are blocking an application deployment

gauravgopi123:
@PramodSSImmaneni : Can you please add test case for this?

pramodin:
@gauravgopi123 The testss in OffsetManagerTest cover it

pramodin:
The test in OffsetManager may have succeeded earlier if timing was right, it will always succeed now. The tests weren't being called because of this being in contrib.

PR #188, opened 2016-02-10
APEXMALHAR-1948: CassandraStore Should Allow You To Specify Protocol Version.

DT-Priyanka:


ilooner:
@DT-Priyanka was this fix tested in an app on the cluster? I believe the hadoop jars pull in an older version of Guava, when the app is launched on cluster. Also please squash commits.

DT-Priyanka:
@ilooner I haven't yet tested in on cluster. It will take time as our cluster doesn't have cassandra on it yet. This was review only, to see if code changes are fine. Once I test and squash changes would update review request accordingly.
Sorry forgot to add "review only"

contrib/src/main/java/com/datatorrent/contrib/cassandra/CassandraStore.java
```diff
@@ -115,21 +130,35 @@ public Session getSession() {
...
     try {
-
-      cluster = Cluster.builder()
-          .addContactPoint(node).withCredentials(userName, password).build();
-    }
-    catch (DriverException ex) {
+      if (protocolVersion != null && protocolVersion.length() != 0) {
+        ProtocolVersion version = getCassandraProtocolVersion();
+        cluster = Cluster.builder().addContactPoint(node).withCredentials(userName, password).withProtocolVersion(version).build();
+      } else {
```
bhupeshchawda, line 139: which version does it use in the default case? Should we add a specific default? Like for example, the latest version?

contrib/pom.xml
```diff
@@ -477,7 +477,14 @@
     <dependency>
       <groupId>com.datastax.cassandra</groupId>
       <artifactId>cassandra-driver-core</artifactId>
-      <version>2.0.2</version>
+      <version>2.1.8</version>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
```
bhupeshchawda, line 493: Why is this needed?

bhupeshchawda:
Can you add unit tests if possible?

contrib/src/main/java/com/datatorrent/contrib/cassandra/CassandraStore.java
DT-Priyanka, line 139: Default depends on cassandra driver version, with 2.1.x it is V2.

contrib/pom.xml
DT-Priyanka, line 493: protocol version 3 was introduced in New in Java driver 2.1.2, that's why upgrading cassandra driver version.

DT-Priyanka:
Added unit test as well as tested it on server.

bhupeshchawda:
I am fine with the changes. I'll merge this in a couple of days if there are no other comments.

PR #189, opened 2016-02-14
APEXMALHAR-1995 Upgrade Apex Core dependency to 3.3.0

tweise:


tweise:
@vrozov @chandnisingh note checkstyle related changes, setup is global and some work needs to be done separately to make it more useful.

@PramodSSImmaneni merge when discussion on dev@ is concluded

pramodin:
Why did the maxAllowedViolations jump?

tweise:
Not sure, hence asked Vlad or Chandni to streamline the checkstyle
configuration.

## 

sent from mobile

vrozov:
There was a change on the core side between 3.2 and 3.3. In 3.3 tests are included into check style. Before 3.3 they were excluded.

IMO, it may be better to have checkstyle config in malhar not to depend on core as any changes in core will not be visible until malhar upgrades core dependency.

pramodin:
Is it an extension today?

tweise:
Probably there are larger issues with the checkstyle setup in Malhar. Reopened https://issues.apache.org/jira/browse/APEXMALHAR-1859

PR #190, opened 2016-02-20
APEXMALHAR-1859 tracking violation per malhar module

otterc:


tweise:
Should we include the kafka module?

pom.xml
```diff
@@ -90,13 +90,6 @@
           </excludes>
         </configuration>
       </plugin>
-      <plugin>
```
tweise, line 93: Does this mean checkstyle is turned off by default? Wouldn't it be better to have it default to no violations with console output?

otterc:
Addressed review comments

otterc:
Squashed the changes

PR #191, opened 2016-02-20
APEXMALHAR-1998 Set test log level to INFO.

tweise:


tweise:
@davidyan74 please merge, this reduces the logs generated by the build.
@vrozov We see all builds getting killed in the Kafka module. Could this be related to https://issues.apache.org/jira/browse/APEXCORE-335 ?

vrozov:
Yes, it may be related. Travis limits VM size, and https://issues.apache.org/jira/browse/APEXCORE-335 fixes resource leak. There may be other reasons why builds are killed.

davidyan74:
Since this change is mainly for Travis and developers generally want the logging level to be DEBUG, can we set the logging level just for Travis to be INFO and leave the default logging level to be DEBUG?

tweise:
Need to double check whether the Kafka test fails because of log overflow or due to memory limit in Travis environment.

tweise:
Confirmed that build fails in kafka module without excessive logging also.

otterc:
@tweise  this branch has conflicts. Can you please rebase?

tweise:
@chandnisingh rebased and also changed to set INFO threshold in Travis only.

PR #192, opened 2016-02-22
SPOI-7087 #resolve #comment Add JDBC dimension output operator to Malhar

brightchen:


brightchen:

checkstyle-suppressions.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0"?>
```
tweise, line 1: This should not show up as part of this PR. Please rebase.

library/src/main/java/com/datatorrent/lib/db/jdbc/JDBCDimensionalOutputOperator.java
```diff
@@ -0,0 +1,462 @@
...
+import com.datatorrent.lib.appdata.schemas.DimensionalConfigurationSchema;
+import com.datatorrent.lib.appdata.schemas.FieldsDescriptor;
+import com.datatorrent.lib.appdata.schemas.Type;
+import com.datatorrent.lib.db.AbstractPassThruTransactionableStoreOutputOperator;
+import com.datatorrent.lib.dimensions.DimensionsDescriptor;
+import com.datatorrent.lib.dimensions.DimensionsEvent.Aggregate;
+import com.datatorrent.lib.dimensions.DimensionsEvent.EventKey;
+import com.datatorrent.lib.dimensions.aggregator.AggregatorRegistry;
+
+import it.unimi.dsi.fastutil.ints.Int2ObjectMap;
```
DT-Priyanka, line 50: Shouldn't this import listed before datatorrent imports?

library/src/main/java/com/datatorrent/lib/db/jdbc/JDBCDimensionalOutputOperator.java
```diff
@@ -0,0 +1,462 @@
...
+        } catch (SQLException ex) {
+          throw new RuntimeException(ex);
+        }
+      }
+    }
+  }
+
+  private String buildStatement(String tableName, List<String> keyNames, List<String> aggregateNames)
+  {
+    LOG.info("building statement");
```
DT-Priyanka, line 147: Can we have this log at DEBUG level?

library/src/main/java/com/datatorrent/lib/db/jdbc/JDBCDimensionalOutputOperator.java
```diff
@@ -0,0 +1,462 @@
...
+    sb.append(") ON DUPLICATE KEY UPDATE ");
+
+    addOnDuplicate(sb, aggregateNames);
+
+    return sb.toString();
+  }
+
+  private void addOnDuplicate(StringBuilder sb, List<String> names)
+  {
+    LOG.info("add Duplicate");
```
DT-Priyanka, line 178: Can we have this and similar logs at DEBUG level?

library/src/main/java/com/datatorrent/lib/db/jdbc/JDBCDimensionalOutputOperator.java
```diff
@@ -0,0 +1,462 @@
...
+      sb.append(",");
+    }
+  }
+
+  private void addList(StringBuilder sb, List<String> names)
+  {
+    for (int index = 0;; index++) {
+      sb.append(names.get(index));
+
+      if (index == names.size() - 1) {
```
DT-Priyanka, line 201: why not use condition in for loop itself, that would be easy to read.

library/src/main/java/com/datatorrent/lib/db/jdbc/JDBCDimensionalOutputOperator.java
```diff
@@ -0,0 +1,462 @@
...
+   * <li>The value is the name of the output Mysql table</li>
+   * </ol>
+   * 
+   * @param tableNames
+   *          The table names that corresponds to the dimensions combinations
+   *          specified in your {@link DimensionalConfigurationSchema}.
+   */
+  public void setTableNames(Map<Integer, Map<String, String>> tableNames)
+  {
+    this.tableNames = Preconditions.checkNotNull(tableNames);
```
DT-Priyanka, line 250: Why do we need NotNull check here, there is already NotNull annotation on tableNames.
Does this Preconditions.checkNotNull checks for each entry inside the Map for null values?

library/src/main/java/com/datatorrent/lib/db/jdbc/JDBCDimensionalOutputOperator.java
```diff
@@ -0,0 +1,462 @@
...
+   * <li>The second key is the name of an aggregation being performed for that
+   * dimensions combination.</li>
+   * <li>The value is the name of the output Mysql table</li>
+   * </ol>
+   * 
+   * @param tableNames
+   *          The table names that corresponds to the dimensions combinations
+   *          specified in your {@link DimensionalConfigurationSchema}.
+   */
+  public void setTableNames(Map<Integer, Map<String, String>> tableNames)
```
DT-Priyanka, line 248: Why don't we have getters for this and other properties?

library/src/main/java/com/datatorrent/lib/db/jdbc/JDBCDimensionalOutputOperator.java
```diff
@@ -0,0 +1,462 @@
...
+   *
+   * @param batchSize
+   *          size of a batch
+   */
+  public void setBatchSize(int batchSize)
+  {
+    this.batchSize = batchSize;
+  }
+
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractJdbcTransactionableOutputOperator.class);
```
DT-Priyanka, line 461: The class name should be "JDBCDimensionalOutputOperator" to get log messages with right class name.

PR #199, opened 2016-02-24
APEXMALHAR-2003 NPE in blockMetaDataIterator after recovery

otterc:


otterc:
@bhupeshchawda @vrozov please review and merge

bhupeshchawda:
@chandnisingh can we create a unit test for this bug?

PR #201, opened 2016-02-25
APEXMALHAR-2003 NPE in blockMetaDataIterator after recovery

otterc:


otterc:
https://github.com/apache/incubator-apex-malhar/pull/198

PR #204, opened 2016-02-27
APEXMALHAR-1965 Utility classes for Write Ahead Log support

tushargosavi:


tweise:
@chandnisingh please review

tweise:
@chandnisingh any update?

otterc:
I will not be able to review this anytime soon.  If it's urgent someone else could take it up

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWal.java
```diff
@@ -0,0 +1,178 @@
...
+ * WAL implementation which allows writing entries to single file.
+ * the pointer type is the offset in the file. The entry is serialized into byte array and
+ * first length of the entry is written followed by the serialized bytes.
+ * @param <T>
+ */
+public class FSWal<T> implements WAL<T, Long>
+{
+  transient long bucketKey;
+  private transient String name;
+  private transient FileAccess fa;
```
otterc, line 42: These fields are all transient.  Is this intended?

library/src/main/java/org/apache/apex/malhar/lib/wal/WAL.java
```diff
@@ -0,0 +1,130 @@
...
+   * Provides iterator like interface to read entries from the WAL.
+   * @param <T> type of WAL entries
+   * @param <P> type of Pointer in the WAL
+   */
+  interface WALReader<T, P> extends Closeable
+  {
+    /**
+     * Close WAL after read.
+     *
+     * @param offset seek offset.
```
otterc, line 49: there isn't any offset argument in close method

library/src/test/java/org/apache/apex/malhar/lib/wal/WALTest.java
```diff
@@ -0,0 +1,236 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.commons.io.FileUtils;
+
+import com.google.common.collect.Sets;
+
+import com.datatorrent.lib.fileaccess.FileAccessFSImpl;
+
+public class WALTest
```
otterc, line 37: If this is test for FSWal can you please rename to that?

otterc:
There is some inconsistency in FSWal and RollingFSWal. All the members in FSWal are made transient however in RollingFSWal are non-transient. 
If these WAL implementations encapsulate configurations for eg. a wal entry serializer/deserializer, base path, etc then IMO it is better that WAL is made non-transient part of the encompassing operator and its properties can be set through nested config instead of duplicating the same properties in operator and then re-instantiating the WAL. This is because multiple operators/utilities maybe needing this WAL and then they all will have to re-declare such configurations.

otterc:
Also please mark the package as @Evolving

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWal.java
```diff
@@ -0,0 +1,178 @@
...
+  private transient WAL.Serde<T> serde;
+
+  public FSWal(FileAccess fa, WAL.Serde<T> serde, long bucketKey, String name)
```
ishark, line 45: Does it make sense to have WAL with simple File APIs instead of FileAccess? That would make it generic to use. Also, can bucketKey be a string instead of long?

otterc, line 45: I agree with Isha on using simple File API.
I don't think there should be a bucket key here. IMO I was thinking of a WAL interface where you provide a path of the WAL and it automatically create part files as and when a file is complete (configured by size)

tweise, line 45: Another consideration for the WAL interface is to use backend other than FS, for example Kafka.

otterc, line 45: My earlier comment just pertains to FSWal. I think bucket key is leftover from HDHT implementation of File System Wal.

otterc:
As I made some changes (locally so far), I realize that we do not need a RollingFSWal extension of FSWal.  In most of the use-cases FSWal will always be rolling.

IMO an FSWAL should have a property rolling. This rolling is size based now and later we can add time based as well (later). This is just like the AbstractFileOutputOperator we have.

I prefer composition here instead of inheritance because adding time based roll-over in future will create problems (and more implementations will lead to confusion) if we want to provide different combinations- no roll-over, only size based roll over, only time based roll over, time + size roll over.

tweise:
+1 in theory this should be similar to the output operator?

otterc:
Yes it should be similar to it - should write to tmp file which get finalized when a roll-over happens.

tweise:
Any thoughts how we take this to the finish line?

library/src/main/java/org/apache/apex/malhar/lib/wal/WAL.java
```diff
@@ -0,0 +1,130 @@
...
+    void flush() throws IOException;
+
+    /**
+     * Returns size of the WAL, last part of the log may not be persisted on disk.
+     * In case of file backed WAL this will be the size of file, in case of kafka
+     * like log, this will be similar to the message offset.
+     *
+     * @return The log size
+     */
+    P getOffset();
```
otterc, line 117: @tushargosavi the javadoc is not clear here. Can you please tell me the intention of this method? I have changed the implementation of FsWALWriter and I am unable to understand  the significance of this method.

otterc:
Here is the work in progress:
https://github.com/chandnisingh/incubator-apex-malhar/tree/APEXMALHAR-1965/library/src/main/java/org/apache/apex/malhar/lib/wal

otterc:
I have few important questions based on the current WAL abstraction that needs to provide a WAL Reader and WAL Writer:

1) The FSWalWriter will write to temporary files (to avoid lease exceptions). These temp files will be renamed to actual files when a file gets rotated. 

Until a file gets rotated, FSWalReader will have nothing to return. Is this a problem?

Also I don't understand if the same container which has a WAL will use both the WalReader and WalWriter or if different containers will use them?

2) The documentation of `advance()` and `get()` is a little vague. WAL is continuously  being appended to. So there be a time when `advance()` returned false and `get()` returns true. However when new data is appended to the file, the reader can resume iterating.

library/src/main/java/org/apache/apex/malhar/lib/wal/WAL.java
tweise, line 117: @chandnisingh I see it as getSize() used in HDHT for rollover.

otterc, line 117: In the changes I have made roll-over is taken by the FileSystemWAL automatically so I have removed this method from the api of WAL Writer. I think rolling over is specific to File System Wal. 
I
 can add it back but if it make sense for other types of Writers.

tweise:
@chandnisingh The scenario of a reader in a different container applies to recovery. Yes, it will need to read the temporary file, just as we do in the output operator.

tushargosavi:
Closed through PR #242

PR #207, opened 2016-03-09
APEXMALHAR-2008: Create HDFS File Reader module

DT-Priyanka:
Code to add HDFS file reader module. 
1. The module reads file/list of files (directory is also accepted) and emit the file blocks. 
2. The module can be configured to emit blocks in order or out of order.
3. Module reads file blocks in parallel. The number of parallel readers is configurable, if not configured it will increase or decrease readers dynamically as per input data rate.

Also updated code of FileSplitterInput to add some improvements:
1. Tracking last file reference times of each folder differently, to avoid duplicates (duplicates could be due to same relative paths of multiple files/sub dir)
2. Small improvements in code.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/block/BlockReader.java
+++ b/library/src/main/java/com/datatorrent/lib/io/block/BlockReader.java
@@ -1,0 +1,66 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.lib.io.block;
+
+import java.io.IOException;
+import java.net.URI;
+
+import org.apache.hadoop.fs.FileSystem;
+
+import com.datatorrent.api.AutoMetric;
+
+/**
+ * BlockReader extends {@link FSSliceReader} to accept case insensitive uri
+ */
+public class BlockReader extends FSSliceReader
+{
+  @AutoMetric
+  private long bytesRead;
+
+  protected String uri;
+
+  @Override
+  public void beginWindow(long windowId)
+  {
+    super.beginWindow(windowId);
+    bytesRead = 0;
+  }
+
+  @Override
+  protected FileSystem getFSInstance() throws IOException
+  {
+    return FileSystem.newInstance(URI.create(uri), configuration);
+  }
+
+  /**
+   * Sets the uri
+   *
+   * @param uri of form hdfs://hostname:port/path/to/input
+   */
+  public void setUri(String uri)
+  {
+    this.uri = uri;
+  }
+
+  public String getUri()
+  {
+    return uri;
+  }
+
+}
--- a/library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
@@ -292,6 +292,9 @@ public class FileSplitterInput extends AbstractFileSplitter implements InputOper
     private transient ScannedFileInfo lastScannedInfo;
     private transient int numDiscoveredPerIteration;
 
+    @NotNull
+    protected final Map<String, Map<String, Long>> inputDirTolastModifiedTimes;
+
     public TimeBasedDirectoryScanner()
     {
       recursive = true;
@@ -301,6 +304,7 @@ public class FileSplitterInput extends AbstractFileSplitter implements InputOper
       discoveredFiles = new LinkedBlockingDeque<>();
       atomicThrowable = new AtomicReference<>();
       ignoredFiles = Sets.newHashSet();
+      inputDirTolastModifiedTimes = Maps.newHashMap();
     }
 
     @Override
@@ -360,7 +364,9 @@ public class FileSplitterInput extends AbstractFileSplitter implements InputOper
             lastScannedInfo = null;
             numDiscoveredPerIteration = 0;
             for (String afile : files) {
-              scan(new Path(afile), null);
+              Map<String, Long> lastModifiedTimesForInputDir;
+              lastModifiedTimesForInputDir = getLastModifiedTimeMap(afile);
+              scan(new Path(afile), null, lastModifiedTimesForInputDir);
             }
             scanIterationComplete();
           } else {
@@ -375,6 +381,15 @@ public class FileSplitterInput extends AbstractFileSplitter implements InputOper
       }
     }
 
+    private Map<String, Long> getLastModifiedTimeMap(String key)
+    {
+      if (inputDirTolastModifiedTimes.get(key) == null) {
+        Map<String, Long> modifiedTimeMap = Maps.newHashMap();
+        inputDirTolastModifiedTimes.put(key, modifiedTimeMap);
+      }
+      return inputDirTolastModifiedTimes.get(key);
+    }
+
     /**
      * Operations that need to be done once a scan is complete.
      */
@@ -386,6 +401,13 @@ public class FileSplitterInput extends AbstractFileSplitter implements InputOper
 
     protected void scan(@NotNull Path filePath, Path rootPath)
     {
+      Map<String, Long> lastModifiedTimesForInputDir;
+      lastModifiedTimesForInputDir = getLastModifiedTimeMap(filePath.toUri().getPath());
+      scan(filePath, rootPath, lastModifiedTimesForInputDir);
+    }
+
+    private void scan(Path filePath, Path rootPath, Map<String, Long> lastModifiedTimesForInputDir)
+    {
       try {
         FileStatus parentStatus = fs.getFileStatus(filePath);
         String parentPathStr = filePath.toUri().getPath();
@@ -393,27 +415,22 @@ public class FileSplitterInput extends AbstractFileSplitter implements InputOper
         LOG.debug("scan {}", parentPathStr);
 
         FileStatus[] childStatuses = fs.listStatus(filePath);
-        for (FileStatus status : childStatuses) {
-          Path childPath = status.getPath();
-          ScannedFileInfo info = createScannedFileInfo(filePath, parentStatus, childPath, status, rootPath);
 
-          if (skipFile(childPath, status.getModificationTime(), referenceTimes.get(info.getFilePath()))) {
-            continue;
-          }
+        if (childStatuses.length == 0 && rootPath == null && lastModifiedTimesForInputDir.get(parentPathStr) == null) { // empty input directory copy as is
+          ScannedFileInfo info = new ScannedFileInfo(null, filePath.toString(), parentStatus.getModificationTime());
+          processDiscoveredFile(info);
+          lastModifiedTimesForInputDir.put(parentPathStr, parentStatus.getModificationTime());
+        }
 
-          if (status.isDirectory()) {
-            if (recursive) {
-              scan(childPath, rootPath == null ? parentStatus.getPath() : rootPath);
-            }
-          }
-
+        for (FileStatus childStatus : childStatuses) {
+          Path childPath = childStatus.getPath();
           String childPathStr = childPath.toUri().getPath();
-          if (ignoredFiles.contains(childPathStr)) {
-            continue;
-          }
-          if (acceptFile(childPathStr)) {
-            LOG.debug("found {}", childPathStr);
-            processDiscoveredFile(info);
+
+          if (childStatus.isDirectory() && isRecursive()) {
+            addToDiscoveredFiles(rootPath, parentStatus, childStatus, lastModifiedTimesForInputDir);
+            scan(childPath, rootPath == null ? parentStatus.getPath() : rootPath, lastModifiedTimesForInputDir);
+          } else if (acceptFile(childPathStr)) {
+            addToDiscoveredFiles(rootPath, parentStatus, childStatus, lastModifiedTimesForInputDir);
           } else {
             // don't look at it again
             ignoredFiles.add(childPathStr);
@@ -424,6 +441,31 @@ public class FileSplitterInput extends AbstractFileSplitter implements InputOper
       } catch (IOException e) {
         throw new RuntimeException("listing files", e);
       }
+    }
+
+    private void addToDiscoveredFiles(Path rootPath, FileStatus parentStatus, FileStatus childStatus,
+        Map<String, Long> lastModifiedTimesForInputDir) throws IOException
+    {
+      Path childPath = childStatus.getPath();
+      String childPathStr = childPath.toUri().getPath();
+      // Directory by now is scanned forcibly. Now check for whether file/directory needs to be added to discoveredFiles.
+      Long oldModificationTime = lastModifiedTimesForInputDir.get(childPathStr);
+      lastModifiedTimesForInputDir.put(childPathStr, childStatus.getModificationTime());
+
+      if (skipFile(childPath, childStatus.getModificationTime(), oldModificationTime) || // Skip dir or file if no timestamp modification
+          (childStatus.isDirectory() && (oldModificationTime != null))) { // If timestamp modified but if its a directory and already present in map, then skip.
+        return;
+      }
+
+      if (ignoredFiles.contains(childPathStr)) {
+        return;
+      }
+
+      ScannedFileInfo info = createScannedFileInfo(parentStatus.getPath(), parentStatus, childPath, childStatus,
+          rootPath);
+
+      LOG.debug("Processing file: " + info.getFilePath());
+      processDiscoveredFile(info);
     }
 
     protected void processDiscoveredFile(ScannedFileInfo info)
@@ -619,4 +661,5 @@ public class FileSplitterInput extends AbstractFileSplitter implements InputOper
   }
 
   private static final Logger LOG = LoggerFactory.getLogger(FileSplitterInput.class);
+
 }
--- a/library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
@@ -1,0 +1,120 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.lib.io.fs;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.apache.hadoop.fs.Path;
+
+import com.datatorrent.api.Context.OperatorContext;
+import com.datatorrent.lib.io.block.BlockMetadata.FileBlockMetadata;
+
+/**
+ * HDFSFileSplitter extends {@link FileSplitterInput} to,
+ * 1. Add relative path to file metadata.
+ * 2. Ignore HDFS temp files (files with extensions _COPYING_).
+ * 3. Set sequencial read option on readers.
+ */
+public class HDFSFileSplitter extends FileSplitterInput
+{
+  private boolean sequencialFileRead;
+
+  public HDFSFileSplitter()
+  {
+    super();
+    super.setScanner(new HDFSScanner());
+  }
+
+
+  @Override
+  protected FileBlockMetadata createBlockMetadata(FileMetadata fileMetadata)
+  {
+    FileBlockMetadata blockMetadta = new FileBlockMetadata(fileMetadata.getFilePath());
+    blockMetadta.setReadBlockInSequence(sequencialFileRead);
+    return blockMetadta;
+  }
+
+  public boolean isSequencialFileRead()
+  {
+    return sequencialFileRead;
+  }
+
+  public void setSequencialFileRead(boolean sequencialFileRead)
+  {
+    this.sequencialFileRead = sequencialFileRead;
+  }
+
+  /**
+   * HDFSScanner extends {@link TimeBasedDirectoryScanner} to ignore HDFS temporary files
+   * and files containing unsupported characters. 
+   */
+  public static class HDFSScanner extends TimeBasedDirectoryScanner
+  {
+    protected static final String UNSUPPORTED_CHARACTOR = ":";
+    private String ignoreFilePatternRegularExp = ".*._COPYING_";
+    private transient Pattern ignoreRegex;
+
+    @Override
+    public void setup(OperatorContext context)
+    {
+      super.setup(context);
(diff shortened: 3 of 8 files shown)


sandeepdeshmukh:
All the white space related changes should reverted. There are number of them.

library/src/main/java/com/datatorrent/lib/io/block/HDFSBlockReader.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /**
+   * Converts Scheme part of the URI to lower case. Multiple URI can be comma separated. If no scheme is there, no
+   * change is made.
+   * 
+   * @param
+   * @return String with scheme part as lower case
+   */
+  private static String convertSchemeToLowerCase(String uri)
```
sandeepdeshmukh: Is there a test case covering  this?

library/src/main/java/com/datatorrent/lib/io/block/HDFSBlockMetadata.java
```diff
@@ -0,0 +1,63 @@
...
+ * under the License.
+ */
+package com.datatorrent.lib.io.block;
+
+/**
+ * HDFSBlockMetadata extends {@link BlockMetadata} to provide an option if blocks of a file should be read in-sequence
+ * or in-parallel
+ */
+public class HDFSBlockMetadata extends BlockMetadata.FileBlockMetadata
+{
```
chaithu14: Is this class is specific to only HDFS? If not, change the class name.

library/src/main/java/com/datatorrent/lib/io/block/HDFSBlockMetadata.java
```diff
@@ -0,0 +1,63 @@
...
+  public boolean isReadBlockInSequence()
+  {
+    return readBlockInSequence;
+  }
+
+  public void setReadBlockInSequence(boolean readBlockInSequence)
+  {
+    this.readBlockInSequence = readBlockInSequence;
+  }
+
```
chaithu14: Add java docs for setters and getters.

sandeepdeshmukh:
Java doc is missing at many places.

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -384,7 +399,14 @@ protected void scanIterationComplete()
       lastScanMillis = System.currentTimeMillis();
     }
 
-    protected void scan(@NotNull Path filePath, Path rootPath)
+    protected void scan(Path filePath, Path rootPath)
```
sandeepdeshmukh: Why is NotNull removed?

library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterInputTest.java
```diff
@@ -215,7 +213,7 @@ public void testTimeScan() throws InterruptedException, IOException, TimeoutExce
     testMeta.blockMetadataSink.clear();
 
     Thread.sleep(1000);
-    //added a new relativeFilePath
+    // added a new relativeFilePath
```
sandeepdeshmukh: Example of white space change. Please revert this and all such changes.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSInputModule.java
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * HDFSInputModule is used to read files from HDFS. <br/>
+ * Module emits FileMetadata, BlockMetadata and the block bytes.
+ */
+public class HDFSInputModule implements Module
+{
+
+  @NotNull
+  @Size(min = 1)
+  private String files;
```
chaithu14, line 56: This is specific to HDFS. So, Split the files property to host, port and files/directories to copy.

yogidevendra, line 56: For splitting property to host, port, directory there should be discussion on dev@apex. Else, stick to original convention. We would split it later only if required.

sandeepdeshmukh, line 56: The purpose of Modules is to ease configuration of operators and hence it is good to separate host, port, directory etc.

yogidevendra, line 56: @sandeepdeshmukh If we decide to do this; this should be discussed explicitly on discussion forum. Because, the discussion conclusion may impact many other modules. This point is not specific to this PR and should be conveyed to wider audience.

chinmaykolhatkar:
Can you please check why travis build is failing?

Failed tests: 
  HDFSInputModuleAppTest.testApplication:126->verifyFileContents:140 File data doesn't contain expected text

chinmaykolhatkar:
I suggest all the new files should be moved under org.apache.apex namespace.

library/src/main/java/com/datatorrent/lib/io/block/HDFSBlockReader.java
```diff
@@ -0,0 +1,78 @@
...
+package com.datatorrent.lib.io.block;
+
+import java.io.IOException;
+import java.net.URI;
+
+import org.apache.hadoop.fs.FileSystem;
+
+import com.google.common.base.Splitter;
+
+public class HDFSBlockReader extends FSSliceReader
```
chinmaykolhatkar: Javadoc for class is missing. Please add it.

library/src/main/java/com/datatorrent/lib/io/block/HDFSBlockReader.java
```diff
@@ -0,0 +1,78 @@
...
+  public String getUri()
+  {
+    return uri;
+  }
+
+  /**
+   * Converts Scheme part of the URI to lower case. Multiple URI can be comma separated. If no scheme is there, no
+   * change is made.
+   * 
+   * @param
```
chinmaykolhatkar: Please give description for param uri in javadoc.

library/src/main/java/com/datatorrent/lib/io/block/HDFSBlockReader.java
chinmaykolhatkar: What is the reason for this method to be static? If this is a utility type method can this be moved to somewhere else more accessible?

library/src/main/java/com/datatorrent/lib/io/block/HDFSBlockReader.java
```diff
@@ -0,0 +1,78 @@
...
+   * @return String with scheme part as lower case
+   */
+  private static String convertSchemeToLowerCase(String uri)
+  {
+    if (uri == null) {
+      return null;
+    }
+    StringBuilder inputMod = new StringBuilder();
```
chinmaykolhatkar: A better name than "inputMod"?

library/src/main/java/com/datatorrent/lib/io/block/HDFSBlockReader.java
```diff
@@ -0,0 +1,78 @@
...
+    for (String f : Splitter.on(",").omitEmptyStrings().split(uri)) {
```
chinmaykolhatkar: Considering this is a utility method, I suggest not to take a comma seperated URIs here. What if someone has some other delimiter? 
I think from design perspective its better for this method to take a single URI and let the caller of this method take care of seperator.

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
DT-Priyanka: By Mistake, will revert this.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
```diff
@@ -0,0 +1,180 @@
...
+import javax.validation.constraints.NotNull;
+
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.Path;
+
+import com.datatorrent.api.Context.OperatorContext;
+import com.datatorrent.lib.io.block.BlockMetadata.FileBlockMetadata;
+import com.datatorrent.lib.io.block.HDFSBlockMetadata;
+
+public class HDFSFileSplitter extends FileSplitterInput
```
chinmaykolhatkar, line 35: Class level javadoc is missing. Not having this causes problem in release process. Please add it.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
```diff
@@ -0,0 +1,180 @@
...
+  {
+    return sequencialFileRead;
+  }
+
+  public void setSequencialFileRead(boolean sequencialFileRead)
+  {
+    this.sequencialFileRead = sequencialFileRead;
+  }
+
+  public static class HDFSScanner extends TimeBasedDirectoryScanner
```
chinmaykolhatkar, line 68: Can you please provide class level javadoc?

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
```diff
@@ -0,0 +1,180 @@
...
+    {
+      return new Path(filePathStr).toUri().getPath().contains(UNSUPPORTED_CHARACTOR);
+    }
+  }
+
+  public static class HDFSFileMetaData extends FileMetadata
+  {
+    private String relativePath;
+
+    protected HDFSFileMetaData()
```
chinmaykolhatkar: Any reason why empty constructor is protected but other constructor is private?

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
```diff
@@ -0,0 +1,180 @@
...
+
+    public void setRelativePath(String relativePath)
+    {
+      this.relativePath = relativePath;
+    }
+
+    @Override
+    public String toString()
+    {
+      return "HDFSFileMetaData [relativePath=" + relativePath + ", getNumberOfBlocks()=" + getNumberOfBlocks() + ", getFileName()=" + getFileName() + ", getFileLength()=" + getFileLength() + ", isDirectory()=" + isDirectory() + "]";
```
chinmaykolhatkar: Please have variable names in toString and not method names.
For example change getNumberOfBlocks with numberOfBlocks etc..

library/src/main/java/com/datatorrent/lib/io/fs/HDFSInputModule.java
```diff
@@ -0,0 +1,236 @@
...
+import com.datatorrent.lib.io.block.HDFSBlockReader;
+import com.datatorrent.lib.io.fs.AbstractFileSplitter.FileMetadata;
+import com.datatorrent.lib.io.fs.HDFSFileSplitter.HDFSScanner;
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * HDFSInputModule is used to read files from HDFS. <br/>
+ * Module emits FileMetadata, BlockMetadata and the block bytes.
+ */
+public class HDFSInputModule implements Module
```
chinmaykolhatkar, line 51: Can you please describe what you've mentioned in this PR description as a javadoc in HDFSInputModule?
Its good to mention in default what all this module is responsible for and also how it'll work.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSInputModule.java
```diff
@@ -0,0 +1,236 @@
...
+    dag.setAttribute(blockReader, Context.OperatorContext.METRICS_AGGREGATOR, blockReaderMetrics);
+    dag.setAttribute(blockReader, Context.OperatorContext.COUNTERS_AGGREGATOR, new BasicCounters.LongAggregator<MutableLong>());
+  }
+
+  /**
+   * A comma separated list of directories to scan. If the path is not fully qualified the default file system is used.
+   * A fully qualified path can be provided to scan directories in other filesystems.
+   *
+   * @param files
+   *          files
```
chinmaykolhatkar, line 107: Can we please have javadoc comments in proper format?

library/src/main/java/com/datatorrent/lib/io/fs/HDFSInputModule.java
```diff
@@ -0,0 +1,236 @@
...
+{
+
+  @NotNull
+  @Size(min = 1)
+  private String files;
+  private String filePatternRegularExp;
+  @Min(0)
+  private long scanIntervalMillis;
+  private boolean recursive = true;
+  private long blockSize;
```
chinmaykolhatkar, line 61: Any default for the blockSize?

library/src/main/java/com/datatorrent/lib/io/fs/HDFSInputModule.java
chinmaykolhatkar, line 51: As a part of this comment, can you provide what are the mandatory properties for this modules and what are optional?
OR atleast in the javadoc for respective setter method.

library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterInputTest.java
```diff
@@ -110,8 +110,7 @@ protected void starting(org.junit.runner.Description description)
       fileSplitterInput.setIdempotentStorageManager(new IdempotentStorageManager.NoopIdempotentStorageManager());
 
       Attribute.AttributeMap.DefaultAttributeMap attributes = new Attribute.AttributeMap.DefaultAttributeMap();
-      attributes.put(Context.DAGContext.APPLICATION_PATH,
-          "target/" + className + "/" + methodName + "/" + Long.toHexString(System.currentTimeMillis()));
+      attributes.put(Context.DAGContext.APPLICATION_PATH, "target/" + className + "/" + methodName + "/" + Long.toHexString(System.currentTimeMillis()));
```
chinmaykolhatkar: Please revert all the changes in this file which are not related to this PR. Only take care of Checkstyle violations which appears in new code.

library/src/main/java/com/datatorrent/lib/io/block/HDFSBlockReader.java
DT-Priyanka: Won't keep it as utility method as no one else is going to use if as of now.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
DT-Priyanka: super class is doing that, I am just carrying over the same thing.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
DT-Priyanka: This is auto generated code, would prefer not to change it.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSInputModule.java
DT-Priyanka, line 56: Won't split the properties for now, this could create problems if someone integrates operators/modules with external UI.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSInputModule.java
DT-Priyanka, line 61: Default will be set by operator based on HDFS block size.

library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterInputTest.java
DT-Priyanka: There are no checkstyle violations in new code, still the violation count was more than limit, so fixed some violations in this file.

library/src/main/java/com/datatorrent/lib/io/block/HDFSBlockMetadata.java
DT-Priyanka: removing the class and moving property to parent class.

library/src/main/java/com/datatorrent/lib/io/block/HDFSBlockReader.java
DT-Priyanka: Adding unit test.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
chinmaykolhatkar: auto generated by what?

library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterInputTest.java
chinmaykolhatkar: That's very wierd considering no code gets into master without rat check AFAIK.
 So without these changes does checkstyle validation fails?

library/src/main/java/com/datatorrent/lib/io/fs/HDFSInputModule.java
```diff
@@ -0,0 +1,253 @@
...
+  private boolean sequencialFileRead = false;
+  private int readersCount;
+
+  public final transient ProxyOutputPort<FileMetadata> filesMetadataOutput = new ProxyOutputPort();
```
ishark: Please use <> for generic type in instantiation: new ProxyOutputPort<>()

library/src/main/java/com/datatorrent/lib/io/fs/HDFSInputModule.java
```diff
@@ -0,0 +1,253 @@
...
+
+    blockReader.setUri(files);
+    if (readersCount != 0) {
+      dag.setAttribute(blockReader, Context.OperatorContext.PARTITIONER, new StatelessPartitioner<BlockReader>(readersCount));
+    }
+
+    MetricsAggregator blockReaderMetrics = new MetricsAggregator();
+    blockReaderMetrics.addAggregators("bytesReadPerSec", new SingleMetricAggregator[] {new LongSumAggregator()});
+    dag.setAttribute(blockReader, Context.OperatorContext.METRICS_AGGREGATOR, blockReaderMetrics);
+    dag.setAttribute(blockReader, Context.OperatorContext.COUNTERS_AGGREGATOR, new BasicCounters.LongAggregator<MutableLong>());
```
ishark: Can you use METRICS_AGGREGATOR here? COUNTERS_AGGREGATOR is marked deprecated.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
```diff
@@ -0,0 +1,184 @@
...
+
+/**
+ * HDFSFileSplitter extends {@link FileSplitterInput} to,
+ * 1. Add relative path to file metadata.
+ * 2. Ignore HDFS temp files (files with extensions _COPYING_).
+ * 3. Set sequencial read option on readers.
+ */
+public class HDFSFileSplitter extends FileSplitterInput
+{
+  private boolean sequencialFileRead;
```
ishark, line 37: Does it make sense to have this property in FileSplitterInput operator itself if it is not specific to HDFS?

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
ishark: Can the relativePath property be added to super class? It seems generic enough.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
```diff
@@ -0,0 +1,184 @@
...
+
+  /**
+   * HDFSScanner extends {@link TimeBasedDirectoryScanner} to ignore HDFS temporary files
+   * and files containing unsupported characters. 
+   */
+  public static class HDFSScanner extends TimeBasedDirectoryScanner
+  {
+    protected static final String HDFS_TEMP_FILE = ".*._COPYING_";
+    protected static final String UNSUPPORTED_CHARACTOR = ":";
+    private transient Pattern ignoreRegex;
```
ishark, line 72: One way to handle this could be to expose ignoreRegex as a property for module, with default being _COPYING_. Could be useful ignore other file extensions as well. It can be also be part of superclass if it applies to other type of file readers.

library/src/main/java/com/datatorrent/lib/io/block/BlockReader.java
```diff
@@ -0,0 +1,81 @@
...
+   * @param uri
+   * @return String with scheme part as lower case
+   */
+  private String convertSchemeToLowerCase(String uri)
+  {
+    if (uri == null) {
+      return null;
+    }
+    StringBuilder uriList = new StringBuilder();
+    for (String f : Splitter.on(",").omitEmptyStrings().split(uri)) {
```
ishark: Is it safe to assume the name of directory in a url cannot contain ',' ? Can list<uri> be a parameter instead?

library/src/main/java/com/datatorrent/lib/io/block/BlockReader.java
```diff
@@ -0,0 +1,81 @@
...
+      String scheme = URI.create(f).getScheme();
+      if (scheme != null) {
+        uriList.append(f.replaceFirst(scheme, scheme.toLowerCase()));
```
ishark: Just a thought: Instead of converting scheme to lower case, can we throw an exception to indicate user to give correct uri?

DT-Priyanka: That's what happens by default we don't do anything. But there was a jira suggesting, we should not error out for such problem. we should be able to accept scheme irrespective of it's case.

library/src/main/java/com/datatorrent/lib/io/block/BlockReader.java
DT-Priyanka: We expect uses to configure input file(s)/directories separated by ",". Should we parse it in list during set method? Does it have any advantage?

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
DT-Priyanka, line 37: It is not HDFS specific, but FileSplitterInput is too generic. If we decide this is good property for all purpose, I can move it to parent class.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
DT-Priyanka, line 72: We already expose a property called "filePatternRegularExp" for patterns to accept. Having one more property for patterns to reject, could be confusing so didn't expose it.
Is provided reason good enough not to expose it as property or I should expose it?

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
DT-Priyanka: Yes, definitely, I was just concerned if that would be accepted in generic class or not.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
DT-Priyanka: By editor. In previous PR I had got a comment that better to use auto generated code for such methods.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSInputModule.java
DT-Priyanka: okay.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSInputModule.java
DT-Priyanka: will update.

library/src/main/java/com/datatorrent/lib/io/block/BlockReader.java
ishark: A corner case, but what about the case when input directory is as below: 
hdfs://node18.morado.com:8020/user/isha/trial/comma,name

library/src/main/java/com/datatorrent/lib/io/fs/HDFSFileSplitter.java
ishark, line 72: I think ignoreFileRegex could be useful for non-hdfs file inputs as well.

DT-Priyanka, line 72: Okay I will expose this as property.

library/src/main/java/com/datatorrent/lib/io/block/BlockReader.java
DT-Priyanka: good catch, how to we normally accept the multiple input then? Provided our primary source of properties is file based. Trying to figure out what else can be used as separator?

library/src/main/java/com/datatorrent/lib/io/fs/HDFSInputModule.java
DT-Priyanka, line 56: I have created a Malhar jira for same: APEXMALHAR-2016, we would do it later and not part of this review request.

PR #211, opened 2016-03-15
APEXMALHAR-2011-2012 Avro to Pojo & Pojo to Avro converter

devtagare:
Please review changes for Avro to Pojo conversion & back.

Thanks,
Dev

### DIFF
--- a/contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
@@ -1,0 +1,168 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.contrib.avro;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.avro.AvroRuntimeException;
+import org.apache.avro.file.DataFileStream;
+import org.apache.avro.generic.GenericDatumReader;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.io.DatumReader;
+import org.apache.hadoop.classification.InterfaceStability;
+import org.apache.hadoop.fs.Path;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import com.datatorrent.api.AutoMetric;
+import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.lib.io.IdempotentStorageManager;
+import com.datatorrent.lib.io.fs.AbstractFileInputOperator;
+
+/**
+ * <p>
+ * Avro File Input Operator
+ * </p>
+ * A specific implementation of the AbstractFileInputOperator to read Avro
+ * container files.<br>
+ * No need to provide schema,its inferred from the file<br>
+ * This operator emits a GenericRecord based on the schema derived from the
+ * input file<br>
+ * Users can add the {@link IdempotentStorageManager.FSIdempotentStorageManager}
+ * to ensure exactly once semantics with a HDFS backed WAL.
+ * 
+ * @displayName AvroFileInputOperator
+ * @category Input
+ * @tags fs, file,avro, input operator
+ */
+@InterfaceStability.Evolving
+public class AvroFileInputOperator extends AbstractFileInputOperator<GenericRecord>
+{
+
+  private transient long offset = 0L;
+
+  @AutoMetric
+  @VisibleForTesting
+  int recordCount = 0;
+
+  @AutoMetric
+  @VisibleForTesting
+  int errorCount = 0;
+
+  private transient DataFileStream<GenericRecord> avroDataStream;
+
+  public final transient DefaultOutputPort<GenericRecord> output = new DefaultOutputPort<GenericRecord>();
+
+  public final transient DefaultOutputPort<String> completedFilesPort = new DefaultOutputPort<String>();
+
+  public final transient DefaultOutputPort<String> errorRecordsPort = new DefaultOutputPort<String>();
+
+  /**
+   * Returns a input stream given a file path
+   * 
+   * @param path
+   * @return InputStream
+   * @throws IOException
+   */
+  @Override
+  protected InputStream openFile(Path path) throws IOException
+  {
+    InputStream is = super.openFile(path);
+    if (is != null) {
+      DatumReader<GenericRecord> datumReader = new GenericDatumReader<GenericRecord>();
+      avroDataStream = new DataFileStream<GenericRecord>(is, datumReader);
+      datumReader.setSchema(avroDataStream.getSchema());
+    }
+    return is;
+  }
+
+  /**
+   * Reads a GenericRecord from the given input stream<br>
+   * Emits the FileName,Offset,Exception on the error port if its connected
+   * 
+   * @return GenericRecord
+   */
+  @Override
+  protected GenericRecord readEntity() throws IOException
+  {
+    GenericRecord record = null;
+
+    record = null;
+
+    try {
+      if (avroDataStream != null && avroDataStream.hasNext()) {
+        offset++;
+        record = avroDataStream.next();
+        recordCount++;
+        return record;
+      }
+    } catch (AvroRuntimeException are) {
+      LOG.error("Exception in parsing record for file - " + super.currentFile + " at offset - " + offset, are);
+      if (errorRecordsPort.isConnected()) {
+        errorRecordsPort.emit("FileName:" + super.currentFile + ", Offset:" + offset);
+      }
+      errorCount++;
+      throw new AvroRuntimeException(are);
+    }
+    return record;
+  }
+
+  /**
+   * Closes the input stream If the completed files port is connected, the
+   * completed file is emitted from this port
+   */
+  @Override
+  protected void closeFile(InputStream is) throws IOException
+  {
+    String fileName = super.currentFile;
+
+    if (avroDataStream != null) {
+      avroDataStream.close();
+    }
+
+    super.closeFile(is);
+
+    if (completedFilesPort.isConnected()) {
+      completedFilesPort.emit(fileName);
+    }
+    offset = 0;
+  }
+
+  @Override
+  protected void emit(GenericRecord tuple)
+  {
+    if (tuple != null) {
+      output.emit(tuple);
+    }
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
+    errorCount = 0;
+    recordCount = 0;
+  }
+
+  private static final Logger LOG = LoggerFactory.getLogger(AvroFileInputOperator.class);
+
+}
--- a/contrib/src/main/java/com/datatorrent/contrib/avro/AvroRecordHelper.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/avro/AvroRecordHelper.java
@@ -1,0 +1,123 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.contrib.avro;
+
+import java.text.ParseException;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.avro.Schema;
+import org.apache.avro.Schema.Type;
+
+/**
+ * This is an utility class for reading Avro converted records.<br>
+ * This class can be used with the {@link PojoToAvro} or in isolation to get Avro values.
+ */
+public class AvroRecordHelper
+{
+
+  /**
+   * Convert a passed String value to the given type for the key as per Schema
+   */
+  public static Object convertValueStringToAvroKeyType(Schema schema, String key, String value) throws ParseException
+  {
+    Type type = null;
+
+    if (schema.getField(key) != null) {
+      type = schema.getField(key).schema().getType();
+    } else {
+      return value;
+    }
+
+    Object convertedValue = null;
+
+    if (type == Type.UNION) {
+      convertedValue = convertAndResolveUnionToPrimitive(schema, key, value);
+    } else {
+      convertedValue = convertValueToAvroPrimitive(type, key, value);
+    }
+
+    return convertedValue;
+
+  }
+
+  private static Object convertValueToAvroPrimitive(Type type, String key, String value) throws ParseException
+  {
+    Object newValue = value;
+    switch (type) {
+      case BOOLEAN:
+        newValue = Boolean.parseBoolean(value);
+        break;
+      case DOUBLE:
+        newValue = Double.parseDouble(value);
+        break;
+      case FLOAT:
+        newValue = Float.parseFloat(value);
+        break;
+      case INT:
+        newValue = Integer.parseInt(value);
+        break;
+      case LONG:
+        newValue = Long.parseLong(value);
+        break;
+      case BYTES:
+        newValue = value.getBytes();
+        break;
+      case STRING:
+        newValue = value;
+        break;
+      case NULL:
+        newValue = null;
+        break;
+      default:
+        newValue = value;
+    }
+    return newValue;
+  }
+
+  private static Object convertAndResolveUnionToPrimitive(Schema schema, String key, String value) throws ParseException
+  {
+    Schema unionSchema = schema.getField(key).schema();
+    List<Schema> types = unionSchema.getTypes();
+    Object convertedValue = null;
+    for (int i = 0; i < types.size(); i++) {
+      try {
+        if (types.get(i).getType() == Type.NULL) {
+          if (value == null || value.equals("null")) {
+            convertedValue = null;
+            break;
+          } else {
+            continue;
+          }
+        }
+        convertedValue = convertValueToAvroPrimitive(types.get(i).getType(), key, value);
+      } catch (RuntimeException e) {
+        LOG.error("Could not handle schema resolution", e);
+        continue;
+      }
+      break;
+    }
+
+    return convertedValue;
+  }
+
+  private static final Logger LOG = LoggerFactory.getLogger(AvroRecordHelper.class);
+}
--- a/contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
@@ -1,0 +1,411 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.contrib.avro;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.StringTokenizer;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.avro.AvroRuntimeException;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.commons.lang3.ClassUtils;
+import org.apache.hadoop.classification.InterfaceStability;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Lists;
+
+import com.datatorrent.api.AutoMetric;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.Context.PortContext;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.api.annotation.OutputPortFieldAnnotation;
+import com.datatorrent.common.util.BaseOperator;
+import com.datatorrent.lib.util.FieldInfo;
+import com.datatorrent.lib.util.FieldInfo.SupportType;
+import com.datatorrent.lib.util.PojoUtils;
+
+/**
+ * <p>
+ * AvroToPojo
+ * </p>
+ * A generic implementation for conversion from Avro to POJO. The POJO class
+ * name & field mapping should be provided by the user.<br>
+ * If this mapping is not provided then reflection is used to determine this
+ * mapping.<br>
+ * As of now only primitive types are supported.<br>
+ * Error records are emitted on the errorPort if connected
+ *
+ * @displayName Avro To Pojo
+ * @category Converter
+ * @tags avro
+ */
+@InterfaceStability.Evolving
+public class AvroToPojo extends BaseOperator
+{
+
+  private transient Class<?> pojoClass;
+
+  private static final String FIELD_SEPARATOR = ":";
+  private static final String RECORD_SEPARATOR = ",";
+
+  private String genericRecordToPOJOFieldsMapping = null;
+
+  private List<FieldInfo> fieldInfos;
+
+  private List<ActiveFieldInfo> columnFieldSetters;
+
+  @AutoMetric
+  @VisibleForTesting
+  int recordCount = 0;
+
+  @AutoMetric
+  @VisibleForTesting
+  int errorCount = 0;
+
+  @AutoMetric
+  @VisibleForTesting
+  int fieldErrorCount = 0;
+
+  public final transient DefaultOutputPort<GenericRecord> errorPort = new DefaultOutputPort<GenericRecord>();
+
+  /**
+   * Returns a string representing mapping from generic record to POJO fields
+   */
+  public String getGenericRecordToPOJOFieldsMapping()
+  {
+    return genericRecordToPOJOFieldsMapping;
+  }
+
+  /**
+   * Comma separated list mapping a field in Avro schema to POJO field eg :
+   * orderId:orderId:INTEGER,total:total:DOUBLE
+   */
+  public void setGenericRecordToPOJOFieldsMapping(String genericRecordToPOJOFieldsMapping)
+  {
+    this.genericRecordToPOJOFieldsMapping = genericRecordToPOJOFieldsMapping;
+  }
+
+  public final transient DefaultInputPort<GenericRecord> data = new DefaultInputPort<GenericRecord>()
+  {
+    @Override
+    public void process(GenericRecord tuple)
+    {
+      processTuple(tuple);
+    }
+  };
+
+  /**
+   * Converts given Generic Record and to a POJO and emits it
+   */
+  protected void processTuple(GenericRecord tuple)
+  {
+    try {
+      Object obj = getPOJOFromGenericRecord(tuple);
+
+      if (obj != null) {
+        output.emit(obj);
+        recordCount++;
+      } else if (errorPort.isConnected()) {
+        errorPort.emit(tuple);
+        errorCount++;
+      }
+
+    } catch (InstantiationException | IllegalAccessException e) {
+      LOG.error("Could not initialize object of class -" + getClass().getName(), e);
+      errorCount++;
+    }
+  }
+
+  /**
+   * Returns a POJO from a Generic Record
+   * 
+   * @return Object
+   */
+  @SuppressWarnings("unchecked")
+  private Object getPOJOFromGenericRecord(GenericRecord tuple) throws InstantiationException, IllegalAccessException
+  {
+    Object newObj = getPojoClass().newInstance();
+
+    try {
+      for (int i = 0; i < columnFieldSetters.size(); i++) {
+
+        AvroToPojo.ActiveFieldInfo afi = columnFieldSetters.get(i);
+        SupportType st = afi.fieldInfo.getType();
+        Object val = null;
+
+        try {
+          val = tuple.get(afi.fieldInfo.getColumnName());
+        } catch (Exception e) {
+          LOG.error("Could not find field -" + afi.fieldInfo.getColumnName() + "- in the generic record", e);
+          val = null;
+          fieldErrorCount++;
+        }
+
+        if (val == null) {
+          continue;
+        }
+
+        try {
+          switch (st) {
+            case BOOLEAN:
+              ((PojoUtils.SetterBoolean<Object>)afi.setterOrGetter).set(newObj,
+                  (boolean)tuple.get(afi.fieldInfo.getColumnName()));
+              break;
+
+            case DOUBLE:
+              ((PojoUtils.SetterDouble<Object>)afi.setterOrGetter).set(newObj,
+                  (double)tuple.get(afi.fieldInfo.getColumnName()));
+              break;
+
+            case FLOAT:
+              ((PojoUtils.SetterFloat<Object>)afi.setterOrGetter).set(newObj,
+                  (float)tuple.get(afi.fieldInfo.getColumnName()));
+              break;
+
+            case INTEGER:
+              ((PojoUtils.SetterInt<Object>)afi.setterOrGetter).set(newObj,
+                  (int)tuple.get(afi.fieldInfo.getColumnName()));
+              break;
+
+            case STRING:
+              ((PojoUtils.Setter<Object, String>)afi.setterOrGetter).set(newObj,
+                  new String(tuple.get(afi.fieldInfo.getColumnName()).toString()));
+              break;
+
+            case LONG:
+              ((PojoUtils.SetterLong<Object>)afi.setterOrGetter).set(newObj,
+                  (long)tuple.get(afi.fieldInfo.getColumnName()));
+              break;
+
+            default:
+              throw new AvroRuntimeException("Invalid Support Type");
+
+          }
+        } catch (AvroRuntimeException e) {
+          LOG.error("Exception in setting value", e);
+          fieldErrorCount++;
+        }
+
+      }
+    } catch (Exception ex) {
+      LOG.error("Generic Exception in setting value" + ex.getMessage());
+      errorCount++;
+      newObj = null;
+    }
+    return newObj;
+  }
+
+  /**
+   * Use reflection to generate field info values if the user has not provided
+   * the inputs mapping
+   * 
+   * @return String representing the POJO field to Avro field mapping
+   */
+  private String generateFieldInfoInputs(Class<?> cls)
+  {
+    java.lang.reflect.Field[] fields = cls.getDeclaredFields();
+    StringBuilder sb = new StringBuilder();
+
+    for (int i = 0; i < fields.length; i++) {
+      java.lang.reflect.Field f = fields[i];
+      Class<?> c = ClassUtils.primitiveToWrapper(f.getType());
+      sb.append(f.getName()).append(FIELD_SEPARATOR).append(f.getName()).append(FIELD_SEPARATOR)
+          .append(c.getSimpleName().toUpperCase()).append(RECORD_SEPARATOR);
+    }
+    return sb.substring(0, sb.length() - 1);
+  }
+
+  /**
+   * Creates a map representing fieldName in POJO:field in Generic Record:Data
+   * type
+   * 
+   * @return List of FieldInfo
+   */
+  private List<FieldInfo> createFieldInfoMap(String str)
+  {
+    fieldInfos = new ArrayList<FieldInfo>();
+    StringTokenizer strtok = new StringTokenizer(str, RECORD_SEPARATOR);
+
+    while (strtok.hasMoreTokens()) {
+      String[] token = strtok.nextToken().split(FIELD_SEPARATOR);
+      try {
+        fieldInfos.add(new FieldInfo(token[0], token[1], SupportType.valueOf(token[2])));
+      } catch (Exception e) {
+        LOG.error("Invalid support type", e);
+      }
+    }
+    return fieldInfos;
+  }
+
+  @OutputPortFieldAnnotation(schemaRequired = true)
+  public final transient DefaultOutputPort<Object> output = new DefaultOutputPort<Object>()
+  {
+    public void setup(PortContext context)
+    {
+      setPojoClass(context.getValue(Context.PortContext.TUPLE_CLASS));
+
+      columnFieldSetters = Lists.newArrayList();
+
+      /**
+       * Check if the mapping of Generic record fields to POJO is given, else
+       * use reflection
+       */
+      if (getGenericRecordToPOJOFieldsMapping() == null) {
+        setFieldInfos(createFieldInfoMap(generateFieldInfoInputs(getPojoClass())));
+      } else {
+        setFieldInfos(createFieldInfoMap(getGenericRecordToPOJOFieldsMapping()));
+      }
+
+      initColumnFieldSetters(getFieldInfos());
+      initializeActiveFieldSetters();
+    }
+  };
+
+  @Override
+  public void endWindow()
+  {
+    errorCount = 0;
+    fieldErrorCount = 0;
+    recordCount = 0;
+
+  }
+
+  private Class<?> getPojoClass()
+  {
+    return pojoClass;
+  }
+
+  public void setPojoClass(Class<?> pojoClass)
+  {
+    this.pojoClass = pojoClass;
+  }
+
+  /**
+   * Class that maps fieldInfo to its getters or setters
+   */
+  protected static class ActiveFieldInfo
+  {
+    final FieldInfo fieldInfo;
+    Object setterOrGetter;
+
+    ActiveFieldInfo(FieldInfo fieldInfo)
+    {
+      this.fieldInfo = fieldInfo;
+    }
+
+  }
+
+  /**
+   * A list of {@link FieldInfo}s where each item maps a column name to a pojo
+   * field name.
+   */
+  private List<FieldInfo> getFieldInfos()
+  {
+    return fieldInfos;
+  }
+
+  /**
+   * Add the Active Fields to the columnFieldSetters {@link ActiveFieldInfo}s
+   */
+  private void initColumnFieldSetters(List<FieldInfo> fieldInfos)
+  {
+    for (FieldInfo fi : fieldInfos) {
+      if (columnFieldSetters == null) {
+        columnFieldSetters = Lists.newArrayList();
+      }
+      columnFieldSetters.add(new AvroToPojo.ActiveFieldInfo(fi));
+    }
+  }
+
+  /**
+   * Sets the {@link FieldInfo}s. A {@link FieldInfo} maps a store column to a
+   * pojo field name.<br/>
+   * The value from fieldInfo.column is assigned to
+   * fieldInfo.pojoFieldExpression.
+   *
+   * @description $[].columnName name of the Output Field in POJO
+   * @description $[].pojoFieldExpression expression to get the respective field
+   *              from generic record
+   * @useSchema $[].pojoFieldExpression outputPort.fields[].name
+   */
+  private void setFieldInfos(List<FieldInfo> fieldInfos)
+  {
+    this.fieldInfos = fieldInfos;
+  }
+
+  /**
+   * Initialize the setters for generating the POJO
+   */
+  private void initializeActiveFieldSetters()
+  {
+    for (int i = 0; i < columnFieldSetters.size(); i++) {
+      ActiveFieldInfo activeFieldInfo = columnFieldSetters.get(i);
+
+      SupportType st = activeFieldInfo.fieldInfo.getType();
+
+      switch (st) {
+
+        case BOOLEAN:
+
+          activeFieldInfo.setterOrGetter = PojoUtils.createSetterBoolean(getPojoClass(),
+              activeFieldInfo.fieldInfo.getPojoFieldExpression());
+          break;
+
+        case DOUBLE:
+          activeFieldInfo.setterOrGetter = PojoUtils.createSetterDouble(getPojoClass(),
+              activeFieldInfo.fieldInfo.getPojoFieldExpression());
+          break;
+
+        case FLOAT:
+          activeFieldInfo.setterOrGetter = PojoUtils.createSetterFloat(getPojoClass(),
+              activeFieldInfo.fieldInfo.getPojoFieldExpression());
+          break;
+
+        case INTEGER:
+          activeFieldInfo.setterOrGetter = PojoUtils.createSetterInt(getPojoClass(),
+              activeFieldInfo.fieldInfo.getPojoFieldExpression());
+          break;
+
+        case STRING:
+          activeFieldInfo.setterOrGetter = PojoUtils.createSetter(getPojoClass(),
+              activeFieldInfo.fieldInfo.getPojoFieldExpression(), activeFieldInfo.fieldInfo.getType().getJavaType());
+          break;
+
+        case LONG:
+          activeFieldInfo.setterOrGetter = PojoUtils.createSetterLong(getPojoClass(),
+              activeFieldInfo.fieldInfo.getPojoFieldExpression());
+          break;
+
+        default:
+          activeFieldInfo.setterOrGetter = PojoUtils.createSetter(getPojoClass(),
+              activeFieldInfo.fieldInfo.getPojoFieldExpression(), Byte.class);
+          break;
+      }
+
+      columnFieldSetters.get(i).setterOrGetter = activeFieldInfo.setterOrGetter;
+    }
+  }
+
+  private static final Logger LOG = LoggerFactory.getLogger(AvroToPojo.class);
+
+}
--- a/contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
@@ -1,0 +1,273 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.contrib.avro;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.avro.AvroRuntimeException;
+import org.apache.avro.Schema;
+import org.apache.avro.Schema.Field;
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.commons.lang3.ClassUtils;
+import org.apache.hadoop.classification.InterfaceStability;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import com.datatorrent.api.AutoMetric;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.Context.PortContext;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.api.annotation.InputPortFieldAnnotation;
+import com.datatorrent.common.util.BaseOperator;
+import com.datatorrent.lib.util.PojoUtils;
+import com.datatorrent.lib.util.PojoUtils.Getter;
+
+/**
+ * <p>
+ * PojoToAvro
+ * </p>
+ * A generic implementation for POJO to Avro conversion. A POJO is converted to
+ * a GenericRecord based on the schema provided.<br>
+ * As of now only primitive types are supported.<br>
+ * Error records are emitted on the errorPort if connected
+ *
+ * @displayName Pojo To Avro
+ * @category Converter
+ * @tags avro
+ */
+@InterfaceStability.Evolving
+public class PojoToAvro extends BaseOperator
+{
+
+  private List<Field> columnNames;
+
+  private Class<?> cls;
+
+  private List<Getter> keyMethodMap;
+
+  private transient String schemaString;
+
+  private transient Schema schema;
+
+  @AutoMetric
+  @VisibleForTesting
+  int recordCount = 0;
+
+  @AutoMetric
+  @VisibleForTesting
+  int errorCount = 0;
+
+  @AutoMetric
+  @VisibleForTesting
+  int fieldErrorCount = 0;
+
+  public final transient DefaultOutputPort<GenericRecord> output = new DefaultOutputPort<GenericRecord>();
+
+  public final transient DefaultOutputPort<Object> errorPort = new DefaultOutputPort<Object>();
+
+  private void parseSchema() throws IOException
+  {
+    setSchema(new Schema.Parser().parse(getSchemaString()));
+  }
+
+  /**
+   * Returns the schema string for Avro Generic Record
+   * 
+   * @return schemaString
+   */
+  public String getSchemaString()
+  {
+    return schemaString;
+  }
+
+  /**
+   * Sets the schema string
+   */
+  public void setSchemaString(String schemaString)
+  {
+    this.schemaString = schemaString;
+  }
+
+  /**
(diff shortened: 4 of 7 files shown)


contrib/src/test/java/com/datatorrent/contrib/avro/AvroReaderWriterTest.java
```diff
@@ -0,0 +1,219 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
```
tushargosavi: formatting is off for this file.

contrib/src/test/java/com/datatorrent/contrib/avro/SimpleOrder.java
```diff
@@ -0,0 +1,93 @@
+/**
```
tushargosavi: It is possible to move it as a static class inside the test file, than creating a separate file?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+ * primitive types are supported.
+ *
+ * @displayName Avro To Pojo
+ * @category Converter
+ * @tags avro
+ * @since 3.3.0
+ */
+
+@InterfaceStability.Evolving
+public class AvroToPojo implements Operator
```
tushargosavi: can you extend it from the Parser abstract class, and override convert method which will convert to GenericObject to Pojo.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+  }
+
+  public void setGenericRecordToPOJOFieldsMapping(String genericRecordToPOJOFieldsMapping)
+  {
+    this.genericRecordToPOJOFieldsMapping = genericRecordToPOJOFieldsMapping;
+  }
+
+  private String schemaFile;
+
+  private Schema schema;
```
tushargosavi: Is the schema serializable by Kryo, It does not have default constructor. What is the different between schemaFile and schema? Can schema be created from schemaFile? If yes you can make schema as transient and initialize it in the setup.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+          continue;
+        }
+
+        try {
+
+          switch (st) {
+
+            case BOOLEAN:
+
+              afi.setterOrGetter = PojoUtils.createSetterBoolean(getCls(), afi.fieldInfo.getPojoFieldExpression());
```
tushargosavi: Can you generate setters and getters only once during setup or first time tuple is received, no need to generated them for each record.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+    for (int i = 0; i < fields.length; i++) {
+
+      java.lang.reflect.Field f = fields[i];
+      Class<?> c = ClassUtils.primitiveToWrapper(f.getType());
+      sb.append(f.getName() + FIELD_SEPARATOR + f.getName() + FIELD_SEPARATOR + c.getSimpleName().toUpperCase()
+          + RECORD_SEPARATOR);
+
+    }
+
+    return sb.substring(0, sb.length() - 1).toString();
```
tushargosavi: toString() not needed

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,296 @@
...
+ * are supported.
+ *
+ * @displayName Pojo To Avro
+ * @category Converter
+ * @tags avro
+ * @since 3.3.0
+ */
+
+@InterfaceStability.Evolving
+public class PojoToAvro implements Operator
```
tushargosavi: extends from Parser abstract class.

tushargosavi:
CI build has failed because of checkstyle errors, can you remove checkstyles error added by these commits. Also avoid not required new lines in the code.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroRecordHelper.java
```diff
@@ -0,0 +1,124 @@
...
+import java.text.ParseException;
+import java.util.List;
+
+import org.apache.avro.Schema;
+import org.apache.avro.Schema.Type;
+
+/**
+ * This is an utility class for Avro converted records
+ */
+
```
chinmaykolhatkar: Please remove this space. The maven cannot find where to put @since tag then.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroRecordHelper.java
```diff
@@ -0,0 +1,124 @@
...
+  /**
+   * Convert a passed String value to the given type for the key as per Schema
+   * 
+   * @param schema
+   * @param key
+   * @param value
+   * @return
+   */
+
```
chinmaykolhatkar: Please remove spaces between start of function line and comment. The comment may not get picked up for javadocs.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+ * name & field mapping should be provided by the user. If this mapping is not
+ * provided then reflection is used to determine this mapping. As of now only
+ * primitive types are supported.
+ *
+ * @displayName Avro To Pojo
+ * @category Converter
+ * @tags avro
+ * @since 3.3.0
+ */
+
```
chinmaykolhatkar: Please remove the space.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+  private static final String RECORD_SEPARATOR = ",";
+
+  private String genericRecordToPOJOFieldsMapping = null;
+
+  public String getGenericRecordToPOJOFieldsMapping()
+  {
+    return genericRecordToPOJOFieldsMapping;
+  }
+
+  public void setGenericRecordToPOJOFieldsMapping(String genericRecordToPOJOFieldsMapping)
```
chinmaykolhatkar: Can you add javadoc for setter methods?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+  {
+    setSchema(new Schema.Parser().parse(getSchemaString()));
+  }
+
+  public String getSchemaString()
+  {
+    return schemaFile;
+  }
+
+  public void setSchemaString(String schemaFile)
```
chinmaykolhatkar: javadoc for Setter method

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+  {
+    return schema;
+  }
+
+  public void setSchema(Schema schema)
+  {
+    this.schema = schema;
+  }
+
+  public byte[] serialize(Object obj) throws IOException
```
chinmaykolhatkar: Does this need to be public? Is this used from outside of this class?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
devtagare: This operator is not working like a parser. The intent is to have a module for Avro of which this will be a part.

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
devtagare: This operator is not working like a parser. The intent is to have a module for Avro of which this will be a part.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+  @InputPortFieldAnnotation(optional = false)
+  public final transient DefaultInputPort<GenericRecord> data = new DefaultInputPort<GenericRecord>()
+  {
+
+    @Override
+    public void process(GenericRecord tuple)
+    {
+
+      try {
+        Object obj = getPOJOFromGenericRecord(tuple, getCls());
```
chinmaykolhatkar: Please move this code from process method at the class level and then call that method from here.
This is so that if one wants to extend this, he/she can override the process functionality too.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
chinmaykolhatkar: Yes please. Doing this everytime is very costly.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+  @AutoMetric
+  int recordCnt = 0;
+
+  @AutoMetric
+  int errorCnt = 0;
+
+  @AutoMetric
+  int fieldErrorCnt = 0;
+
+  public void parseSchema() throws IOException
```
chinmaykolhatkar: Any reason for this method to be public? If its not meant to be accessed outside of the class, please make this private or protected appropriately.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+          val = null;
+          fieldErrorCnt++;
+        }
+
+        if (val == null) {
+          continue;
+        }
+
+        try {
+
```
chinmaykolhatkar: CHeckstyle is off here.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+            case BOOLEAN:
+
+              afi.setterOrGetter = PojoUtils.createSetterBoolean(getCls(), afi.fieldInfo.getPojoFieldExpression());
+              ((PojoUtils.SetterBoolean<Object>)afi.setterOrGetter).set(newObj,
+                  (boolean)tuple.get(afi.fieldInfo.getColumnName()));
+              break;
+
+            case DOUBLE:
+              afi.setterOrGetter = PojoUtils.createSetterDouble(getCls(), afi.fieldInfo.getPojoFieldExpression());
+              ((PojoUtils.SetterDouble<Object>)afi.setterOrGetter).set(newObj,
```
chinmaykolhatkar, line 176: I'm not sure if this would work if the output POJO has wrapper classes.
Can you please add such a test?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+              break;
+
+            case LONG:
+              afi.setterOrGetter = PojoUtils.createSetterLong(getCls(), afi.fieldInfo.getPojoFieldExpression());
+              ((PojoUtils.SetterLong<Object>)afi.setterOrGetter).set(newObj,
+                  (long)tuple.get(afi.fieldInfo.getColumnName()));
+              break;
+
+            default:
+              afi.setterOrGetter = PojoUtils.createSetter(getCls(), afi.fieldInfo.getPojoFieldExpression(), Byte.class);
```
chinmaykolhatkar: What is this default case meant for?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+      return b.toByteArray();
+    }
+  }
+
+  @InputPortFieldAnnotation(optional = false)
+  public final transient DefaultInputPort<GenericRecord> data = new DefaultInputPort<GenericRecord>()
+  {
+
+    @Override
+    public void process(GenericRecord tuple)
```
chinmaykolhatkar, line 111: I see tuple being received by this Operator is a GenericRecord which is a avro type.
Who is expected to create that? Upstream operator?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+    columnFieldSetters = Lists.newArrayList();
+
+    try {
+      parseSchema();
+    } catch (IOException e) {
+      LOG.error("Exception in parsing schema");
+    }
+
+    if (getFieldInfos().isEmpty() || getGenericRecordToPOJOFieldsMapping() == null) {
+      setFieldInfos(createFieldInfoMap(generateFieldInfoInputs(getCls())));
```
chinmaykolhatkar: Is this working?

cls variable is set in output Port's setup method which gets called after this method. So getCls would return null here?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+  public void teardown()
+  {
+    // TODO Auto-generated method stub
+
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
+    // TODO Auto-generated method stub
```
chinmaykolhatkar: I think you need to reset your autometric variables here.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
devtagare, line 111: Yes.Thats the use case.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroRecordHelper.java
```diff
@@ -0,0 +1,124 @@
...
+          if (value == null || value.equals("null")) {
+            convertedValue = null;
+            break;
+          } else {
+            continue;
+          }
+        }
+        convertedValue = convertValueToAvroPrimitive(types.get(i).getType(), key, value);
+      } catch (RuntimeException e) {
+        continue;
```
DT-Priyanka, line 114: Shouldn't we at least log the exception?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+  {
+    return columnNames;
+  }
+
+  public void setColumnNames(List<Field> columnNames)
+  {
+    this.columnNames = columnNames;
+  }
+
+  public Class<?> getCls()
```
chinmaykolhatkar: Why is this externall required variables?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroRecordHelper.java
```diff
@@ -0,0 +1,124 @@
...
+      try {
+        if (types.get(i).getType() == Type.NULL) {
+          if (value == null || value.equals("null")) {
+            convertedValue = null;
+            break;
+          } else {
+            continue;
+          }
+        }
+        convertedValue = convertValueToAvroPrimitive(types.get(i).getType(), key, value);
```
DT-Priyanka, line 111: Can you directly return from here and skip return on line 122? code would be more readable.

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,296 @@
...
+  }
+
+  /**
+   * Adding this as a plug for being able to serialize non primitive types
+   * 
+   * @param -
+   *          object to serialize Returns a byte array
+   */
+
+  public byte[] serialize(Object obj) throws IOException
```
chinmaykolhatkar: why public?

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,296 @@
...
+    java.lang.reflect.Field f = cls.getDeclaredField(inputFieldName);
+    Class<?> c = ClassUtils.primitiveToWrapper(f.getType());
+
+    Getter<?, ?> classGetter = PojoUtils.createGetter(cls, inputFieldName, c);
+
+    return classGetter;
+  }
+
+  /**
+   * @param -
```
chinmaykolhatkar: Please have a valid javadoc.

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,296 @@
...
+   *          schema of the generic record Assumption is that the name of a
+   *          field in POJO is the same as the name in Avro schema
+   */
+
+  public void initializeColumnMap(Schema schema)
```
chinmaykolhatkar: Why public?

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,296 @@
...
+    for (int i = 0; i < columnNames.size(); i++) {
+      try {
+        keyMethodMap.add(generateGettersForField(cls, columnNames.get(i).name()));
+      } catch (NoSuchFieldException | SecurityException e) {
+        throw new RuntimeException("Failed to initialize pojo class getters for field: ", e);
+      }
+    }
+  }
+
+  @InputPortFieldAnnotation(optional = true, schemaRequired = true)
```
chinmaykolhatkar: Not sure if I understand the purpose of input port being optional.

If this is the only input port and optional to connect, why would anyone add this operator?

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,296 @@
...
+    public void setup(PortContext context)
+    {
+      cls = context.getValue(Context.PortContext.TUPLE_CLASS);
+    }
+
+    @Override
+    public void process(Object tuple)
+    {
+
+      GenericRecord record = null;
```
chinmaykolhatkar: Please create a protected process method at operator class level and call that method from this process method.
This is to ensure that one can extend processing of tuple.

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,296 @@
...
+      } else {
+        errorCnt++;
+      }
+
+    }
+
+  };
+
+  /**
+   * @param -Object
```
chinmaykolhatkar: Please have a complete javadoc.

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,296 @@
...
+   *          Returns a generic record mapping the POJO fields to provided
+   *          schema
+   */
+
+  public GenericRecord getGenericRecord(Object tuple) throws Exception
```
chinmaykolhatkar: Why public?

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    return rec;
+
+  }
+
+  @Override
+  public void teardown()
+  {
+    // TODO Auto-generated method stub
```
chinmaykolhatkar: Please remove this comment. I suggest to extend the class BaseOperator class instead of implementing Operator interface in which case you would not need to have any of these empty method.

contrib/src/test/java/com/datatorrent/contrib/avro/AvroReaderWriterTest.java
```diff
@@ -0,0 +1,219 @@
...
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.generic.GenericRecord;
+
+import junit.framework.Assert;
+
+@SuppressWarnings("deprecation")
```
chinmaykolhatkar: Please have warning suppression at specific places if that's needed and not put this as a general warning suppression at class level.

contrib/src/test/java/com/datatorrent/contrib/avro/AvroReaderWriterTest.java
```diff
@@ -0,0 +1,219 @@
...
+
+		getWriter().initializeColumnMap(writer.getSchema());
+
+	}
+
+	public void createReaderInput() {
+		int cnt = 3;
+
+		while (cnt > 0) {
+
```
chinmaykolhatkar: Indentation is off.

contrib/src/test/java/com/datatorrent/contrib/avro/AvroReaderWriterTest.java
```diff
@@ -0,0 +1,219 @@
...
+	public PojoToAvro getWriter() {
+		return writer;
+	}
+
+	public void setWriter(PojoToAvro writer) {
+		this.writer = writer;
+	}
+
+	@Test
+	public void testWriter() {
```
chinmaykolhatkar: You need to make operator go through proper lifecycle in unit tests.
op.setup()
op.beingWindow(x)
op.port.put(<tuple>)
op.endWindow();
op.teardown()

contrib/src/test/java/com/datatorrent/contrib/avro/Order.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package com.datatorrent.contrib.avro;
+
+import java.io.Serializable;
+
+public class Order implements Serializable
+{
+
+  /**
```
chinmaykolhatkar: Please remove unnecessary comments.

contrib/src/test/java/com/datatorrent/contrib/avro/Order.java
```diff
@@ -0,0 +1,49 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.datatorrent.contrib.avro;
+
+import java.io.Serializable;
+
+public class Order implements Serializable
```
chinmaykolhatkar: Class level comments are required. This causes problem otherwise during release cycle.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,428 @@
...
+  public Object getPOJOFromGenericRecord(GenericRecord tuple, Class<?> cls)
+      throws InstantiationException, IllegalAccessException
+  {
+    Object newObj = getCls().newInstance();
+
+    try {
+
+      for (int i = 0; i < columnFieldSetters.size(); i++) {
+
+        AvroToPojo.ActiveFieldInfo afi = columnFieldSetters.get(i);
```
DT-Priyanka, line 152: Can you rename the variables to have meaningful names e.g. afi can be fieldInfo or activeFieldInfo? This applies to many variable names.

chinmaykolhatkar:
@devtagare In general, please take care of following things:
1. Make sure travis build passes after the PR is created.
2. Take care of all the checkstyle violations are taken care of.
3. Methods other than the lifecycle method & getter/setter should be made only public if required. Otherwise take a best call on whether the method should be private or protected based on possibility of extending.
4. Please add class level javadoc. This is required so that "@since" tags gets properly added during release cycle.
5. About unit tests, please have a look at other unit tests in malhar and how they make the operator to go through the lifecycle. If possible it would be great if the test not only checks for process method but also takes operator to whole lifecycle.
6. For tests, as this is going to contrib, these unit tests won't be run by travis (please correct me if I'm wrong), please make sure that the tests passes in your local machine.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
DT-Priyanka: Do we need this getter setter provided the class is initialized from PortContext.TUPLE_CLASS

contrib/src/main/java/com/datatorrent/contrib/avro/AvroRecordHelper.java
```diff
@@ -114,6 +107,7 @@ private static Object convertAndResolveUnionToPrimitive(Schema schema, String ke
         }
         convertedValue = convertValueToAvroPrimitive(types.get(i).getType(), key, value);
       } catch (RuntimeException e) {
+        LOG.error("Could not handle schema resolution" + e.getMessage());
```
chinmaykolhatkar: Should this not actually throw to the caller instead of hiding here?

devtagare: Actually failure for reading a single field inside a union should not cause a failure to read the rest of the values inside the union.

devtagare:
Re-opening, had closed it due to travis build issues.

devtagare:
Re-opening

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,159 @@
...
+ * A specific implementation of the AbstractFileInputOperator to read Avro
+ * container files.<br>
+ * No need to provide schema,its inferred from the file<br>
+ * Users can add the {@link IdempotentStorageManager.FSIdempotentStorageManager}
+ * to ensure exactly once semantics with a HDFS backed WAL.
+ * 
+ * @displayName AvroFileInputOperator
+ * @category Input
+ * @tags fs, file,avro, input operator
+ * @since 3.3.0
```
chinmaykolhatkar: Please don't add @since tag in code.. It would be added automatically during release process.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,159 @@
...
+  {
+    InputStream is = super.openFile(path);
+    if (is != null) {
+      try {
+        DatumReader<GenericRecord> datumReader = new GenericDatumReader<GenericRecord>();
+        avroDataStream = new DataFileStream<GenericRecord>(is, datumReader);
+        datumReader.setSchema(avroDataStream.getSchema());
+        return is;
+      } catch (NullPointerException npe) {
+        LOG.error("Schemaless file - " + npe.getMessage());
```
chinmaykolhatkar: Please don't hide any exception with logs... You could log the exception as well which would show callstack for debugging purpose...

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,159 @@
...
+  @AutoMetric
+  int recordCnt = 0;
+
+  @AutoMetric
+  int errorCnt = 0;
+
+  private transient DataFileStream<GenericRecord> avroDataStream;
+  private transient GenericRecord record = null;
+
+  public final transient DefaultOutputPort<GenericRecord> output = new DefaultOutputPort<GenericRecord>();
```
chinmaykolhatkar, line 74: Is it possible to utilize output port from AbstractFileInputOperator class instead of creating a new one?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,159 @@
...
+
+  @OutputPortFieldAnnotation(optional = true)
```
chinmaykolhatkar: No need to set output port as optional... Its optional by default.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,159 @@
...
+  public final transient DefaultOutputPort<String> completedFilesPort = new DefaultOutputPort<String>();
+
+  @OutputPortFieldAnnotation(optional = true)
```
chinmaykolhatkar: No need to set output port as optional... Its optional by default.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,159 @@
...
+  public final transient DefaultOutputPort<String> errorRecordsPort = new DefaultOutputPort<String>();
+
+  @Override
```
chinmaykolhatkar, line 87: Can you please add the method javadoc for openFile and describe what it does?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,159 @@
...
+      try {
+        DatumReader<GenericRecord> datumReader = new GenericDatumReader<GenericRecord>();
+        avroDataStream = new DataFileStream<GenericRecord>(is, datumReader);
+        datumReader.setSchema(avroDataStream.getSchema());
+        return is;
+      } catch (NullPointerException npe) {
+        LOG.error("Schemaless file - " + npe.getMessage());
+      }
+    }
+    return null;
```
chinmaykolhatkar: What happens when this method returns null?
You take care of following cases:
1. How is null handled?
2. Is there a point in returning null or probably just throw this as a RuntimeException if the situation is not recoverable?

devtagare: Failure to read a single record should not cause a failure to read the rest of the records from the data stream.So ideally we should not be throwing this exception and return a null instead to allow further reading of the records from the stream after skipping this record.

However the emitTuples() from AbstractFileInputOperator marks the File as closed if there is a null emitted.
   try {
        int counterForTuple = 0;
        while (counterForTuple++ < emitBatchSize) {
          T line = readEntity();
          if (line == null) {
            LOG.info("done reading file ({} entries).", offset);
            closeFile(inputStream);
            break;
          }

So no further reading of record from the file will happen anyways.
In this case we can throw a RuntimeException or return null effect would be the same.

Changing it to throwing an exception.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
devtagare, line 87: We should add that DOC to the AbstractFileInputOperator as well.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
devtagare, line 74: There isn't  any.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
devtagare: Done

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
devtagare: Removed it.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
devtagare: Done

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,164 @@
...
+  protected InputStream openFile(Path path) throws IOException
+  {
+    InputStream is = super.openFile(path);
+    if (is != null) {
+      try {
+        DatumReader<GenericRecord> datumReader = new GenericDatumReader<GenericRecord>();
+        avroDataStream = new DataFileStream<GenericRecord>(is, datumReader);
+        datumReader.setSchema(avroDataStream.getSchema());
+      } catch (NullPointerException npe) {
+        LOG.error("Schemaless file", npe);
```
chinmaykolhatkar: If you anyway to throw exception here.. Then why catch it? Is there specific advantage that you see with logging this instead of throwing to parent?

chinmaykolhatkar: Suggesting to remove the try/catch.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * Reads a GenericRecord from the given input stream<br>
+   * Emits the FileName,Offset,Exception on the error port if its connected
+   * 
+   * @return GenericRecord
+   */
+  private GenericRecord readRecord() throws IOException
+  {
+    record = null;
```
chinmaykolhatkar, line 110: I don't see "record" variable being used anywhere else other than this method.
Can you make this in local scope instead of class level?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,164 @@
...
+        throw new NullPointerException();
+      }
+    }
+    return is;
+  }
+
+  @Override
+  protected GenericRecord readEntity() throws IOException
+  {
+    return readRecord();
```
chinmaykolhatkar: Why have another level method redirection if nothing else is in it?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,164 @@
...
+      throw new AvroRuntimeException(are);
+    }
+    return record;
+  }
+
+  @Override
+  protected void closeFile(InputStream is) throws IOException
+  {
+    String fileName = super.currentFile;
+    super.closeFile(is);
```
chinmaykolhatkar: Shouldn't suport.closeFile be done later and avroDataStream.close() first? This is because avroStream is created from super.inputStream.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,164 @@
...
+ * @tags fs, file,avro, input operator
+ */
+@InterfaceStability.Evolving
+public class AvroFileInputOperator extends AbstractFileInputOperator<GenericRecord>
+{
+
+  private transient long offset = 0L;
+
+  @AutoMetric
+  int recordCnt = 0;
```
chinmaykolhatkar: Can you give a complete name recordCount as this name will appear as metric. Samefor errorCnt.. Please make it recordCount and errorCount.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * <p>
+ * Avro File Input Operator
+ * </p>
+ * A specific implementation of the AbstractFileInputOperator to read Avro
+ * container files.<br>
+ * No need to provide schema,its inferred from the file<br>
+ * Users can add the {@link IdempotentStorageManager.FSIdempotentStorageManager}
+ * to ensure exactly once semantics with a HDFS backed WAL.
```
chinmaykolhatkar, line 52: Can you please add a line saying the GenericRecord is emitted from this operator?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,164 @@
...
+  {
+    if (tuple != null) {
+      output.emit(tuple);
+    }
+  }
+
+  @Override
+  public void endWindow()
+  {
+    errorCnt = 0;
```
chinmaykolhatkar: Can you please check with @chandnisingh about what is the best practice to reset autometric variables? In beginWindow or endWindow?

If values of AutoMetric variables are been picked up after endWindow is called, the values will always be 0.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,164 @@
...
+    recordCnt = 0;
+  }
+
+  final Logger LOG = LoggerFactory.getLogger(AvroFileInputOperator.class);
```
chinmaykolhatkar: "private static"

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+  private static final String RECORD_SEPARATOR = ",";
+
+  private String genericRecordToPOJOFieldsMapping = null;
+
+  private List<FieldInfo> fieldInfos;
+
+  private List<ActiveFieldInfo> columnFieldSetters = null;
+
+  @AutoMetric
+  int recordCnt = 0;
```
chinmaykolhatkar: Can you give a complete name like recordCount and errorCount?

chinmaykolhatkar: Same for fieldErrorCount

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+  /**
+   * Comma separated list mapping a field in Avro schema to POJO field eg :
+   * orderId:orderId:INTEGER
+   */
+  public void setGenericRecordToPOJOFieldsMapping(String genericRecordToPOJOFieldsMapping)
+  {
+    this.genericRecordToPOJOFieldsMapping = genericRecordToPOJOFieldsMapping;
+  }
+
+  @InputPortFieldAnnotation(optional = false)
```
chinmaykolhatkar: setting input port to optional=false is not required. Its false by default.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+    try {
+      Object obj = getPOJOFromGenericRecord(tuple, getCls());
+
+      if (obj != null) {
+        output.emit(obj);
+        recordCnt++;
+      }
+
+    } catch (InstantiationException | IllegalAccessException e) {
+      LOG.error("Could not initialize object of class - " + getClass().getName());
```
chinmaykolhatkar: Please do not hide the exception, atleast log the exception.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+
+      initColumnFieldSetters(getFieldInfos());
+      initializeActiveFieldSetters();
+    }
+  };
+
+  @Override
+  public void endWindow()
+  {
+    errorCnt = 0;
```
chinmaykolhatkar: @chandnisingh Can you please comment if this is ok place to reset autometric variables considering when they're picked up? And what is the best practice?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+    fieldErrorCnt = 0;
+    recordCnt = 0;
+
+  }
+
+  public Class<?> getCls()
```
chinmaykolhatkar: Is it expected by user to set cls? If not, can you please remove this? If yes, can you please give a more intuitive name than cls?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+  {
+    return cls;
+  }
+
+  public void setCls(Class<?> cls)
```
chinmaykolhatkar: How is user expected to set cls from properties file if that's expected? Maybe take qualified class name and revolve it?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+      columnFieldSetters.get(i).setterOrGetter = activeFieldInfo.setterOrGetter;
+    }
+  }
+
+  /**
+   * Serializes a given object, Added as a plug to serialize non-primitive types
+   * 
+   * @return byte[]
+   */
+  private byte[] serialize(Object obj) throws IOException
```
chinmaykolhatkar: Can you please add class level javadoc comment, if this comes in action and how would the POJO get populated with it?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+              break;
+          }
+        } catch (AvroRuntimeException | IOException e) {
+          LOG.error("Exception in setting value" + e.getMessage());
+          fieldErrorCnt++;
+        }
+
+      }
+    } catch (Exception ex) {
+      LOG.error("Generic Exception in setting value" + ex.getMessage());
```
chinmaykolhatkar, line 211: You could just add ex to LOG.error and it'll call its toString method. getMessage might not print callstack.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+
+  /**
+   * Converts given Generic Record and to a POJO and emits it
+   */
+  protected void processTuple(GenericRecord tuple)
+  {
+    try {
+      Object obj = getPOJOFromGenericRecord(tuple, getCls());
+
+      if (obj != null) {
```
chinmaykolhatkar, line 125: Should there be a error port where GenericRecord is emitted in case of any failure?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+    }
+  }
+
+  /**
+   * Returns a POJO from a Generic Record
+   * 
+   * @return Object
+   */
+  @SuppressWarnings("unchecked")
+  private Object getPOJOFromGenericRecord(GenericRecord tuple, Class<?> cls)
```
chinmaykolhatkar: cls is unused.. Please remove it.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
chinmaykolhatkar: private int recordCount... Same for other autometrics too..

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+      throws InstantiationException, IllegalAccessException
+  {
+    Object newObj = getCls().newInstance();
+
+    try {
+      for (int i = 0; i < columnFieldSetters.size(); i++) {
+
+        AvroToPojo.ActiveFieldInfo afi = columnFieldSetters.get(i);
+        afi = columnFieldSetters.get(i);
```
chinmaykolhatkar: Same assignment done twice.. Can you please remove one?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+              break;
+
+            case INTEGER:
+              ((PojoUtils.SetterInt<Object>)afi.setterOrGetter).set(newObj,
+                  (int)tuple.get(afi.fieldInfo.getColumnName()));
+              break;
+
+            case STRING:
+              ((PojoUtils.Setter<Object, String>)afi.setterOrGetter).set(newObj,
+                  new String(tuple.get(afi.fieldInfo.getColumnName()).toString()));
```
chinmaykolhatkar, line 192: new String(<getfield>).toString() shouldn't be needed? Isn't it? You could just do <getfield>

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+    return newObj;
+  }
+
+  /**
+   * Use reflection to generate field info values if the user has not provided
+   * the inputs mapping
+   * 
+   * @return String representing the POJO field to Avro field mapping
+   */
+  public String generateFieldInfoInputs(Class<?> cls)
```
chinmaykolhatkar: Is this method expected to be used outside of the class? If not please make it private.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+  {
+    java.lang.reflect.Field[] fields = cls.getDeclaredFields();
+    StringBuilder sb = new StringBuilder();
+
+    for (int i = 0; i < fields.length; i++) {
+      java.lang.reflect.Field f = fields[i];
+      Class<?> c = ClassUtils.primitiveToWrapper(f.getType());
+      sb.append(f.getName() + FIELD_SEPARATOR + f.getName() + FIELD_SEPARATOR + c.getSimpleName().toUpperCase()
```
chinmaykolhatkar: This is very picky... But you already are working with StringBuilder.
you can just do sb.append(...).append(...)...

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+    return sb.substring(0, sb.length() - 1);
+  }
+
+  /**
+   * Creates a map representing fieldName in POJO:field in Generic Record:Data
+   * type
+   * 
+   * @return List of FieldInfo
+   */
+  public List<FieldInfo> createFieldInfoMap(String str)
```
chinmaykolhatkar: Can this be made private?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+   */
+  private List<FieldInfo> getFieldInfos()
+  {
+    return fieldInfos;
+  }
+
+  /**
+   * Add the Active Fields to the columnFieldSetters {@link ActiveFieldInfo}s
+   */
+  public void initColumnFieldSetters(List<FieldInfo> fieldInfos)
```
chinmaykolhatkar: Can this be private?

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,264 @@
...
+  private Class<?> cls;
+
+  private List<Getter> keyMethodMap;
+
+  private transient String schemaString;
+
+  private transient Schema schema;
+
+  @AutoMetric
+  int recordCnt = 0;
```
chinmaykolhatkar: Can you make this private?
Same for other AutoMetrics too.

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,264 @@
...
+ * PojoToAvro
+ * </p>
+ * A generic implementation for POJO to Avro conversion. A POJO is converted to
+ * a GenericRecord based on the schema provided. As of now only primitive types
+ * are supported.
+ *
+ * @displayName Pojo To Avro
+ * @category Converter
+ * @tags avro
+ * @since 3.3.0
```
chinmaykolhatkar: Please remove @since... This will be added during release process.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
```diff
@@ -0,0 +1,415 @@
...
+ * </p>
+ * A generic implementation for conversion from Avro to POJO. The POJO class
+ * name & field mapping should be provided by the user. If this mapping is not
+ * provided then reflection is used to determine this mapping. As of now only
+ * primitive types are supported.
+ *
+ * @displayName Avro To Pojo
+ * @category Converter
+ * @tags avro
+ * @since 3.3.0
```
chinmaykolhatkar: Please remove @since.

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,264 @@
...
+    for (int i = 0; i < getColumnNames().size(); i++) {
+      try {
+        keyMethodMap.add(generateGettersForField(cls, getColumnNames().get(i).name()));
+      } catch (NoSuchFieldException | SecurityException e) {
+        throw new RuntimeException("Failed to initialize pojo class getters for field: ", e);
+      }
+    }
+  }
+
+  @InputPortFieldAnnotation(optional = false, schemaRequired = true)
```
chinmaykolhatkar: No need fo explicitly mentioning optional=false for input port. That's default.

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,264 @@
...
+   * Converts incoming tuples into Generic records
+   */
+  protected void processTuple(Object tuple)
+  {
+    GenericRecord record = null;
+
+    try {
+      record = getGenericRecord(tuple);
+    } catch (Exception e) {
+      LOG.error("Exception in parsing record");
```
chinmaykolhatkar, line 218: Please don't hide the exception callstack.. Please log it.

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,264 @@
...
+      errorCnt++;
+    }
+
+    if (record != null) {
+      output.emit(record);
+      recordCnt++;
+    } else {
+      errorCnt++;
```
chinmaykolhatkar: If conversion fails, shouldn't this be going to error output port?

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,264 @@
...
+  {
+    int writeErrorCount = 0;
+    GenericRecord rec = new GenericData.Record(getSchema());
+
+    for (int i = 0; i < columnNames.size(); i++) {
+      try {
+        rec.put(columnNames.get(i).name(), AvroRecordHelper.convertValueStringToAvroKeyType(getSchema(),
+            columnNames.get(i).name(), keyMethodMap.get(i).get(tuple).toString()));
+      } catch (AvroRuntimeException e) {
+        LOG.error("Could not set Field [" + columnNames.get(i).name() + "] in the generic record");
```
chinmaykolhatkar: Please log exception.

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,264 @@
...
+        fieldErrorCnt++;
+      } catch (Exception e) {
+        LOG.error("Parse Exception");
```
chinmaykolhatkar: Please log exception.

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,264 @@
...
+      return null;
+    } else {
+      return rec;
+    }
+  }
+
+  @Override
+  public void endWindow()
+  {
+    recordCnt = 0;
```
chinmaykolhatkar: @chandnisingh , Can you please check if this is right place to reset autometrics variable OR should this be done in beginWindow.

contrib/src/main/java/com/datatorrent/contrib/avro/PojoToAvro.java
```diff
@@ -0,0 +1,264 @@
...
+    errorCnt = 0;
+    fieldErrorCnt = 0;
+  }
+
+  public static final Logger LOG = LoggerFactory.getLogger(PojoToAvro.class);
```
chinmaykolhatkar: Should be private.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroRecordHelper.java
```diff
@@ -0,0 +1,120 @@
...
+{
+
+  /**
+   * Convert a passed String value to the given type for the key as per Schema
+   */
+  public static Object convertValueStringToAvroKeyType(Schema schema, String key, String value) throws ParseException
+  {
+    Type type = null;
+
+    if (schema.getField(key) != null) {
```
chinmaykolhatkar, line 44: Can scheme be null here?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroRecordHelper.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.apache.avro.Schema;
+import org.apache.avro.Schema.Type;
+
+/**
+ * This is an utility class for reading Avro converted records
+ */
+public class AvroRecordHelper
```
chinmaykolhatkar, line 34: I see that this class is used only in PojoToAvro.getGenericRecord. Please correct me if I'm wrong.
In that case, can you move this to that class itself as a private method?

contrib/src/main/java/com/datatorrent/contrib/avro/AvroRecordHelper.java
```diff
@@ -0,0 +1,120 @@
...
+        LOG.error("Could not handle schema resolution" + e.getMessage());
+        continue;
+      }
+      break;
+    }
+
+    return convertedValue;
+  }
+
+  static final Logger LOG = LoggerFactory.getLogger(AvroRecordHelper.class);
```
chinmaykolhatkar: private

contrib/src/test/java/com/datatorrent/contrib/avro/AvroToPojoTest.java
```diff
@@ -0,0 +1,294 @@
...
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.generic.GenericRecord;
+
+import com.datatorrent.api.Attribute;
+import com.datatorrent.api.Context;
+import com.datatorrent.lib.helper.TestPortContext;
+import com.datatorrent.lib.testbench.CollectorTestSink;
+
+public class AvroToPojoTest
```
chinmaykolhatkar, line 40: Can you also add application test for this to check if any serialization issues exist?

chinmaykolhatkar:
@devtagare If I remember correctly, all contrib tests are skipped by default. If your tests does not depend on any external system, can only your test be included and exclude everything else for travis?

I see there is includes/excluded/tests parameter of surefire plugin which can control what test to execute. Instead of skipTest=true, can these parameters be used to control which tests to run ?

@tweise Thoughts?

tweise:
Chinmay, that's a good suggestion. There are several other tests in contrib
that run without external setup. We should whitelist them if possible.

## 

sent from mobile

chinmaykolhatkar:
@tweise Creating a tracking Jira... https://issues.apache.org/jira/browse/APEXMALHAR-2038
Will ask on dev for inputs on what all such test are there in their area of work.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
devtagare: Changed the variable names.Not sure about marking autometrics as private.Can you give a rationale & point to the places where its marked private.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
devtagare: Noticed that autometric variables are not marked private in several operators.see CsvParser, JsonParser. They will not be accessible in the test class unless getters are provided.
Has the usage changed since they were first implemented ?
beginWindow seems a better place for resetting the values, moving it there.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
devtagare: Done

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
devtagare: Had a chat with @chandnisingh. Marking them private & marking @VisibleForTesting

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
devtagare: Had a chat with @chandnisingh. Marking them private & marking @VisibleForTesting
 Add a line note

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
devtagare, line 192: its on tuple.get(afi.fieldInfo.getColumnName()) not on new String

contrib/src/main/java/com/datatorrent/contrib/avro/AvroRecordHelper.java
devtagare, line 34: This can be used by other classes for resolving avro schema.Its expected to be used as a helper to get values associated with all Avro types both primitive & non-primitive.A user can use it directly as a part of this operator to resolve the same.

contrib/src/test/java/com/datatorrent/contrib/avro/AvroToPojoTest.java
devtagare, line 40: Added in the AvroFileInputOperator to test AvroFileInputOperator + AvroToPojo

contrib/src/main/java/com/datatorrent/contrib/avro/AvroToPojo.java
chinmaykolhatkar, line 192: Ok then new String() is not required as toString() returns String object itself.

contrib/src/main/java/com/datatorrent/contrib/avro/AvroRecordHelper.java
chinmaykolhatkar: Ok.. Please log the complete exception then instead of just e.getMessage()..

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileInputOperator.java
chinmaykolhatkar, line 87: Well yes it should be there too. But considering the review is for this file I suggest to put it here because the default functionality is overridden or extended.

chinmaykolhatkar:
The changes looks good to me. 
@devtagare Please squash the commits into a single comment. 
Will wait for a day after squash for any other comments and then I can merge the changes.

tweise:
@chinmaykolhatkar @devtagare Did this change pass `mvn verify`? I see that the checkstyle violation count went up. Changes should not be pushed without running verify, especially when the travis build did not succeed.

devtagare:
Yes, had run mvn verfiy before committing the final changes on the pull
request.All checks had passed.

Thanks,
Dev

chinmaykolhatkar:
Yes.. I checked the same.. Before merging.. I ran travis build command and it passed successfully.

PR #217, opened 2016-03-17
APEXMALHAR-2015: Projection Operator & its unit tests

pradeepdalvi:
- Projection Operator
- Unit tests for select/drop fields & projection operator

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/projection/ProjectionOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/projection/ProjectionOperator.java
@@ -1,0 +1,310 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.datatorrent.lib.projection;
+
+import java.lang.reflect.Field;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.commons.lang3.ClassUtils;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import com.datatorrent.api.AutoMetric;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.Context.PortContext;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.DefaultOutputPort;
+
+import com.datatorrent.api.Operator;
+import com.datatorrent.api.annotation.InputPortFieldAnnotation;
+import com.datatorrent.api.annotation.OutputPortFieldAnnotation;
+
+import com.datatorrent.common.util.BaseOperator;
+
+import com.datatorrent.lib.util.PojoUtils;
+
+/**
+ * <b>ProjectionOperator</b>
+ * Projection Operator projects defined set of fields from given selectFields/dropFields
+ *
+ * <b>Parameters</b>
+ * - selectFields: comma separated list of fields to be selected from input tuples
+ * - dropFields: comma separated list of fields to be dropped from input tuples
+ * selectFields and dropFields are optional and either of them shall be specified
+ * When both are not specified, all fields shall be projected to downstream operator
+ * When both are specified, selectFields shall be given the preference
+ *
+ * <b>Input Port</b> takes POJOs as an input
+ *
+ * <b>Output Ports</b>
+ * - projected port emits POJOs with projected fields from input POJOs
+ * - remainder port, if connected, emits POJOs with remainder fields from input POJOs
+ * - error port emits input POJOs as is upon error situations
+ * 
+ * <b>Examples</b>
+ * For {a, b, c} type of input tuples
+ *  - when selectFields = "" and dropFields = "", projected port shall emit {a, b, c}
+ *  - when selectFields = "a" and dropFields = "b", projected port shall emit {a}, remainder {b, c}
+ *  - when selectFields = "b", projected port shall emit {b} and remainder port shall emit {a, c}
+ *  - when dropFields = "b", projected port shall emit {a, c} and remainder port shall emit {b}
+ *
+ */
+public class ProjectionOperator extends BaseOperator implements Operator.ActivationListener<Context>
+{
+  protected String selectFields;
+  protected String dropFields;
+  protected String condition;
+
+  static class TypeInfo
+  {
+    String name;
+    Class type;
+    PojoUtils.Setter setter;
+    PojoUtils.Getter getter;
+
+    public TypeInfo(String name, Class<?> type)
+    {
+      this.name = name;
+      this.type = type;
+    }
+
+    public String toString()
+    {
+      String s = new String("'name': " + name + " 'type': " + type);
+      return s;
+    }
+  }
+
+  private transient List<TypeInfo> projectedFields = new ArrayList<>();
+  private transient List<TypeInfo> remainderFields = new ArrayList<>();
+
+  @VisibleForTesting
+  List<TypeInfo> getProjectedFields()
+  {
+    return projectedFields;
+  }
+
+  @VisibleForTesting
+  List<TypeInfo> getRemainderFields()
+  {
+    return remainderFields;
+  }
+
+  @AutoMetric
+  protected long projectedTuples;
+
+  @AutoMetric
+  protected long remainderTuples;
+
+  @AutoMetric
+  protected long errorTuples;
+
+  protected Class<?> inClazz = null;
+  protected Class<?> projectedClazz = null;
+  protected Class<?> remainderClazz = null;
+
+  @InputPortFieldAnnotation(schemaRequired = true)
+  public transient DefaultInputPort<Object> input = new DefaultInputPort<Object>()
+  {
+    public void setup(PortContext context)
+    {
+      inClazz = context.getValue(Context.PortContext.TUPLE_CLASS);
+    }
+
+    @Override
+    public void process(Object t)
+    {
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/datatorrent/lib/projection/ProjectionOperator.java
```diff
@@ -0,0 +1,303 @@
...
+/**
+ * <b>ProjectionOperator</b>
+ * Projection Operator projects defined set of fields from given selectFields/dropFields
+ *
+ * <b>Parameters</b>
+ * - selectFields: comma separated list of fields to be selected from input tuples
+ * - dropFields: comma separated list of fields to be dropped from input tuples
+ * selectFields and dropFields are optional and either of them shall be specified
+ * When both are not specified, all fields shall be projected to downstream operator
+ *
```
yogidevendra, line 59: Add a comment about what happens in case both of them are specified.

library/src/main/java/com/datatorrent/lib/projection/ProjectionOperator.java
```diff
@@ -0,0 +1,303 @@
...
+
+  @AutoMetric
+  protected long errorTuples;
+
+  protected Class<?> inClazz = null;
+  protected Class<?> projectedClazz = null;
+  protected Class<?> remainderClazz = null;
+
+  @InputPortFieldAnnotation(schemaRequired = true)
+  public transient DefaultInputPort<Object> in = new DefaultInputPort<Object>()
```
yogidevendra: better to name port as 'input' rather than 'in'

library/src/main/java/com/datatorrent/lib/projection/ProjectionOperator.java
```diff
@@ -0,0 +1,303 @@
...
+      Field f = inClazz.getDeclaredField(s);
+      TypeInfo t = new TypeInfo(f.getName(), ClassUtils.primitiveToWrapper(f.getType()));
+      t.getter = PojoUtils.createGetter(inClazz, t.name, t.type);
+      t.setter = PojoUtils.createSetter(remainderClazz, t.name, t.type);
+      remainderFields.add(t);
+    } catch (NoSuchFieldException e) {
+      throw new RuntimeException("Field " + s + " not found in class " + inClazz, e);
+    }
+  }
+
```
yogidevendra, line 198: How about moving common code from addProjectedField, addRemainderField into common method?

library/src/test/java/com/datatorrent/lib/projection/ActivateTest.java
```diff
@@ -0,0 +1,147 @@
...
+    logger.debug("start round 0");
+    projection.selectFields = "";
+    projection.dropFields = "str,date";
+    projection.activate(null);
+    Assert.assertEquals("projected fields", 1, projection.getProjectedFieldsSize());
+    Assert.assertEquals("remainder fields", 2, projection.getRemainderFieldsSize());
+    projection.deactivate();
+    logger.debug("start round 0");
+  }
+
```
yogidevendra, line 144: Should we add test case for where both selectFields, dropFields are specified.

library/src/test/java/com/datatorrent/lib/projection/DummyProjection.java
```diff
@@ -0,0 +1,33 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.datatorrent.lib.projection;
+
+class DummyProjection extends ProjectionOperator
```
yogidevendra: How about having this as inner class instead of separate file?

library/src/test/java/com/datatorrent/lib/projection/ProjectionTest.java
```diff
@@ -0,0 +1,264 @@
...
+    private long remainder;
+
+    public long getProjected()
+    {
+      return projected;
+    }
+
+    public void setProjected(long _projected)
+    {
+      projected = _projected;
```
yogidevendra: https://docs.oracle.com/javase/tutorial/java/nutsandbolts/variables.html
The convention, is to always begin your variable names with a letter, not "$" or "_".

this.projected = projected is more popular way of doing it.

library/src/test/java/com/datatorrent/lib/projection/ProjectionTest.java
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  @Test
+  public void testProjection()
+  {
+    logger.debug("start round 0");
+    projection.beginWindow(0);
+
+    projection.in.process(data);
+    Assert.assertEquals("projected tuples", 1, projection.projectedTuples);
```
yogidevendra, line 223: Can we add some more scenarios for this test.

library/src/main/java/com/datatorrent/lib/projection/ProjectionOperator.java
```diff
@@ -0,0 +1,303 @@
...
+    {
+      remainderClazz = context.getValue(Context.PortContext.TUPLE_CLASS);
+    }
+  };
+
+
+  @OutputPortFieldAnnotation(schemaRequired = true)
+  public final transient DefaultOutputPort<Object> error = new DefaultOutputPort<Object>()
+  {
+    public void setup(PortContext context)
```
chinmaykolhatkar: Not sure if this is required to be done.
inClass is already set by this point or will be set in the future for sure.

library/src/test/java/com/datatorrent/lib/projection/DummyProjection.java
chinmaykolhatkar: I agree.

library/src/main/java/com/datatorrent/lib/projection/ProjectionOperator.java
```diff
@@ -0,0 +1,303 @@
...
+ * - error port emits input POJOs as is upon error situations
+ * 
+ * <b>Examples</b>
+ * For {a, b, c} type of input tuples
+ *  - when selectFields = "" and dropFields = "", projected port shall emit {a, b, c}
+ *  - when selectFields = "b", projected port shall emit {b} and remainder port shall emit {a, c}
+ *  - when dropFields = "b", projected port shall emit {a, c} and remainder port shall emit {b}
+ * 
+ */
+
```
chinmaykolhatkar: Please remove space between classlevel javadoc comment and public class ProjectOperator.
This might cause problem during release for adding @since tag.

library/src/main/java/com/datatorrent/lib/projection/ProjectionOperator.java
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    public String toString()
+    {
+      String s = new String("'name': " + name + " 'type': " + type);
+      return s;
+    }
+  }
+
+  protected transient List<TypeInfo> projectedFields = new ArrayList<>();
```
chinmaykolhatkar: Unless there is a immediate need for a projectedFields & remainderFields to be used by subclass, I'll suggest to make these fields private.

library/src/main/java/com/datatorrent/lib/projection/ProjectionOperator.java
```diff
@@ -0,0 +1,303 @@
...
+    {
+      inClazz = context.getValue(Context.PortContext.TUPLE_CLASS);
+    }
+  };
+
+  /**
+   * addProjectedField: Add field details (name, type, getter and setter) for field with given name
+   * in projectedFields list
+   */
+  protected void addProjectedField(String s)
```
chinmaykolhatkar, line 169: Suggesting to make the method private unless there is a immediate need where these methods can be overridden by subclass.

library/src/main/java/com/datatorrent/lib/projection/ProjectionOperator.java
```diff
@@ -0,0 +1,303 @@
...
+    } catch (NoSuchFieldException e) {
+      throw new RuntimeException("Field " + s + " not found in class " + inClazz, e);
+    }
+  }
+
+  /**
+   * addRemainderField: Add field details (name, type, getter and setter) for field with given name
+   * in remainderFields list
+   */
+  protected void addRemainderField(String s)
```
chinmaykolhatkar, line 186: Suggesting to make the method private unless there is a immediate need where these methods can be overridden by subclass.

library/src/test/java/com/datatorrent/lib/projection/ProjectionTest.java
```diff
@@ -0,0 +1,264 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Tests for ProjectionOperator
+ */
+
```
chinmaykolhatkar: Extra space present between class-level javadoc and class..

library/src/test/java/com/datatorrent/lib/projection/ProjectionTest.java
chinmaykolhatkar: There are other places also this is done in this file. Please take care of it.

library/src/test/java/com/datatorrent/lib/projection/ProjectionTest.java
```diff
@@ -0,0 +1,264 @@
...
+public class ProjectionTest
```
chinmaykolhatkar, line 37: Can you also add a test with source and sink?
For example you can refer here:
https://github.com/apache/incubator-apex-malhar/blob/master/library/src/test/java/com/datatorrent/lib/math/AverageTest.java

There are bunch of others too.

chinmaykolhatkar:
@pradeepdalvi Can you also add a applicationtest.. This will verify the serialization too and would act as an example how projection operator should be used.

library/src/main/java/com/datatorrent/lib/projection/ProjectionOperator.java
pradeepdalvi, line 198: projectedFields and remainderFields are separate and extending class might be interested only in one of them. Hence populating these lists have been kept differently.

library/src/main/java/com/datatorrent/lib/projection/ProjectionOperator.java
```diff
@@ -0,0 +1,311 @@
...
+  public final transient DefaultOutputPort<Object> remainder = new DefaultOutputPort<Object>()
+  {
+    public void setup(PortContext context)
+    {
+      remainderClazz = context.getValue(Context.PortContext.TUPLE_CLASS);
+    }
+  };
+
+
+  @OutputPortFieldAnnotation(schemaRequired = true)
```
chinmaykolhatkar: Please remove schemaRequired=true from this.

chinmaykolhatkar:
Looks good to me..If there any no further comments, I'll merge this in a day.

PR #218, opened 2016-03-17
APEXMALHAR-2017 Use pre checkpoint notification to optimize operator IO

pramodin:
@chandnisingh @siyuanh @DT-Priyanka please review as it touches many operators

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
@@ -110,7 +110,7 @@ import com.datatorrent.lib.counters.BasicCounters;
  * @since 2.0.0
  */
 @OperatorAnnotation(checkpointableWithinAppWindow = false)
-public abstract class AbstractFileOutputOperator<INPUT> extends BaseOperator implements Operator.CheckpointListener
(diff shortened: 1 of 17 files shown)


library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
```diff
@@ -1195,6 +1188,24 @@ public void close() throws IOException
...
   @Override
+  public void beforeCheckpoint(long l)
+  {
+    try {
+      Map<String, FSFilterStreamContext> openStreams = streamsCache.asMap();
+      for (FSFilterStreamContext streamContext: openStreams.values()) {
+        long start = System.currentTimeMillis();
+        streamContext.finalizeContext();
+        totalWritingTime += System.currentTimeMillis() - start;
+        //streamContext.resetFilter();
```
otterc, line 1200: why this commented out code?

otterc:
changes to the file output operator look good to me other than the minor comment.

gauravgopi123:
@PramodSSImmaneni: Say this PR goes in 3.5. Will user be able to use this with apex version < 3.3 as new interface CheckpointNotificationListener is add in apex 3.3?

pramodin:
@gauravgopi123 That would be forward compatibility isn't it (old apex version supporting newer malhar versions). There would almost always be newer features added in newer minor versions and we cannot expect old versions to be forward compatible in that fashion.

PR #222, opened 2016-03-28
APEXMALHAR-2031 enable WindowDataManager to save window files under u…

otterc:
…ser configured path


otterc:

PR #228, opened 2016-04-02
APEXMALHAR-2005 fix for timebased directory scanner's paths passing v…

otterc:
…alidation even when empty


otterc:

PR #229, opened 2016-04-02
APEXMALHAR-2043 removed local copies of checkstyle config in Malhar a…

otterc:
…nd updated checkstyle plugin dependency


otterc:
Should go after 
https://github.com/apache/incubator-apex-core/pull/290

@tweise please review

PR #231, opened 2016-04-05
APEXMALHAR-2043 using property logViolationsToConsole instead of cons…

otterc:
…oleOutput

### DIFF
--- a/library/pom.xml
+++ b/library/pom.xml
@@ -35,6 +35,7 @@
   <name>Apache Apex Malhar (incubating) Library</name>
 
   <properties>
+    <checkstyle.console>false</checkstyle.console>
   </properties>
 
   <build>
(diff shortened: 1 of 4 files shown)


otterc:
@vrozov can you please review this.

library/pom.xml
```diff
@@ -35,6 +35,7 @@
   <name>Apache Apex Malhar (incubating) Library</name>
 
   <properties>
+    <checkstyle.console>false</checkstyle.console>
```
tweise, line 38: Why this property if the setting is local and final for the module?

otterc, line 38: This is how it is in Apex core as well by @vrozov so was mimicking it.
I think the reason is that by default we don't want to print all the existing violations to console so we set the user property `checkstyle.console` to `false`. However we give the user the option to see the existing violations by `mvn checkstyle:check -Dcheckstyle.console`  if they want to see the existing violations.

If `logViolationsToConsole` is set to false, then user will not be able to spit out violations to the console.

tweise:
The user property does not override the setting in pom.xml?

https://maven.apache.org/plugins/maven-checkstyle-plugin/check-mojo.html

otterc:
It doesn't seem to override.  This is how I tried and it doesn't work.

```
<logViolationsToConsole>false</logViolationsToConsole>
```

I ran 
`mvn checkstyle:check -Dcheckstyle.console=true` 
it doesn't print the violations.

PR #233, opened 2016-04-06
Added concrete implementation for HDFS output operator with unit tests.

yogidevendra:


library/src/main/java/com/datatorrent/lib/io/fs/HDFSOutputOperator.java
```diff
@@ -0,0 +1,393 @@
...
+
+  /**
+   * Name of the file to write the output. Directory for the output file should
+   * be specified in the filePath
+   */
+  @NotNull
+  private String fileName;
+
+  /**
+   * currentPartName is of the form fileName_physicalPartionId.partNumber
```
DT-Priyanka: I guess it can be of form "fileName_physicalPartionId.partNumber" (or it's a default), user can give any other format right?

library/src/main/java/com/datatorrent/lib/io/fs/HDFSOutputOperator.java
```diff
@@ -0,0 +1,393 @@
...
+   */
+  private transient String currentPartName;
+
+  /**
+   * Physical partition id for the current partition.
+   */
+  protected transient int physicalPartitionId;
```
DT-Priyanka: why protected?

library/src/main/java/com/datatorrent/lib/io/fs/HDFSOutputOperator.java
```diff
@@ -0,0 +1,393 @@
...
+  /**
+   * No. of bytes received in current application window
+   */
+  @AutoMetric
+  private long byteCount;
+
+  /**
+   * No. of tuples present in current output file
+   */
+  private long currentPartTupleCount;
```
DT-Priyanka: may be rename to currentFileTupleCount?

library/src/main/java/com/datatorrent/lib/io/fs/HDFSOutputOperator.java
```diff
@@ -0,0 +1,393 @@
...
+      rotateCall(currentPartName);
+    }
+  }
+
+  /**
+   * Rollover check at the endWindow
+   */
+  private boolean checkEndWindowFinalization()
+  {
+    if ((currentPartIdleWindows == maxIdleWindows) && !endOffsets.isEmpty()) {
```
DT-Priyanka: what is this condition about endOffsets?

library/src/main/java/com/datatorrent/lib/io/fs/HDFSOutputOperator.java
```diff
@@ -0,0 +1,393 @@
...
+   */
+  protected void rotateCall(String lastFile)
+  {
+    try {
+      this.rotate(lastFile);
+      currentPartName = String.format(currentPartNameformat, fileName, physicalPartitionId);
+      currentPartIdleWindows = 0;
+      currentPartTupleCount = 0;
+    } catch (IOException ex) {
+      LOG.debug(ex.getMessage());
```
DT-Priyanka: log it at INFO or ERROR level?

library/src/test/java/com/datatorrent/lib/io/fs/HDFSOutputOperatorTest.java
```diff
@@ -0,0 +1,113 @@
...
+  public void testIdleWindowsFinalize() throws IOException
+  {
+    HDFSOutputOperator writer = new HDFSOutputOperator();
+    writer.setFileName("output.txt");
+    writer.setFilePath(testMeta.getDir());
+    writer.setAlwaysWriteToTmp(false);
+    writer.setMaxIdleWindows(5);
+    writer.setup(testMeta.testOperatorContext);
+
+    String[][] tuples = {{"0a", "0b" }, {"1a", "1b" }, {}, {}, {}, {}, {"6a", "6b" }, {"7a", "7b" }, {}, {}, {},
```
DT-Priyanka: You can reduce test input here.
Good test cases though :+1:

yogidevendra: I tried to cover different scenarios with long input. Hence, would like to keep it unchanged.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSOutputOperator.java
devtagare: why not log the entire exception ? LOG.debug("Exception in file rotation",ex);.
Also, error level log seems more appropriate.

library/src/main/java/com/datatorrent/lib/io/fs/HDFSOutputOperator.java
```diff
@@ -0,0 +1,393 @@
...
+      DTThrowable.rethrow(ex);
+    } catch (ExecutionException ex) {
+      LOG.debug(ex.getMessage());
```
devtagare: Full exception log and at ERROR level

library/src/test/java/com/datatorrent/lib/io/fs/HDFSOutputOperatorTest.java
```diff
@@ -0,0 +1,113 @@
...
+   * 
+   * @throws IOException
+   */
+  @Test
+  public void testIdleWindowsFinalize() throws IOException
+  {
+    HDFSOutputOperator writer = new HDFSOutputOperator();
+    writer.setFileName("output.txt");
+    writer.setFilePath(testMeta.getDir());
+    writer.setAlwaysWriteToTmp(false);
```
devtagare: This would not trigger a file rolling since setAlwaysWriteToTmp is false.Could you please add a test case with it set to true and check the file rolling ? You can repeat both the test cases with it set to true and split the input across files and then check the results.
- 1 for the test cases though, well thought of.

yogidevendra:
Incorporated review comments. @DT-Priyanka Can you please merge this?

yogidevendra:
Moving new files under org.apache.apex. Updated the PR.  @DT-Priyanka Could you please review and merge this?

otterc:
Have few questions
1. Why does the new operator does not extend AbstractSingleFileOutputOperator? 
2. Why is it called HDFS ? Will it not work with NFS?

PR #240, opened 2016-04-14
SPOI-7818 #resolve #comment Remove dimensions code duplication betwee…

brightchen:
…n megh and malhar

### DIFF
--- a/library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterBaseTest.java
+++ b/library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterBaseTest.java
@@ -97,7 +97,7 @@ public class FileSplitterBaseTest
     @Override
     protected void finished(Description description)
     {
-      TestUtils.deleteTargetTestClassFolder(description);
+      TestUtils.deleteTargetTestClassFolderQuietly(description);
     }
   }
 
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ npm-debug.log
 nb-configuration.xml
 hadoop.log
 site/
+.checkstyle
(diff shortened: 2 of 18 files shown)


library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterBaseTest.java
```diff
@@ -97,7 +97,13 @@ protected void starting(org.junit.runner.Description description)
     @Override
     protected void finished(Description description)
     {
-      TestUtils.deleteTargetTestClassFolder(description);
+      try {
+        //got RuntimeException: java.io.IOException: Unable to delete directory target/com.datatorrent.lib.io.fs.FileSplitterBaseTest/testSplitterInApp
+        //in build server and sometime in local machine
+        TestUtils.deleteTargetTestClassFolder(description);
```
tweise: How about a deleteQuietly function if this is frequently needed?

tweise:
Please create a JIRA that describes what this change is about (topN support?) and change the PR and commit message.

library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterBaseTest.java
brightchen: deleteQuietly is method of FileUtils, which means we need to duplicate the logic of "new File("target/" + description.getClassName())"

brightchen: This is in fact is not related to this task.
Maybe it related to "APEXMALHAR-1991"
But I got the exception at least twice in build server, and got once in my local machine. 
It's better to separate this issue. but this task is out of time.
May I can create a bug and merge this task even if this are problem in build server

PR #242, opened 2016-04-15
APEXMALHAR-1965 WAL Utility

otterc:


otterc:
@tushargosavi @tweise @PramodSSImmaneni @ilooner 
Hi All,

Please review this. If follows the same semantics as AbstractFileOutputOperator. Also some questions are still open which are posted here:
https://github.com/apache/incubator-apex-malhar/pull/204

library/src/main/java/org/apache/apex/malhar/lib/wal/WAL.java
```diff
@@ -0,0 +1,100 @@
...
+     */
+    T get();
+  }
+
+  /**
+   * Provide method to write entries to the WAL.
+   * @param <T>
+   * @param <P>
+   */
+  interface WALWriter<T, P>
```
ishark: Is P needed for writer?

library/src/main/java/org/apache/apex/malhar/lib/wal/WAL.java
```diff
@@ -0,0 +1,100 @@
...
+  {
+    /**
+     * Write an entry to the WAL
+     */
+    int append(T entry) throws IOException;
```
ishark: Would it be useful to have an API for bulk append of bytes?

library/src/main/java/org/apache/apex/malhar/lib/wal/WAL.java
otterc: No. will remove unless someone comes up with a suggestion

library/src/main/java/org/apache/apex/malhar/lib/wal/WAL.java
otterc: For a Write ahead log,  IMO appending an entry at a time will be most common. However bulk append support can be added later.

library/src/main/java/org/apache/apex/malhar/lib/wal/WAL.java
```diff
@@ -0,0 +1,100 @@
...
+
+  }
+
+  /**
+   * Serializer interface used while reading and writing entries to the WAL.
+   * @param <T>
+   */
+  interface Serde<T>
```
ilooner: There is already a Serde interface here https://github.com/apache/incubator-apex-malhar/blob/master/library/src/main/java/org/apache/apex/malhar/lib/spillable/Serde.java

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
```diff
@@ -0,0 +1,504 @@
...
+public class FileSystemWAL<T> implements WAL<FileSystemWAL.FileSystemWALReader, FileSystemWAL.FileSystemWALWriter>
+{
+  @NotNull
+  private Serde<T> serde;
+
+  @NotNull
+  private String filePath;
+
+  //max length of the file
+  private long maxLength;
```
ilooner, line 72: Min(0)

library/src/main/java/org/apache/apex/malhar/lib/wal/WAL.java
otterc: Can you please move it to utils so that it can be reused

otterc:
Had to separate out the commits where the file is renamed and the changes after that because tushar's attribution was getting lost

otterc:
I added a recovery test. Line coverage by tests is 81%. 
As suggested by @tweise it will be better if we integrate it to an operator and test it on the cluster. I will do that and hopefully that can complete testing.

tweise:
Would it make sense to make this a sub-package under org.apache.apex.malhar.lib.state?

library/src/main/java/org/apache/apex/malhar/lib/utils/IOUtils.java
```diff
@@ -0,0 +1,56 @@
...
+public class IOUtils
+{
+  /**
+   * Utility method to copy partial data from input stream to output stream.
+   * @param inputStream        input stream
+   * @param inputStreamOffset  offset in the input stream till which the data is copied.
+   * @param outputStream       output stream
+   * @throws IOException
+   */
+  public static void copyPartial(InputStream inputStream, long inputStreamOffset, OutputStream outputStream)
```
tweise: org.apache.commons.io.IOUtils.copyLarge won't work?

otterc: IOUtils.copyLarge will copy all the data from input to output stream.
Needed to copy data till a particular offset from the input to output. This is common in AbstractFileOutput operator which can use this util function.

otterc:
If WAL is always used with respect to the state then yes we can move. However I think WAL is more generic and may have other use cases outside the state. Also it may be easy to find it when it is directly under lib.

I have an application running application_1459879799578_8496 and I have tested fault-tolerance of the WAL with it and added some more test cases.

This is how wal is used in WALOperator

```
public class WALOperator extends BaseOperator implements Operator.CheckpointNotificationListener
{
  @Valid
  private FileSystemWAL<String> fileSystemWAL = new FileSystemWAL<>();

  public WALOperator()
  {
    fileSystemWAL.setSerde(new SimpleStringSerde());
  }

  public final transient DefaultInputPort<Double> input = new DefaultInputPort<Double>()
  {
    @Override public void process(Double aDouble)
    {
      try {
        fileSystemWAL.getWriter().append(Double.toString(aDouble) + "/n");
      } catch (IOException e) {
      }
    }
  };

  @Override
  {
    fileSystemWAL.setup();
  }

  @Override public void beforeCheckpoint(long l)
  {
    fileSystemWAL.beforeCheckpoint(l);
  }

  @Override public void checkpointed(long l)
  {

  }

  @Override public void committed(long l)
  {
    fileSystemWAL.committed(l);
  }

  @Override public void teardown()
  {
    fileSystemWAL.teardown();
  }

  public FileSystemWAL<String> getFileSystemWAL()
  {
    return fileSystemWAL;
  }

  public void setFileSystemWAL(FileSystemWAL<String> fileSystemWAL)
  {
    this.fileSystemWAL = fileSystemWAL;
  }
}

```

library/src/main/java/org/apache/apex/malhar/lib/utils/IOUtils.java
tweise: Yes, only the newer version has the support to specify a range.

library/src/main/java/org/apache/apex/malhar/lib/wal/WAL.java
```diff
@@ -0,0 +1,88 @@
...
+  interface WALReader<T, P>
+  {
+    /**
+     * Seek to middle of the WAL. This is used primarily during recovery,
+     * when we need to start recovering data from middle of WAL file.
+     */
+    void seek(P pointer) throws IOException;
+
+    /**
+     * Advance WAL by one entry, returns true if it can advance, else false
```
tweise: Why separate advance and get? How about combining it into next() instead.

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
```diff
@@ -0,0 +1,598 @@
...
+        outputStream.close();
+        outputStream = null;
+        LOG.debug("closed {}", currentPointer.partNum);
+      }
+    }
+
+    @Override
+    public int append(T entry) throws IOException
+    {
+      byte[] slice = fileSystemWAL.serde.serialize(entry);
```
tweise: Is it good to mix in serialization at this level. This requires you to make a copy when you have Slice as source.

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
```diff
@@ -0,0 +1,598 @@
...
+
+      return entryLength;
+    }
+
+    protected void flush() throws IOException
+    {
+      if (outputStream != null) {
+        outputStream.flush();
+        if (outputStream instanceof FSDataOutputStream) {
+          ((FSDataOutputStream)outputStream).hflush();
```
tweise: Reminder on why we need both calls?

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
otterc: Will make the change.

library/src/main/java/org/apache/apex/malhar/lib/wal/WAL.java
otterc: I think that will be better. Will do.

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
chandnisingh: We do not need both ```outputStream.flush()``` and ```outputStream.hfush()```.
I will change this to 
```
    protected void flush() throws IOException
    {
      if (outputStream != null) {
        if (outputStream instanceof Syncable) {
          Syncable syncableOutputStream = (Syncable)outputStream;
          syncableOutputStream.hflush();
          syncableOutputStream.hsync();
        } else {
          outputStream.flush();
        }
      }
    }
```

chandnisingh:
```
      if (outputStream instanceof Syncable) {
          Syncable syncableOutputStream = (Syncable)outputStream;
          syncableOutputStream.hflush();
          syncableOutputStream.hsync();
        } else {
          //On local file system the write stream needs to be closed for the reader to see any data
          outputStream.close();
        }
```
Does the above look good. On local file system, the reader wasn't seeing data unless the output stream gets closed

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
```diff
@@ -0,0 +1,598 @@
...
+
+  private long lastCheckpointedWindow = Stateless.WINDOW_ID;
+
+  @Override
+  public void setup()
+  {
+    try {
+      FileContext fileContext = FileContextUtils.getFileContext(filePath);
+      if (maxLength == 0) {
+        maxLength = fileContext.getDefaultFileSystem().getServerDefaults().getBlockSize();
```
davidyan74, line 93: From my understanding, a file in HDFS always takes a multiple of the block size, even if the content of the file is just 1 byte, which is the main reason why HDFS is not suitable for many small files. If that's true, how about instead of maxLength, we let the user specify the multiple of file system's block size? This would also disallow the user from having a too low value for maxLength.

http://blog.cloudera.com/blog/2009/02/the-small-files-problem/

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
otterc: Please take a look at the new changes. I had to make some changes in order to read from temporary files. On local file system until the stream is closed, the data to readers isn't available.

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
tweise, line 93: These are not long living files. They will be purged as soon as the data is flushed. Hence, you have a small number of small files.

otterc, line 93: Currently purge support in WAL is missing. I think I will add a `delete(P pointer)` to the WALWriter api.

tweise, line 93: Yes, once the data is materialized, completed WAL segments need to be deleted.

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
```diff
@@ -0,0 +1,594 @@
...
+
+      int entryLength = entry.length + 4;
+
+      // rotate if needed
+      if (shouldRotate(entryLength)) {
+        rotate(true);
+      }
+
+      outputStream.writeInt(entry.length);
+      outputStream.write(entry.toByteArray());
```
tweise: Should use write(byte b[], int off, int len) instead

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
```diff
@@ -0,0 +1,594 @@
...
+
+    @Override
+    public Slice next() throws IOException
+    {
+      do {
+        if (inputStream == null) {
+          inputStream = getInputStream(currentPointer);
+        }
+
+        if (inputStream != null && !fileContext.util().exists(currentOpenPath)) {
```
tweise: This is very expensive. Why not invalidate the reader when the file rename occurs?

otterc: Ok will do that. Was trying to minimize the overlap between reader and writer.

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
davidyan74, line 93: @tweise But does it make sense to have a maxLength not a multiple of the FS block size?

otterc:
Added support for purging WAL and addressed other review comments. Test coverage has been increased to 92% lines covered :-)

Will change the example application as well and test it on the cluster

otterc:
- I tested the new changes by running application_1459879799578_8853.
- squashed the commits
- this is ready to merge (unless no more review comments)

tweise:
@chandnisingh can you please rebase so that we get the travis related fixes into this PR?

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
tweise, line 93: Yes, because we don't make assumptions about the FS implementation.

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
```diff
@@ -0,0 +1,688 @@
...
+    public FileSystemWALPointer(int partNum, long offset)
+    {
+      this.partNum = partNum;
+      this.offset = offset;
+    }
+
+    @Override
+    public int compareTo(@NotNull FileSystemWALPointer o)
+    {
+      if (this.partNum < o.partNum) {
```
tweise: Integer.compare would keep this short :-)

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
```diff
@@ -0,0 +1,688 @@
...
+    try {
+      lastCheckpointedWindow = window;
+      fileSystemWALWriter.flush();
+    } catch (IOException e) {
+      throw new RuntimeException("during before cp", e);
+    }
+  }
+
+  @Override
+  public void committed(long window)
```
tweise, line 130: Why does it happen in committed. Shouldn't the file be renamed as soon as it is closed?

otterc, line 130: This is because after a failure when the operator is restored to a checkpoint it might expect to write to a tmp file which will not be found if it was renamed just after it was closed.

For eg. the operator starts to write to WAL_x_tmp in window 59. It gets check-pointed.
In window 60 it completes writing to WAL_x_tmp  and so it closes it. If we rename it write then and there is a failure then the operator when gets restored to window 59 state will expect WAL_x_tmp.

The above could be handled programmatically  but we avoided writing to the final target file unless we were absolutely sure it will not be open again which was because HDFS had a bug where rename operation could leave a dangling lease on the target.

 This is how it was done in AbstractFileOutputOperator as well. Finalizing of files is deferred to committed because until then we can write to the same file.

otterc:
@tweise travis is failing because of  license check. I have verified that all the files that I have added have have correct license

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
tweise, line 130: Got it. FYI this isn't releated to the dangling lease issue, which occurs only when the file were left open.

tweise:
`
[WARNING] Missing header in: /home/travis/build/apache/incubator-apex-malhar/library/src/test/java/org/apache/apex/malhar/lib/utils/IOUtilsTest.java
`

Strange, I see the header.

otterc:
Sorry my mistake. Some spaces are missing in that header. Will fix it.

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
tweise, line 130: @chandnisingh It may be useful to document above. It's not immediately obvious why we need committed() otherwise.

otterc, line 130: Ok I will.
This is HDFS issue with the rename+overwrite option that I was referring to
https://issues.apache.org/jira/browse/HDFS-6757
It was advised to avoid  rename + overwrite option on under construction file.

Should I document all this as well here?

otterc:
addressed all the review comments

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
tweise, line 130: No, the HDFS issue doesn't apply here (file is closed and not under construction). Looks good, will merge it.

PR #250, opened 2016-04-20
APEXMALHAR-2061 make unit test options configurable and default max h…

tweise:
…eap 2g


tweise:
@davidyan74 please merge

PR #253, opened 2016-04-22
APEXMALHAR-2064 moved WindowDataManager to org.apache.apex.malhar.lib…

otterc:
….wal

### DIFF
--- a/apps/pom.xml
+++ b/apps/pom.xml
@@ -168,6 +168,13 @@
         </plugin>
       </plugins>
     </pluginManagement>
+    <plugins>
+      <plugin>
+        <!--Enables checkstyle for all sub-module by default-->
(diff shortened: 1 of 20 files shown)


apps/pom.xml
```diff
@@ -168,6 +168,13 @@
         </plugin>
       </plugins>
     </pluginManagement>
+    <plugins>
```
tweise, line 171: Should be enabled through parent.

tweise:
The diff shows unrelated changes. Can you please rebase.

apps/pom.xml
chandnisingh: apps is not a child module of the main module

otterc, line 171: Ok I see.

PR #254, opened 2016-04-22
APEXMALHAR-2064 Move WindowDataManager to org.apache.apex.malhar.lib.wal

otterc:


otterc:
@tweise please review and merge

PR #255, opened 2016-04-25
APEXMALHAR-2065 #resolve #comment Added getWindowIds method to Window…

ilooner:
…Datamanager

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/wal/WindowDataManager.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/wal/WindowDataManager.java
@@ -24,6 +24,7 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.SortedSet;
 
 import javax.validation.constraints.NotNull;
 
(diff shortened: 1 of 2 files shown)


ilooner:
@chandnisingh please review

library/src/main/java/org/apache/apex/malhar/lib/wal/WindowDataManager.java
```diff
@@ -227,6 +235,22 @@ public void delete(int operatorId, long windowId) throws IOException
...
+
+      int index = 0;
+
+      for (Long windowId: windowIds) {
+        windowIdsArray[index] = windowId;
+        index++;
+      }
+
+      return windowIdsArray;
+    }
```
otterc, line 254: Can you please document that windows ids are sorted?

ilooner, line 254: @chandnisingh Added to the documentation

otterc:
@ilooner this uses `replayState` and returns window ids which are sorted, however the `getWindowIds(int operatorId)` utilizes FSStorageAgent. Can you please looking into making the older method use `replayState` as well and return a sorted array. If there are any operators that are using WindowDataManager you might have to check that we didn't break them. 

Thanks, 
Chandni

ilooner:
@chandnisingh since the implementation of WindowDataManager will be changing soon anyway I'd like to avoid making additional changes to it.

otterc:
@ilooner there is a check style issue at line 111. There needs to be an empty line between class/method definitions. 
Please fix it and then I will merge it.

otterc:
@ilooner please close this. I merged it and fixed the check style.
I needed to change WindowDataManager to use Wal which is why I took the change.

Thanks,
Chandni

PR #259, opened 2016-04-29
APEXMALHAR-2068 made noop default in FileSplitter

otterc:
@tweise @vrozov made noop default in FileSplitterInput

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
@@ -96,7 +96,7 @@ public class FileSplitterInput extends AbstractFileSplitter implements InputOper
   {
     super();
     currentWindowRecoveryState = Lists.newLinkedList();
-    idempotentStorageManager = new IdempotentStorageManager.FSIdempotentStorageManager();
+    idempotentStorageManager = new IdempotentStorageManager.NoopIdempotentStorageManager();
     referenceTimes = Maps.newHashMap();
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -96,7 +96,7 @@ public FileSplitterInput()
   {
     super();
     currentWindowRecoveryState = Lists.newLinkedList();
-    idempotentStorageManager = new IdempotentStorageManager.FSIdempotentStorageManager();
+    idempotentStorageManager = new IdempotentStorageManager.NoopIdempotentStorageManager();
```
pramodin, line 99: I understand why we don't want to use FSIdempotentStorageManger in FileSplitter as FileSplitter is more generic than FS. However I think setting Noop disables idempotency by default. How about letting the user knowingly make that choice by not creating it in the operator and set to @NotNull so the user will have to set it.

otterc, line 99: The reason to make this change is that if multiple `FSIdempotentStorageManager`s  are used in an application then their state (on disk) collide.  The jira APEXMALHAR-2068 is opened to address that. 

The problem in having FSIdempotentStorageManager as default is that the users get bewildered because of in-correct behavior manifested by the bug mentioned above.

If the user sets FSIdempotentStorageManage explicitly then they can ensure to make the relative path unique as well.

Not setting any default entails additional step of setting an Idempotent storage manager even when the user doesn't care about idempotency.

PR #260, opened 2016-04-29
APEXMALHAR-1701 An abstract deduper implementation [Review Only]

otterc:
This is an abstract deduplicator that uses managed state.
@bhupeshchawda if you like you can use this base implementation and add features to it.


bhupeshchawda:
@chandnisingh Is it possible for you to add a concrete impl and a unit test for this? If not, then I can start working on it and create another PR. I can change the attribution for the abstract class in my PR.

otterc:
You can work on top of this commit. Here is how to do that - 
Pull the change in locally  
create another commit on top of it.

PR #266, opened 2016-05-06
Fix for LF/CR

vrozov:
@tweise Please review


gauravgopi123:
@vrozov : Is there any Jira for this?

tweise:
Let's consider this self-explanatory :-)

PR #267, opened 2016-05-09
APEXMALHAR-2081 removed HDFS specific classes and moved useful proper…

otterc:
…ties to the base class


otterc:
@tweise @PramodSSImmaneni @DT-Priyanka  please review and merge.

Please Note: there weren't any test cases for the operators that are being removed so the test cases for the existing operators aren't modified.

library/src/main/java/com/datatorrent/lib/io/block/BlockMetadata.java
```diff
@@ -227,37 +226,10 @@ public String getFilePath()
       return filePath;
     }
 
-    /**
-     * Get if blocks should be read in sequence
-     * @return readBlockInSequence
-     */
-    public boolean isReadBlockInSequence()
```
DT-Priyanka, line 234: Why are we removing this functionality? I remember that you mentioned javadoc is not clear, so we need to improve that. But why remove the feature?

library/src/main/java/com/datatorrent/lib/io/block/BlockReader.java
```diff
@@ -1,66 +0,0 @@
...
-import org.apache.hadoop.fs.FileSystem;
-
-import com.datatorrent.api.AutoMetric;
-
-/**
- * BlockReader extends {@link FSSliceReader} to accept case insensitive uri
- */
-public class BlockReader extends FSSliceReader
-{
-  @AutoMetric
```
DT-Priyanka, line 33: shouldn't we move the metrics to parent class?

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -266,6 +266,7 @@ public TimeBasedDirectoryScanner getScanner()
   public static class TimeBasedDirectoryScanner implements Runnable, Component<Context.OperatorContext>
   {
     private static long DEF_SCAN_INTERVAL_MILLIS = 5000;
+    private static String FILE_BEING_COPIED = "_COPYING_";
```
DT-Priyanka, line 269: Couple of questions:
1. Why do we have it as independent field and then check it as extension? We can put it as default in ignore regex part right?
2. This is generic class applicable to file systems other than HDFS, it is safe to ignore these files on other file systems? This file may not mean same on other file systems.

library/src/main/java/com/datatorrent/lib/io/block/BlockReader.java
otterc, line 33: The metrics wasn't doing anything in this class. If was just being initialized to 0 in begin window. Wasn't getting updated

library/src/main/java/com/datatorrent/lib/io/block/BlockMetadata.java
otterc, line 234: I mentioned in the discussion that the way this was accomplished is hacky.
The correct way of doing this is by setting a StreamCodec which I changed in FSInputModule

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
otterc, line 269: 1. Ignore regex is a property that is a property is set by the user so it can be changed by the user anytime.
   IMO this is a clean/explicit way of doing it.  
2. That is correct. However it is highly unlikely that on other file systems, the  actual files are called something that ends with _COPYING_. If its important I will make this run for only HDFS

library/src/main/java/com/datatorrent/lib/io/block/BlockReader.java
otterc, line 33: Added this metric in the AbstractBlockReader. It gets updated there with bytes.

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
otterc, line 269: Made this specific to hdfs.

gauravgopi123:
Should the files be marked as deprecated instead of deleting for backward compatibility?

otterc:
@gauravgopi123 
These files haven't been released yet otherwise semantic versioning plugin would have failed.

otterc:
@DT-Priyanka I have addressed all the comments. Can you please merge?

DT-Priyanka:
sure will do that today.

otterc:
Thanks Priyanka!

PR #268, opened 2016-05-09
APEXMALHAR-2078: Thread issue due to data structures at thread level.

DT-Priyanka:


gauravgopi123:
@DT-Priyanka : I think this change is not the right fix. What happens when Operator is checkpointing while scanner is updating inputDirTolastModifiedTimes. Will this not cause issues?

otterc:
@DT-Priyanka please look at the flow with `referenceTimes`. The state that is checkpointed by the operator thread is updated by the operator thread. 

Also why do we need another map?  why can't the entries of `inputDirTolastModifiedTimes` be added to the same data-structure `referenceTimes`

DT-Priyanka:
Let me check if I can merge two data structures (referenceTimes and inputDirTolastModifiedTimes) by changing data structure of referenceTimes.

otterc:
Can you please explain what `inputDirTolastModifiedTimes` is for ?

DT-Priyanka:
The reason behind introducing this data structure is,
When we ask filesplitter to scan multiple directories. We do maintain time stamps of seen files using their relative path. Now provided you have asked to scan dir1 and dir1/dir2 having following structure,
dir1/myDir
dir1/dir2/myDir

Now the /myDir, though lie in different source directories will be considered as duplicates by operator and one of the entries will be skipped.

DT-Priyanka:
I guess I can move the logic to detect these duplicates from scan time to the process time, so that I can reuse referenceTimes. Will update on this soon.

otterc:
Can the problem be solved by the key in `referenceTimes` be absolute path for all the files?

DT-Priyanka:
Ya I am actually trying to evaluate that, will get back to you.

DT-Priyanka:
we can't do it with absolute path because, say in above example dir to scan was dir1 and dir2 in output you will expect something like,
Output
- dir1/myDir
- dir1/dir2/myDir
- dir2/myDir

If we use absolute path output won't have dir2/myDir after copying "dir1"

DT-Priyanka:
Here is one proposed solution,
1. We do keep the "inputDirTolastModifiedTimes" in operator thread.
2. We pass a copy of inputDirTolastModifiedTimes to scan thread and mark scan thread copy transient. 
3. We do the filtering twice, during scanning and then processing. So if operator is running from long time scanner filtering should take care of avoiding duplicates. If operator is redeployed the scan thread won't remove duplicates (as it didn't save inputDirTolastModifiedTimes) and this time the operator thread will take care of filtering. 
This sounds like some processing overhead, but that's not much. I will still think further to optimize it. Meantime if you, @chandnisingh , have any suggestion let me know.

otterc:
If you use the abs path 

will it not be as - 
- /user/priyanka/dir1/myDir
- /user/priyanka/dir1/dir2/myDir

Since /user/priyank/dir1/dir2/myDir will already be there, it will be ignored?

otterc:
@DT-Priyanka I may have misread the problem.

Is the problem that if the user provides 2 directories to scan but they are duplicates then we do want to have duplicated output as well?

tweise:
@DT-Priyanka @chandnisingh this is marked for 3.4.0, can it be wrapped up or moved out of the release?

DT-Priyanka:
@chandnisingh  updated the code to address threading issue, please review and merge.

DT-Priyanka:
@tweise  have updated code, hopefully it will make it in current release.

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -194,7 +194,12 @@ protected boolean processFileInfo(FileInfo fileInfo)
 
   protected void updateReferenceTimes(ScannedFileInfo fileInfo)
   {
-    referenceTimes.put(fileInfo.getFilePath(), fileInfo.modifiedTime);
+    Map<String, Long> referenceTimeMap;
```
otterc: Nitpicking: please call the local variable `referenceTimePerInputDir`. Will help to remember why it was changed.

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -461,8 +465,11 @@ private void addToDiscoveredFiles(Path rootPath, FileStatus parentStatus, FileSt
       Path childPath = childStatus.getPath();
       String childPathStr = childPath.toUri().getPath();
       // Directory by now is scanned forcibly. Now check for whether file/directory needs to be added to discoveredFiles.
-      Long oldModificationTime = lastModifiedTimesForInputDir.get(childPathStr);
-      lastModifiedTimesForInputDir.put(childPathStr, childStatus.getModificationTime());
+      Long oldModificationTime = null;
+      if (lastModifiedTimesForInputDir != null) {
+        oldModificationTime = lastModifiedTimesForInputDir.get(childPathStr);
+      }
+//      lastModifiedTimesForInputDir.put(childPathStr, childStatus.getModificationTime());
```
otterc: please remove this commented out code

library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterInputTest.java
```diff
@@ -107,7 +108,7 @@ protected void starting(org.junit.runner.Description description)
       fileSplitterInput.setScanner(scanner);
       fileSplitterInput.getScanner().setScanIntervalMillis(500);
       fileSplitterInput.getScanner().setFilePatternRegularExp(".*[.]txt");
-      fileSplitterInput.getScanner().setFiles(dataDirectory);
+      fileSplitterInput.getScanner().setFiles(new File(dataDirectory).getAbsolutePath());
```
otterc: @DT-Priyanka 
The user can add relative or absolute path. We shouldn't be imposing adding absolute path on the user.

DT-Priyanka:
Addressed review comments, also added a test to check if empty input directory gets copied.

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -515,6 +521,9 @@ protected ScannedFileInfo createScannedFileInfo(Path parentPath, FileStatus pare
     protected static boolean skipFile(@SuppressWarnings("unused") @NotNull Path path, @NotNull Long modificationTime,
         Long lastModificationTime) throws IOException
     {
+      if(lastModificationTime == null) {
```
otterc: checkstyle violation here. space before opening parenthesis

otterc: The check lastModificationTime == null becomes redundant in the next line.
Why was this change necessary? 
Please revert it.

PR #270, opened 2016-05-11
APEXMALHAR-2082 Data Filter Operator

pradeepdalvi:
- Added Data Filter Operator
- Marked it as Evolving
- Added test cases around variations

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
@@ -1,0 +1,213 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.datatorrent.lib.filter;
+
+import java.util.LinkedList;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.hadoop.classification.InterfaceStability;
+
+import com.datatorrent.api.AutoMetric;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.Context.PortContext;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.DefaultOutputPort;
+
+import com.datatorrent.api.Operator;
+import com.datatorrent.api.annotation.InputPortFieldAnnotation;
+
+import com.datatorrent.common.util.BaseOperator;
+import com.datatorrent.lib.expression.Expression;
+import com.datatorrent.lib.util.PojoUtils;
+
+/**
+ * <b>FilterOperator</b>
+ * Filter Operator filter out tuples based on defined condition
+ *
+ * <b>Parameters</b>
+ * - condition: condition based on expression language
+ *
+ * <b>Input Port</b> takes POJOs as an input
+ *
+ * <b>Output Ports</b>
+ * - truePort emits POJOs meeting the given condition
+ * - falsePort emits POJOs not meeting the given condition
+ * - error port emits any error situation while evaluating expression
+ * 
+ */
+@InterfaceStability.Evolving
+public class FilterOperator extends BaseOperator implements Operator.ActivationListener
+{
+  private String condition;
+  private List<String> expressionFunctions = new LinkedList<>();
+
+  private transient Class<?> inClazz = null;
+  private transient Expression<Boolean> expr = null;
+
+  @AutoMetric
+  private long trueTuples;
+
+  @AutoMetric
+  private long falseTuples;
+
+  @AutoMetric
+  private long errorTuples;
+
+  public final transient DefaultOutputPort<Object> truePort = new DefaultOutputPort<Object>();
+
+  public final transient DefaultOutputPort<Object> falsePort = new DefaultOutputPort<Object>();
+
+  public final transient DefaultOutputPort<Object> error = new DefaultOutputPort<Object>();
+
+  public FilterOperator()
+  {
+    expressionFunctions.add("java.lang.Math.*");
+    expressionFunctions.add("org.apache.commons.lang3.StringUtils.*");
+    expressionFunctions.add("org.apache.commons.lang3.StringEscapeUtils.*");
+    expressionFunctions.add("org.apache.commons.lang3.time.DurationFormatUtils.*");
+    expressionFunctions.add("org.apache.commons.lang3.time.DateFormatUtils.*");
+  }
+
+  @InputPortFieldAnnotation(schemaRequired = true)
+  public final transient DefaultInputPort<Object> input = new DefaultInputPort<Object>()
+  {
+    public void setup(PortContext context)
+    {
+      inClazz = context.getValue(Context.PortContext.TUPLE_CLASS);
+    }
+
+    @Override
+    public void process(Object t)
+    {
+      processTuple(t);
+    }
+  };
+
+  @Override
+  public void activate(Context context)
+  {
+    createExpression();
+  }
+
+  @Override
+  public void deactivate()
+  {
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
+    errorTuples = trueTuples = falseTuples = 0;
+  }
+
+  /**
+   * createExpression: create an expression from condition of POJO fields
+   * Override this function for custom field expressions
+   */
+  protected void createExpression()
+  {
+    logger.info("Creating an expression for condition {}", condition);
+    expr = PojoUtils.createExpression(inClazz, condition, Boolean.class,
+        expressionFunctions.toArray(new String[expressionFunctions.size()]));
+  }
+
+  /**
+   * evalExpression: Evaluate condition/expression
+   * Override this function for custom condition evaluation
+   */
+  protected Boolean evalExpression(Object t)
+  {
+    return expr.execute(t);
+  }
+
+  /**
+   * handleFilter: emit POJO meeting condition on truePort
+   * and if it did not meet condition then on falsePort
+   */
+  private void processTuple(Object t)
+  {
+    try {
+      if (evalExpression(t)) {
+        truePort.emit(t);
+        trueTuples++;
+      } else {
+        falsePort.emit(t);
+        falseTuples++;
+      }
+    } catch (Exception ex) {
+      logger.error("Error in expression eval: {}", ex.getMessage());
+      logger.debug("Exception: ", ex);
+      error.emit(t);
+      errorTuples++;
+    }
+  }
+
+  /**
+   * Returns condition/expression with which Filtering is done
+   *
+   * @return condition parameter of Filter Operator
+   */
+  public String getCondition()
+  {
+    return condition;
+  }
+
+  /**
+   * Set condition/expression with which Filtering operation would be applied
+   *
+   * @param condition parameter of Filter Operator
+   */
+  public void setCondition(String condition)
+  {
+    logger.info("Changing condition from {} to {}", this.condition, condition);
+    this.condition = condition;
+  }
+
+  /**
+   * Returns the list of expression function which would be made available to
+   * expression to use.
+   *
+   * @return List of functions available in expression.
+   */
+  public List<String> getExpressionFunctions()
+  {
+    return expressionFunctions;
+  }
+
+  /**
+   * Set list of import classes/method should should be made statically available
+   * to expression to use.
+   * For ex. org.apache.apex.test1.Test would mean that "Test" method will be
+   * available in the expression to be used directly.
+   * This is an optional property. See constructor to see defaults that are included.
+   *
+   * @param expressionFunctions List of qualified class/method that needs to be
+   *                            imported to expression.
+   */
+  public void setExpressionFunctions(List<String> expressionFunctions)
+  {
+    this.expressionFunctions = expressionFunctions;
+  }
+
+  private static final Logger logger = LoggerFactory.getLogger(FilterOperator.class);
+}
--- a/library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
+++ b/library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
@@ -1,0 +1,188 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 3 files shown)


library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
```diff
@@ -0,0 +1,177 @@
...
+ * 
+ */
+@InterfaceStability.Evolving
+public class FilterOperator extends BaseOperator implements Operator.ActivationListener
+{
+  protected String condition;
+
+  protected Class<?> inClazz = null;
+
+  private Expression<Boolean> expr = null;
```
gauravgopi123: Shouldn't expr and inClazz be marked as transient?

gauravgopi123:
http://apex.apache.org/contributing.html page says that branch name should be named with Jira Number

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
```diff
@@ -0,0 +1,177 @@
...
+ * <b>Output Ports</b>
+ * - truePort emits POJOs meeting the given condition
+ * - falsePort emits POJOs not meeting the given condition
+ * - error port emits any error situation while evaluating expression
+ * 
+ */
+@InterfaceStability.Evolving
+public class FilterOperator extends BaseOperator implements Operator.ActivationListener
+{
+  protected String condition;
```
chinmaykolhatkar: Please following this order of access:
1) default private
2) Protected if there is a valid usecases where this class can be overridden and that variable/function will need to be accessed from subclass.
3) Public - Avoid as much as possible IMO unless its a constant.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
```diff
@@ -0,0 +1,177 @@
...
+
+  protected Class<?> inClazz = null;
```
chinmaykolhatkar: Should be private

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
chinmaykolhatkar: Yes it should be transient.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
```diff
@@ -0,0 +1,177 @@
...
+
+  private Expression<Boolean> expr = null;
+
+  @AutoMetric
+  protected long trueTuples;
```
chinmaykolhatkar: No need of this to be protected.. Please make it private.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
```diff
@@ -0,0 +1,177 @@
...
+
+  @AutoMetric
+  protected long falseTuples;
```
chinmaykolhatkar: If its a anticipation under which these variables are made protected, I suggest to make them private now and later on need basis change the access modifier.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
```diff
@@ -0,0 +1,177 @@
...
+  @Override
+  public void activate(Context context)
+  {
+    createExpression();
+  }
+
+  @Override
+  public void deactivate()
+  {
+    expr = null;
```
chinmaykolhatkar: No need of this..

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
```diff
@@ -0,0 +1,177 @@
...
+      if (evalExpression(t)) {
+        truePort.emit(t);
+        trueTuples++;
+      } else {
+        falsePort.emit(t);
+        falseTuples++;
+      }
+    } catch (Exception ex) {
+      logger.error("Error in expression eval: {}", ex.getMessage());
+      logger.debug("Exception: ", ex);
```
chinmaykolhatkar, line 159: I'll suggest to put the exception also in log.error. In production environment, the app won't be run with debug level and in such case, please it might be hard to reproduce prod issues, hence its best to get all error information right away.

With exception being logged in debug, most important information about error is logs. ex.message, often doesn't have too much information.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
```diff
@@ -0,0 +1,177 @@
...
+ */
+@InterfaceStability.Evolving
+public class FilterOperator extends BaseOperator implements Operator.ActivationListener
+{
+  protected String condition;
+
+  protected Class<?> inClazz = null;
+
+  private Expression<Boolean> expr = null;
+
```
chinmaykolhatkar, line 66: You need to add one more variable here.. Please check Transformation Operator.. There is a variable called as expressionFunctions... Idea is this variable is to allow user to add any static method in classpath directly usable in expression. Please add such a variable.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
```diff
@@ -0,0 +1,177 @@
...
+      errorTuples++;
+    }
+  }
+
+  /**
+   * setCondition: setter method for condition parameter
+   */
+  public void setCondition(String condition)
+  {
+    logger.info("Changing condition from {} to {}", this.condition, condition);
```
chinmaykolhatkar, line 182: No logging required. There are ample ways in platform to know what is the value of property.

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+  private static FilterOperator filter;
+  private static DummyPrivatePOJO data;
+  private static DummyPublicPOJO pdata;
+
+  @Test
+  public void testFilterPrivate()
+  {
+    filter.inClazz = DummyPrivatePOJO.class;
+    filter.setCondition("({$}.getVal() == 100)");
```
chinmaykolhatkar: $.val should also work.

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
```diff
@@ -0,0 +1,165 @@
...
+    filter.activate(null);
+
+    /* when condition is true */
+    logger.debug("start round 0");
+    filter.beginWindow(0);
+
+    data.setVal(100);
+    filter.input.process(data);
```
chinmaykolhatkar: instead of directly calling process, please do filter.input.put(data)

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
```diff
@@ -0,0 +1,165 @@
...
+    Assert.assertEquals("true tuples", 1, filter.trueTuples);
+    Assert.assertEquals("false tuples", 0, filter.falseTuples);
+
+    filter.endWindow();
+    logger.debug("end round 0");
```
chinmaykolhatkar: Try not to log in test cases... You can have as much asserts as you want.. But that should indicate the information and not the logging.

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+    /* when condition is not true */
+    logger.debug("start round 1");
+    filter.beginWindow(1);
+
+    data.setVal(1000);
+    filter.input.process(data);
```
chinmaykolhatkar: filter.input.put(data);

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
```diff
@@ -0,0 +1,165 @@
...
+    public long val;
+  }
+
+
+  private static FilterOperator filter;
+  private static DummyPrivatePOJO data;
+  private static DummyPublicPOJO pdata;
+
+  @Test
+  public void testFilterPrivate()
```
chinmaykolhatkar, line 98: Please have setup and teardown also called for this object.

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+    filter.deactivate();
+  }
+
+  @Test
+  public void testFilterPublic()
+  {
+    filter.inClazz = DummyPublicPOJO.class;
+    filter.setCondition("({$}.val == 100)");
+    filter.activate(null);
```
chinmaykolhatkar: Please call setup and tear down of the operator.

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
```diff
@@ -0,0 +1,165 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Tests for FilterOperator
+ */
+public class FilterTest
```
chinmaykolhatkar, line 34: Can you please add applicationTest for Filter operator?
This is an important test to see whether serialization and de-serialization happens correctlly.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
pradeepdalvi, line 159: Basically the idea is to limit extended error messages in data path. This exception is on incorrect tuples, hence stacktrace would probably fill logs quickly if we start hitting the issue and make it hard to analyze debug logs. Hence we have added single line message in error and if required entire stacktrace can be retrieved in debug mode.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
pradeepdalvi, line 182: Conditions shall not change frequently and this debug message would immediately help understand that the condition has been changed at Operator level.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
pradeepdalvi: Changed both to transient.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
pradeepdalvi: Changed the order.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
pradeepdalvi: Done

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
pradeepdalvi: Done

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
pradeepdalvi: Done

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
pradeepdalvi: Done

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
pradeepdalvi, line 98: Done

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
pradeepdalvi: Removed debug logs

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
pradeepdalvi: This test is specifying Private member variable in condition. {$}.val raises following exception: Private member cannot be accessed from type "SC".

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
pradeepdalvi: Done

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
pradeepdalvi: These AutoMetrics were being referred in Unit Tests. Made them private and changed Unit Tests to refer CountTestSink.getCount()

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
pradeepdalvi: Made AutoMetrics as private

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
pradeepdalvi, line 66: Added expressionFunctions

library/src/test/java/com/datatorrent/lib/filter/FilterTest.java
pradeepdalvi, line 34: Added FilterAppTest

pradeepdalvi:
I ran travis command couple of times locally and each time it completed successfully.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
```diff
@@ -0,0 +1,210 @@
...
+  {
+    expressionFunctions.add("java.lang.Math.*");
+    expressionFunctions.add("org.apache.commons.lang3.StringUtils.*");
+    expressionFunctions.add("org.apache.commons.lang3.StringEscapeUtils.*");
+    expressionFunctions.add("org.apache.commons.lang3.time.DurationFormatUtils.*");
+    expressionFunctions.add("org.apache.commons.lang3.time.DateFormatUtils.*");
+  }
+
+  @InputPortFieldAnnotation(schemaRequired = true)
+  public transient DefaultInputPort<Object> input = new DefaultInputPort<Object>()
```
chinmaykolhatkar: Please make this port final.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
```diff
@@ -0,0 +1,210 @@
...
+  @AutoMetric
+  private long trueTuples;
+
+  @AutoMetric
+  private long falseTuples;
+
+  @AutoMetric
+  private long errorTuples;
+
+  protected String condition;
```
chinmaykolhatkar: Can this be made private? You already have setter and getter for this.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
```diff
@@ -0,0 +1,210 @@
...
+      logger.debug("Exception: ", ex);
+      error.emit(t);
+      errorTuples++;
+    }
+  }
+
+  /**
+   * Returns condition/expression with which Filtering is done
+   *
+   * @return Condition parameter of Filter Operator
```
chinmaykolhatkar: C in Condition should be small. Otherwise javadoc won't be able to identify it.

library/src/main/java/com/datatorrent/lib/filter/FilterOperator.java
```diff
@@ -0,0 +1,210 @@
...
+import com.datatorrent.common.util.BaseOperator;
+import com.datatorrent.lib.expression.Expression;
+import com.datatorrent.lib.util.PojoUtils;
+
+/**
+ * <b>FilterOperator</b>
+ * Filter Operator filter out tuples based on defined condition
+ *
+ * <b>Parameters</b>
+ * - condition: condition based on expression langugage
```
chinmaykolhatkar: spelling: language.

PR #273, opened 2016-05-13
APEXMALHAR-2077 AbstractSingleFileOutputOperator appending partitionID

yogidevendra:
1. Renamed HDFSOutputOperator to BytesFileOutputOperator
2. Moved partitionID related logic into AbstractSingleFileOutputOperator
3. BytesFileOutputOperator extends AbstractSingleFileOutputOperator
4. renamed currentPartName to partitionedFileName
5. update partitionedFileName from field setters

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/AbstractSingleFileOutputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/AbstractSingleFileOutputOperator.java
@@ -20,6 +20,10 @@ package com.datatorrent.lib.io.fs;
 
 import javax.validation.constraints.NotNull;
 
+import org.apache.commons.lang.StringUtils;
+
+import com.datatorrent.api.Context.OperatorContext;
+
 /**
  * This is a simple class that output all tuples to a single file.
  *
@@ -39,10 +43,41 @@ public abstract class AbstractSingleFileOutputOperator<INPUT> extends AbstractFi
   @NotNull
   protected String outputFileName;
 
+  /**
+   * partitionedFileName string format specifier 
+      e.g. fileName_physicalPartionId -> %s_%d 
+   */
+  private String partitionedFileNameformat = "%s_%d";
+
+  /**
+   * Derived name for file based on physicalPartitionId
+   */
+  private transient String partitionedFileName;
+
+  /**
+   * Physical partition id for the current partition.
+   */
+  private transient int physicalPartitionId;
+
+  /**
+   * Initializing current partition id, partitionedFileName etc. {@inheritDoc}
+   */
+  @Override
+  public void setup(OperatorContext context)
+  {
+    super.setup(context);
+    physicalPartitionId = context.getId();
+    if (StringUtils.isEmpty(partitionedFileNameformat)) {
+      partitionedFileName = outputFileName;
+    } else {
+      partitionedFileName = String.format(partitionedFileNameformat, outputFileName, physicalPartitionId);
+    }
+  }
+
   @Override
   protected String getFileName(INPUT tuple)
   {
-    return outputFileName;
+    return partitionedFileName;
   }
 
   /**
@@ -62,4 +97,32 @@ public abstract class AbstractSingleFileOutputOperator<INPUT> extends AbstractFi
   {
     return outputFileName;
   }
+
+  /**
+   * @return string format specifier for the partitioned file name
+   */
+  public String getPartitionedFileNameformat()
+  {
+    return partitionedFileNameformat;
+  }
+  
+  /**
+   * @param partitionedFileNameformat
+   *          string format specifier for the partitioned file name. It should have one %s and one %d.
+   *          e.g. fileName_physicalPartionId -> %s_%d 
+   */
+  public void setPartitionedFileNameformat(String partitionedFileNameformat)
+  {
+    this.partitionedFileNameformat = partitionedFileNameformat;
+  }
+  
+  /**
+   * @return
+   * Derived name for file based on physicalPartitionId
+   */
+  public String getPartitionedFileName()
+  {
+    return partitionedFileName;
+  }
+  
 }
--- a/library/src/main/java/org/apache/apex/malhar/lib/fs/BytesFileOutputOperator.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/fs/BytesFileOutputOperator.java
@@ -23,16 +23,13 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.concurrent.ExecutionException;
 
-import javax.validation.constraints.NotNull;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.datatorrent.api.AutoMetric;
-import com.datatorrent.api.Context.OperatorContext;
 import com.datatorrent.api.DefaultInputPort;
 import com.datatorrent.api.StreamCodec;
-import com.datatorrent.lib.io.fs.AbstractFileOutputOperator;
+import com.datatorrent.lib.io.fs.AbstractSingleFileOutputOperator;
 import com.datatorrent.netlet.util.DTThrowable;
 
 /**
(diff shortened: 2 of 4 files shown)


yogidevendra:
@chandnisingh Would you please review this and merge?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractSingleFileOutputOperator.java
```diff
@@ -62,4 +92,32 @@ public String getOutputFileName()
...
+  }
+  
+  /**
+   * @param partitionedFileNameformat
+   *          string format specifier for the partitioned file name
+   */
+  public void setPartitionedFileNameformat(String partitionedFileNameformat)
+  {
+    this.partitionedFileNameformat = partitionedFileNameformat;
+    partitionedFileName = String.format(partitionedFileNameformat, outputFileName, physicalPartitionId);
```
amberarrow: See note above.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractSingleFileOutputOperator.java
```diff
@@ -62,4 +92,32 @@ public String getOutputFileName()
...
+   */
+  public String getPartitionedFileNameformat()
+  {
+    return partitionedFileNameformat;
+  }
+  
+  /**
+   * @param partitionedFileNameformat
+   *          string format specifier for the partitioned file name
+   */
```
amberarrow, line 113: The JavaDoc comment should state the expected form of the format string.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractSingleFileOutputOperator.java
```diff
@@ -52,6 +81,7 @@ protected String getFileName(INPUT tuple)
   public void setOutputFileName(String outputFileName)
   {
     this.outputFileName = outputFileName;
+    partitionedFileName = String.format(partitionedFileNameformat, outputFileName, physicalPartitionId);
```
amberarrow: Since partitionedFileName is assigned in setup() there is no need to do it here; in fact, since these setters are typically called before the operator is serialized, physicalPartitionId will always be 0 here.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractSingleFileOutputOperator.java
```diff
@@ -39,10 +41,37 @@
...
+  /**
+   * Physical partition id for the current partition.
+   */
+  private transient int physicalPartitionId;
+
+  /**
+   * Initializing current partition id, partitionedFileName etc. {@inheritDoc}
+   */
+  @Override
+  public void setup(OperatorContext context)
```
amberarrow, line 66: I'm wondering why we even need this class; can we not simply add a boolean property to the base class called "appendPartitionId" ?

Also, the ticket description says "If SingleFileOutputOperator is partitioned into **more than one** physical instances then it should append partitionId to file name" (my emphasis), I don't see any check for that condition before appending the partition id.

otterc, line 66: @amberarrow AbstractSingleFileOutputOperator is not a new class. It was there from before.

amberarrow, line 66: Ok. Can we at least provide a way to turn off this behavior -- e.g. if the format string is null or the empty string, the operatorId is not appended.

otterc, line 66: Yes I think that should be the behavior

library/src/main/java/com/datatorrent/lib/io/fs/AbstractSingleFileOutputOperator.java
yogidevendra: My assumption was when we set property values for the operators in the running application using apexcli; then it will only invoke setter method and setup method will not be invoked. 
Hence, I am initializing derived field values in the setter of original property.

yogidevendra: Let me know if this assumption is incorrect.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractSingleFileOutputOperator.java
yogidevendra, line 66: My assumption is : if we have multiple partitions then each partition must write to different file. Thus giving this choice to the end use may not be good idea. 

Regarding adding operatorId suffix only in case when there are more than one partitions. This is good suggestion, but there is no direct way of getting number of physical instances for the operator in the setup method. We might need to add custom partitioner for doing this. IMO, it is over-complicating the thing for the delta value addition. 

Can you think of simpler way of knowing no. of physical partitions? OR do you feel that it should be OK without worrying too much about it?

amberarrow, line 66: The **user** knows whether there are multiple partitions or just one, we don't need to figure it out. In the latter case, they need to have a way of turning off this behavior (e.g. can set the format string to null). There is no harm in doing this if it is properly documented.

yogidevendra, line 66: OK. So, I will add check which says that if physicalPartitionId is 0 and format string is (empty or null) then do not append partition id.

Does that sounds good?

yogidevendra, line 66: OR you prefer to have explicit flag to turn this off? But, this flag will be honored only for single partition.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractSingleFileOutputOperator.java
amberarrow: This assumption is correct, my suggestion then is: Has this capability been tested with an actual application on a cluster (not just local mode) where the file name is changed mid-stream ? If so, suggest we document that this feature is supported and add a unit test that tests it.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractSingleFileOutputOperator.java
amberarrow, line 66: People have different preferences on this topic; my preference is a separate flag but I'm OK either way.

otterc, line 66: @yogidevendra IMO you can check if format string is empty string or null and decide whether to append partition id. A separate flag I think is redundant.

amberarrow, line 66: Yogi, if you can just implement the check if the format string is null or empty, we can merge this.  We can defer the issue of whether dynamically changing the file name is supported for the next release. We have no guarantee on the value of the operator id (not guaranteed to start at 0).

yogidevendra:
@amberarrow @chandnisingh  Added isEmpty check for the format string as per your suggestion.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractSingleFileOutputOperator.java
```diff
@@ -39,10 +43,42 @@
...
+  {
+    super.setup(context);
+    physicalPartitionId = context.getId();
+    if(StringUtils.isEmpty(partitionedFileNameformat)){
+      partitionedFileName = outputFileName;
+    }
+    else{
```
amberarrow: else on same line as the 2 braces with a space around it.

yogidevendra: Good catch. I will fix this. But, just wondering why checkstyle did not catch this?
Anything else?

amberarrow:
Looks good.

yogidevendra:
Build successful. 
@amberarrow Would you please merge this?

amberarrow:
Merged.

PR #274, opened 2016-05-14
SPOI-8049-Fixed getters and setters for topics and clusters

bhupeshchawda:


kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaInputOperator.java
```diff
@@ -413,33 +413,39 @@ public int getInitialPartitionCount()
...
   /**
-   * The topics the operator consumes, separate by','
+   * The topics the operator consumes
    * Topic name can only contain ASCII alphanumerics, '.', '_' and '-'
    */
-  public String getTopics()
+  public void setTopics(String[] topics)
+  {
+    ArrayList<String> temp = new ArrayList<>();
+    for (String topic: Arrays.asList(topics)) {
```
amberarrow: Why not iterate topics directly ?

bhupeshchawda: Yeah, will change it.
Was trying out something to do it without the loop, but forgot to revert the list.

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaInputOperator.java
```diff
@@ -413,33 +413,39 @@ public int getInitialPartitionCount()
...
+      if (topic != null && topic.length() != 0) {
+        temp.add(topic);
+      }
+    }
+    temp.toArray(this.topics);
```
amberarrow: Won't **this.topics** be null here since this function is supposed to initialize it ? If so, the **toArray()** call will throw a NPE.

tweise:
@siyuanh this looks like reverse of https://github.com/apache/incubator-apex-malhar/pull/176 - please have a look.

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaInputOperator.java
```diff
@@ -413,33 +413,39 @@ public int getInitialPartitionCount()
...
    *  Same setting as bootstrap.servers property to KafkaConsumer
    *  refer to http://kafka.apache.org/documentation.html#newconsumerconfigs
-   *  To support multi cluster, you can have multiple bootstrap.servers separated by ";"
+   *  To support multi cluster, you can have multiple elements in the array
    */
-  public String getClusters()
+  public void setClusters(String[] clusters)
   {
-    return Joiner.on(';').join(clusters);
+    this.clusters = clusters;
```
siyuanh, line 423: How do you separate nodes in each cluster string?

bhupeshchawda, line 423: Clusters are separated by comma. If we have multiple host:port pairs in each cluster, they can be specified as follows: "host1:port1, host2:port2", "host3:port3" where we use quotes to separate clusters.

bhupeshchawda:
Closing the pull request due to wrong branch name. Will reopen with the correct one.

PR #275, opened 2016-05-16
[Review Only] APEXMALHAR-2084-Fixed getters and setters for topics and clusters

bhupeshchawda:


kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaInputOperator.java
```diff
@@ -413,33 +413,39 @@ public int getInitialPartitionCount()
...
   }
 
   /**
-   * The topics the operator consumes, separate by','
+   * The topics the operator consumes
    * Topic name can only contain ASCII alphanumerics, '.', '_' and '-'
    */
-  public String getTopics()
+  public void setTopics(String[] topics)
+  {
```
amberarrow, line 436: You can avoid creation of the intermediate ArrayList and compact the array in place with something like this but
feel free to ignore if you like the original better:
{code}
int j = 0, len = topics.length;
for (int i = 0; i < len; ++i) {
    if (null == topics[i] || 0 == topics[i].length()) continue;
    if (i != j) topics[j] = topics[i];
    ++j;
}
this.topics = Arrays.copyOfRange(topics, 0, j);
{code}

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaInputOperator.java
```diff
@@ -413,33 +413,39 @@ public int getInitialPartitionCount()
...
    *  Same setting as bootstrap.servers property to KafkaConsumer
    *  refer to http://kafka.apache.org/documentation.html#newconsumerconfigs
-   *  To support multi cluster, you can have multiple bootstrap.servers separated by ";"
+   *  To support multi cluster, you can have multiple elements in the array
    */
-  public String getClusters()
+  public void setClusters(String[] clusters)
   {
-    return Joiner.on(';').join(clusters);
+    this.clusters = clusters;
```
siyuanh, line 423: How does user separate nodes in each cluster string?

bhupeshchawda, line 423: The clusters are separated by commas. In case the clusters themselves have multiple host:port pairs, the string can be specified using quotes as follows: "host1:port1, host2:port2", "host3:port3".

siyuanh, line 423: But does it work in properties.xml file? Can you give an example?

bhupeshchawda, line 423: Yes, it should work in xml. This is an example configuration for clusters: `"localhost:9092, localhost:9093", "localhost2:9092"`

tweise, line 423: Have you tested it? If so, would it make sense to include it into the javadoc? 

@siyuanh does this address the issue you were trying to fix previously?

siyuanh, line 423: @tweise This issue is reported by @bhupeshchawda that the operator won't show up in gateway because it violates the BEAN convention. I'm ok with this change, as long as it is configurable through configuration file and probably test the json application also to make sure the operator  can be used everywhere

bhupeshchawda, line 423: I have tested whether the configuration works in xml. This is verified using a application test. 
For UI component, I have just verified that it allows for one text box per element in the array. However, I think we should wait on this pull request until I can test it out end to end for the UI.

bhupeshchawda:
Closing this for now.

PR #277, opened 2016-05-17
APEXMALHAR-2093 Removed usages of Idempotent Storage Manager

otterc:
This is a backward incompatible change so added exclusions to japicmp plugin configuration.


sandeshh:
Is there a strong reason to make the backward incompatible change? 

How about following
1. Mark the idempotent Storage manager api deprecated.
2. IdempotentStorageManager will use the WindowDataManager ( Proxy pattern )
3. Expose WindowDataManager api

otterc:
@sandeshh 
- We have already marked the idempotent storage manager api deprecated.
- The 2nd point can be done but there will be so much of unused code because IdempotentStorageManager is not Evolving and has to be backward compatible.
- WindowDataManager API is already exposed.

tweise:
@chandnisingh please rebase

otterc:
@tweise rebased

PR #279, opened 2016-05-18
APEXMALHAR-2045: Adding bandwidth control feature to limit Input Oper…

DT-Priyanka:
…ator bandwidth limit

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthLimitingOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthLimitingOperator.java
@@ -1,0 +1,29 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.lib.bandwidth;
+
+import com.datatorrent.api.Operator;
+
+/**
+ * Operator which limits bandwidth consumption. It should have instance of BandwidthManager.
+ */
+public interface BandwidthLimitingOperator extends Operator
+{
+  BandwidthManager getBandwidthManager();
+}
--- a/library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthManager.java
+++ b/library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthManager.java
@@ -1,0 +1,131 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.lib.bandwidth;
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.datatorrent.api.Component;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.Context.OperatorContext;
+
+/**
+ * BandwidthManager keeps track of bandwidth consumption and provides limit on maximum bandwidth that can be consumed at
+ * any moment. This accumulates bandwidth upto certain limits so that accumulated bandwidth can be used over a period of
+ * time.
+ */
+public class BandwidthManager implements Component<Context.OperatorContext>
+{
+  private static final Logger LOG = LoggerFactory.getLogger(BandwidthManager.class);
+  /**
+   * Maximum bandwidth that can be consumed in bytes/sec
+   */
+  private long bandwidthLimit = Long.MAX_VALUE;
+  private transient long currentBandwidthConsumption;
+  private final transient ScheduledExecutorService scheduler;
+  private final transient Object lock = new Object();
+
+  public BandwidthManager()
+  {
+    scheduler = Executors.newScheduledThreadPool(1);
+  }
+
+  BandwidthManager(ScheduledExecutorService scheduler)
+  {
+    this.scheduler = scheduler;
+  }
+
+  @Override
+  public void setup(OperatorContext context)
+  {
+    scheduler.scheduleAtFixedRate(new Runnable()
+    {
+      @Override
+      public void run()
+      {
+        if (isBandwidthRestricted()) {
+          synchronized (lock) {
+            if (currentBandwidthConsumption < 0) {
+              currentBandwidthConsumption += bandwidthLimit;
+            } else {
+              currentBandwidthConsumption = bandwidthLimit;
+            }
+            LOG.debug("Available bandwidth: " + currentBandwidthConsumption);
+          }
+        }
+      }
+    }, 1, 1, TimeUnit.SECONDS);
+  }
+
+  public boolean canConsumeBandwidth()
+  {
+    if (!isBandwidthRestricted()) {
+      return true;
+    }
+    synchronized (lock) {
+      return currentBandwidthConsumption >= 0;
+    }
+  }
+
+  public void consumeBandwidth(long sentTupleSize)
+  {
+    if (isBandwidthRestricted()) {
+      synchronized (lock) {
+        currentBandwidthConsumption -= sentTupleSize;
+      }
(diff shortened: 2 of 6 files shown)


DT-Priyanka:
Reopening pull request for review after rebasing with latest master code.

library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthManager.java
```diff
@@ -0,0 +1,137 @@
...
+    LOG.info("Bandwidth limit is set to: " + bandwidth + " bytes/sec");
+  }
+
+  @Override
+  public void teardown()
+  {
+    scheduler.shutdownNow();
+  }
+
+  class BandwidthAccumulator implements Runnable
```
vrozov: Use anonymous class.

library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthManager.java
```diff
@@ -0,0 +1,137 @@
...
+  {
+    @Override
+    public void run()
+    {
+      if (isBandwidthRestricted()) {
+        synchronized (lock) {
+          if (currentBandwidthConsumption < 0) {
```
vrozov: This condition does not seem to be correct. Should not currentBandwidthConsumption be set to the limit when this method is invoked independently of how much bandwidth were consumed during the previous period?

DT-Priyanka: The idea here is to average out the bandwidth consumption over the time period. So when we start we do send the first batch assuming we have bandwidth. After that we wait till we pass some time to accumulate bandwidth, so overall bandwidth consumption comes to x bytes/sec on average.

vrozov: This will work when operator consumes more bandwidth than it is allowed to use. For cases where  operator does use all allowed bandwidth this approach does not work.

DT-Priyanka: Got it, updated as per your comments, setting currentBandwidth to "bandwidthLimit" when currentBandwidth is already +ve

DT-Priyanka:
@vrozov if changes looks good, can you please merge the PR?

library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthLimitingOperator.java
```diff
@@ -0,0 +1,29 @@
...
+package com.datatorrent.lib.bandwidth;
+
+import com.datatorrent.api.Operator;
+
+/**
+ * Operator which limits bandwidth consumption. It should have instance of BandwidthManager.
+ */
+public interface BandwidthLimitingOperator extends Operator
+{
+  public BandwidthManager getBandwidthManager();
```
vrozov: `public` is redundant.

library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthManager.java
vrozov: Is there a reason to make this class visible outside of BandwidthManager? Please make it either anonymous (preferable) or private.

library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthPartitioner.java
```diff
@@ -0,0 +1,78 @@
...
+  {
+    long currentBandwidth = partitions.iterator().next().getPartitionedInstance().getBandwidthManager().getBandwidth()
+        * partitions.size();
+    Collection<Partition<T>> newpartitions = super.definePartitions(partitions, context);
+    return updateBandwidth(newpartitions, currentBandwidth);
+  }
+
+  public Collection<Partition<T>> updateBandwidth(Collection<Partition<T>> newpartitions, long currentBandwidth)
+  {
+    LOG.info("Updating bandwidth of partitions.");
```
vrozov: Is this logging necessary? It does not seem to add much value.

library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthManager.java
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * Set maximum bandwidth that can be consumed in bytes/sec
+   *
+   * @param bandwidth
+   */
+  public void setBandwidth(long bandwidth)
+  {
+    this.bandwidthLimit = bandwidth;
+    LOG.info("Bandwidth limit is set to: " + bandwidth + " bytes/sec");
```
vrozov: Is INFO level necessary or it can be changed to DEBUG?

library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthManager.java
```diff
@@ -0,0 +1,139 @@
...
+    scheduler.scheduleAtFixedRate(new BandwidthAccumulator(), 1, 1, TimeUnit.SECONDS);
+  }
+
+  public boolean canConsumeBandwidth()
+  {
+    if (!isBandwidthRestricted()) {
+      return true;
+    }
+    synchronized (lock) {
+      if (currentBandwidthConsumption >= 0) {
```
vrozov: `return currentBandwidthConsumption >= 0`?

library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthManager.java
```diff
@@ -0,0 +1,139 @@
...
+    if (isBandwidthRestricted()) {
+      synchronized (lock) {
+        currentBandwidthConsumption -= sentTupleSize;
+      }
+    }
+  }
+
+  public boolean isBandwidthRestricted()
+  {
+    if (bandwidthLimit == Long.MAX_VALUE) {
```
vrozov: `return bandwidthLimit != Long.MAX_VALUE`?

library/src/test/java/com/datatorrent/lib/bandwidth/BandwidthManagerTest.java
```diff
@@ -0,0 +1,150 @@
...
+  public TestMeta testMeta = new TestMeta();
+
+  @Test
+  public void testBandwidthForLargeBlocks() throws InterruptedException
+  {
+    String data = "Tuple: test data to be emitted.";
+    long startTime = System.currentTimeMillis();
+    testMeta.underTest.consumeBandwidth(data.length());
+    while (!testMeta.underTest.canConsumeBandwidth()) {
+      Thread.sleep(1000);
```
vrozov: waiting for a second is not necessary here. Can it be replaced with a count instead?

library/src/test/java/com/datatorrent/lib/bandwidth/BandwidthManagerTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+public class BandwidthManagerTest
+{
+  private static class TestMeta extends TestWatcher
+  {
+    private String applicationPath;
+    private BandwidthManager underTest;
+    private Context.OperatorContext context;
+    private long bandwidthLimit = 10L;
+    private ScheduledExecutorTestService mockschedular;
```
vrozov: Consider making ScheduledExecutorTestService private or better use anonymous ScheduledExecutorService interface implementation.

library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthPartitioner.java
DT-Priyanka: Updating message to add new bandwidth value. This will be useful when operator undergoes partitioning and the bandwidth value is updated.

library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthManager.java
DT-Priyanka: changing it to debug level.

library/src/test/java/com/datatorrent/lib/bandwidth/BandwidthManagerTest.java
DT-Priyanka: I had introduced this sleep as schedular is suppose to update bandwidth after every one second. And I am mocking same behavior here.
I will introduce a constant than using hardcoded value.

library/src/test/java/com/datatorrent/lib/bandwidth/BandwidthManagerTest.java
DT-Priyanka: I guess it will be more readable if we keep it as indipendent class as size of class is pretty big. Also it is accessed from withing private class, so I am moving it at "default" access level in same file.

DT-Priyanka:
@vrozov  updated code to address review comments.

library/src/test/java/com/datatorrent/lib/bandwidth/BandwidthManagerTest.java
DT-Priyanka: I am getting some rat check errors, so moving back the class to it's own file. Can we leave it this way? Having around 90 lines of class as annonymous class won't be readable code.

library/src/main/java/com/datatorrent/lib/bandwidth/BandwidthPartitioner.java
vrozov: There is already logging in `setBandwidth()` method that is called for every partition from `updateBandwidth()`.

library/src/test/java/com/datatorrent/lib/bandwidth/BandwidthManagerTest.java
vrozov: There is no second thread and waiting for a second here only adds unnecessary wait time to overall test time. Using count instead of relying on System.currentTimeMillis() will not change test logic assuming that Thread.sleep(1000) correctly sleeps for 1 sec.

library/src/test/java/com/datatorrent/lib/bandwidth/BandwidthManagerTest.java
vrozov: Using anonymous interface implementation will be more readable as it indicates that 
1. ScheduledExecutorTestService is used in a single location only for `mockschedular` and it is not necessary to switch between files.
2. Only ScheduledExecutorService API is exposed by the ScheduledExecutorTestService.

rat only checks for apache license, please check that there are no files without the license header.

DT-Priyanka: updated the code to use anonymous classes.

vrozov:
Please address unnecessary sleep in the unit test. The rest looks good.

DT-Priyanka:
replaced sleep with counter.

PR #282, opened 2016-05-18
APEXMALHAR-2066 JdbcPolling,idempotent,partitionable

devtagare:
JdbcPolling,idempotent,partitionable #282

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
@@ -1,0 +1,661 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.lib.db.jdbc;
+
+import java.io.IOException;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.atomic.AtomicReference;
+
+import javax.validation.constraints.Min;
+
+import org.slf4j.LoggerFactory;
+
+import org.apache.apex.malhar.lib.wal.FSWindowDataManager;
+import org.apache.apex.malhar.lib.wal.WindowDataManager;
+import org.apache.commons.lang3.tuple.MutablePair;
+import org.apache.hadoop.classification.InterfaceStability.Evolving;
+
+import com.datatorrent.api.Context;
+import com.datatorrent.api.Context.OperatorContext;
+import com.datatorrent.api.DefaultPartition;
+import com.datatorrent.api.Operator.ActivationListener;
+import com.datatorrent.api.Operator.IdleTimeHandler;
+import com.datatorrent.api.Partitioner;
+import com.datatorrent.api.annotation.Stateless;
+import com.datatorrent.lib.db.AbstractStoreInputOperator;
+import com.datatorrent.lib.util.KeyValPair;
+import com.datatorrent.lib.util.KryoCloneUtils;
+import com.datatorrent.netlet.util.DTThrowable;
+
+/**
+ * Abstract operator for for consuming data using JDBC interface<br>
+ * User needs User needs to provide
+ * tableName,dbConnection,setEmitColumnList,look-up key <br>
+ * Optionally batchSize,pollInterval,Look-up key and a where clause can be given
+ * <br>
+ * This operator uses static partitioning to arrive at range queries for exactly
+ * once reads<br>
+ * This operator will create a configured number of non-polling static
+ * partitions for fetching the existing data in the table. And an additional
+ * single partition for polling additive data. Assumption is that there is an
+ * ordered column using which range queries can be formed<br>
+ * If an emitColumnList is provided, please ensure that the keyColumn is the
+ * first column in the list<br>
+ * Range queries are formed using the {@link JdbcMetaDataUtility}} Output -
+ * comma separated list of the emit columns eg columnA,columnB,columnC<br>
+ * Only newly added data which has increasing ids will be fetched by the polling
+ * jdbc partition
+ * 
+ * In the next iterations this operator would support an in-clause for
+ * idempotency instead of having only range query support to support non ordered
+ * key columns
+ * 
+ * 
+ * @displayName Jdbc Polling Input Operator
+ * @category Input
+ * @tags database, sql, jdbc, partitionable,exactlyOnce
+ */
+@Evolving
+public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInputOperator<T, JdbcStore>
+    implements ActivationListener<Context>, IdleTimeHandler, Partitioner<AbstractJdbcPollInputOperator<T>>
+{
+  /**
+   * poll interval in milliseconds
+   */
+  private static int pollInterval = 10000;
+
+  @Min(1)
+  private int partitionCount = 1;
+  protected transient int operatorId;
+  protected transient boolean isReplayed;
+  protected transient boolean isPollable;
+  protected int batchSize;
+  protected static int fetchSize = 20000;
+  /**
+   * Map of windowId to <lower bound,upper bound> of the range key
+   */
+  protected transient MutablePair<String, String> currentWindowRecoveryState;
+
+  /**
+   * size of the emit queue used to hold polled records before emit
+   */
+  private static int queueCapacity = 4 * 1024 * 1024;
+  private transient volatile boolean execute;
+  private transient AtomicReference<Throwable> cause;
+  protected transient int spinMillis;
+  private transient OperatorContext context;
+  protected String tableName;
+  protected String key;
+  protected long currentWindowId;
+  protected KeyValPair<String, String> rangeQueryPair;
+  protected String lower;
+  protected String upper;
+  protected boolean recovered;
+  protected boolean isPolled;
+  protected String whereCondition = null;
+  protected String previousUpperBound;
+  protected String highestPolled;
+  private static final String user = "";
+  private static final String password = "";
+  /**
+   * thread to poll database
+   */
+  private transient Thread dbPoller;
+  protected transient ArrayBlockingQueue<List<T>> emitQueue;
+  protected transient PreparedStatement ps;
+  protected WindowDataManager windowManager;
+  private String emitColumnList;
+
+  /**
+   * Returns the where clause
+   */
+  public String getWhereCondition()
+  {
+    return whereCondition;
+  }
+
+  /**
+   * Sets the where clause
+   */
+  public void setWhereCondition(String whereCondition)
+  {
+    this.whereCondition = whereCondition;
+  }
+
+  /**
+   * Returns the list of columns to select from the query
+   */
+  public String getEmitColumnList()
+  {
+    return emitColumnList;
+  }
+
+  /**
+   * Comma separated list of columns to select from the given table
+   */
+  public void setEmitColumnList(String emitColumnList)
+  {
+    this.emitColumnList = emitColumnList;
+  }
+
+  /**
+   * Returns the fetchsize for getting the results
+   */
+  public int getFetchSize()
+  {
+    return fetchSize;
+  }
+
+  /**
+   * Sets the fetchsize for getting the results
+   */
+  public void setFetchSize(int fetchSize)
+  {
+    this.fetchSize = fetchSize;
+  }
+
+  protected abstract void pollRecords(PreparedStatement ps);
+
+  /**
+   * Returns the interval for polling the queue
+   */
+  public int getPollInterval()
+  {
+    return pollInterval;
+  }
+
+  /**
+   * Sets the interval for polling the emit queue
+   */
+  public void setPollInterval(int pollInterval)
+  {
+    this.pollInterval = pollInterval;
+  }
+
+  /**
+   * Returns the capacity of the emit queue
+   */
+  public int getQueueCapacity()
+  {
+    return queueCapacity;
+  }
+
+  /**
+   * Sets the capacity of the emit queue
+   */
+  public void setQueueCapacity(int queueCapacity)
+  {
+    this.queueCapacity = queueCapacity;
+  }
+
+  /**
+   * Returns the ordered key used to generate the range queries
+   */
+  public String getKey()
+  {
+    return key;
+  }
+
+  /**
+   * Sets the ordered key used to generate the range queries
+   */
+  public void setKey(String key)
+  {
+    this.key = key;
+  }
+
+  /**
+   * Returns the tableName which would be queried
+   */
+  public String getTableName()
+  {
+    return tableName;
+  }
+
+  /**
+   * Sets the tableName to query
+   */
+  public void setTableName(String tableName)
+  {
+    this.tableName = tableName;
+  }
+
+  /**
+   * Returns rangeQueryPair - <lowerBound,upperBound>
+   */
+  public KeyValPair<String, String> getRangeQueryPair()
+  {
+    return rangeQueryPair;
+  }
+
+  /**
+   * Sets the rangeQueryPair <lowerBound,upperBound>
+   */
+  public void setRangeQueryPair(KeyValPair<String, String> rangeQueryPair)
+  {
+    this.rangeQueryPair = rangeQueryPair;
+  }
+
+  /**
+   * Returns batchSize indicating the number of elements in emitQueue
+   */
+  public int getBatchSize()
+  {
+    return batchSize;
+  }
+
+  /**
+   * Sets batchSize for number of elements in the emitQueue
+   */
+  public void setBatchSize(int batchSize)
+  {
+    this.batchSize = batchSize;
+  }
+
+  public AbstractJdbcPollInputOperator()
+  {
+    currentWindowRecoveryState = new MutablePair<>();
+    windowManager = new FSWindowDataManager();
+  }
+
+  @Override
+  public void setup(OperatorContext context)
+  {
+    super.setup(context);
+    spinMillis = context.getValue(Context.OperatorContext.SPIN_MILLIS);
+    execute = true;
+    cause = new AtomicReference<Throwable>();
+    emitQueue = new ArrayBlockingQueue<List<T>>(queueCapacity);
+    this.context = context;
+    operatorId = context.getId();
+
+    try {
+
+      //If its a range query pass upper and lower bounds
+      //If its a polling query pass only the lower bound
+      if (getRangeQueryPair().getValue() != null) {
+        ps = store.getConnection()
+            .prepareStatement(
+                JdbcMetaDataUtility.buildRangeQuery(getTableName(), getKey(), rangeQueryPair.getKey(),
+                    rangeQueryPair.getValue()),
+                java.sql.ResultSet.TYPE_FORWARD_ONLY, java.sql.ResultSet.CONCUR_READ_ONLY);
+      } else {
+        ps = store.getConnection().prepareStatement(
+            JdbcMetaDataUtility.buildPollableQuery(getTableName(), getKey(), rangeQueryPair.getKey()),
+            java.sql.ResultSet.TYPE_FORWARD_ONLY, java.sql.ResultSet.CONCUR_READ_ONLY);
+        isPollable = true;
+      }
+
+    } catch (SQLException e) {
+      LOG.error("Exception in initializing the range query for a given partition", e);
+      throw new RuntimeException(e);
+    }
+
+    windowManager.setup(context);
+    LOG.debug("super setup done...");
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
+    currentWindowId = windowId;
+
+    isReplayed = false;
+
+    if (currentWindowId <= windowManager.getLargestRecoveryWindow()) {
+      try {
+        replay(currentWindowId);
+      } catch (SQLException e) {
+        LOG.error("Exception in replayed windows", e);
+        throw new RuntimeException(e);
+      }
+    }
+
+    if (isReplayed && currentWindowId == windowManager.getLargestRecoveryWindow()) {
+      try {
+        if (!isPollable && rangeQueryPair.getValue() != null) {
+
+          ps = store.getConnection().prepareStatement(
+              JdbcMetaDataUtility.buildGTRangeQuery(getTableName(), getKey(), previousUpperBound,
+                  rangeQueryPair.getValue()),
+              java.sql.ResultSet.TYPE_FORWARD_ONLY, java.sql.ResultSet.CONCUR_READ_ONLY);
+        } else {
+          String bound = null;
+          if (previousUpperBound == null) {
+            bound = getRangeQueryPair().getKey();
+          } else {
+            bound = previousUpperBound;
+          }
+          ps = store.getConnection().prepareStatement(
+              JdbcMetaDataUtility.buildPollableQuery(getTableName(), getKey(), bound),
+              java.sql.ResultSet.TYPE_FORWARD_ONLY, java.sql.ResultSet.CONCUR_READ_ONLY);
+          isPollable = true;
+        }
+        isReplayed = false;
+        LOG.debug("Prepared statement after re-initialization - {} ", ps.toString());
+      } catch (SQLException e) {
+        // TODO Auto-generated catch block
+        throw new RuntimeException(e);
+      }
+    }
+
+    //Reset the pollable query with the updated upper and lower bounds
+    if (isPollable) {
+      try {
+        String bound = null;
+        if (previousUpperBound == null && highestPolled == null) {
+          bound = getRangeQueryPair().getKey();
+        } else {
+          bound = highestPolled;
+        }
+        ps = store.getConnection().prepareStatement(
+            JdbcMetaDataUtility.buildPollableQuery(getTableName(), getKey(), bound),
+            java.sql.ResultSet.TYPE_FORWARD_ONLY, java.sql.ResultSet.CONCUR_READ_ONLY);
+        LOG.debug("Polling query {} {}", ps.toString(), currentWindowId);
+        isPolled = false;
+      } catch (SQLException e) {
+        throw new RuntimeException(e);
+      }
+    }
+
+    lower = null;
+    upper = null;
+
+    //Check if a thread is already active and start only if its no
+    //Do not start the thread from setup, will conflict with the replay
+    if (dbPoller == null && !isReplayed) {
+      //If this is not a replayed state, reset the ps to highest read offset + 1, 
+      //keep the upper bound as the one that was initialized after static partitioning
+      LOG.info("Statement when re-initialized {}", ps.toString());
+      dbPoller = new Thread(new DBPoller());
+      dbPoller.start();
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


devtagare:
JdbcPolling,idempotent,partitionable

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,592 @@
...
+   */
+  private int pollInterval;
+
+  @Min(1)
+  private int partitionCount = 1;
+  protected transient int operatorId;
+  protected transient boolean isReplayed;
+  protected transient boolean isPollable;
+  protected int batchSize;
+  protected int fetchSize;
```
gauravgopi123: add javadocs for properties

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,592 @@
...
+  private static final String user = "user";
+  private static final String password = "password";
+  /**
+   * thread to poll database
+   */
+  private transient Thread dbPoller;
+  protected transient ArrayBlockingQueue<List<T>> emitQueue;
+  protected transient PreparedStatement ps;
+  protected WindowDataManager windowManager;
+  private String emitColumnList;
```
gauravgopi123, line 131: Same as above.. Add java docs for each of the properties

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,592 @@
...
+  protected transient MutablePair<String, String> currentWindowRecoveryState;
+
+  /**
+   * size of the emit queue used to hold polled records before emit
+   */
+  private int queueCapacity = 4 * 1024 * 1024;
+  private transient volatile boolean execute;
+  private transient AtomicReference<Throwable> cause;
+  protected transient int spinMillis;
+  private transient OperatorContext context = null;
```
gauravgopi123: no need to initialize to null

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,592 @@
...
+  protected String tableName;
+  protected String key;
+  protected transient long currentWindowId;
```
gauravgopi123: What happens when checkpoint happens between Application window? currentWindowId should not be transient in that case

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,592 @@
...
+    super.setup(context);
+    spinMillis = context.getValue(Context.OperatorContext.SPIN_MILLIS);
+    execute = true;
+    cause = new AtomicReference<Throwable>();
+    emitQueue = new ArrayBlockingQueue<List<T>>(queueCapacity);
+    this.context = context;
+    operatorId = context.getId();
+
+    try {
+      if (ps == null) {
```
gauravgopi123: When is ps not null?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,592 @@
...
+              java.sql.ResultSet.TYPE_FORWARD_ONLY, java.sql.ResultSet.CONCUR_READ_ONLY);
+        } else {
+          ps = store.getConnection().prepareStatement(
+              JdbcMetaDataUtility.buildPollableQuery(getTableName(), getKey(), rangeQueryPair.getKey()),
+              java.sql.ResultSet.TYPE_FORWARD_ONLY, java.sql.ResultSet.CONCUR_READ_ONLY);
+          isPollable = true;
+        }
+      }
+    } catch (SQLException e) {
+      LOG.error("Exception in initializing the range query for a given partition", e);
```
gauravgopi123, line 314: why swallow exception? Throw exception

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,592 @@
...
+  {
+    currentWindowId = windowId;
+
+    isReplayed = false;
+
+    if (currentWindowId <= windowManager.getLargestRecoveryWindow()) {
+      try {
+        replay(currentWindowId);
+      } catch (SQLException e) {
+        LOG.error("Exception in replayed windows");
```
gauravgopi123: No need to log exception

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,592 @@
...
+        throw new RuntimeException("saving recovery", e);
```
gauravgopi123: Is it "saving recovery" or replaying old data?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,592 @@
...
+    this.context = context;
+    operatorId = context.getId();
+
+    try {
+      if (ps == null) {
+        //If its a range query pass upper and lower bounds
+        //If its a polling query pass only the lower bound
+        if (getRangeQueryPair().getValue() != null) {
+          ps = store.getConnection().prepareStatement(
+              JdbcMetaDataUtility.buildRangeQuery(getTableName(), getKey(), rangeQueryPair.getKey(),
```
gauravgopi123: Do you need to initialize "ps" here? I see that it is being done in beginWindow() again

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcMetaDataUtility.java
```diff
@@ -0,0 +1,344 @@
...
+public class JdbcMetaDataUtility
+{
+  private static String DB_DRIVER = "com.mysql.jdbc.Driver";
+  private static String DB_CONNECTION = "";
+  private static String DB_USER = "";
+  private static String DB_PASSWORD = "";
+  private static String TABLE_NAME = "";
+  private static String KEY_COLUMN = "";
+  private static String WHERE_CLAUSE = null;
+  private static String COLUMN_LIST = null;
```
gauravgopi123, line 58: why are these fields static?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
devtagare: Yes.In begin window, a current partition starts with the most recent lower and upper bounds in-case of failure.In setup, the ps is initialized to the lower and upper bounds returned by the JdbcMetaDataUtility which is invoked from definePartitions().

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,652 @@
...
+      } else {
+        ps = store.getConnection().prepareStatement(
+            JdbcMetaDataUtility.buildPollableQuery(getTableName(), getKey(), rangeQueryPair.getKey()),
+            java.sql.ResultSet.TYPE_FORWARD_ONLY, java.sql.ResultSet.CONCUR_READ_ONLY);
+        isPollable = true;
+      }
+
+    } catch (SQLException e) {
+      LOG.error("Exception in initializing the range query for a given partition", e);
+      throw new RuntimeException();
```
gauravgopi123: throw new RuntimeException(e);

No need to log error

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,652 @@
...
+  {
+    currentWindowId = windowId;
+
+    isReplayed = false;
+
+    if (currentWindowId <= windowManager.getLargestRecoveryWindow()) {
+      try {
+        replay(currentWindowId);
+      } catch (SQLException e) {
+        LOG.error("Exception in replayed windows", e);
```
gauravgopi123, line 333: no need to log

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,652 @@
...
+          }
+          ps = store.getConnection().prepareStatement(
+              JdbcMetaDataUtility.buildPollableQuery(getTableName(), getKey(), bound),
+              java.sql.ResultSet.TYPE_FORWARD_ONLY, java.sql.ResultSet.CONCUR_READ_ONLY);
+          isPollable = true;
+        }
+        isReplayed = false;
+        LOG.debug("Prepared statement after re-initialization - " + ps.toString());
+      } catch (SQLException e) {
+        // TODO Auto-generated catch block
```
gauravgopi123, line 361: remove this

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,652 @@
...
+      throw new RuntimeException();
+    }
+
+    windowManager.setup(context);
+    LOG.debug("super setup done...");
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
```
gauravgopi123, line 324: begin window will not be called when checkpoint happens within Applicationwindow so how would replay happen?

May be you want to set @OperatorAnnotation(checkpointableWithinAppWindow = false) on the operator

devtagare, line 324: Good point.will add.

Thanks,
Dev

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,652 @@
...
+ * If an emitColumnList is provided, please ensure that the keyColumn is the
+ * first column in the list<br>
+ * Range queries are formed using the {@link JdbcMetaDataUtility}} Output -
+ * comma separated list of the emit columns eg columnA,columnB,columnC
+ * 
+ * @displayName Jdbc Polling Input Operator
+ * @category Input
+ * @tags database, sql, jdbc, partitionable,exactlyOnce
+ */
+public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInputOperator<T, JdbcStore>
```
gauravgopi123, line 83: How is AbstractJdbcPollInputOperator different from AbstractJdbcInputOperator?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
tweise: Should also wrap original exception, use Throwables.propagate

library/src/test/resources/log4j.properties
```diff
@@ -39,5 +39,5 @@ log4j.appender.SYSLOG.Facility=LOCAL1
 
 log4j.logger.org=info
 #log4j.logger.org.apache.commons.beanutils=warn
-log4j.logger.com.datatorrent=debug
-log4j.logger.org.apache.apex=debug
+log4j.logger.com.datatorrent=info
```
tweise: Why this change?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
devtagare, line 83: This operator supports polling,partitioning & idempotency.

gauravgopi123, line 83: why write another and not enhance existing? Can you please point me to the discussion that happened over @dev?

pramodin, line 83: @gauravgopi123 @devtagare Looks like this operator is not a drop in replacement for the existing one as it makes certain assumptions about the organization of data and expects certain extra information to be made available by the implementation. In that case it would break compatibility if we try to merge it into existing. Are there use cases where the current one works correctly considering fault tolerance etc? If not we can consider deprecating it after having a discussion on the dev list.

gauravgopi123, line 83: @PramodSSImmaneni : I could not find these assumptions documented in JIRA. For everyone's benefit JIRA should have detailed use case information and why existing (if any) operators can't address this usecase

devtagare, line 83: @gauravgopi123 agreed.Apologies for the same.I will add the assumptions and the use-case.

sandeepdeshmukh:
What are the challenges in extending AbstractJdbcInputOperator to support new features ? As AbstractJdbcInputOperator is Evolving, there is no backward compatibility issue.

pramodin:
@sandeepdeshmukh I will let @devtagare comment but I believe this operator makes assumptions on how data is organized in the database and expects the implementations to provide that information whereas the original operator makes no such assumptions.

gauravgopi123:
Any updates?

devtagare:
I have updated the JIRA with assumptions for the operator - https://issues.apache.org/jira/browse/APEXMALHAR-2066.

Waiting for comments from more folks before making final changes.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,652 @@
...
+  private transient OperatorContext context;
+  protected String tableName;
+  protected String key;
+  protected long currentWindowId;
+  protected KeyValPair<String, String> rangeQueryPair;
+  protected String lower;
+  protected String upper;
+  protected boolean recovered;
+  protected boolean isPolled;
+  protected String whereCondition = null;
```
bhupeshchawda, line 119: Can you also clarify which subset of SQL will be supported here? How complex can the where clause be?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,652 @@
...
+   * Sets batchSize for number of elements in the emitQueue
+   */
+  public void setBatchSize(int batchSize)
+  {
+    this.batchSize = batchSize;
+  }
+
+  public AbstractJdbcPollInputOperator()
+  {
+    this.pollInterval = 10 * 1000;
```
bhupeshchawda: Can these be declared as private static constants as defaults?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,652 @@
...
+  protected int fetchSize;
+  /**
+   * Map of windowId to <lower bound,upper bound> of the range key
+   */
+  protected transient MutablePair<String, String> currentWindowRecoveryState;
+
+  /**
+   * size of the emit queue used to hold polled records before emit
+   */
+  private int queueCapacity = 4 * 1024 * 1024;
```
bhupeshchawda: private static final int QUEUE_CAPACITY_DEFAULT?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,652 @@
...
+  private transient volatile boolean execute;
+  private transient AtomicReference<Throwable> cause;
+  protected transient int spinMillis;
+  private transient OperatorContext context;
+  protected String tableName;
+  protected String key;
+  protected long currentWindowId;
+  protected KeyValPair<String, String> rangeQueryPair;
```
bhupeshchawda, line 114: @NotNull?

gauravgopi123:
@devtagare : I wen through the Jira but it still doesn't mention how is this different from existing version and why can't existing version be enhanced?

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPollInputOperator.java
```diff
@@ -0,0 +1,226 @@
...
+        if (isPollable) {
+          highestPolled = key.toString();
+          isPolled = true;
+        }
+      }
+      /*Flush the remaining records once the result set is over and batch-size is not reached,
+       * Dont flush if its pollable*/
+      if (!hasNext) {
+        if ((isPollable && isPolled) || !isPollable) {
+          emitQueue.add(metaList);
```
bhupeshchawda: This might throw an exception if the queue is full. Use offer() instead?

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcMetaDataUtility.java
```diff
@@ -0,0 +1,344 @@
...
+    PreparedStatement preparedStatement = null;
+
+    try {
+      dbConnection = getDBConnection();
+      preparedStatement = dbConnection.prepareStatement(query);
+
+      ResultSet rs = preparedStatement.executeQuery();
+
+      while (rs.next()) {
+        rowCount = Long.parseLong(rs.getString("RowCount"));
```
bhupeshchawda: Suggest to use index instead of a column alias. Also seems like this is a simple count utility. Can you rename it to something like "getTupleCount()" ?

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcMetaDataUtility.java
```diff
@@ -0,0 +1,344 @@
...
+    PreparedStatement psUpperBound = null;
+
+    StringBuilder lowerBound = new StringBuilder();
+    StringBuilder upperBound = new StringBuilder();
+
+    KeyValPair<String, String> boundedQUeryPair = null;
+
+    try {
+      dbConnectionLower = getDBConnection();
+      dbConnectionUpper = getDBConnection();
```
bhupeshchawda: Do we need two database connections?

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcMetaDataUtility.java
```diff
@@ -0,0 +1,344 @@
...
+
+      lowerBound.append("SELECT " + KEY_COLUMN + " FROM " + TABLE_NAME);
+      upperBound.append("SELECT " + KEY_COLUMN + " FROM " + TABLE_NAME);
+
+      if (WHERE_CLAUSE != null) {
+        lowerBound.append(" WHERE " + WHERE_CLAUSE);
+        upperBound.append(" WHERE " + WHERE_CLAUSE);
+      }
+
+      lowerBound.append(" LIMIT " + (0 + lower) + ",1");
```
bhupeshchawda, line 178: LIMIT clause might not work with all databases. Need to document this.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,652 @@
...
+        if (recoveredData.right.equals(rangeQueryPair.getValue()) || recoveredData.right.equals(previousUpperBound)) {
+          LOG.info("Matched so returning");
+          return;
+        }
+
+        JdbcPollInputOperator jdbcPoller = new JdbcPollInputOperator();
+        jdbcPoller.setStore(store);
+        jdbcPoller.setKey(getKey());
+        jdbcPoller.setPartitionCount(getPartitionCount());
+        jdbcPoller.setPollInterval(50);
```
bhupeshchawda: Should this be a constant? If so, can you declare this as a _DEFAULT at the top?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,652 @@
...
+        jdbcPoller.setTableName(getTableName());
+        jdbcPoller.setBatchSize(getBatchSize());
+        isPollable = false;
```
bhupeshchawda: Is pollInterval needed if this is not pollable?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,652 @@
...
+
+  /**
+   * Replays the tuples in sync mode for replayed windows
+   */
+  public void emitReplayedTuples(PreparedStatement ps)
+  {
+    LOG.debug("Emitting replayed statement is -" + ps.toString());
+    ResultSet rs = null;
+    try (PreparedStatement pStat = ps;) {
+      pStat.setFetchSize(10000);
```
bhupeshchawda: Can this be picked up from some property. Else declare it as a constant at the top.

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcMetaDataUtility.java
```diff
@@ -0,0 +1,344 @@
...
+
+  private static Connection getDBConnection()
+  {
+
+    Connection dbConnection = null;
+
+    try {
+      Class.forName(DB_DRIVER);
+    } catch (ClassNotFoundException e) {
+      LOG.error("Driver not found", e);
```
bhupeshchawda, line 87: Rethrow the exception

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcMetaDataUtility.java
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    try {
+      dbConnection = DriverManager.getConnection(DB_CONNECTION, DB_USER, DB_PASSWORD);
+      return dbConnection;
+    } catch (SQLException e) {
+      LOG.error("Exception in getting connection handle", e);
```
bhupeshchawda, line 95: Re throw

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcMetaDataUtility.java
```diff
@@ -0,0 +1,344 @@
...
+
+      ResultSet rs = preparedStatement.executeQuery();
+
+      while (rs.next()) {
+        rowCount = Long.parseLong(rs.getString("RowCount"));
+        LOG.info("# Rows - " + rowCount);
+      }
+
+    } catch (SQLException e) {
+      LOG.error("Exception in retreiving result set", e);
```
bhupeshchawda, line 134: Re throw

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
bhupeshchawda, line 83: I suggest we mark these classes as `@Evolving` since these are new and likely to change.

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcMetaDataUtility.java
```diff
@@ -0,0 +1,344 @@
...
+
+    try {
+      dbConnection = getDBConnection();
+      preparedStatement = dbConnection.prepareStatement(query);
+
+      ResultSet rs = preparedStatement.executeQuery();
+
+      while (rs.next()) {
+        rowCount = Long.parseLong(rs.getString("RowCount"));
+        LOG.info("# Rows - " + rowCount);
```
bhupeshchawda: DOn't log this.

bhupeshchawda:
@devtagare Apart from the existing comments I am fine with the changes. 

The only major concern I have is related to the way polling is done might be dependent on a lot of things like the target database implementation and the organization of data in the database.
Request you to please mention the design of the operator (the core part on how you are planning to achieve polling functionality) on dev@apex. I guess, this will be done sooner or later as part of documentation :-) 
Apologies for the delay.

devtagare:
@bhupeshchawda - review comments are incorporated.Could you please check.

library/src/test/java/com/datatorrent/lib/db/jdbc/JdbcPollerTest.java
```diff
@@ -0,0 +1,246 @@
...
+    insertEventsInTable(10, 10);
+
+    AbstractJdbcPollInputOperator<Object> pollableInstance = itr.next().getPartitionedInstance();
+
+    pollableInstance.outputPort.setSink(sink1);
+
+    pollableInstance.beginWindow(1);
+    Thread.sleep(700);
+    pollableInstance.endWindow();
+    pollableInstance.emitTuples();
```
bhupeshchawda, line 215: This is emitting outside the window. Needs to be between begin and end window.

library/src/test/java/com/datatorrent/lib/db/jdbc/JdbcPollerTest.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    sink1.collectedTuples.clear();
+
+    insertEventsInTable(10, 20);
+
+    pollableInstance.beginWindow(2);
+    Thread.sleep(700);
+    pollableInstance.endWindow();
+    pollableInstance.emitTuples();
```
bhupeshchawda, line 232: Same here.

bhupeshchawda:
@devtagare I can still see a lot of open comments. Can you please address them?
Also, as I suggested, please provide the description of the design on the dev mailing list. I think this will serve the larger purpose of letting the community know about this operator and its use, in contrast to the existing Jdbc operator in Malhar.

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcMetaDataUtility.java
devtagare, line 87: Done

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcMetaDataUtility.java
devtagare, line 95: Done

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcMetaDataUtility.java
devtagare, line 178: Done

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
devtagare, line 114: At the start of the operator the rangeQueryPair would be null so a NotNull check will not work

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
devtagare, line 83: Done

devtagare:
@bhupeshchawda - fixed the latest comments and pushed the changes.Mail sent to dev@.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -0,0 +1,656 @@
...
+      while (execute) {
+        try {
+          long startTs = System.currentTimeMillis();
+          if ((isPollable && !isPolled) || !isPollable) {
+            pollRecords(ps);
+          }
+          long endTs = System.currentTimeMillis();
+          long ioTime = endTs - startTs;
+          long sleepTime = pollInterval - ioTime;
+          LOG.debug("pollInterval = " + pollInterval + " I/O time = " + ioTime + " sleepTime = " + sleepTime);
```
tushargosavi: use {} syntax for log message, the string gets evaluated even if log level is not enabled.

tweise:
Why would this not be part of the javadoc?

## 

sent from mobile

bhupeshchawda:
Yes, this should be part of javadoc as well.

PR #283, opened 2016-05-19
APEXMALHAR-2096: Add property blocksThreshold to limit input rate

DT-Priyanka:
The blocksThreshold is already exposed by FileSplitter operator, exposing same using module.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/FSInputModule.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/FSInputModule.java
@@ -58,7 +58,6 @@ public class FSInputModule implements Module
   private long blockSize;
   private boolean sequencialFileRead = false;
   private int readersCount;
-  @Min(1)
   protected int blocksThreshold;
 
   public final transient ProxyOutputPort<AbstractFileSplitter.FileMetadata> filesMetadataOutput = new ProxyOutputPort<>();
@@ -108,7 +107,6 @@ public class FSInputModule implements Module
     blockReader.setBasePath(files);
     if (readersCount != 0) {
       dag.setAttribute(blockReader, Context.OperatorContext.PARTITIONER, new StatelessPartitioner<FSSliceReader>(readersCount));
-      fileSplitter.setBlocksThreshold(readersCount);
     }
     if (blocksThreshold != 0) {
       fileSplitter.setBlocksThreshold(blocksThreshold);
@@ -257,8 +255,8 @@ public class FSInputModule implements Module
   }
 
   /**
-   * Sets number of blocks to be emitted per window. A lot of blocks emitted
-   * per window can overwhelm the downstream operators. This setting helps to control that.
+   * Sets number of blocks to be emitted per window.<br/>
+   * A lot of blocks emitted per window can overwhelm the downstream operators. This setting helps to control that.
    * @param threshold
    */
   public void setBlocksThreshold(int threshold)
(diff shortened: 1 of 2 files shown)


DT-Priyanka:
@chandnisingh  can you please review this?

library/src/main/java/com/datatorrent/lib/io/fs/FSInputModule.java
```diff
@@ -107,6 +110,9 @@ public void populateDAG(DAG dag, Configuration conf)
       dag.setAttribute(blockReader, Context.OperatorContext.PARTITIONER, new StatelessPartitioner<FSSliceReader>(readersCount));
       fileSplitter.setBlocksThreshold(readersCount);
     }
+    if (blocksThreshold != 0) {
```
gauravgopi123, line 111: Do you need this check as blocksThreshold is always >=1 according to constraint?

library/src/main/java/com/datatorrent/lib/io/fs/FSInputModule.java
```diff
@@ -57,6 +58,8 @@
   private long blockSize;
   private boolean sequencialFileRead = false;
   private int readersCount;
+  @Min(1)
+  protected int blocksThreshold;
```
gauravgopi123, line 61: why not initialize it to 1 so that user don't have to set anything if user goes with default value

gauravgopi123:
Test case?

otterc:
@DT-Priyanka 
Can we change the default value of `blocksThreshold` in AbstractFileSplitter. 
Currently it is Integer.MAX_VALUE which is too large and we have seen many a times the users forget to change this and the blockReaders are unable to handle the number of blocks.
Maybe we should reduce it to 10.
@vrozov can you please suggest a reasonable default value

DT-Priyanka:
As per my experience so far, 2 or 4 looks like a good number. @vrozov  or anyone has any other idea please let me know. I would remove min=1 constraint and put default in AbstractFileSplitter.

otterc:
@DT-Priyanka 
Why will you remove min=1 constraint?

DT-Priyanka:
Need to find a good default value and also update tests if required, closing this PR for time being.

vrozov:
I don't think there is a good default value. It depends on a run-time environment and the amount of processing necessary. In general, it will be better to change the behavior of the downstream operator to process as many records as it can in a single window, instead of trying to process all tuples emitted by FileSplitter.

otterc:
That breaks idempotency of the downstream.
Also not having a good default  value will still be applicable to Block reader. 

File splitter and block reader work in conjunction so to keep things simple and not break idempotency, this property IMO should be on splitter. 

Priyanka,  how about setting the default to  16?

DT-Priyanka:
If user end up setting static partitioning on Readers, and sets readers count to a really small value say 1 or 2, the readers will be overloaded, reading 16 blocks in a window, and we can still see this problem.
Couple of options:
1. Force user to set blocksThreshold (by marking it as notNull). Here we will have to put good documentation for what this property actually means and how other parameters like readers count, block size etc can affect this parameter value. (Vlad also suggested we can make this property compulsory parameter).
2. Looking at number of readers and block size we do some calculations and set the blocksThreshold ourselves. Will need to handle dynamic partitioning scenarios gracefully here.

otterc:
Priyanka, 
Point 2 is already handled.  

I agree with point 1.

vrozov:
@chandnisingh can you please elaborate how item 2 is already handled.

otterc:
@vrozov dynamic partition of block readers is explained in the documentation of this operator. For changing the block threshold dynamically on splitter there is a ticket open.

PR #287, opened 2016-05-23
APEXMALHAR-2096: Add property blocksThreshold to limit input rate to FSInputModule

DT-Priyanka:


DT-Priyanka:
Created pull request based on discussions on #283

DT-Priyanka:
@chandnisingh  can you please review it?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileSplitter.java
```diff
@@ -53,6 +53,7 @@
    * This is a threshold on the no. of blocks emitted per window. A lot of blocks emitted
    * per window can overwhelm the downstream operators. This setting helps to control that.
    */
+  @NotNull
```
gauravgopi123: Is this required?

library/src/main/java/com/datatorrent/lib/io/fs/FSInputModule.java
```diff
@@ -59,6 +58,9 @@
   private long blockSize;
   private boolean sequencialFileRead = false;
   private int readersCount;
+  @NotNull
```
gauravgopi123: Is this required?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileSplitter.java
DT-Priyanka: As per discussions on pull request #283 we decided to make blocksThreshold as compulsory parameter, so user always sets it.

otterc: @DT-Priyanka 
It is of type primitive int. So default value will be 0. Since the min is 1, it will be necessary for the user to set it. @NotNull is not relevant here.

DT-Priyanka: Yes right, let me change datatype to Integer.

otterc: Please don't change the type.  Having the constraint min(1) is sufficient

DT-Priyanka: As per my understanding that won't make it as compulsory parameter.

library/src/main/java/com/datatorrent/lib/io/fs/FSInputModule.java
```diff
@@ -59,6 +58,9 @@
...
+  @Min(1)
+  protected Integer blocksThreshold;
```
gauravgopi123: why Integer type and not primitive int type?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileSplitter.java
otterc: It will because min constraint is 1. By default the field will be assigned 0. So the user will have to change the value.  Please don't change types

DT-Priyanka:
Got it updated the datatype.

PR #289, opened 2016-05-25
APEXMALHAR-2087 Hive output module

yogidevendra:
1. Added Hive output module
2. Minor enhancements for getters, setters
3. Fixing some import order, checkstyle violations


yogidevendra:
@ashwinchandrap Could you please review this?

contrib/src/main/java/com/datatorrent/contrib/hive/FSPojoToHiveOperator.java
```diff
@@ -40,14 +43,14 @@
  * @tags fs, hive, database
  * @since 3.0.0
  */
-public class FSPojoToHiveOperator extends AbstractFSRollingOutputOperator<Object>
+public class FSPojoToHiveOperator<T> extends AbstractFSRollingOutputOperator<T>
```
tweise: What's the purpose of the type parameter?

contrib/src/main/java/com/datatorrent/contrib/hive/HiveOutputModule.java
```diff
@@ -0,0 +1,476 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.datatorrent.contrib.hive;
```
tweise: New classes should be created in appropriate package.

tweise:
Consider moving this to a separate module with clean dependency setup.

tweise:
Meant to say move all the Hive support to that separate module.

yogidevendra:
If we decide to move hive support to separate pom project:
1. Should it be under malhar/hive similar to malhar/kafka?
There will be too many top level folders under malhar if we decide to do it for all connectors.
 OR it should be malhar/contrib/hive? 
2. Should we mark existing classes for Hive under contrib as deprecated in favor of new module? (to maintain backward compatibility?)
3. Should new module have copy of the existing code with needed modifications? OR we should refer to existing classes from new module?

tweise:
It should move out of contrib. For now, I would suggest to put it top level, we can move the entire module easily once we break out more of the connectors.

I recommend to move the existing code to new module, and not to make copies. Also please do it in a way that preserves attribution.

yogidevendra:
But, then existing applications having dependency on malhar-contrib-3.4.0 would expect hive support would continue as it is in malhar-contrib-3.5.0. They would have to change pom dependency to malhar-hive-3.5.0 to get the hive support. Is this OK? 
I observed that some kafka related classes are still residing under malhar/contrib even if we have malhar/kafka. I thought that this was done to retain original classes for backward compatibility. Am I missing something?

tweise:
It is OK for existing apps (are there any?) to change the dependency in pom.xml when moving to a newer version. We have separate implementations for Kafka 0.8 and 0.9. We left the 0.8 code where it was.

hive/pom.xml
```diff
@@ -0,0 +1,265 @@
...
+  <packaging>jar</packaging>
+
+  <properties>
+    <checkstyle.console>true</checkstyle.console>
+  </properties>
+
+  <build>
+    <plugins>
+      <!-- Publish tests jar -->
+      <plugin>
```
tweise, line 42: Do we need this here?

hive/pom.xml
```diff
@@ -0,0 +1,265 @@
...
+          <execution>
+            <goals>
+              <goal>test-jar</goal>
+            </goals>
+            <phase>package</phase>
+          </execution>
+        </executions>
+      </plugin>
+     <!-- create resource directory for xml javadoc-->
+     <plugin>
```
tweise, line 56: It would be terrific if we can centralize this doclet boilerplate code that repeats in very module. Related to https://issues.apache.org/jira/browse/APEXCORE-435

hive/pom.xml
```diff
@@ -0,0 +1,265 @@
...
+     </plugin>
+    </plugins>
+  </build>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.hive</groupId>
+      <artifactId>hive-exec</artifactId>
+      <version>0.13.1</version>
+      <optional>true</optional>
```
tweise: Why optional? Dependencies should be defined so that user of module does not have to repeat them (different from what was done in contrib).

hive/pom.xml
yogidevendra, line 56: I would need some guidance on what needs to be done. I suggest to defer this in a separate PR.

yogidevendra:
@tweise incorporated review comments except for the test-jar related bolierplate code. Could you please put appropriate comments in APEXCORE-435 to mention what needs to be done for that?

If everything else looks OK; could you please merge this PR?

yogidevendra:
@tweise could you please review this?

yogidevendra:
@tweise Do you prefer to get this reviewed from some else? Would you please mention best person to review this change?

PR #292, opened 2016-05-25
APEXMALHAR-1957: Updated HBase Input Operator

bhupeshchawda:
Added threading for reading data from hbase. Added support for progressive read. Added unit tests.

Reopened an old PR to correct the branch name. Here is the old PR: https://github.com/apache/incubator-apex-malhar/pull/212

### DIFF
--- a/contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseFieldValueGenerator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseFieldValueGenerator.java
@@ -1,0 +1,59 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.contrib.hbase;
+
+import java.util.List;
+
+import com.datatorrent.lib.util.FieldValueGenerator;
+import com.datatorrent.lib.util.PojoUtils;
+
+/**
+ * A {@link FieldValueGenerator} implementation for {@link HBaseFieldInfo}
+ */
+public class HBaseFieldValueGenerator extends FieldValueGenerator<HBaseFieldInfo>
+{
+  public static final String COLON = ":";
+
+  @SuppressWarnings("unchecked")
+  protected HBaseFieldValueGenerator(final Class<?> clazz, List<HBaseFieldInfo> fieldInfos)
+  {
+    for (HBaseFieldInfo fieldInfo : fieldInfos) {
+      fieldInfoMap.put(fieldInfo.getFamilyName() + COLON + fieldInfo.getColumnName(), fieldInfo);
+
+      PojoUtils.Getter<Object, Object> getter =
+          PojoUtils.createGetter(clazz, fieldInfo.getPojoFieldExpression(), fieldInfo.getType().getJavaType());
+      fieldGetterMap.put(fieldInfo, getter);
+    }
+
+    for (HBaseFieldInfo fieldInfo : fieldInfos) {
+      PojoUtils.Setter<Object, Object> setter =
+          PojoUtils.createSetter(clazz, fieldInfo.getPojoFieldExpression(), fieldInfo.getType().getJavaType());
+      fieldSetterMap.put(fieldInfo, setter);
+    }
+  }
+
+  public void setColumnValue(Object instance, String columnName, String columnFamily, Object value,
+      ValueConverter<HBaseFieldInfo> valueConverter)
+  {
+    HBaseFieldInfo fieldInfo = fieldInfoMap.get(columnFamily + COLON + columnName);
+    PojoUtils.Setter<Object, Object> setter = fieldSetterMap.get(fieldInfo);
+    setter.set(instance, valueConverter == null ? value : valueConverter.convertValue(fieldInfo, value));
+  }
+
+}
--- a/contrib/src/main/java/com/datatorrent/contrib/hbase/HBasePOJOInputOperator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/hbase/HBasePOJOInputOperator.java
@@ -18,14 +18,14 @@
  */
 package com.datatorrent.contrib.hbase;
 
-import java.io.IOException;
 import java.util.List;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.apache.hadoop.classification.InterfaceStability.Evolving;
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.CellUtil;
 import org.apache.hadoop.hbase.client.Result;
-import org.apache.hadoop.hbase.client.ResultScanner;
 import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.util.Bytes;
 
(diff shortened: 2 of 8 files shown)


contrib/src/main/java/com/datatorrent/contrib/hbase/HBasePOJOInputOperator.java
```diff
@@ -37,20 +37,21 @@
...
 @Evolving
-public class HBasePOJOInputOperator extends HBaseInputOperator<Object>
+public class HBasePOJOInputOperator extends HBaseScanOperator<Object>
 {
   private TableInfo<HBaseFieldInfo> tableInfo;
-  protected HBaseStore store;
   private String pojoTypeName;
-  private String startRow;
-  private String lastReadRow;
+  private Scan scan;
```
bhupeshchawda: Transient

contrib/src/main/java/com/datatorrent/contrib/hbase/HBasePOJOInputOperator.java
```diff
@@ -69,116 +70,105 @@ public Object convertValue( HBaseFieldInfo fieldInfo, Object value)
   public void setup(OperatorContext context)
   {
     try {
-      store.connect();
+      super.setup(context);
       pojoType = Class.forName(pojoTypeName);
```
bhupeshchawda: use attrib
ute of port

contrib/src/main/java/com/datatorrent/contrib/hbase/HBasePOJOInputOperator.java
```diff
@@ -69,116 +70,105 @@ public Object convertValue( HBaseFieldInfo fieldInfo, Object value)
...
-        String readRow = Bytes.toString(result.getRow());
-        if( readRow.equals( lastReadRow ))
-          continue;
-
-        Object instance = pojoType.newInstance();
-        rowSetter.set(instance, readRow);
-
-        List<Cell> cells = result.listCells();
+      String readRow = Bytes.toString(result.getRow());
+      if( readRow.equals( getLastReadRow() )) {
```
bhupeshchawda, line 99: Do this in base class

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseScanOperator.java
```diff
@@ -40,27 +51,107 @@
...
+  {
+    super.beginWindow(windowId);
+    tuplesRead = 0;
+  }
 
   @Override
   public void emitTuples()
   {
+    if (!readThread.isAlive() && threadFailed) {
+      throw new RuntimeException("Exception in scan thread");
```
bhupeshchawda: capture exception from thread and throw the right onw

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseScanOperator.java
```diff
@@ -40,27 +51,107 @@
...
     } catch (Exception e) {
-      e.printStackTrace();
+      throw new RuntimeException(e);
     }
   }
 
+  @Override
+  public void deactivate()
+  {
+    readThread.interrupt();
```
bhupeshchawda, line 166: Make sure thread knows when it is a interrupt from deactivate

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseFieldValueGenerator.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @SuppressWarnings("unchecked")
+  protected HBaseFieldValueGenerator(final Class<?> clazz, List<HBaseFieldInfo> fieldInfos)
+  {
+    for (HBaseFieldInfo fieldInfo : fieldInfos) {
+      fieldInfoMap.put(fieldInfo.getFamilyName() + COLON + fieldInfo.getColumnName(), fieldInfo);
+
+      PojoUtils.Getter<Object, Object> getter =
+          PojoUtils.createGetter(clazz, fieldInfo.getPojoFieldExpression(), fieldInfo.getType().getJavaType());
+      fieldGetterMap.put(fieldInfo, getter);
```
chinmaykolhatkar, line 41: Would it be more efficient if the key of fieldGetterMap is made as familyName:columnName instead of any object?

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseFieldValueGenerator.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    for (HBaseFieldInfo fieldInfo : fieldInfos) {
+      PojoUtils.Setter<Object, Object> setter =
+          PojoUtils.createSetter(clazz, fieldInfo.getPojoFieldExpression(), fieldInfo.getType().getJavaType());
+      fieldSetterMap.put(fieldInfo, setter);
```
chinmaykolhatkar, line 47: Would it be more efficient if the key of fieldSetterMap is made as familyName:columnName instead of any object?

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseScanOperator.java
chinmaykolhatkar: Should we not ideally restart the thread and let the operator continue running?

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseScanOperator.java
chinmaykolhatkar, line 166: Please don't interrupt I would suggest. Have a variable in thread which is set to true for running and false for existing. Then wait for the operation to finish.

chinmaykolhatkar, line 166: Also scanner needs to be closed on exit of thread.

chinmaykolhatkar:
@bhupeshchawda , In general can you describe, what is the advantage of scanning in single another thread instead of main operator thread? Does this approach shows performance improvements?

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseScanOperator.java
```diff
@@ -79,4 +170,79 @@ public void emitTuples()
...
+   * Sets the last read row key from the hbase table. After the failures, the new scan will start from this row key
+   * @param lastReadRow
+   */
+  public void setLastReadRow(String lastReadRow)
+  {
+    this.lastReadRow = lastReadRow;
+  }
+
+  /**
+   * Returns the Scan HINT_LOOKAHEAD parameter as configured. Default is {@value #DEF_HINT_SCAN_LOOKAHEAD}
```
chinmaykolhatkar: Can you please check once that HINT_LOOKAHEAD feature is available in all HBase client and server versions?
If not, can you please describe this operator as works best for certain version and above for all other versions this field will be ignored?

contrib/src/main/java/com/datatorrent/contrib/hbase/HBasePOJOInputOperator.java
chinmaykolhatkar: Yes please.

contrib/src/main/java/com/datatorrent/contrib/hbase/HBasePOJOInputOperator.java
```diff
@@ -69,116 +70,105 @@ public Object convertValue( HBaseFieldInfo fieldInfo, Object value)
   public void setup(OperatorContext context)
   {
     try {
-      store.connect();
+      super.setup(context);
       pojoType = Class.forName(pojoTypeName);
       pojoType.newInstance();   //try create new instance to verify the class.
       rowSetter = PojoUtils.createSetter(pojoType, tableInfo.getRowOrIdExpression(), String.class);
-      fieldValueGenerator = FieldValueGenerator.getFieldValueGenerator(pojoType, tableInfo.getFieldsInfo() );
+      fieldValueGenerator = HBaseFieldValueGenerator.getHBaseFieldValueGenerator(pojoType, tableInfo.getFieldsInfo() );
```
chinmaykolhatkar: You could just use the constructor of HBaseFieldValueGenerator... Do you see any advantage is having another public static entry point for generating this object?

contrib/src/main/java/com/datatorrent/contrib/hbase/HBasePOJOInputOperator.java
```diff
@@ -37,20 +37,21 @@
...
 {
   private TableInfo<HBaseFieldInfo> tableInfo;
-  protected HBaseStore store;
   private String pojoTypeName;
-  private String startRow;
-  private String lastReadRow;
+  private Scan scan;
 
+  // Transients
   protected transient Class pojoType;
```
chinmaykolhatkar: Can you make pojoType, fieldValueGenerator & valueConverter private?
I don't see this being used outside of this class.

contrib/src/main/java/com/datatorrent/contrib/hbase/HBasePOJOInputOperator.java
bhupeshchawda, line 99: This will need changes to the signature of getTuple. Here we are assuming that getTuple will extract the Hbase row and then convert it to the output type. However, the lastReadRow is stored in the raw form. In order to move this to base class, we'll have to map the tuple to a raw type and then check if the record is the one we saw last.

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseScanOperator.java
bhupeshchawda: I think if we should not just restart the thread without understanding what the issue was. In any case, this exception will ultimately kill the operator and restart it from a consistent state instead.

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseScanOperator.java
bhupeshchawda, line 166: I have introduced a control variable for running and pausing the thread. However, the interrupt in deactivate would not cause any problems, since it is not affecting the state of the application at all. In normal functioning, we never stop the running thread, just pause it.

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseScanOperator.java
bhupeshchawda: Agreed.

bhupeshchawda: It seems this is now deprecated and optimization will happen automatically. Will remove this from the changes.

contrib/src/main/java/com/datatorrent/contrib/hbase/HBasePOJOInputOperator.java
bhupeshchawda: Done

contrib/src/main/java/com/datatorrent/contrib/hbase/HBasePOJOInputOperator.java
bhupeshchawda: Did not change this as not related to the current PR.

bhupeshchawda:
@chinmaykolhatkar the purpose was to make sure that the lags due to hbase scan do not affect the operator thread. Discussions on dev@apex here: http://mail-archives.apache.org/mod_mbox/apex-dev/201512.mbox/%3CCABAipVZWG5zWuD4ne89mU0ZR7_jQE1e2zg50e2H2fLF4mWpzMA@mail.gmail.com%3E

chinmaykolhatkar:
Changes looks functionally fine. But this operator is not idempotent and so was previous implementation before this PR.
As this PR is dangling for a long time, I suggest to merge this PR and open a new Jira to make Hbase Input operator idempotent.

Please share your thoughts on the same.

I'll wait for couple of days to get opinion on above approach or else will merge this and open a new Jira for idempotency.

PR #227, opened 2016-04-01
APEXMALHAR-1966: Update casandra output opreator

DT-Priyanka:
Changes:
1. Accept user queries (can be insert/update/delete)
2. Use default field info, as per pojo fields if fieldInfos is not set.
3. Add metrics.

PR #295, opened 2016-05-26
APEXMALHAR-1966: Update cassandra output operator

DT-Priyanka:
1. By default populating field info from POJO
2. Option to add update/insert query


DT-Priyanka:
#227 was closed and this one is open with same code changes to address rebase issues.

DT-Priyanka:
@bhupeshchawda  please review again and merge the changes.

benchmark/src/main/java/com/datatorrent/benchmark/cassandra/CassandraOutputOperator.java
```diff
@@ -48,4 +48,9 @@ protected Statement setStatementParameters(PreparedStatement updateCommand,
     return boundStmnt.bind(id++,tuple);
   }
 
+  @Override
+  public void deactivate()
```
bhupeshchawda: Can you move this to the Abstract parent class? Otherwise every implementation would have to include this method.

contrib/src/main/java/com/datatorrent/contrib/cassandra/CassandraPOJOOutputOperator.java
```diff
@@ -162,35 +160,75 @@ public void activate(Context.OperatorContext context)
...
+
+  private PreparedStatement prepareStatementFromFieldsAndTableName()
+  {
+    if (tablename == null || tablename.length() == 0) {
+      throw new RuntimeException("Please sepcify query or table name.");
+    }
     StringBuilder queryfields = new StringBuilder();
     StringBuilder values = new StringBuilder();
-    for (FieldInfo fieldInfo: fieldInfos) {
+    for (FieldInfo fieldInfo : fieldInfos) {
```
bhupeshchawda: Please remove checkstyle related changes from this PR

contrib/src/main/java/com/datatorrent/contrib/cassandra/CassandraPOJOOutputOperator.java
```diff
@@ -162,35 +160,75 @@ public void activate(Context.OperatorContext context)
...
         values.append(",").append("?");
       }
     }
-    String statement
-            = "INSERT INTO " + store.keyspace + "."
-            + tablename
-            + " (" + queryfields.toString() + ") "
-            + "VALUES (" + values.toString() + ");";
-    LOG.debug("statement is {}", statement);
+    String statement = "INSERT INTO " + store.keyspace + "." + tablename + " (" + queryfields.toString() + ") " + "VALUES (" + values.toString() + ");";
```
bhupeshchawda: Checkstyle.
Also why remove logging?

contrib/src/main/java/com/datatorrent/contrib/cassandra/CassandraPOJOOutputOperator.java
```diff
@@ -162,35 +160,75 @@ public void activate(Context.OperatorContext context)
...
+  {
+    fieldInfos = Lists.newArrayList();
+    Field[] fields = pojoClass.getDeclaredFields();
+    for (int i = 0; i < rsMetaData.size(); i++) {
+      String columnName = rsMetaData.getName(i);
+      String pojoField = getMatchingField(fields, columnName);
+      if (pojoField != null && pojoField.length() != 0) {
+        fieldInfos.add(new FieldInfo(columnName, pojoField, null));
+      } else {
+        LOG.error("Couldn't find corrosponding pojo field for column: " + columnName);
```
bhupeshchawda: This will insert null for the column which is not found in POJO. This might cause issues with any not null constraints in cassandra tables?

library/src/main/java/com/datatorrent/lib/db/AbstractTransactionableStoreOutputOperator.java
```diff
@@ -67,6 +69,9 @@ public void process(T t)
 
   };
 
+  @OutputPortFieldAnnotation(error = true)
+  public final transient DefaultOutputPort<T> error = new DefaultOutputPort<>();
```
bhupeshchawda: Can you move this error port to the class where it is actually used?

contrib/src/main/java/com/datatorrent/contrib/cassandra/CassandraPOJOOutputOperator.java
```diff
@@ -50,17 +48,21 @@
  * @since 2.1.0
  */
 @Evolving
-public class CassandraPOJOOutputOperator extends AbstractCassandraTransactionableOutputOperatorPS<Object> implements Operator.ActivationListener<Context.OperatorContext>
+public class CassandraPOJOOutputOperator extends AbstractCassandraTransactionableOutputOperatorPS<Object>
```
bhupeshchawda: Can you check why we have almost similar abstract classes: AbstractCassandraTransactionableOutputOperator and AbstractCassandraTransactionableOutputOperatorPS ?

contrib/src/main/java/com/datatorrent/contrib/cassandra/CassandraPOJOOutputOperator.java
DT-Priyanka: I haven't tried it yet, but if not null constraint is defined it should throw exception. Ideally the app developer should populate the fields as per the table schema.
We can do a quick schema check and throw exception before hand but I wonder will that do much of value addition?

contrib/src/main/java/com/datatorrent/contrib/cassandra/CassandraPOJOOutputOperator.java
DT-Priyanka: removing PS operator and updating the base operator.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractCassandraTransactionableOutputOperator.java
```diff
@@ -48,29 +54,58 @@
...
    * Sets the parameter of the insert/update statement with values from the tuple.
    *
    * @param tuple     tuple
    * @return statement The statement to execute
    * @throws DriverException
+   *
+   * @deprecated since 3.5.0.0 use {@link #setStatementParameters(PreparedStatement, T)} instead.<br/>
    */
+  @Deprecated
   protected abstract Statement getUpdateStatement(T tuple) throws DriverException;
```
bhupeshchawda: Will removing this cause any issues with backward compatibility. Keeping this (even deprecated) may require us to ensure that calls to `getUpdateStatement()` are successful.
@PramodSSImmaneni What is the right way to get rid of this method?

DT-Priyanka: I will remove this method for now, if we want to keep it I will have to provide method implementation for setStatementParameters and call "getUpdateStatement" from within setStatementParameters. That means I can't mark setStatementParameters as abstract.

We are anyway removing AbstractCassandraTransactionableOutputOperatorPS. Also the class is marked as evolving. So I guess it would be okay to remove "getUpdateStatement" method.

@PramodSSImmaneni  @tweise  please revert with right option to avoid backword compatibility issues.

DT-Priyanka:
failed due to Kafka test failure.

PR #297, opened 2016-05-26
[APEXMALHAR-2102] Add ClonePartitioner

ilganeli:
**Note:** This is still marked as an Apex issue even though it should live in Malhar. Once the JIRA is updated I'll update title and reference. 
- Created a Clone partitioner similar to the StatelessPartitioner that assigns all data to all partitions
- Added a simple unit test suite to test scale up and scale down

**Questions**
Is there a good place/way to add a test to verify that data is partitioned appropriately? For example, the stateless partitioner assigns keys based on the nearest power of two - is that validated anywhere?

How is the serialVersionUUID generated?

Does the manual assignment of ports to PartitionKeys need to happen for each Partition? There is a comment in the Partition class that by default all data is sent to all partitions. However, digging into the implementation in depth doesn't show that behavior. Instead, it seems data is assigned based on the mask associated with each Partition. 

This references: https://issues.apache.org/jira/browse/APEXMALHAR-2102


ilganeli:
retest this please

ilganeli:
Hi all - any chance of a review?

gauravgopi123:
Please add test case where this Partitioner is used in a DAG and make sure that all the partitions receive all the tuples..You can refer this https://github.com/apache/apex-core/blob/master/engine/src/test/java/com/datatorrent/stram/StramLocalClusterTest.java

ilganeli:
@gauravgopi123 Thanks, how could I actually validate that partitions receive individual tuples? Is there a preferred way to do this?

gauravgopi123:
Have an input operator that sends n number of tuples every window then all the partitions of downstream partitioned operator should receive same n tuples in every window.

ilganeli:
@gauravgopi123 My question was in regards to how can I actually capture the tuples seen by each partition? I'm not aware of any mechanism that lets you do this and I haven't been able to find a comparable example in the unit tests. The only thing I've seen is tests using "sinks" which don't quite work here.

gauravgopi123:
@ilganeli : The dummy operator that will be partitioned can cache the tuples coming to it..
For Eg. Input operator always generates say 26 tuples from A..Z in a given window.. Dummy operator should in EndWindow make sure that tuples received are 26 and they are in order A..Z o/w throw exception.. You can capture this error in a static variable in DummyOperator. This will code will run in all instances of Dummy operator and after DAG runs make sure that Exception variable is null....

chinmaykolhatkar:
Hi @ilganeli ,

I see this PR is closed without taking the Jira to a resolution.

Would you please opening a new PR for this?

Thanks,
Chinmay.

PR #298, opened 2016-05-27
[APEXMALHAR-2086] Kafka output operator: 0.9.0

sandeshh:
Kafka output exactly once operator and the regular output operator.

### DIFF
--- a/kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaOutputOperator.java
+++ b/kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaOutputOperator.java
@@ -1,0 +1,122 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.apex.malhar.kafka;
+
+import java.util.Properties;
+import javax.validation.constraints.NotNull;
+
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.Producer;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.Operator;
+
+/**
+ * This is the base implementation of a Kafka output operator(0.9.0), which writes data to the Kafka message bus.
+ *
+ * @displayName Abstract Kafka Output
+ * @category Messaging
+ * @tags output operator
+ *
+ */
+@org.apache.hadoop.classification.InterfaceStability.Evolving
+public abstract class AbstractKafkaOutputOperator<K, V> implements Operator
+{
+  private transient Producer<K, V> producer;
+  @NotNull
+  private String topic;
+  private Properties properties = new Properties();
+
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    producer = new KafkaProducer<K, V>(properties);
+  }
+
+  /**
+   * Implement Component Interface.
+   */
+  @Override
+  public void teardown()
+  {
+    producer.close();
+  }
+
+  /**
+   * Implement Operator Interface.
+   */
+  @Override
+  public void beginWindow(long windowId)
+  {
+  }
+
+  /**
+   * Implement Operator Interface.
+   */
+  @Override
+  public void endWindow()
+  {
+  }
+
+  public Properties getProperties()
+  {
+    return properties;
+  }
+
+  /**
+   * Set the Kafka producer properties.
+   *
+   * @param properties Producer properties
+   */
+  public void setProperties(Properties properties)
+  {
+    this.properties.putAll(properties);
+  }
+
+  /**
+   * Set the Kafka producer property.
+   *
+   * @param key Producer Property name
+   * @param val Producer Property value
+   */
+  public void setProperty(Object key, Object val)
+  {
+    properties.put(key,val);
+  }
+
+  public String getTopic()
+  {
+    return topic;
+  }
+
+  /**
+   * Set the Kafka topic
+   * @param  topic  Kafka topic for which the data is sent
+   */
+  public void setTopic(String topic)
+  {
+    this.topic = topic;
+  }
+
+  protected Producer<K, V> getProducer()
+  {
+    return producer;
+  }
+}
+
--- a/kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
+++ b/kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
@@ -1,0 +1,401 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.apex.malhar.kafka;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.ExecutionException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.apex.malhar.lib.wal.FSWindowDataManager;
+import org.apache.apex.malhar.lib.wal.WindowDataManager;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.clients.producer.Callback;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.clients.producer.RecordMetadata;
+
+import org.apache.kafka.common.PartitionInfo;
+import org.apache.kafka.common.TopicPartition;
+
+import com.datatorrent.api.Context;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.Operator;
+
+import static org.apache.kafka.clients.consumer.ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG;
+import static org.apache.kafka.clients.consumer.ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG;
+import static org.apache.kafka.clients.consumer.ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG;
+
+/**
+ * Kafka output operator with exactly once processing semantics.
+ *<br>
+ *
+ * <p>
+ * <b>Recovery handling</b>
+ * <li> Offsets of the Kafka partitions are stored in the WindowDataManager at the endWindow</li>
+ * <li> During recovery,
+ * <ul>
+ *    <li>Partially written Streaming Window before the crash is constructed. ( Explained below ) </li>
+ *    <li>Tuples from the completed Streaming Window's are skipped </li>
+ *    <li>Tuples coming for the partially written Streaming Window are skipped.
+ *       (No assumption is made on the order and the uniqueness of the tuples) </li>
+ *  </ul>
+ *  </li>
+ *</p>
+ *
+ * <p>
+ * <b>Partial Window Construction</b>
+ * <li> Operator uses the Key in the Kafka message, which is not available for use by the operator users.</li>
+ * <li> Key is used to uniquely identify the message written by the particular instance of this operator.</li>
+ *    This allows multiple writers to same Kafka partitions. Format of the key is "APPLICATTION_ID#OPERATOR_ID".
+ * <li>During recovery Kafka partitions are read between the latest offset and the last written offsets.</li>
+ *<li>All the tuples written by the particular instance is kept in the Map</li>
+ *</p>
+ *
+ * <p>
+ * <b>Limitations</b>
+ * <li> Key in the Kafka message is reserved for Operator's use </li>
+ * <li> During recovery, operator needs to read tuples between 2 offsets, if there are lot of data to be read, Operator may
+ *    appear to be blocked to the Stram and can kill the operator. </li>
+ *</p>
+ *
+ * @displayName Kafka Single Port Exactly Once Output(0.9.0)
+ * @category Messaging
+ * @tags output operator
+ *
+ */
+@org.apache.hadoop.classification.InterfaceStability.Evolving
+public class KafkaSinglePortExactlyOnceOutputOperator<T> extends AbstractKafkaOutputOperator<String, T>
+    implements Operator.CheckpointNotificationListener
+{
+  private transient String key;
+  private transient String appName;
+  private transient Integer operatorId;
+  private transient Long windowId;
+  private transient Map<T, Integer> partialWindowTuples = new HashMap<>();
+  private transient KafkaConsumer consumer;
+
+  private WindowDataManager windowDataManager = new FSWindowDataManager();
+  private final int KAFKA_CONNECT_ATTEMPT = 10;
+  private final String KEY_SEPARATOR = "#";
+
+  private final String KEY_DESERIALIZER = "org.apache.kafka.common.serialization.StringDeserializer";
+  private final String VALUE_DESERIALIZER = "org.apache.kafka.common.serialization.StringDeserializer";
+
+  public final transient DefaultInputPort<T> inputPort = new DefaultInputPort<T>()
+  {
+    @Override
+    public void process(T tuple)
+    {
+      sendTuple(tuple);
+    }
+  };
+
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    super.setup(context);
+
+    this.operatorId = context.getId();
+    this.windowDataManager.setup(context);
+    this.appName = context.getValue(Context.DAGContext.APPLICATION_NAME);
+    this.key = appName + KEY_SEPARATOR + (new Integer(operatorId));
+    this.consumer = KafkaConsumerInit();
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
+    this.windowId = windowId;
+
+    if (windowId == windowDataManager.getLargestRecoveryWindow()) {
+      rebuildPartialWindow();
+    }
+  }
+
+  @Override
+  public void checkpointed(long windowId)
+  {
+  }
+
+  @Override
+  public void committed(long windowId)
+  {
+    try {
+      windowDataManager.deleteUpTo(operatorId, windowId);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Override
+  public void beforeCheckpoint(long windowId)
+  {
+  }
+
+  @Override
+  public void teardown()
+  {
+    consumer.close();
+    super.teardown();
+  }
+
+  @Override
+  public void endWindow()
+  {
+    if (!partialWindowTuples.isEmpty() && windowId > windowDataManager.getLargestRecoveryWindow()) {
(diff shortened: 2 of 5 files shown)


sandeshh:
@siyuanh 

Please comment on the overall approach

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortOutputOperator.java
```diff
@@ -0,0 +1,38 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.apex.malhar.kafka;
+
+import org.apache.kafka.clients.producer.ProducerRecord;
+import com.datatorrent.api.DefaultInputPort;
+
+public class KafkaSinglePortOutputOperator<K,V> extends AbstractKafkaOutputOperator
```
chinmaykolhatkar, line 30: Can you make this evolving?

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaOutputOperator.java
```diff
@@ -0,0 +1,141 @@
...
+import com.datatorrent.api.Context;
+import com.datatorrent.api.Operator;
+
+public abstract class AbstractKafkaOutputOperator<K, V> implements Operator
+{
+  @SuppressWarnings("unused")
+  private static final Logger logger = LoggerFactory.getLogger(AbstractKafkaOutputOperator.class);
+  private transient Producer<K, V> producer;  // K is key partitioner, V is value type
+  @NotNull
+  private String topic = "topic1";
```
chinmaykolhatkar: Why is default value of "topic" property set to "topic1"... That invalidates the use of "@NotNull"...

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaOutputOperator.java
```diff
@@ -0,0 +1,141 @@
...
+import javax.validation.constraints.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.Producer;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.Operator;
+
+public abstract class AbstractKafkaOutputOperator<K, V> implements Operator
```
chinmaykolhatkar, line 39: Can you add a javadoc for this class?

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaOutputOperator.java
```diff
@@ -0,0 +1,141 @@
...
+{
+  @SuppressWarnings("unused")
+  private static final Logger logger = LoggerFactory.getLogger(AbstractKafkaOutputOperator.class);
+  private transient Producer<K, V> producer;  // K is key partitioner, V is value type
+  @NotNull
+  private String topic = "topic1";
+  private String producerProperties = "";
+  private Properties configProperties = new Properties();
+  public Properties getConfigProperties()
```
chinmaykolhatkar: We don't have a constraint like this, but would be good if you could organize the property access method at one place.

chinmaykolhatkar: And also add javadoc for each of those.

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaOutputOperator.java
```diff
@@ -0,0 +1,141 @@
...
+      String[] keyVal = StringUtils.trim(propString).split("=");
+      prop.put(StringUtils.trim(keyVal[0]), StringUtils.trim(keyVal[1]));
+    }
+
+    configProperties.putAll(prop);
+
+    return configProperties;
+  }
+
+  public Producer<K, V> getProducer()
```
chinmaykolhatkar: Any reason why this is made public? Its being accessed from subclass.. So can be made protected.

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaOutputOperator.java
```diff
@@ -0,0 +1,141 @@
...
+    this.configProperties.putAll(configProperties);
+  }
+
+  /**
+   * setup producer configuration.
+   * @return ProducerConfig
+   */
+  protected Properties createKafkaProducerConfig()
+  {
+    Properties prop = new Properties();
```
chinmaykolhatkar: If someone has set configProperties directly, call to createKafkaProproducerConfig will override this.
Any reason why producerProperties should override what is being set in configProperties?

Suggesting to do a null+empty check before before overriding.

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaOutputOperator.java
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.Producer;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.Operator;
+
+public abstract class AbstractKafkaOutputOperator<K, V> implements Operator
+{
+  @SuppressWarnings("unused")
+  private static final Logger logger = LoggerFactory.getLogger(AbstractKafkaOutputOperator.class);
```
chinmaykolhatkar: Can you remove the logger is unused?

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractExactlyOnceKafkaOutputOperator.java
```diff
@@ -0,0 +1,385 @@
...
+      e.printStackTrace();
+    }
+
+    return storedOffsets;
+  }
+
+  private KafkaConsumer KafkaConsumerInit()
+  {
+    Properties props = new Properties();
+    props.put("bootstrap.servers", getConfigProperties().get("bootstrap.servers"));
```
chinmaykolhatkar: Can you use Java Constants instead of hardcoding property strings?
You can find the public static string constants in org.apache.kafka.clients.consumer.ConsumerConfig

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractExactlyOnceKafkaOutputOperator.java
```diff
@@ -0,0 +1,366 @@
...
+ *   Every message within the Window is unique
+ *
+ *  This operator uses *Key* to distinguish the messages written by particular instance of the Output operator.
+ *  Operator users can only use *value* for storing the data.
+ *
+ * @displayName Abstract Exactly Once Kafka Output(0.9.0)
+ * @category Messaging
+ * @tags output operator
+ *
+ * @since 3.5
```
chinmaykolhatkar: Can you remove @since? They'll get added during release time.

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractExactlyOnceKafkaOutputOperator.java
```diff
@@ -0,0 +1,366 @@
...
+ */
+@org.apache.hadoop.classification.InterfaceStability.Evolving
+public abstract class AbstractExactlyOnceKafkaOutputOperator<T> extends AbstractKafkaOutputOperator<String, T>
+    implements Operator.CheckpointNotificationListener
```
chinmaykolhatkar: I see this class is extended by KafkaSinglePortExactlyOnceOutputOperator.. The subclass does following 2 things:
1. Add a port
2. call sendTuple method of parent class.

I see there is no need to have this one as abstract operator. This can become concreate and add the port here itself

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractExactlyOnceKafkaOutputOperator.java
devtagare: Why unique records in a window ? From this implementation looks like thats not a required criterion.

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractExactlyOnceKafkaOutputOperator.java
```diff
@@ -0,0 +1,366 @@
...
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    super.setup(context);
+
+    this.operatorId = context.getId();
+    this.windowDataManager.setup(context);
+    this.appId = context.getValue(Context.DAGContext.APPLICATION_ID);
+    this.key = appId + KEY_SEPARATOR + (new Integer(operatorId));
+    this.consumer = KafkaConsumerInit();
```
devtagare: this.key = (new Integer(operatorId)) + partitionId (kafkaPartition) can be a more deterministic key.Since you want to peek into only the partitions to which this operator had written.Also APP_ID is a constant and does not add to arriving at uniqueness

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractExactlyOnceKafkaOutputOperator.java
```diff
@@ -0,0 +1,366 @@
...
+
+    setProperty(ACKS_CONFIG, "all");
+  }
+
```
devtagare: can't we have leader only acknowledgement as the default . acks =1 ?

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractExactlyOnceKafkaOutputOperator.java
```diff
@@ -0,0 +1,366 @@
...
+  private boolean doesKeyBelongsToThisInstance(Integer operatorId, String key)
+  {
+    String[] split = key.split(KEY_SEPARATOR);
+
+    if (split.length != 2) {
+      return false;
+    }
+
+    if ((Integer.parseInt(split[1]) == operatorId) && (split[0].equals(appId))) {
```
devtagare: why not have a partitionId instead of appId.A re-launch will give a new applicationId so the second check will not work.

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractExactlyOnceKafkaOutputOperator.java
```diff
@@ -0,0 +1,366 @@
...
+
+  private KafkaConsumer KafkaConsumerInit()
+  {
+    Properties props = new Properties();
+
+    props.put(BOOTSTRAP_SERVERS_CONFIG, getProperties().get(BOOTSTRAP_SERVERS_CONFIG));
+    props.put(KEY_DESERIALIZER_CLASS_CONFIG, KEY_SERIALIZER);
+    props.put(VALUE_DESERIALIZER_CLASS_CONFIG, VALUE_SERIALIZER);
+
+    return new KafkaConsumer<>(props);
```
devtagare: producer connection seems to be open all the time.Can you add connections.max.idle.ms as a property here so the idle connections are closed after the given time limit.

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
```diff
@@ -0,0 +1,369 @@
...
+    if (!partialWindowTuples.isEmpty() && windowId > windowDataManager.getLargestRecoveryWindow()) {
+      throw new RuntimeException("Violates Exactly once. Not all the tuples received after operator reset.");
+    }
+
+    try {
+      this.windowDataManager.save(getPartitionsAndOffsets(), operatorId, windowId);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    } catch (InterruptedException e) {
+      e.printStackTrace();
```
siyuanh: Don't swallow exception here

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
```diff
@@ -0,0 +1,369 @@
...
+    }
+  }
+
+  private Map<Integer,Long> getCurrentOffsets()
+  {
+    Map<Integer, Long> currentOffsets;
+
+    try {
+      currentOffsets = getPartitionsAndOffsets();
+    } catch (ExecutionException e) {
```
siyuanh: Group into one catch?

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
```diff
@@ -0,0 +1,369 @@
...
+  {
+    return windowDataManager;
+  }
+
+  public void setWindowDataManager(WindowDataManager windowDataManager)
+  {
+    this.windowDataManager = windowDataManager;
+  }
+
+  private boolean doesKeyBelongsToThisInstance(Integer operatorId, String key)
```
siyuanh: If possible, int is always better than Integer

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    return parttionsAndOffset;
+  }
+
+  private void rebuildPartialWindow()
+  {
+    logger.info("Rebuild the partial window after " + windowDataManager.getLargestRecoveryWindow());
+
+    Map<Integer,Long> storedOffsets = getStoredOffsets();
```
siyuanh: You already load storedOffset in the getStoredOffsets method. You can either just return here, or rename it to loadStoredOffsets and return nothing

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
```diff
@@ -0,0 +1,369 @@
...
+    Map<Integer,Long> currentOffsets = getCurrentOffsets();
```
siyuanh: Same problem here

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
```diff
@@ -0,0 +1,369 @@
...
+
+    if (storedOffsets == null || currentOffsets == null) {
```
siyuanh: storedOffsets can be null if operators fails before the first checkpoint. And in this case the stored offsets should be 0 so the operator can load messages from beginning

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
```diff
@@ -0,0 +1,369 @@
...
+
+      if (storedOffsets.containsKey(currentPartition)) {
+        storedOffset = storedOffsets.get(currentPartition);
+      }
+
+      if (storedOffset >= currentOffset) {
+        continue;
+      }
+
+      consumer.seek(new TopicPartition(getTopic(), currentPartition), storedOffset);
```
siyuanh: This line could possibly throw an exception if the storedOffset is invalid because of corrupted windowed data or maybe the messages are purged from kafka. We can log an warn line saying "exactly once" can't be guaranteed and seek to the end of the queue

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
```diff
@@ -0,0 +1,369 @@
...
+
+      int kafkaAttempt = 0;
+
+      while ( true ) {
+
+        ConsumerRecords<String, String> consumerRecords = consumer.poll(100);
+
+        if (consumerRecords.count() == 0) {
+          if (kafkaAttempt++ == KAFKA_CONNECT_ATTEMPT) {
```
siyuanh, line 333: If some other operator are pushing data into same partition, it goes to infinite loop. The ConsumerRecord would give you an offset, you should break the while loop if the offset is bigger than the current offset you load before.

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
```diff
@@ -0,0 +1,369 @@
...
+  }
+
+  protected void sendTuple(T tuple)
+  {
+    if ( alreadyInKafka(tuple) ) {
+      return;
+    }
+
+    try {
+      getProducer().send(new ProducerRecord<>(getTopic(), key, tuple)).get();
```
siyuanh: Use the send(ProducerRecord<K,V> record, Callback callback) method, In callback if there is any exception, you can simply throw exception to fail the operator.

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
sandeshh, line 333: Already there is a check for that in line 300 ( crossedBoundary )

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
sandeshh: get() throws exception and is rethrown

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
sandeshh: So the assumption is that Topic assigned to this operator has the offset as 0. Is that always the case?

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
sandeshh: Get current offsets reads the latest offsets from Kafka and stored anywhere.

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
sandeshh: loading happens from WindowDataManager, will make that inplace.

tweise:
Why do we need this operator, how is it different from the existing operator?

siyuanh:
@tweise  There is no 0.9 output operator.

sandeshh:
@siyuanh I have addressed the review comments that we discussed.

tweise:
@siyuanh that does not answer my question. If added a comment to the JIRA.

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
```diff
@@ -0,0 +1,374 @@
...
+import com.datatorrent.api.Operator;
+
+import static org.apache.kafka.clients.consumer.ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG;
+import static org.apache.kafka.clients.consumer.ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG;
+import static org.apache.kafka.clients.consumer.ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG;
+
+/**
+ * Kafka output operator with exactly once processing semantics under certain conditions.,
+ *
+ *  This operator uses *Key* to distinguish the messages written by particular instance of the Output operator.
```
tweise: I think this needs more explanation. What makes it exactly once, are there other restrictions or assumptions?

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
```diff
@@ -0,0 +1,401 @@
...
+import com.datatorrent.api.Context;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.Operator;
+
+import static org.apache.kafka.clients.consumer.ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG;
+import static org.apache.kafka.clients.consumer.ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG;
+import static org.apache.kafka.clients.consumer.ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG;
+
+/**
+ * Kafka output operator with exactly once processing semantics.
```
sandeshh, line 55: @tweise & @siyuanh updated the JavaDoc. Please take a look.

siyuanh:
@sandeshh  Test is failing, can you change the log level and run it again so it won't fail because of too much logs

sandeshh:
@siyuanh Travis build is failing as it reached log limit of 4mb. 

Here is the log file,
https://s3.amazonaws.com/archive.travis-ci.org/jobs/143218837/log.txt 

What do you recommend? Disable the Kafka unit tests?

siyuanh:
@sandeshh  Looks like the travis doesn't pick up the log4j settings. Can you try to run the travis build on your local machine?

kafka/src/test/resources/log4j.properties
```diff
@@ -47,3 +47,4 @@ log4j.logger.org.apache.apex=INFO
 log4j.logger.org.apacke.kafka=WARN
```
siyuanh: Typo here, this is the reason why log4j doesn't pick up lines after this

PR #300, opened 2016-05-27
APEXMALHAR-2103 Fixed the scanner issue in FileSplitterInput Class

chaithu14:


chaithu14:
@DT-Priyanka @chandnisingh Please review the changes.

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -375,11 +374,11 @@ public void run()
             lastScannedInfo = null;
             numDiscoveredPerIteration = 0;
             for (String afile : files) {
-              String filePath = new File(afile).getAbsolutePath();
-              LOG.debug("Scan started for input {}", filePath);
+              Path filePath = new Path(afile);
+              LOG.debug("Scan started for input {}", filePath.toString());
```
gauravgopi123: toString() not required.. why not use afile?

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -375,11 +374,11 @@ public void run()
...
               Map<String, Long> lastModifiedTimesForInputDir;
-              lastModifiedTimesForInputDir = referenceTimes.get(filePath);
-              scan(new Path(afile), null, lastModifiedTimesForInputDir);
+              lastModifiedTimesForInputDir = referenceTimes.get(fs.getFileStatus(filePath).getPath().toString());
```
gauravgopi123: can filePath.toString() be used here?

chaithu14: No. In case of LocalFileSystem, if the filePath consists of relative path then again the keys are different. 
Setting the directory path in createScannedFileInfo() and accessing the value from referenceTimes must be sync.

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
chaithu14: Yeah. afile and filePath.toString() represents the same value.

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -375,11 +374,18 @@ public void run()
             lastScannedInfo = null;
             numDiscoveredPerIteration = 0;
             for (String afile : files) {
-              String filePath = new File(afile).getAbsolutePath();
-              LOG.debug("Scan started for input {}", filePath);
-              Map<String, Long> lastModifiedTimesForInputDir;
-              lastModifiedTimesForInputDir = referenceTimes.get(filePath);
-              scan(new Path(afile), null, lastModifiedTimesForInputDir);
+              Path filePath = new Path(afile);
+              LOG.debug("Scan started for input {}", filePath.toString());
```
gauravgopi123: @chaithu14 : I see that you updated PR but it is still using toString()?

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -375,11 +374,18 @@ public void run()
...
+              Map<String, Long> lastModifiedTimesForInputDir = null;
+              if (fs.exists(filePath)) {
+                FileStatus fileStatus = fs.getFileStatus(filePath);
+                if (fileStatus.isDirectory()) {
+                  lastModifiedTimesForInputDir = referenceTimes.get(fileStatus.getPath().toString());
+                } else {
+                  lastModifiedTimesForInputDir = referenceTimes.get(fileStatus.getPath().getParent().toString());
```
DT-Priyanka: This is not right, in case user has given input as,
/home/myDir, /home/myDir/file1.txt, the scan of second input i.e. /home/myDir/file1.txt will overwrite the reference times for input /home/myDir.

otterc:
@chaithu14 @DT-Priyanka 
- Please add test cases that validate the changes.
- This commit consists of adding stray files.

DT-Priyanka:
@chaithu14  Let's not modify the metadata. After updating refereceTimes in last changes, I forgot to modify code to consider files along with directories. I would suggest we do something like this: https://github.com/apache/incubator-apex-malhar/compare/master...DT-Priyanka:APEXMALHAR-2103-fix-for-file-input
I will update changes with test case soon.

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
chaithu14: @Priyanka: If we maintain 2 different keys, then fileSplitter will emit multiple filemetadata's for /home/myDir/file1.txt. Is it expected behavior. Please correct it, if I am wrong.

chaithu14:
I Updated the code as per Priyanka suggestions and also added unit test case.

library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterInputTest.java
```diff
@@ -155,6 +155,38 @@ public void testFileMetadata() throws InterruptedException
   }
 
   @Test
+  public void testFileMetadataForSingleFile() throws InterruptedException
```
DT-Priyanka: there is already a test with name "testSingleFile" can we rename this to something like, "testScannerFilterForDuplicates"

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -375,11 +375,14 @@ public void run()
...
-              String filePath = new File(afile).getAbsolutePath();
+              Path filePath = new Path(afile);
               LOG.debug("Scan started for input {}", filePath);
-              Map<String, Long> lastModifiedTimesForInputDir;
-              lastModifiedTimesForInputDir = referenceTimes.get(filePath);
-              scan(new Path(afile), null, lastModifiedTimesForInputDir);
+              Map<String, Long> lastModifiedTimesForInputDir = null;
+              if (fs.exists(filePath)) {
+                FileStatus fileStatus = fs.getFileStatus(filePath);
+                lastModifiedTimesForInputDir = referenceTimes.get(fileStatus.getPath().toUri().getPath());
```
DT-Priyanka, line 383: can you use "filePath" instead of fileStatus.getPath().

chaithu14, line 383: No. In createScannedFileInfo(), the directory/file path specified as absolute path. But, In case of local file system, files might be relative path.

PR #302, opened 2016-05-31
[APEXMALHAR-139] Make Sigma Operator Partitionable

ilganeli:
- Added unifiers on all output ports for Sigma operator to support using it as a Partitionable operator.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/math/AbstractOutput.java
+++ b/library/src/main/java/com/datatorrent/lib/math/AbstractOutput.java
@@ -19,6 +19,7 @@
 package com.datatorrent.lib.math;
 
 import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.api.Operator;
 import com.datatorrent.api.annotation.OutputPortFieldAnnotation;
 import com.datatorrent.common.util.BaseOperator;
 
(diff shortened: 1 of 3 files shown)


ilganeli:
Note: commits are squashed even though GitHub showing multiple entries...

ilganeli:
Retest this please.

ilganeli:
retest this please

library/src/main/java/com/datatorrent/lib/math/AbstractOutput.java
```diff
@@ -38,23 +38,23 @@
    * Double type output.
    */
   @OutputPortFieldAnnotation(optional = true)
-  public final transient DefaultOutputPort<Double> doubleResult = new DefaultOutputPort<Double>();
+  public transient DefaultOutputPort<Double> doubleResult = new DefaultOutputPort<Double>();
```
chinmaykolhatkar: @ilganeli This won't work. The purpose of keeping port variable final is because this variable is used in Application.java to connect the ports. Once its used in addStream in application.java, the reference to this variable is kept and the same object will be used. Hence even if the object is recreated in child class, that won't take effect.

Suggested approach for adding unifier to super class's port is as follows:
1. Do following in parent class:

```
public final transient DefaultOutputPort<Double> doubleResult = new DefaultOutputPort<>()
{
   @Override
  public Operator.Unifier getUnifier()
  {
     return getUnifier();
  }
}

protected Operator.Unifier getUnifier()
{
   return null;
}
```
1. Do following in child class:

```
@Override
protected Operator.Unifier<Double> getUnifier()
{
    return new UnifierSumNumber<>();
}
```

ilganeli:
@chinmaykolhatkar I'm having trouble with the approach you've proposed. It seems that all data is emitted from the port of the unifier (rather than from the the partitioned operator) so data isn't actually sent. I'm not clear how to resolve this, but I've added the updated code.

chinmaykolhatkar:
@ilganeli That looks fine to me.

When sigma operator is partitioned, there will be a unifier (kind of like an operator instance) which will come after that stage and just before the downstream operator in the series. 
Unifier will merge the output of all the partitioned Sigma operator instances under the logic provided by UnifierSumNumber and forward the output to downstream operator.

So your physical DAG will look like following:

```
                    -> Sigma1   -> 
UpstreamOperator -> Sigma2   ->  Unifier -> Downstream operator
                    -> Sigma3   ->
```

Please let me know if I've not described the case here properly OR if I misunderstood the problem, in which case it would be great if you could share the sample code where you are seeing the issue.

chinmaykolhatkar:
@ilganeli Can you please check why the travis build is failing?

chinmaykolhatkar:
Hi @ilganeli , When I try to run the tests for this diff against latest master, I get test failure for Sigma Test locally:

Failed tests: 
  SigmaTest.testApplication:134 Total sum expected:<2595> but was:<0>

Can you please take a look?

Also, thomas has mentioned a better way to write application test in his mail:
http://mail-archives.apache.org/mod_mbox/apex-users/201606.mbox/%3CCA%2B5xAo3XPThwG1oVt42QXZ1%3DaZMNG8OAZpVFuG_y%3Dt0f6QHxeg%40mail.gmail.com%3E

Can you please following that?

Let me know if you need any help in completing this PR.

Thanks,
Chinmay.

chinmaykolhatkar:
@ilganeli Any update on this PR?

ilganeli:
No update / I got frustrated with the multiple requested test updates and the spurious test failure after the first couple of iterations. 

The code was functional and the tests I had at the time indicated it was working but then something else changed. I don't presently have the cycles to complete this PR. May be able to do so in a few weeks.

vrozov:
@ilganeli @chinmaykolhatkar closed due to inactivity.

PR #303, opened 2016-05-31
[APEXMALHAR-175] KeyValPair isn't automatically sticky partitioned for key

ilganeli:
- Updated KeyValPair to override hashCode to support sticky partitioning by key.


gauravgopi123:
@ilganeli : This is not required. It is intentional. For your use case you can use https://github.com/apache/incubator-apex-malhar/blob/master/library/src/main/java/com/datatorrent/lib/util/KeyHashValPair.java

ilganeli:
@gauravgopi123 should the relevant JIRA be closed then?

gauravgopi123:
@ilganeli Yes..

PR #304, opened 2016-06-01
APEXMALHAR-2104-BytesFileOutputOperator refactoring

yogidevendra:
1. Marking BytesFileOutputOperator as @evolving
2. Removed input port stringInput
3. Added japicmp exclusion for this class
4. Added Converter field for pluging-in custom conversion to byte[]
5. Provided default converters for no-op, String
6. Renamed BytesFileOutputOperator-> GenericFileOutputOperator
7. Introduced BytesFileOutputOperator, String FileOutputOperator
8. Added javadocs

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/fs/GenericFileOutputOperator.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/fs/GenericFileOutputOperator.java
@@ -23,12 +23,14 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.concurrent.ExecutionException;
 
+import javax.validation.constraints.NotNull;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.datatorrent.api.AutoMetric;
-import com.datatorrent.api.DefaultInputPort;
 import com.datatorrent.api.StreamCodec;
+import com.datatorrent.lib.converter.Converter;
 import com.datatorrent.lib.io.fs.AbstractSingleFileOutputOperator;
 import com.datatorrent.netlet.util.DTThrowable;
 
@@ -39,8 +41,8 @@ import com.datatorrent.netlet.util.DTThrowable;
  *
  * @since 3.4.0
  */
-
-public class BytesFileOutputOperator extends AbstractSingleFileOutputOperator<byte[]>
+@org.apache.hadoop.classification.InterfaceStability.Evolving
+public abstract class GenericFileOutputOperator<INPUT> extends AbstractSingleFileOutputOperator<INPUT>
 {
 
   /**
@@ -81,6 +83,12 @@ public class BytesFileOutputOperator extends AbstractSingleFileOutputOperator<by
   private long currentPartIdleWindows;
 
   /**
+   * Converter for conversion of input tuples to byte[]
+   */
+  @NotNull
+  private Converter<INPUT, byte[]> converter;
+
+  /**
    * Max number of idle windows for which no new data is added to current part
    * file. Part file will be finalized after these many idle windows after last
    * new data.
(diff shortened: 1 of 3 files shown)


tweise:
I don't think this operator should have multiple input ports. That's also confusing to the user. Why not create separate operators for String[] and byte[] input instead? You can have a generic base that has a T -> byte[] converter.

yogidevendra:
So, suggestion is to remove stringInput port from this operator. Is my understanding correct?
Is this OK from backward compatibility perspective? We have released this operator in 3.4.0 without @ evolving. 

Following sentence is not clear to me:

> You can have a generic base that has a T -> byte[] converter.

Are you suggesting to have a common base class with generic type (<INPUT>) and then extend it to concrete implementations for byte[], String?

tweise:
Adding it without evolving annotation was a mistake too. Before anyone starts using it, let's fix it.

The only difference between the 2 operators is how they convert input to bytes. If you want a generic port, then you would need a configurable conversion. Not a base class, but perhaps a 3rd option and not needed as part of this change.

pramodin:
@tweise why not go with your original suggestion of abstract base class with generic port. When you say "a 3rd option" do you mean a concrete class that doesn't have any port but has a processTuple method and the sub-class has the specific port.

yogidevendra:
Here is my proposal:
1. Add a field converter which is com.datatorrent.lib.converter.Converter<Object,byte[]>.
2. Have static inner classes with two implementations for converter: NoOpConverter, StringToBytesConverter.
3. Use converter in getBytesForTuple to convert given tuple to byte[].
With this change: single java file can serve both the usecases with single input port.
4. Tuples incompatible with specified converter would throw ClassCastException

For adding @ evolving
1. I will add the annotation to BytesFileOutputOperator
2. Add entry in pom.xml to exclude BytesFileOutputOperator for japicmp.

yogidevendra:
@tweise @PramodSSImmaneni Incorporated review comments to the best of my understanding.
Could you please review this?

tweise:
@yogidevendra the converter is something the user will have to configure. Instead of setting a default that assumes byte[], why not have 2 nested classes for byte[] and String that fix the type parameter and also set the converter?

yogidevendra:
@tweise How about having separate java file for StringFileOutputOperator instead of inner class? 
In the past, I noticed that operators in inner classes are hard to discover. [e.g. LineByLineFileInputOperators is introduced even if inner class AbstractFileInputOperator.FileLineInputOperator was already present. ]

Also, if we decide to have inner classes for Bytes, String it would difficult to give appropriate name to the outer class. [One option I thought of for inner class names is BytesFileOutputOperator, StringFileOutputOperator. But, then what about outer class?]

Let me know your feedback. Also, class naming suggestions if you feel that inner classes is the way to go ahead.

yogidevendra:
@tweise awaiting for your feedback on my comment. I will update the PR accordingly. 
A OR B:
A: Separate Java file for StringFileOutputOperator
B: Inner classes for Bytes, String 
For option B, class name suggestions please.

tweise:
@yogidevendra the discussion about the line input operator was slightly different. It was inside an abstract class and actually provides added functionality. Here we are talking about convenience classes (for the non-Java tool user?) that set a converter so that user does not have to configure it. Otherwise the developer would pick the top level class and set the converter, which should also be required field which causes a validation error when null.

yogidevendra:
OK. So your vote is for option B. I will update the PR with these changes.

I am planning to rename outer class as GenericFileOutputOperator and inner classes as BytesFileOutputOperator, StringFileOutputOperator. 

Let me know if you have any other naming suggestion.

yogidevendra:
@tweise Added inner classes for BytesFileOutputOperator, StringFileOutputOperator.

Would you please review this and merge if things are OK?

library/src/main/java/org/apache/apex/malhar/lib/fs/GenericFileOutputOperator.java
```diff
@@ -39,8 +41,8 @@
  *
  * @since 3.4.0
  */
-
-public class BytesFileOutputOperator extends AbstractSingleFileOutputOperator<byte[]>
+@org.apache.hadoop.classification.InterfaceStability.Evolving
+public abstract class GenericFileOutputOperator<INPUT> extends AbstractSingleFileOutputOperator<INPUT>
```
pramodin, line 45: Why abstract?

PR #307, opened 2016-06-01
[APEXMALHAR-2106] Support multiple streams in StreamMerger

ilganeli:
- Added ability to support more than one stream in the operator

Note: I am unclear whether this approach will work if the operator or the DAG restarts, will the ports be reconnected properly? If not, what mechanism prevents this?

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/stream/StreamMerger.java
+++ b/library/src/main/java/com/datatorrent/lib/stream/StreamMerger.java
@@ -18,13 +18,19 @@
  */
 package com.datatorrent.lib.stream;
 
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+
+import com.datatorrent.api.Context;
 import com.datatorrent.api.DefaultInputPort;
 import com.datatorrent.api.DefaultOutputPort;
 import com.datatorrent.api.annotation.Stateless;
 import com.datatorrent.common.util.BaseOperator;
 
 /**
- * An implementation of BaseOperator that merges two streams with identical schema and emits the tuples to the output port in order.
+ * An implementation of BaseOperator that merges multiple streams with identical schema and emits the tuples
+ * to the output port in order.
  *
  * <p>
  * This is a pass through operator<br>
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/datatorrent/lib/stream/StreamMerger.java
```diff
@@ -41,35 +47,57 @@
 @Stateless
 public class StreamMerger<K> extends BaseOperator
 {
-  /**
-   * Data input port 1.
-   */
-  public final transient DefaultInputPort<K> data1 = new DefaultInputPort<K>()
+  private int portCount = 0;
+  private final transient ArrayList<DefaultInputPort<K>> ports = new ArrayList<>(portCount);
```
gauravgopi123, line 51: Did you use this operator in a DAG and see if it works?

ilganeli, line 51: A great question, was in the process of doing so. Short answer is it fails wonderfully:

```
java.lang.IllegalArgumentException: Port is not associated to any operator in the DAG: com.datatorrent.lib.stream.StreamMerger$1@3b084709
    at com.datatorrent.stram.plan.logical.LogicalPlan.assertGetPortMeta(LogicalPlan.java:1501)
    at com.datatorrent.stram.plan.logical.LogicalPlan.access$1000(LogicalPlan.java:119)
```

The relevant question is thus is there any way to define ports in this way, e.g. an array of ports, such that the DAG is aware of the ports?

tweise, line 51: No, there isn't. But it is possible to chain multiple mergers with 2 ports and accomplish the same.

ilganeli:
@tweise what is the internal mechanism that prevents creating port connections in this fashion?

Chaining together operators that only take two ports seems cludgy. This could not be implemented within its own operator, I think it could be implemented as a module (and operators could all be local but it seems to add unnecessary complexity to do so.

tweise:
Only fields of type InputPort are recognized, not collections. However, there is nothing stopping you from adding more fields, that are annotated as optional.

Chaining is a workaround, but using thread local it can be done with little overhead in cases such as the beam translation (which is similar to the expansion of a module).

ilganeli:
Opened a new PR here with cascading oeprators: 
https://github.com/apache/incubator-apex-malhar/pull/309

PR #309, opened 2016-06-02
[APEXMALHAR-2106] Support multiple streams in StreamMerger

ilganeli:
- Created a module which creates a binary tree of StreamMerger operators to support merging multiple streams
- StreamMerger operators are presently allocated as CONTAINER_LOCAL. I would like for them to be THREAD_LOCAL but I was getting the following error when attempting this:

```
Caused by: javax.validation.ValidationException: Locality THREAD_LOCAL invalid for operator OperatorMeta{name=Merger_Tier_0_#_0, operator=StreamMerger{name=null}, attributes={}} with multiple input streams as they origin from different owner OIO operators
```
- An example application is included

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/stream/MultipleStreamMerger.java
+++ b/library/src/main/java/com/datatorrent/lib/stream/MultipleStreamMerger.java
@@ -1,0 +1,210 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.lib.stream;
+
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.Queue;
+
+import org.apache.hadoop.conf.Configuration;
+
+import com.datatorrent.api.DAG;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.DefaultOutputPort;
+
+/**
+ * A helper class  that adds functionality to bypass limitations preventing us from combining more than two
+ * streams at a time with the existing Stream Merger.
+ *
+ * Usage:
+ *
+ * dag.addOperator("Stream_1", op1);
+ * dag.addOperator("Stream_2", op2);
+ * dag.addOperator("Stream_3", op3);
+ *
+ * MultipleStreamMerger merger = new MultipleStreamMerger();
+ * DefaultOutputPort streamOut = merger.merge(op1.out)
+ * .merge(op2.out)
+ * .merge(op3.out)
+ * .mergeStreams(dag, conf);
+ *
+ * dag.addStream("merger-counter", streamOut, counterOp.counter);
+ *
+ * @param <K>
+ */
+public class MultipleStreamMerger<K>
+{
+  public class Stream
+  {
+    DefaultInputPort<K> destPort;
+    SourcedOutputPort sourcePort;
+    String name;
+
+    public Stream(String name, SourcedOutputPort sourcePort, DefaultInputPort<K> destPort)
+    {
+      this.destPort = destPort;
+      this.sourcePort = sourcePort;
+      this.name = name;
+    }
+  }
+
+  public class NamedMerger
+  {
+    StreamMerger<K> merger;
+    String name;
+
+    public NamedMerger(String name, StreamMerger<K> merger)
+    {
+      this.merger = merger;
+      this.name = name;
+    }
+  }
+
+  /**
+   * A simple class to allow us to track whether the port to be merged is internal (allowing it to be thread local)
+   * or external
+   */
+  public class SourcedOutputPort
+  {
+    boolean internal;
+    DefaultOutputPort<K> port;
+
+    public SourcedOutputPort(DefaultOutputPort<K> port)
+    {
+      this.internal = false;
+      this.port = port;
+    }
+
+    public SourcedOutputPort(boolean internal, DefaultOutputPort<K> port)
+    {
+      this.internal = internal;
+      this.port = port;
+    }
+  }
+
+  ArrayList<DefaultOutputPort<K>> streamsToMerge = new ArrayList<>();
+
+  private DefaultOutputPort<K> streamOutput = new DefaultOutputPort<>();
+
+  /**
+   * Used to define all the sources to be merged into a single stream.
+   *
+   * @param sourcePort - The output port from the upstream operator that provides data
+   * @return The updated MultipleStreamMerger object that tracks which streams should be unified.
+   */
+  public MultipleStreamMerger<K> merge(DefaultOutputPort<K> sourcePort)
+  {
+    streamsToMerge.add(sourcePort);
+    return this;
+  }
+
+
+  /**
+   * Given the streams to merge have been selected with {@link #merge(DefaultOutputPort)}, create a subDAG and add it
+   * to an existing DAG.
+   *
+   * To merge more than two streams at a time, we construct a tiered hierarchy of thread-local StreamMerger operators
+   * E.g.
+   *
+   * Stream 0 ->
+   *            StreamMerger_1 ->
+   * Stream 1 ->
+   *                               StreamMerger_Final -> Out
+   * Stream 2 ->
+   *            StreamMerger_2 ->
+   * Stream 3 ->
+   *
+   * @param dag - The DAG to update
+   * @param conf - The configuration
+   *
+   */
+  public DefaultOutputPort<K> mergeStreams(DAG dag, Configuration conf)
+  {
+    if (streamsToMerge.size() < 2) {
+      throw new IllegalArgumentException("Not enough streams to merge, at least two streams must be selected for " +
+          "merging with `.merge()`.");
+    }
+
+    ArrayList<Stream> streamsToAddToDag = new ArrayList<>();
+    ArrayList<NamedMerger> operatorsToAdd = new ArrayList<>();
+
+    // Determine operators and streams to add to the DAG
+    constructMergeTree(streamsToAddToDag, operatorsToAdd);
+
+    for (NamedMerger m : operatorsToAdd) {
+      dag.addOperator(m.name, m.merger);
+    }
+
+    for (Stream s : streamsToAddToDag) {
+      DAG.StreamMeta stream = dag.addStream(s.name, s.sourcePort.port, s.destPort);
+      if (s.sourcePort.internal) {
+        stream.setLocality(DAG.Locality.CONTAINER_LOCAL);
+      }
+    }
+
+    return streamOutput;
+  }
+
+  /**
+   * Given a set of streams to be merged (defined via {@link #merge(DefaultOutputPort)}), compute the optimal
+   * structure of cascading mergers that need to be instantiated, added to the dag, and linked together.
+   * @param streamsToAddToDag - (output)  A list that is populated with streams that should be added to the  DAG
+   * @param operatorsToAdd - (output) A list that is populated with operators to be added to the DAG
+   */
+  public void constructMergeTree(
+      ArrayList<Stream> streamsToAddToDag,
+      ArrayList<NamedMerger> operatorsToAdd)
+  {
+    if (streamsToMerge.size() < 2) {
+      throw new IllegalArgumentException("Not enough streams to merge. Ensure `.merge` was called for each stream " +
+          "to be added.");
+    }
+
+    /**
+     * We can unify all unmerged streams by using a Queue. Unconnected ports are pushed to a queue (FIFO) and then for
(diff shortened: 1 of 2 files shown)


ilganeli:
retest this please

library/src/main/java/com/datatorrent/lib/stream/MultipleStreamMerger.java
```diff
@@ -0,0 +1,277 @@
...
+
+      // When assigning streams, we will switch between ports 1 and 2 as we use successive mergers.
+      boolean usePort1;
+
+      // For each tier, create the mergers in that tier, and connect the relevant streams
+      for (int i = 0; i < numTiers - 1; i++) {
+        int streamIdx = 0;
+        usePort1 = true;
+
+        int numMergers = (int)Math.ceil(streamsToMerge.size() / Math.pow(2, i + 1));
```
bhupeshchawda: Should this be floor instead of ceil? The one odd stream can be directly fed to the final tier.

library/src/main/java/com/datatorrent/lib/stream/MultipleStreamMerger.java
```diff
@@ -0,0 +1,277 @@
...
+ * dag.addModule("Merger", merger);
+ *
+ * @param <K>
+ */
+public class MultipleStreamMerger<K> implements Module
+{
+  public class Stream
+  {
+    DefaultInputPort destPort;
+    DefaultOutputPort sourcePort;
```
bhupeshchawda: Parametrize these ports or have it as a generic <?>

library/src/main/java/com/datatorrent/lib/stream/MultipleStreamMerger.java
```diff
@@ -0,0 +1,277 @@
...
+   *            StreamMerger_2 ->
+   * Stream 3 ->
+   *
+   * @param dag
+   * @param conf Note that we don't use the populateDAG function because that is only used to flatten the
+   *             module when what we really need to do is to define the connections to the parent DAG.
+   *             The populateDAG does not contain operators outside of the module and thus cannot connect
+   *             to the external operators.
+   */
+  public void insertInto(DAG dag, Configuration conf)
```
bhupeshchawda: Is this needed? Can you directly call mergeStreams() from populateDag()?

library/src/main/java/com/datatorrent/lib/stream/MultipleStreamMerger.java
```diff
@@ -0,0 +1,277 @@
...
+
+    // Determine operators and streams to add to the DAG
+    constructMergeTree(streamsToAddToDag, operatorsToAdd);
+
+    for (NamedMerger m : operatorsToAdd) {
+      dag.addOperator(m.name, m.merger);
+    }
+
+    for (Stream s : streamsToAddToDag) {
+      dag.addStream(s.name, s.sourcePort, s.destPort).setLocality(DAG.Locality.CONTAINER_LOCAL);
```
bhupeshchawda: Making all streams container local will force all operators upstream to the merger as well as down stream to the merger also to be in the same container. Only the streams which are within the merger (the ones between two tiers of the mergers ) must be made CONTAINER_LOCAL. Even THREAD_LOCAL should work here.

library/src/main/java/com/datatorrent/lib/stream/MultipleStreamMerger.java
```diff
@@ -0,0 +1,277 @@
...
+      assert (streamsToMerge.size() == 2);
+      streamsToAddToDag.add(new Stream("FinalMerge_Stream_0", streamsToMerge.get(0), finalMerger.data1));
+      streamsToAddToDag.add(new Stream("FinalMerge_Stream_1", streamsToMerge.get(1), finalMerger.data2));
+
+      // We don't need to add any operators since we've already added the final merger
+    } else {
+      Iterator<DefaultOutputPort<K>> streams = streamsToMerge.iterator();
+
+      // When assigning streams, we will switch between ports 1 and 2 as we use successive mergers.
+      boolean usePort1;
```
bhupeshchawda: better name?

library/src/main/java/com/datatorrent/lib/stream/MultipleStreamMerger.java
```diff
@@ -0,0 +1,277 @@
...
+
+        ArrayList<StreamMerger<K>> mergersTierI = new ArrayList<>(numMergers);
+
+        // For each merger in the tier, assign the appropriate streams to that merger
+        for (int mergerIdx = 0; mergerIdx < numMergers; mergerIdx++) {
+          StreamMerger<K> merger = new StreamMerger<>();
+          operatorsToAdd.add(new NamedMerger("Merger_Tier_" + i + "_#_" + mergerIdx, merger));
+
+          // Each operator has two ports so add a simple inner loop
+          for (int port = 0; port < 2; port++) {
```
bhupeshchawda: Do we need a loop here? Can we open up the loop?

library/src/main/java/com/datatorrent/lib/stream/MultipleStreamMerger.java
```diff
@@ -0,0 +1,277 @@
...
+
+      // We don't need to add any operators since we've already added the final merger
+    } else {
+      Iterator<DefaultOutputPort<K>> streams = streamsToMerge.iterator();
+
+      // When assigning streams, we will switch between ports 1 and 2 as we use successive mergers.
+      boolean usePort1;
+
+      // For each tier, create the mergers in that tier, and connect the relevant streams
+      for (int i = 0; i < numTiers - 1; i++) {
```
bhupeshchawda: A suggestion: Can the overall logic in this loop be simplified? This will help understand the code better.

library/src/main/java/com/datatorrent/lib/stream/StreamMerger.java
```diff
@@ -44,6 +45,7 @@
   /**
    * Data input port 1.
    */
+  @InputPortFieldAnnotation(optional = true)
```
bhupeshchawda: This won't be needed if we don't use mergers with just one input. See comment above on number of mergers.

library/src/main/java/com/datatorrent/lib/stream/StreamMerger.java
```diff
@@ -59,6 +61,7 @@ public void process(K tuple)
   /**
    * Data input port 2.
    */
+  @InputPortFieldAnnotation(optional = true)
```
bhupeshchawda: Same here

library/src/test/java/com/datatorrent/lib/stream/MultipleStreamMergerTest.java
```diff
@@ -0,0 +1,145 @@
...
+      dag.addModule("Merger", merger);
+
+      // And then we should see the output
+      ConsoleOutputOperator consoleOperator = dag.addOperator("console", new ConsoleOutputOperator());
+      dag.addStream("merger-console", merger.streamOutput, consoleOperator.input);
+    }
+  }
+
+  @Test
+  public void testApplication() throws IOException, Exception {
```
bhupeshchawda: Can you also add validation to this test? A very basic test could be done by controlling the number of objects sent by each RandomWordGenerator and counting them at the console operator.

bhupeshchawda:
In addition, I suggest testing this out on a real cluster. Some issues related to serialization (if any) may not be detected in a unit test.

ilganeli:
@bhupeshchawda Thanks for the review! I'll work through these suggestions.

library/src/test/java/com/datatorrent/lib/stream/MultipleStreamMergerTest.java
ilganeli: Done - i've added a test that confirms all inputs received at output match the input.

ilganeli:
@bhupeshchawda What's the easiest way to launch an app on the cluster if I've made changes to Malhar?

bhupeshchawda:
I think you will need to create an application package. See https://apex.apache.org/docs/apex/apex_development_setup/#creating-new-apex-project for more details

ilganeli:
@bhupeshchawda I've created a new demo module but I'm having a lot of issues referencing the updated version of Malhar using Maven. Even when I successfully build the demo, if I attempt to launch the generated JAR, I'm getting an error at runtime: `java.lang.NoClassDefFoundError: com/datatorrent/api/Module`.

I'm able to build and run through IntelliJ

bhupeshchawda:
@ilganeli After some discussion, it seems that this functionality does not really need a module. This can still be a utility function which people can directly use in their populateDag() method as illustrated by you.

bhupeshchawda:
Additionally, the demo project is not really needed as it just replicates what is already there in the test you have added. I suggested to run it on the cluster because it would allow us to uncover certain bugs which remain undetected otherwise. 
To your earlier query about easily running some code on the cluster, the following approach is suggested: once you have the application in malhar-library, you can create a new project using the archetype command. In the Application.java, simply remove all the code and just extend the Application by the test application in malhar, as follows: 
`public class Application extends MultipleStreamMergerTest.Application
{}`
Note that the malhr dependency should be defined in the pom.xml. Once you compile the new project, the apa generated would allow you to run the test application on the cluster.

ilganeli:
The main advantage of using a module is that it gives a clearly defined output port for the stream. I guess the stream builder could just return the final output port to connect to?

ilganeli:
@bhupeshchawda What's the best way to resolve the run-time dependency on the new Malhar library (with merger)? 

I get the following error when running on the cluster after building and deploying the App:

```
./datatorrent/current/bin/dtcli -e "launch merger-1.0-SNAPSHOT.apa"
java.lang.NoClassDefFoundError: com/datatorrent/lib/stream/MultipleStreamMerger
```

Local execution works fine.

library/src/main/java/com/datatorrent/lib/stream/MultipleStreamMerger.java
```diff
@@ -0,0 +1,215 @@
...
+import org.eclipse.jetty.util.ArrayQueue;
+
+import org.apache.hadoop.conf.Configuration;
+
+import com.datatorrent.api.DAG;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.DefaultOutputPort;
+
+/**
+ * A helper class  that adds functionality to bypass the platform limitations of combining more than two
```
bhupeshchawda: Not really the limitation of the platform. May be limitation of the existing StreamMerger?

library/src/main/java/com/datatorrent/lib/stream/MultipleStreamMerger.java
```diff
@@ -0,0 +1,215 @@
...
+   * E.g.
+   *
+   * Stream 0 ->
+   *            StreamMerger_1 ->
+   * Stream 1 ->
+   *                               StreamMerger_Final -> Out
+   * Stream 2 ->
+   *            StreamMerger_2 ->
+   * Stream 3 ->
+   * Note that we don't use the populateDAG function because that is only used to flatten the
```
bhupeshchawda: Need to change the docs since it is no more a module.

library/src/main/java/com/datatorrent/lib/stream/MultipleStreamMerger.java
```diff
@@ -0,0 +1,215 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.lib.stream;
+
+import java.util.ArrayList;
+import java.util.Queue;
+
+import org.eclipse.jetty.util.ArrayQueue;
```
bhupeshchawda: Why not use the java.util implementations instead? May be LinkedList?

library/src/main/java/com/datatorrent/lib/stream/MultipleStreamMerger.java
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * We can unify all unmerged streams by using a Queue. Unconnected ports are pushed to a queue (FIFO) and then for
+     * every pair of unconnected ports, we create a Merger. If only one port remains, we simply use the last added
+     * merger.
+     *
+     * We use SourcedOutputPorts to track whether a given port comes from inside or outside the module, allowing us to
+     * set operator locality for those ports that are strictly from inside the module.
+     */
+    Queue<SourcedOutputPort> unconnectedPorts = new ArrayQueue<>();
```
bhupeshchawda: Why using this implementation of queue?

library/src/test/java/com/datatorrent/lib/stream/MultipleStreamMergerTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+  @Test
+  public void testApplication() throws IOException, Exception
+  {
+    try {
+      LocalMode lma = LocalMode.newInstance();
+      Configuration conf = new Configuration(false);
+      lma.prepareDAG(new Application(), conf);
+      LocalMode.Controller lc = lma.getController();
+      lc.run(15000); // runs for 10 seconds and quits
```
bhupeshchawda: This approach may run the app shorter or longer than required.
May be this is okay for now, but the suggested approach for testing in local mode is to use lc.runAsync(). See https://github.com/apache/apex-malhar/tree/79eeff782f5cba09bda633dc760cd34190b694c4/demos/iteration/src/test/java/com/datatorrent/demos/iteration for an example.

bhupeshchawda:
@ilganeli Apologies for the delay in replying.
Was the malhar dependency included in the pom dependencies? I think you need to include the SNAPSHOT version as a dependency in your application pom and then create the apa.

ilganeli:
@bhupeshchawda I've included the Malhar dependency from the updated Malhar version, compiled the APA, and uploaded it to the cluster. However, when I try to execute it I still get the error that the MutlipleStreamMerger.class is not found.

bhupeshchawda:
@ilganeli Did the suggestion on the mailing list help?

vrozov:
@ilganeli @bhupeshchawda closed due to inactivity

PR #318, opened 2016-06-15
APEXMALHAR-2119 add setters for partition count and index.

tushargosavi:


library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -1056,10 +1056,15 @@ protected Pattern getRegex()
       return pathSet;
     }
 
+    protected int getParition(String filePathStr)
```
amberarrow: getParition => getPartition

amberarrow:
Suggest adding a unit test for a custom DirectoryScanner that invokes these new methods.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
tushargosavi: done ..

tushargosavi:
Added test case

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -1110,6 +1116,16 @@ public String toString()
...
+
+    protected void setPartitionIndex(int partitionIndex)
+    {
+      this.partitionIndex = partitionIndex;
+    }
+
+    protected void setPartitionCount(int partitionCount)
+    {
+      this.partitionCount = partitionCount;
+    }
```
amberarrow, line 1128: Don't see a test for the newly added methods setPartitionIndex and setPartitionCount (since the rationale in the ticket is "Inherited DirectoryScanner can not set them", we should have a test that shows that this situation has
been remedied).

tushargosavi, line 1128: With the kryo.clone change in createPartition, it is getting set on the derived object from there. Initially createPartition was creating object of type DirectoryScanner, so derived class had no option to override it and create object of own type, and then set the count and index value, but it could not  as those were private without setters in base class.

tushargosavi, line 1128: Added test for setPartition and verified that the base function usage the value set.

PR #319, opened 2016-06-15
APEXMALHAR-2085: Operator supporting the Beam concepts of windowing, watermarks, triggering and accumulation

davidyan74:
This review-only PR contains the interfaces and a first rough draft of the implementation of the operator that supports the concepts of windowing, watermarks, triggering and accumulation mode specified by the Apache Beam API. 

When you review this PR, please:
- note the TODO comments
- look at how the pi and wordcount samples work
- try running the pi and wordcount samples (there is a main method that you can run in your IDE)

thank you.

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/window/Accumulation.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/window/Accumulation.java
@@ -1,0 +1,76 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.lib.window;
+
+import org.apache.hadoop.classification.InterfaceStability;
+
+/**
+ * This interface is for the processing part of the WindowedOperator.
+ * We can assume that all stateful processing of the WindowedOperator is a form of accumulation.
+ *
+ * In most cases, AccumT is the same as OutputT. But in some cases, the accumulated type and the output type may be
+ * different. For example, if we are doing the AVERAGE of doubles, InputT will be double, and we need the SUM and the
+ * COUNT stored as type AccumT, and AccumT will be a pair of double and long, in which double is the sum of the inputs,
+ * and long is the number of inputs. OutputT will be double, because it represents the average of the inputs.
+ */
+@InterfaceStability.Evolving
+public interface Accumulation<InputT, AccumT, OutputT>
+{
+  /**
+   * Returns the default accumulated value when nothing has been accumulated
+   *
+   * @return
+   */
+  AccumT defaultAccumulatedValue();
+
+  /**
+   * Accumulates the input to the accumulated value
+   *
+   * @param accumulatedValue
+   * @param input
+   * @return
+   */
+  AccumT accumulate(AccumT accumulatedValue, InputT input);
+
+  /**
+   * Merges two accumulated value into one
+   *
+   * @param accumulatedValue1
+   * @param accumulatedValue2
+   * @return
+   */
+  AccumT merge(AccumT accumulatedValue1, AccumT accumulatedValue2);
+
+  /**
+   * Gets the output of the accumulated value. This is used for generating the data for triggers
+   *
+   * @param accumulatedValue
+   * @return
+   */
+  OutputT getOutput(AccumT accumulatedValue);
+
+  /**
+   * Gets the retraction of the value. This is used for retracting previous panes in
+   * ACCUMULATING_AND_RETRACTING accumulation mode
+   *
+   * @param value
+   * @return
+   */
+  OutputT getRetraction(OutputT value);
+}
--- a/library/src/main/java/org/apache/apex/malhar/lib/window/TriggerOption.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/window/TriggerOption.java
@@ -1,0 +1,278 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.lib.window;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import org.joda.time.Duration;
+
+import org.apache.hadoop.classification.InterfaceStability;
+
+import com.esotericsoftware.kryo.serializers.FieldSerializer;
+import com.esotericsoftware.kryo.serializers.JavaSerializer;
+
+/**
+ * This class describes how triggers should be fired for each window.
+ * For each window, a trigger can be fired before the watermark (EARLY), at the watermark (ON_TIME), or after the watermark (LATE).
+ * If a LATE trigger is specified and the accumulation mode is ACCUMULATING, it is important for the WindowOption to
+ * specify the allowed lateness because otherwise, all states must be kept in storage.
+ *
+ */
+@InterfaceStability.Evolving
+public class TriggerOption
+{
+
+  public enum AccumulationMode
+  {
+    DISCARDING,
+    ACCUMULATING,
+    ACCUMULATING_AND_RETRACTING
+  }
+
+  private AccumulationMode accumulationMode = AccumulationMode.DISCARDING;
+  private boolean firingOnlyUpdatedPanes = false;
+
+  /**
+   * Whether the trigger should be fired before the watermark, at the watermark, or after the watermark
+   */
+  public enum Type
+  {
+    EARLY,
+    ON_TIME,
+    LATE
+  }
+
+  /**
+   * This class represents the individual trigger spec.
+   */
+  public static class Trigger
+  {
+    protected Type type;
+
+    private Trigger()
+    {
+      // for kryo
+    }
+
+    Trigger(Type type)
+    {
+      this.type = type;
+    }
+
+    public Type getType()
+    {
+      return type;
+    }
+  }
+
+  /**
+   * This class represents a trigger spec in which triggers are fired at regular time intervals.
+   */
+  public static class TimeTrigger extends Trigger
+  {
+    @FieldSerializer.Bind(JavaSerializer.class)
+    Duration duration;
+
+    private TimeTrigger()
+    {
+      // for kryo
+    }
+
+    public TimeTrigger(Type type, Duration duration)
+    {
+      super(type);
+      this.duration = duration;
+    }
+
+    public Duration getDuration()
+    {
+      return duration;
+    }
+  }
+
+  /**
+   * This class represents a trigger spec in which triggers are fired every n tuples
+   */
+  public static class CountTrigger extends Trigger
+  {
+    private long count;
+    private CountTrigger()
+    {
+      //for kryo
+    }
+
+    public CountTrigger(Type type, long count)
+    {
+      super(type);
+      this.count = count;
+    }
+
+    public long getCount()
+    {
+      return count;
+    }
+  }
+
+  List<Trigger> triggerList = new ArrayList<>();
+
+  /**
+   * Creates a TriggerOption with an initial trigger that should be fired at the watermark
+   *
+   * @return
+   */
+  public static TriggerOption AtWatermark()
+  {
+    TriggerOption triggerOption = new TriggerOption();
+    Trigger trigger = new Trigger(Type.ON_TIME);
+    triggerOption.triggerList.add(trigger);
+    return triggerOption;
+  }
+
+  /**
+   * A trigger should be fired before the watermark once for every specified duration
+   *
+   * @param duration
+   * @return
+   */
+  public TriggerOption withEarlyFiringsAtEvery(Duration duration)
+  {
+    TimeTrigger trigger = new TimeTrigger(Type.EARLY, duration);
+    triggerList.add(trigger);
+    return this;
+  }
+
+  /**
+   * A trigger should be fired before the watermark once for every n tuple(s)
+   *
+   * @param count
+   * @return
+   */
+  public TriggerOption withEarlyFiringsAtEvery(long count)
+  {
+    CountTrigger trigger = new CountTrigger(Type.EARLY, count);
+    triggerList.add(trigger);
+    return this;
+  }
+
+  /**
+   * A trigger should be fired after the watermark once for every specified duration
+   *
+   * @param duration
+   * @return
+   */
+  public TriggerOption withLateFiringsAtEvery(Duration duration)
+  {
+    TimeTrigger trigger = new TimeTrigger(Type.LATE, duration);
+    triggerList.add(trigger);
+    return this;
+  }
+
+  /**
+   * A trigger should be fired after the watermark once for every n late tuple(s)
+   *
+   * @param count
+   * @return
+   */
+  public TriggerOption withLateFiringsAtEvery(long count)
+  {
+    CountTrigger trigger = new CountTrigger(Type.LATE, count);
+    triggerList.add(trigger);
+    return this;
+  }
+
+  /**
+   * With discarding mode, the state is discarded after each trigger
+   *
+   * @return
+   */
+  public TriggerOption discardingFiredPanes()
+  {
+    this.accumulationMode = AccumulationMode.DISCARDING;
+    return this;
+  }
+
+  /**
+   * With accumulating mode, the state is kept
+   *
+   * @return
+   */
+  public TriggerOption accumulatingFiredPanes()
+  {
+    this.accumulationMode = AccumulationMode.ACCUMULATING;
+    return this;
+  }
+
+  /**
+   * With accumulating and retracting mode, the state is kept, and the snapshot of the state is saved after each trigger
+   * so when new values come in that change the state, a retraction trigger can be fired with the snapshot of the state
+   * when the last trigger was fired
+   *
+   * @return
+   */
+  public TriggerOption accumulatingAndRetractingFiredPanes()
+  {
+    this.accumulationMode = AccumulationMode.ACCUMULATING_AND_RETRACTING;
+    return this;
+  }
+
+  /**
+   * Only fire triggers for data that has changed from the last trigger. This only applies to ACCUMULATING and
+   * ACCUMULATING_AND_RETRACTING accumulation modes.
+   *
+   * @return
+   */
+  public TriggerOption firingOnlyUpdatedPanes()
+  {
+    this.firingOnlyUpdatedPanes = true;
+    return this;
+  }
+
+  /**
+   * Gets the accumulation mode
+   *
+   * @return
+   */
+  public AccumulationMode getAccumulationMode()
+  {
+    return accumulationMode;
+  }
+
+  /**
+   * Gets the trigger list
+   *
+   * @return
+   */
+  public List<Trigger> getTriggerList()
+  {
+    return Collections.unmodifiableList(triggerList);
+  }
+
+  /**
+   * Returns whether we should only fire panes that have been updated since the last trigger.
+   * When this option is set, DISCARDING accumulation mode must not be used.
+   *
+   * @return
+   */
+  public boolean isFiringOnlyUpdatedPanes()
+  {
+    return this.firingOnlyUpdatedPanes;
+  }
+}
--- a/library/src/main/java/org/apache/apex/malhar/lib/window/Window.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/window/Window.java
@@ -1,0 +1,194 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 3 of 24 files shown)


stream/src/main/java/org/apache/apex/malhar/stream/window/Accumulation.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.hadoop.classification.InterfaceStability;
+
+/**
+ * This interface is for the processing part of the WindowedOperator.
+ * We can assume that all processing of the WindowedOperator is a form of accumulation since a filter would not
+ * require windows
+ *
+ */
+@InterfaceStability.Evolving
+public interface Accumulation<InputT, AccumT, OutputT>
```
tweise: Might be helpful to add an example where AccumT is different from OutputT

stream/src/main/java/org/apache/apex/malhar/stream/window/WindowedStorage.java
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.hadoop.classification.InterfaceStability;
+
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * This interface is for a key/value store for storing data for windowed streams.
+ * The key to this key/value store is a pair of (Window, K)
+ *
+ * Note that this interface does not take recovery into consideration.
```
tweise: Maybe state that implementations need to do that and how.

stream/src/main/java/org/apache/apex/malhar/stream/window/WindowedStorage.java
```diff
@@ -0,0 +1,134 @@
...
+   * @return
+   */
+  Set<Map.Entry<K, V>> entrySet(Window window);
+
+  /**
+   * Gets the windows in the storage for a given timestamp
+   *
+   * @param timestamp
+   */
+  Set<Window> windowsIncluding(long timestamp);
```
tweise: Any insight on how these operations can be implemented efficiently with the backing storage for high throughput streams?

stream/src/main/java/org/apache/apex/malhar/stream/window/impl/DefaultWindowedStorageImpl.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
+
+/**
+ * Created by david on 6/14/16.
+ */
+public class DefaultWindowedStorageImpl<K, V> implements WindowedStorage<K, V>
```
tweise: In the spirit of designing an interface that allows for efficient storage operation with large state, can you consider 
https://issues.apache.org/jira/browse/APEXMALHAR-2026
as well as managed state?

stream/src/main/java/org/apache/apex/malhar/stream/window/impl/WindowedOperatorImpl.java
```diff
@@ -0,0 +1,436 @@
...
+    extends BaseOperator implements WindowedOperator<InputT, KeyT, AccumT, OutputT>
+{
+  // TODO: Need further discussion on the type parameters. InputT and OuputT may be a watermark, a KV pair, a WindowedValue, or a plain data object
+
+  private WindowOption windowOption;
+  private Accumulation<InputT, AccumT, OutputT> accumulation;
+  private WindowedStorage<KeyT, AccumT> dataStorage;
+  private WindowedStorage<KeyT, AccumT> retractionStorage;
+
+  private TreeMap<Window, WindowState> windowStateMap = new TreeMap<>();
```
tweise: Assumes that there will only be a few windows?

stream/src/main/java/org/apache/apex/malhar/stream/window/impl/WindowedOperatorImpl.java
```diff
@@ -0,0 +1,436 @@
...
+            break;
+          }
+          default:
+            throw new IllegalStateException("There are more than two sessions matching one timestamp");
+        }
+      }
+    }
+  }
+
+  private List<Window.TimeWindow> getTimeWindowsFromTimestamp(long timestamp)
```
tweise: ForTimestamp? Even though it's private, some javadoc wouldn't hurt here.

stream/src/main/java/org/apache/apex/malhar/stream/window/impl/WindowedOperatorImpl.java
```diff
@@ -0,0 +1,436 @@
...
+    //output.emit(watermark);
+
+    // TODO: if join operation (multiple input ports), we need to keep track of the watermark on all ports, and then propagate the watermark only when a new watermark arrives with the smallest timestamp among all the ports
+  }
+
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    if (this.firstWindowMillis <= 0) {
+      this.firstWindowMillis = System.currentTimeMillis();
```
tweise: Should eventually be the DAG start time. We need to add that in core in the future. Till then we could also derive it from first windowId.

stream/src/main/java/org/apache/apex/malhar/stream/window/WindowedStorage.java
```diff
@@ -0,0 +1,134 @@
...
+   */
+  void put(Window window, K key, V value);
+
+  /**
+   * Gets the key/value pairs associated with the given window
+   *
+   * @param window
+   * @return
+   */
+  Set<Map.Entry<K, V>> entrySet(Window window);
```
tushargosavi: Can it be Iterator<Map.Entry<K,V>>  In case state is huge and stored in fs

stream/src/main/java/org/apache/apex/malhar/stream/api/WindowedStream.java
```diff
@@ -53,4 +60,121 @@
...
+  <O, K, STREAM extends ApexStream<KeyValPair<K, O>>> STREAM foldByKey(Function.FoldFunction<T, KeyValPair<K, O>> fold);
+
+
+  /**
+   * Reduce transformation<br>
+   * Add an operator to the DAG which merge tuple t1, t2 to new tuple
+   * @param reduce reduce function
+   * @return new stream of same type
+   */
+  <STREAM extends ApexStream<T>> STREAM reduce(Function.ReduceFunction<T> reduce);
```
tushargosavi: reduceByKey ??

siyuanh:
@davidyan74 There are some compile errors on Stream API and classes depends on that due to your change, Shall I resolve those issues?

davidyan74:
@siyuanh Yes, please do so. Also review my changes and see if you're okay with them

davidyan74:
@siyuanh Note that I removed the support for count based window (for now), because the concepts of watermarks, timestamp, early/late triggers don't apply any more. let me know if you want to discuss this further.

stream/src/main/java/org/apache/apex/malhar/stream/window/impl/WindowedOperatorImpl.java
davidyan74: done

stream/src/main/java/org/apache/apex/malhar/stream/window/impl/WindowedOperatorImpl.java
davidyan74: I have abstracted out the storage for the window state

stream/src/main/java/org/apache/apex/malhar/stream/window/Accumulation.java
davidyan74: done

stream/src/main/java/org/apache/apex/malhar/stream/window/WindowedStorage.java
davidyan74: done

stream/pom.xml
```diff
@@ -94,6 +94,18 @@
       <artifactId>cglib</artifactId>
       <version>3.2.1</version>
     </dependency>
+    <dependency>
+      <!-- required by twitter demo -->
```
tweise: Why this dependency?

stream/src/main/java/org/apache/apex/malhar/stream/window/Tuple.java
```diff
@@ -0,0 +1,157 @@
...
+      this.windows.add(window);
+    }
+  }
+
+  /**
+   * Tuple that represents a watermark
+   *
+   * @param <T>
+   */
+  class WatermarkTuple<T> extends TimestampedTuple<T> implements Watermark
```
tweise: This won't work. Watermarks need to be sent through a different port and stream to handle partitioning. In the future, this would be a prime candidate for custom control tuple.

davidyan74: Should we add the support for custom control tuple in the core engine sooner to avoid having a separate port for watermark? This will also help with batch support.

chinmaykolhatkar:
Can we have a separate project for windowed operator and related concepts instead of putting in malhar-library? Maybe "malhar-window".

Reason being if one want to write a transformation other than what is present, he/she can just depend on malhar-window and get going instead of adding dependency of malhar-library.

I think this would also provide a good distinction of feature of windowing.

davidyan74:
@chinmaykolhatkar About your point on dependency, malhar-window would depend on malhar-library anyway because of the usage of KeyValPair. But I still think it's a good idea because the concepts here are very different from the rest of malhar library. @siyuanh and @tweise and everyone else, what are your thoughts on that?

chinmaykolhatkar:
KeyValPair is extended from AbstractMap.SimpleEntry which can be directly used. I don't think KeyValPair as such does anything otherwise creating default constructor for deserialization. I think if KeyValPair is the only reason for dependency of malhar-library, then I think we can find a way out.

I uust think this concept should be highlighted separately.

tweise:
I don't see the need for separate module. malhar-library has minimal dependencies and the window functionality can be added as proposed by David. I don't foresee situation where someone would only use the window operator to build an application.

siyuanh:
I also think it should be part of malhar-library.

library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
```diff
@@ -0,0 +1,484 @@
...
+import com.datatorrent.api.annotation.InputPortFieldAnnotation;
+import com.datatorrent.common.util.BaseOperator;
+
+/**
+ * This is the abstract windowed operator class that implements most of the windowing, triggering, and accumulating
+ * concepts. The subclass of this abstract class is supposed to provide the implementation of how the accumulated
+ * values are stored in the storage.
+ *
+ */
+public abstract class AbstractWindowedOperator<InputT, AccumT, OutputT, DataStorageT extends WindowedStorage, AccumulationT extends Accumulation>
```
chinmaykolhatkar: I don't see AccumT used anywhere in this class. If unused can we remove this from Generic type list?
Same for WindowedOperator too.. AccumT & OutputT is unused.

library/src/main/java/org/apache/apex/malhar/lib/window/impl/KeyedWindowedOperatorImpl.java
```diff
@@ -0,0 +1,159 @@
...
+import org.apache.apex.malhar.lib.window.WindowState;
+import org.apache.apex.malhar.lib.window.WindowedKeyedStorage;
+
+import com.datatorrent.lib.util.KeyValPair;
+
+/**
+ * This is an implementation of WindowedOperator that takes in key value pairs as input and gives out key value pairs
+ * as output. If your operation is not key based, please use {@link WindowedOperatorImpl}.
+ */
+public class KeyedWindowedOperatorImpl<KeyT, InputValT, AccumT, OutputValT>
```
chinmaykolhatkar, line 48: Can you make this evolving?

chinmaykolhatkar:
In general, can we make the concrete class and interfaces evolving?

library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
davidyan74: Good point. done.

library/src/main/java/org/apache/apex/malhar/lib/window/impl/KeyedWindowedOperatorImpl.java
davidyan74, line 48: done

library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
```diff
@@ -0,0 +1,486 @@
...
+  public final transient DefaultOutputPort<ControlTuple> controlOutput = new DefaultOutputPort<>();
+
+  /**
+   * Process the incoming data tuple
+   *
+   * @param tuple
+   */
+  public void processTuple(Tuple<InputT> tuple)
+  {
+    long timestamp = extractTimestamp(tuple);
```
chinmaykolhatkar, line 126: extractTimestamp method is called here, then in getWindowsValue and also in assignWindow for the same tuple.
If its a TimestampedTuple its fine as as its simple a getter method but if TimeExtractor is set the operation of extracting timestamp might get costly depending on the logic in extractor.

Can this call be made only once?

library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
```diff
@@ -0,0 +1,486 @@
...
+      } else {
+        return 0;
+      }
+    } else {
+      return timestampExtractor.apply(tuple.getValue());
+    }
+  }
+
+  private void assignWindows(List<Window> windows, Tuple<InputT> inputTuple)
+  {
```
chinmaykolhatkar, line 307: Minor nit: Considering WindowedTuple has a addWindow method, maybe it just make sense to have only single parameter to this method "WindowedTuple<InputT> tuple".. Rest all can be derived from this object.

library/src/main/java/org/apache/apex/malhar/lib/window/Window.java
```diff
@@ -0,0 +1,179 @@
...
+
+import java.util.Comparator;
+
+import org.apache.hadoop.classification.InterfaceStability;
+
+/**
+ * This interface describes the individual window.
+ */
+@InterfaceStability.Evolving
+public interface Window
```
brightchen, line 29: I think for Window, we could only provide one interface isWithinWIndow(long time); So for GlobalWindow, we can simplly return true. And in SessionWindow, instead of using static method to merge the window, we can just extendWindow(). The Window interface looks like assume there has a fixed boundary(it could be the interface of FixedWindow). Both GlobalWindow and SessionWindow don't have fixed boundary.

library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
```diff
@@ -0,0 +1,490 @@
...
+
+
+  @Override
+  public void processWatermark(ControlTuple.Watermark watermark)
+  {
+    currentWatermark = watermark.getTimestamp();
+    long horizon = currentWatermark - allowedLatenessMillis;
+    if (allowedLatenessMillis >= 0) {
+      // purge window that are too late to accept any more input
+      dataStorage.removeUpTo(horizon);
```
chinmaykolhatkar: Before removing from storage, should we check whether triggers are fired and data is send to downstream for window that is too late?

library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
```diff
@@ -0,0 +1,490 @@
...
+      Window window = entry.getKey();
+      WindowState windowState = entry.getValue();
+      if (allowedLatenessMillis >= 0 && window.getBeginTimestamp() + window.getDurationMillis() < horizon) {
+        // discard this window because it's too late now
+        it.remove();
+      } else if (window.getBeginTimestamp() + window.getDurationMillis() < currentWatermark) {
+        // watermark has not arrived for this window before, marking this window late
+        if (windowState.watermarkArrivalTime == -1) {
+          windowState.watermarkArrivalTime = currentDerivedTimestamp;
+          if (triggerAtWatermark) {
```
chinmaykolhatkar: Shouldn't triggerAtWatermark be a default option?
As per beam specification, trigger at watermark always happens.

library/src/main/java/org/apache/apex/malhar/lib/window/TriggerOption.java
```diff
@@ -0,0 +1,278 @@
...
+
+  private AccumulationMode accumulationMode = AccumulationMode.DISCARDING;
+  private boolean firingOnlyUpdatedPanes = false;
+
+  /**
+   * Whether the trigger should be fired before the watermark, at the watermark, or after the watermark
+   */
+  public enum WatermarkOpt
+  {
+    EARLY,
```
chinmaykolhatkar, line 58: The terminology of WatermarkOpt is confusion. can we change this to TriggerType?

library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
davidyan74: This gives an option to not trigger at watermark, which may be useful.

library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
davidyan74: good point, will make the changes

library/src/main/java/org/apache/apex/malhar/lib/window/Window.java
davidyan74, line 29: Session window can be extended, and two session windows can be merged into one.

library/src/main/java/org/apache/apex/malhar/lib/window/TriggerOption.java
davidyan74, line 58: We can. @siyuanh any problem with that?

library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
chinmaykolhatkar: Can we make this property default true in that case?
If we're going by mean specifications, then even if TriggerOption list not having ON_TIME is set, should make sure by default that trigger happens at watermark.

library/src/main/java/org/apache/apex/malhar/lib/window/WindowedStorage.java
```diff
@@ -0,0 +1,88 @@
...
+
+import org.apache.hadoop.classification.InterfaceStability;
+
+/**
+ * WindowedStorage is a key-value store with the key being the window. The implementation of this interface should
+ * make sure checkpointing and recovery will be done correctly.
+ *
+ * @param <T> The type of the data that is stored per window
+ *
+ * TODO: Look at the possibility of integrating spillable data structure: https://issues.apache.org/jira/browse/APEXMALHAR-2026
```
bhupeshchawda, line 34: This seems very similar to the managed state implementations. Can we also look at managed state directly before looking at Spillable data structures?

library/src/main/java/org/apache/apex/malhar/lib/window/Accumulation.java
```diff
@@ -0,0 +1,76 @@
...
+  AccumT defaultAccumulatedValue();
+
+  /**
+   * Accumulates the input to the accumulated value
+   *
+   * @param accumulatedValue
+   * @param input
+   * @return
+   */
+  AccumT accumulate(AccumT accumulatedValue, InputT input);
```
vrozov, line 49: What is the reason to make the interface stateless? Should it be stateful? My concern is with the interface promoting boxing/unboxing pattern and leading to an excessive GC.

davidyan74, line 49: @vrozov Good point. I'm assuming you're asking why not: 

``` java
  void accumulate(AccumT accumulatedValue, InputT input);
```

with accumulatedValue updated in place. But doing it will make it a lot less flexible because the underlying storage might not support this kind of operation. For example, if the storage supports get(key) and put(key, value) with get(key) returning not a reference to the actual object (possibly as a result of deserialization), then it would not work.

library/src/main/java/org/apache/apex/malhar/lib/window/WindowedStorage.java
davidyan74, line 34: @bhupeshchawda We will for sure look at ManagedState.

library/src/main/java/org/apache/apex/malhar/lib/window/Accumulation.java
vrozov, line 49: My suggestion is to delegate to a class that implements Accumulation interface handling of accumulation and change interface to 

```
void accumulate(InputT input);
void merge(Accumulation accumulatedValue);
OutputT getOutput();
```

The implementation class will need to define how it handles accumulation and how AccumT is defined. The implementation may use Collection for AccumT or may use primitive types such as int, long or double to accumulate values.

library/src/test/java/org/apache/apex/malhar/lib/window/SumAccumulation.java
```diff
@@ -0,0 +1,55 @@
...
+  @Override
+  public Long defaultAccumulatedValue()
+  {
+    return 0L;
+  }
+
+  @Override
+  public Long accumulate(Long accumulatedValue, Long input)
+  {
+    return accumulatedValue + input;
```
vrozov: This is an example of the interface promoting an implementation that is subject to boxing and GC. It will be much more efficient to accumulate values in an internal long member variable.

davidyan74:

siyuanh:
@chinmaykolhatkar @brightchen @tweise  If you guys are all ok with the change. I will merge it.
@vrozov  I think you make a good point of the Accumulation interface, but can we discuss and solve the problem in separate jira? If so, could you please create a ticket for that? Thanks!

tweise:
I would prefer a bit more clarity on the storage part. We already have ManagedState and the spillable data structures in-memory implementation, is there a reason why we don't use them here?

chinmaykolhatkar:
@siyuanh I agree with the changes done and is good to go as first version. Further changes to this can follow after this.

davidyan74:
@tweise We don't have an implementation of SpillableByteMap yet in master. Looks like this PR from @ilooner https://github.com/apache/apex-malhar/pull/324 has not provided that yet, but when we have an in-memory spillable map implementation, we will make another PR to change the current in-memory implementation of WindowedStorage. This should not be a blocker for merging this PR.

ilooner:
@davidyan74 This PR may not be the place for it. But it would be good to have discussion about what Spillable datastructures will be needed. To me it looks like your WindowStorage interfaces are basically array list multimaps. If that's the case I would recommend you make a note on your Storage interfaces that they are temporary and will likely disappear very soon, since SpillableDatastructures already provides that abstraction.

As long as we do those things, I think it would be acceptable to pull this in and make the improvements in the next round of changes.

ilooner:
@davidyan74 In fact it may be a good exercise to validate the Spillable Data Structures interfaces by using the InMemorySpillableArrayListMultimap which is already in master as a next step.

davidyan74:
@ilooner Thanks will take a look at InMemorySpillableArrayListMultimap

davidyan74:
@tweise I already marked the WindowedStorage interface _Unstable_ and a note in the javadoc that the WindowedStorage interfaces may change or go away entirely soon and we have plans to integrate with spillable data structures in the very near future. Let me know if you think this PR is good to go.

tweise:
I think this PR is good to go but we need to see the follow up work for scalable checkpointing and benchmarking would be good also.

vrozov:
Filed two JIRAs - one to revisit Accumulation interface (https://issues.apache.org/jira/browse/APEXMALHAR-2145) and one to benchmark new operators and evaluate boxing impact (https://issues.apache.org/jira/browse/APEXMALHAR-2146)

PR #321, opened 2016-06-16
APEXMALHAR-2120 #resolve #comment solve problems of KafkaInputOperato…

brightchen:
…rTest and AbstractKafkaInputOperator

### DIFF
--- a/kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
+++ b/kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
@@ -32,6 +32,7 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -74,7 +75,7 @@ public class KafkaConsumerWrapper implements Closeable
 
   private static final Logger logger = LoggerFactory.getLogger(KafkaConsumerWrapper.class);
 
-  private boolean isAlive = false;
+  private AtomicBoolean isAlive = new AtomicBoolean(false);
 
   private final Map<String, KafkaConsumer<byte[], byte[]>> consumers = new HashMap<>();
 
@@ -129,6 +130,12 @@ public class KafkaConsumerWrapper implements Closeable
         if (meta.getTopicPartition().equals(tp)) {
           kc.resume(tp);
         } else {
+          try {
+            kc.position(tp);
+          } catch (NoOffsetForPartitionException e) {
+            //the poll() method of a consumer will throw exception if any of subscribed consumers not initialized with position
+            handleNoOffsetForPartitionException(e, kc);
+          }
           kc.pause(tp);
         }
       }
@@ -188,7 +195,7 @@ public class KafkaConsumerWrapper implements Closeable
       try {
 
 
-        while (wrapper.isAlive) {
+        while (wrapper.isAlive.get()) {
           if (wrapper.waitForReplay) {
             Thread.sleep(100);
             continue;
@@ -207,19 +214,7 @@ public class KafkaConsumerWrapper implements Closeable
               wrapper.putMessage(Pair.of(cluster, record));
             }
           } catch (NoOffsetForPartitionException e) {
-            // if initialOffset is set to EARLIST or LATEST
-            // and the application is run as first time
-            // then there is no existing committed offset and this error will be caught
-            // we need to seek to either beginning or end of the partition
-            // based on the initial offset setting
-            AbstractKafkaInputOperator.InitialOffset io =
-                AbstractKafkaInputOperator.InitialOffset.valueOf(wrapper.ownerOperator.getInitialOffset());
-            if (io == AbstractKafkaInputOperator.InitialOffset.APPLICATION_OR_EARLIEST
-                || io == AbstractKafkaInputOperator.InitialOffset.EARLIEST) {
-              consumer.seekToBeginning(e.partitions().toArray(new TopicPartition[0]));
-            } else {
-              consumer.seekToEnd(e.partitions().toArray(new TopicPartition[0]));
-            }
+            wrapper.handleNoOffsetForPartitionException(e, consumer);
           } catch (InterruptedException e) {
             throw new IllegalStateException("Consumer thread is interrupted unexpectedly", e);
           }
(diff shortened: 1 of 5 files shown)


kafka/src/test/java/org/apache/apex/malhar/kafka/KafkaInputOperatorTest.java
```diff
@@ -68,26 +70,53 @@
...
   }
 
+
+
   @Before
   public void before()
   {
-    FileUtils.deleteQuietly(new File(APPLICATION_PATH));
-    tupleCollection.clear();
+    //FileUtils.deleteQuietly(new File(APPLICATION_PATH));
```
tweise: Why is this commented out?

kafka/src/test/java/org/apache/apex/malhar/kafka/KafkaInputOperatorTest.java
```diff
@@ -275,10 +324,21 @@ public void testInputOperator(boolean hasFailure, boolean idempotent) throws Exc
     final LocalMode.Controller lc = lma.getController();
     lc.setHeartbeatMonitoringEnabled(false);
 
-    lc.runAsync();
```
tweise, line 278: How is the test running? Why do we need another thread and the cast?

kafka/src/test/java/org/apache/apex/malhar/kafka/KafkaInputOperatorTest.java
brightchen: The different test case/test Scenario use difference directory. So it is not necessary to delete the directory

kafka/src/test/java/org/apache/apex/malhar/kafka/KafkaInputOperatorTest.java
brightchen, line 278: I added a monitor thread to make sure the Engine will be complete shutdown before start next test case

kafka/pom.xml
```diff
@@ -198,7 +198,7 @@
     <dependency>
       <groupId>org.apache.kafka</groupId>
       <artifactId>kafka_2.11</artifactId>
-      <version>0.9.0.0</version>
```
siyuanh, line 201: Why do we change the version?

brightchen, line 201: 0.9.0.1 fixed some bugs, it should be more stable than 0.9.0.0. I think it's better to upgrade before a lot of customers using it. Following is Kafka 0.9.0.1 release notes: http://apache.claz.org/kafka/0.9.0.1/RELEASE_NOTES.html

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
```diff
@@ -74,7 +75,7 @@
 
   private static final Logger logger = LoggerFactory.getLogger(KafkaConsumerWrapper.class);
 
-  private boolean isAlive = false;
```
siyuanh, line 77: This value can only be changed in Operator main thread, right?

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
```diff
@@ -129,6 +130,11 @@ public void emitImmediately(Map<AbstractKafkaPartitioner.PartitionMeta, Pair<Lon
         if (meta.getTopicPartition().equals(tp)) {
           kc.resume(tp);
         } else {
+          try {
```
siyuanh, line 133: Why do we need to catch the exception here? It will always reposition some partition to the beginning which is not correct

kafka/src/test/java/org/apache/apex/malhar/kafka/KafkaInputOperatorTest.java
```diff
@@ -68,26 +70,53 @@
...
   @Parameterized.Parameters(name = "multi-cluster: {0}, multi-partition: {1}, partition: {2}")
   public static Collection<Object[]> testScenario()
   {
-    return Arrays.asList(new Object[][]{{true, false, "one_to_one"},// multi cluster with single partition
+    return Arrays.asList(new Object[][]{
+      {true, false, "one_to_one"},// multi cluster with single partition
       {true, false, "one_to_many"},
       {true, true, "one_to_one"},// multi cluster with multi partitions
-      {true, true, "one_to_many"},
+      {true, true, "one_to_many"},   //test failed, no data received.
```
siyuanh: Comment is confusing

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
brightchen, line 77: In AbstractKafkaInputOperator operator, a new thread of consumerWrapper will be started: 
  @Override
  public void activate(Context.OperatorContext context)
  {
    consumerWrapper.start(isIdempotent());
  }

 which in fact created a thread pool to execute
// create thread pool for consumer threads
    kafkaConsumerExecutor = Executors.newCachedThreadPool(
      new ThreadFactoryBuilder().setNameFormat("kafka-consumer-%d").build());

  @Override
  public void deactivate()
  {
    consumerWrapper.stop();
  }

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
brightchen, line 133: I had got following exception where run unit test:
org.apache.kafka.clients.consumer.NoOffsetForPartitionException: Undefined offset with no reset policy for partition. please see jira https://datatorrent.atlassian.net/browse/SPOI-8461

kafka/src/test/java/org/apache/apex/malhar/kafka/KafkaInputOperatorTest.java
brightchen: Yes, I'll remove the comments

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
siyuanh, line 77: The stop method cannot guarantee immediate stop either unless you synchronize the stop method with the while block in the separate thread. Here as long as there is no concurrent data write operation, I think it should be fine

kafka/pom.xml
siyuanh, line 201: The operator is supposed to work with any 0.9.x.x kafka brokers and clients unless the 0.9.0.0 API can not meet our requirement.

tweise, line 201: It's a backward compatible patch change, I don't see why it would cause a problem. On the contrary there was earlier some code trying to use the new operator that had to redefine the dependency to 0.9.0.1 due to a bug.

kafka/src/test/java/org/apache/apex/malhar/kafka/KafkaInputOperatorTest.java
```diff
@@ -49,13 +50,14 @@
...
 import com.datatorrent.common.util.BaseOperator;
+import com.datatorrent.lib.util.TestUtils;
 import com.datatorrent.stram.StramLocalCluster;
 
 /**
  * A bunch of test to verify the input operator will be automatically partitioned per kafka partition This test is launching its
  * own Kafka cluster.
  */
-@Ignore
+//@Ignore
```
tweise: delete this

kafka/src/test/java/org/apache/apex/malhar/kafka/KafkaInputOperatorTest.java
tweise: Tests should always delete in case it was leftover from previous run.

kafka/src/test/java/org/apache/apex/malhar/kafka/KafkaInputOperatorTest.java
tweise, line 278: I guess I don't understand what that thread is doing that the main thread cannot achieve?

brightchen:
@siyuanh The NullPointerException issue should be fixed, I've tested more than 10 times. Please review and merger

kafka/src/test/java/org/apache/apex/malhar/kafka/KafkaInputOperatorTest.java
```diff
@@ -264,49 +336,57 @@ public void testInputOperator(boolean hasFailure, boolean idempotent) throws Exc
...
+
+      //wait until control thread finished.
+      monitorThread.join();
+    } catch (Exception e) {
+      logger.warn(e.getMessage());
+    }
+    
+    t.join();
+    
+    logger.info("Number of received/expected tuples: {}/{}, testName: {}, tuples: \n{}", tupleCollection.size(), expectedReceiveCount, testName, tupleCollection);
```
siyuanh: Log this only when something goes wrong

brightchen:
@siyuanh Please review and merge

PR #322, opened 2016-06-18
APEXMALHAR-2063 Made window data manager use file system wal

otterc:
@PramodSSImmaneni @tweise  @ilooner 
Please review

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
@@ -19,6 +19,7 @@
 package org.apache.apex.malhar.lib.state.managed;
 
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ExecutorService;
@@ -33,10 +34,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.apex.malhar.lib.wal.FSWindowDataManager;
+import org.apache.apex.malhar.lib.wal.FileSystemWAL;
 
 import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 import com.google.common.collect.Queues;
+import com.google.common.primitives.Longs;
 
 import com.datatorrent.api.Context;
 import com.datatorrent.api.annotation.Stateless;
@@ -78,7 +81,8 @@ public class IncrementalCheckpointManager extends FSWindowDataManager
   public IncrementalCheckpointManager()
   {
     super();
-    setRecoveryPath(WAL_RELATIVE_PATH);
+    setStatePath(WAL_RELATIVE_PATH);
+    setRelyOnCheckpoints(true);
   }
 
   @Override
@@ -132,7 +136,7 @@ public class IncrementalCheckpointManager extends FSWindowDataManager
             managedStateContext.getBucketsFileSystem().writeBucketData(windowId, singleBucket.getKey(),
                 singleBucket.getValue());
           }
-          storageAgent.delete(managedStateContext.getOperatorContext().getId(), windowId);
+          committed(windowId);
         } catch (Throwable t) {
           throwable.set(t);
           LOG.debug("transfer window {}", windowId, t);
@@ -150,7 +154,7 @@ public class IncrementalCheckpointManager extends FSWindowDataManager
   }
 
   @Override
-  public void save(Object object, int operatorId, long windowId) throws IOException
+  public void save(Object object, long windowId) throws IOException
   {
     throw new UnsupportedOperationException("doesn't support saving any object");
   }
@@ -159,15 +163,13 @@ public class IncrementalCheckpointManager extends FSWindowDataManager
    * The unsaved state combines data received in multiple windows. This window data manager persists this data
    * on disk by the window id in which it was requested.
    * @param unsavedData   un-saved data of all buckets.
-   * @param operatorId    operator id.
    * @param windowId      window id.
    * @param skipWriteToWindowFile flag that enables/disables saving the window file.
    *
    * @throws IOException
    */
-  public void save(Map<Long, Map<Slice, Bucket.BucketedValue>> unsavedData, int operatorId, long windowId,
-      boolean skipWriteToWindowFile)
-      throws IOException
+  public void save(Map<Long, Map<Slice, Bucket.BucketedValue>> unsavedData, long windowId,
+      boolean skipWriteToWindowFile) throws IOException
   {
     Throwable lthrowable;
     if ((lthrowable = throwable.get()) != null) {
@@ -177,25 +179,46 @@ public class IncrementalCheckpointManager extends FSWindowDataManager
     savedWindows.put(windowId, unsavedData);
 
     if (!skipWriteToWindowFile) {
-      super.save(unsavedData, operatorId, windowId);
+      super.save(unsavedData, windowId);
     }
+  }
+
+  /**
+   * Retrieves artifacts available for all the windows saved by the enclosing partitions.
+   * @return  artifact saved per window.
+   * @throws IOException
+   */
+  public Map<Long, Object> retrieveAllWindows() throws IOException
+  {
+    Map<Long, Object> artifactPerWindow = new HashMap<>();
+    FileSystemWAL.FileSystemWALReader reader = getWal().getReader();
+    reader.seek(getWal().getWalStartPointer());
+    
+    Slice windowSlice = readNext(reader);
+    while (reader.getCurrentPointer().compareTo(getWal().getWalEndPointerAfterRecovery()) < 0 && windowSlice != null) {
+      long window = Longs.fromByteArray(windowSlice.toByteArray());
+      Object data = fromSlice(readNext(reader));
+      artifactPerWindow.put(window, data);
+      windowSlice = readNext(reader); //null or next window
+    }
+    reader.seek(getWal().getWalStartPointer());
+    return artifactPerWindow;
   }
 
   /**
    * Transfers the data which has been committed till windowId to data files.
    *
-   * @param operatorId operator id
-   * @param windowId   window id
+   * @param committedWindowId   window id
    */
-  @SuppressWarnings("UnusedParameters")
-  protected void committed(int operatorId, long windowId) throws IOException, InterruptedException
+  @Override
+  public void committed(long committedWindowId) throws IOException
   {
-    LOG.debug("data manager committed {}", windowId);
+    LOG.debug("data manager committed {}", committedWindowId);
     for (Long currentWindow : savedWindows.keySet()) {
       if (currentWindow <= largestWindowAddedToTransferQueue) {
         continue;
       }
-      if (currentWindow <= windowId) {
+      if (currentWindow <= committedWindowId) {
         LOG.debug("to transfer {}", currentWindow);
         largestWindowAddedToTransferQueue = currentWindow;
         windowsToTransfer.add(currentWindow);
--- a/library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
@@ -18,324 +18,687 @@
  */
 package org.apache.apex.malhar.lib.wal;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
+import java.util.NavigableSet;
 import java.util.Set;
 import java.util.SortedSet;
+import java.util.TreeSet;
 
 import javax.validation.constraints.NotNull;
 
-import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.FileStatus;
-import org.apache.hadoop.fs.FileSystem;
-import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import org.apache.apex.malhar.lib.state.managed.IncrementalCheckpointManager;
+import org.apache.apex.malhar.lib.utils.FileContextUtils;
+import org.apache.hadoop.fs.FileContext;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.RemoteIterator;
+
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.io.Output;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
-import com.google.common.collect.TreeMultimap;
+import com.google.common.primitives.Longs;
 
 import com.datatorrent.api.Context;
 import com.datatorrent.api.DAG;
 import com.datatorrent.api.annotation.Stateless;
-import com.datatorrent.common.util.FSStorageAgent;
+import com.datatorrent.lib.io.fs.AbstractFileInputOperator;
+import com.datatorrent.lib.util.KryoCloneUtils;
+import com.datatorrent.netlet.util.Slice;
 
 /**
- * An {@link WindowDataManager} that uses FS to persist state.
+ * An {@link WindowDataManager} that uses FS to persist state every completed application window.<p/>
+ *
+ * FSWindowDataManager uses {@link FSWindowReplayWAL} to write to files. While saving an artifact corresponding
+ * to a window, the window date manager saves:
+ * <ol>
+ *   <li>Window id</li>
+ *   <li>Artifact</li>
+ * </ol>
+ * In order to ensure that all the entries corresponding to a window id are appended to the same wal part file, the
+ * wal operates in batch mode. In batch mode, the rotation of a wal part is done only after a batch is complete.<br/>
+ * <p/>
+ *
+ * <b>Replaying data of a completed window</b><br/>
+ * Main support that {@link WindowDataManager} provides to input operators is to be able to replay windows which
+ * were completely processed but not checkpointed. This is necessary for making input operators idempotent.<br/>
+ * The {@link FileSystemWAL}, however, ignores any data which is not checkpointed after failure. Therefore,
+ * {@link FSWindowDataManager} cannot rely solely on the state in wal after failures and so during recovery it modifies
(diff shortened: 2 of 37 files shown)


otterc:
@ilooner Can you please take a look?

ilooner:
@chandnisingh there are conflicts

library/src/main/java/org/apache/apex/malhar/lib/wal/WindowDataManager.java
```diff
@@ -41,15 +41,42 @@
...
+   * Typically it is used to replay tuples of successive windows in input operators after failure.
+   *
+   * @param windowId window id
+   * @return saved state for the window id.
+   * @throws IOException
+   */
+  Object retrieve(long windowId) throws IOException;
+
+  /**
+   * Delete the artifact corresponding to the
```
ilooner: complete javadoc here?

library/src/main/java/org/apache/apex/malhar/lib/wal/WindowDataManager.java
```diff
@@ -126,36 +146,24 @@ public void teardown()
...
       return null;
     }
 
     @Override
-    public void delete(int operatorId, long windowId) throws IOException
-    {
-    }
-
-    @Override
-    public void deleteUpTo(int operatorId, long windowId) throws IOException
```
ilooner, line 145: These are no longer needed?

otterc, line 145: changed the method to be committed(long windowId). WindowDataManager is Evolving so this was ok.

otterc:
@ilooner I have resolved the conflicts. Please have a look.

otterc:
@ilooner @PramodSSImmaneni @tweise 
This change touches all the operators which use WindowDataManager. Any changes to those operators will cause conflicts again so please review it when you get time.

tweise:
Chandni, yes it is on my TODO list.

## 

sent from mobile

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
```diff
@@ -51,291 +66,607 @@
...
+  private final transient Kryo kryo = new Kryo();
+
+  private transient FileContext fileContext;
 
   public FSWindowDataManager()
   {
-    replayState = TreeMultimap.create();
-    largestRecoveryWindow = Stateless.WINDOW_ID;
-    recoveryPath = DEF_RECOVERY_PATH;
+    kryo.setClassLoader(Thread.currentThread().getContextClassLoader());
```
ilooner, line 154: for my learning :) why is this needed?

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
```diff
@@ -51,291 +66,607 @@
...
+
+          if (slice == null || !slice.equals(DELETED)) {
+            //delete entry not found so window was not deleted
+            long window = Longs.fromByteArray(windowSlice.toByteArray());
+
+            if (ceilingWindow != null && window > ceilingWindow) {
+              break;
+            }
+            endOffset = offset;
+            lastWindow = window;
```
ilooner: I'm not sure if I'm understanding how delete works correctly. But should the lastWindow be rolled back to the previous window when we encounter DELETED?

ilooner: Would the endoffset have to be rolled back as well?

tweise:
CONFLICT (content): Merge conflict in library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
tweise, line 154: Wondering about that also?

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
```diff
@@ -51,291 +66,607 @@
...
-    Configuration configuration = new Configuration();
+    operatorId = context.getId();
+
     if (isRecoveryPathRelativeToAppPath) {
-      appPath = new Path(context.getValue(DAG.APPLICATION_PATH) + Path.SEPARATOR + recoveryPath);
+      statePath = context.getValue(DAG.APPLICATION_PATH) + Path.SEPARATOR + recoveryPath;
     } else {
-      appPath = new Path(recoveryPath);
+      statePath = recoveryPath;
     }
```
tweise, line 166: Is statePath shared for all logical operators? What happens if there are 2 operators in a DAG that are using the window data manager?

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
```diff
@@ -51,291 +66,607 @@
...
+  public void deleteLastWindow() throws IOException
   {
-    return largestRecoveryWindow;
+    wal.getWriter().append(DELETED);
+    wal.getWriter().flush();
   }
 
   @Override
-  public void partitioned(Collection<WindowDataManager> newManagers, Set<Integer> removedOperatorIds)
+  public void committed(long windowId) throws IOException
```
tweise: It would be good to summarize below logic in the javadoc block.

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
```diff
@@ -51,291 +66,607 @@
...
+  }
 
-    if (fs.listStatus(new Path(appPath, Integer.toString(operatorId))).length > 0) {
-      long[] windowsAfterReplay = storageAgent.getWindowIds(operatorId);
-      Arrays.sort(windowsAfterReplay);
-      for (long lwindow : windowsAfterReplay) {
-        if (lwindow <= windowId) {
-          storageAgent.delete(operatorId, lwindow);
-        }
+  public Map<Long, Object> retrieveAllWindows() throws IOException
```
tweise: No reference to this method? Is it needed. If yes, please add javadoc.

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
```diff
@@ -51,291 +66,607 @@
 public class FSWindowDataManager implements WindowDataManager
```
tweise, line 109: It may be useful to add a few lines WRT dynamic partitioning support and the relationship to checkpointing vs. commit per window.

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
```diff
@@ -51,291 +66,607 @@
...
+          //the read only wal can be removed.
+          walIterator.remove();
+        }
+      }
     }
+  }
 
-    return windowIdsArray;
+  @Override
+  public void save(Object object, long windowId) throws IOException
```
tweise, line 409: Is this method supposed to be called from endWindow only? It appears that it assumes batch operation?

tweise:
@chandnisingh  any chance you can look into it?

otterc:
@tweise will address the review comments by the weekend

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
otterc: I remember that if we don't have this, then Kryo sometimes is not able to find a Class. All the kryo utils that we have `KryoSerializableStreamCodec` and `KryoCloneUtils` do set  class loader on the kryo instance explicitly.

otterc:
Resolved conflicts. Will address pending comments today as well

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
otterc: Will move this to `IncrementalCheckpointManager`. `AbstractManagedState` uses it.

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
otterc: Yes this is called from  `endWindow()`. FSWindowDataManager does use the wal in batch mode. The wal instance in this class is initialized with `inBatchMode=true`.
The batch mode was added so that multiple entries related to one window go into the same wal file, that is, data of a window is not distributed over multiple wal files.
WindowDataManager provided the support for deleting last saved artifact. Please see `deleteLastWindow()`. `AbstractJMSInputOperator` uses that call to deleted the last artifact in case ack fails.
In the new implementation, when `deleteLastWindow()` is invoked, I just add marker bytes `DELETED` to the file instead of actually truncating the file. 

Therefore, in order to make sure that the wal file doesn't get rotated between save and delete, the batching was added and FSWindowDataManager needs the batch mode to be set to true.

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
otterc: @ilooner 
delete is handled by the else block from 283-284.
lastWindow and  endOffset are not at all modified for a window that is marked deleted.

I fail to see the problem. Can you please have another look?

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
otterc: Yeah this is a problem. By default `statePath` will be shared by all logical operator. There is another ticket APEXMALHAR-2068 opened for that. Also the fix will need the recent change to Apex Core (APEXCORE-448 ) which will not be there until the core version in malhar changes.

otterc:
Addressed comments

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
tweise: @chandnisingh this question isn't directly related to this PR but a potential bug in AbstractJMSInputOperator. If it removes the incomplete state in the finally block and that does not succeed, then what happens on restart? Should this not be done in setup instead?

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
```diff
@@ -18,324 +18,718 @@
...
+   *                                      otherwise.
    */
-  public void setRecoveryPathRelativeToAppPath(boolean recoveryPathRelativeToAppPath)
+  public void setStatePathRelativeToAppPath(boolean statePathRelativeToAppPath)
   {
-    isRecoveryPathRelativeToAppPath = recoveryPathRelativeToAppPath;
+    isStatePathRelativeToAppPath = statePathRelativeToAppPath;
   }
+
+  protected static Slice DELETED = new Slice("DEL".getBytes());
```
tweise: Is this used to indicate that a window segment wasn't complete? If so, then this isn't working because writing of the marker may also fail and we should write the complete marker instead (which is end of window)? Please clarify, maybe I got that wrong?

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
```diff
@@ -227,6 +240,47 @@ public void setMaxLength(long maxLength)
...
+    return inBatchMode;
+  }
+
+  /**
+   * When in batch mode, a file is rotated only when a batch gets completed. This facilitates writing multiple entries
+   * that will all be written to the same part file.
+   *
+   * @param inBatchMode write in batch mode or not.
+   */
+  public void setInBatchMode(boolean inBatchMode)
```
tweise: Should this really be exposed as public setter?

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowReplayWAL.java
```diff
@@ -0,0 +1,188 @@
...
+import javax.validation.constraints.NotNull;
+
+import org.apache.hadoop.fs.Path;
+
+import com.google.common.collect.TreeMultimap;
+
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * A FileSystemWAL that WindowDataManager uses to save state of every window.
```
tweise: Minor nit, use link fro friendly javadoc and preserve during refactor:

```
 * A {@link FileSystemWAL} that WindowDataManager uses to save state of every window.
```

library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterInputTest.java
```diff
@@ -293,51 +324,56 @@ public void testTrigger() throws InterruptedException, IOException, TimeoutExcep
...
-    for (int i = 0; i < 12; i++) {
-      File testFile = new File(testMeta.dataDirectory, "file" + i + ".txt");
-      noOfBlocks += (int)Math.ceil(testFile.length() / (2 * 1.0));
-    }
-
-    testMeta.fileSplitterInput.setBlockSize(2L);
-    testMeta.fileSplitterInput.setBlocksThreshold(10);
     testMeta.fileSplitterInput.beginWindow(1);
-
+
```
tweise: trailing whitespace, please check the IDE settings.

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
```diff
@@ -18,324 +18,718 @@
...
     }
+  }
 
-    return windowIdsArray;
+  @Override
+  public void save(Object object, long windowId) throws IOException
+  {
+    closeReaders();
+    FileSystemWAL.FileSystemWALWriter writer = wal.getWriter();
+    writer.batchCompleted(); //finish previous batch and rotate if needed
```
tweise: Maybe rotateIfNecessary would be more applicable? Also, it may be good to capture some of this in method javadoc?

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
otterc: JMSInput used this to explicitly delete the last saved artifact.
Earlier this used to delete the file  which also could have failed and not have actually deleted the file.  Even though the delete fails, JMSInput will not emit duplicate because if the entry failed to be deleted then in replay the message from file is added to a `pendingAck` set. New messages from consumer are checked against this set and considered duplicate.

otterc: It was just an effort made by JMS Input to delete the last saved artifact because it ack failed. It didn't rely on that call to be successful. If we add an end marker for window then we need to add an additional method to WindowDataManager api and every operator needs to invoke it after saving a window.

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
otterc: JMS Input in its endWindow does the following:
1. saves the artifact
2. acks the session

When acks fails but the artifact got saved, then it tries to delete the artifact.
If the artifact is deleted successfully, the window is considered completed or not.
1. When deleted - window was not completed and the messages will be consumed from JMS
2. When not deleted - window was considered completed so it is replayed from file. However while replaying message ids are added to pendingAck set. When these messages are received from JMS again (since ack failed earlier), they will be considered re-delivered.

Deletion of a window during `setup` is difficult because we do not actually know if ack failed for that window.

tweise:
Build fails due to addition from another PR. Please rebase and squash.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
```diff
@@ -177,25 +179,50 @@ public void save(Map<Long, Map<Slice, Bucket.BucketedValue>> unsavedData, int op
...
+    Map<Long, Object> artifactPerWindow = new HashMap<>();
+    FileSystemWAL.FileSystemWALReader reader = getWal().getReader();
+    reader.seek(getWal().getWalStartPointer());
+    
+    Slice windowSlice = readNext(reader);
+    while (reader.getCurrentPointer().compareTo(getWal().getWalEndPointerAfterRecovery()) < 0 && windowSlice != null) {
+      long window = Longs.fromByteArray(windowSlice.toByteArray());
+      Object data = fromSlice(readNext(reader));
+      windowSlice = readNext(reader); //null, deleted or next window
+      if (windowSlice == DELETED) {
```
tweise: Won't this always be 2 different objects?

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
tweise: So does this mean the delete isn't really needed? What is the benefit of keeping it? It would be good to document it in the JMS operator. Also it sounds like the DELETED logic is very specific to the JMS operator. If it is not needed anywhere else, then how about pulling it into the JMS operator?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
otterc: Yeah wonder why the test didn't catch this. Will check

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
otterc: Will make this change

otterc: Is it okay if this delete logic is removed from JMS as well? 
It is difficult to pull it into JMS because the api for saving is `save(window id,Object artifact)`. 
If JMSInput calls `save(windowId, DELETED)` then that creates multiple entries for the same window id. So in order for JMSInput to figure out if the last window is deleted or not, it has to read that window twice  which I think is hacky.

Either we completely remove this functionality  or we change WindowDataManager Api to have a trigger for `beginWindow(long window)` in which it can record the window id to wal and `save` just saves the artifact not the window.

tweise: What's the downside of removing it completely?

otterc: Don't think there is any. If the ack failed but the state is saved, the messages are re-played from the state. However since the ack failed, they will also be re-delivered by the message bus but we ignore them. So we do not send any duplicates downstream or miss any message .

tweise: Saving the state and detecting the duplicate is how I would expect it to work. So +1 on removing this logic altogether. It would be good to also document this in the JMSInputOperator.

otterc:
Addressed all the comments

PR #323, opened 2016-06-20
APEXMALHAR-2121: Updating KafkaInputOperator to add emitTuple method with access to more kafka parameters

DT-Priyanka:
Added emitTuple method with KafkaMessage as parameter. A user can override this emitTuple to get access to more kafka parameters and choose what he wants to emit along with kafka Message.

### DIFF
--- a/contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
@@ -29,12 +29,12 @@ import com.datatorrent.api.StatsListener;
 import com.datatorrent.api.annotation.OperatorAnnotation;
 import com.datatorrent.api.annotation.Stateless;
 import com.datatorrent.lib.util.KryoCloneUtils;
-
 import com.google.common.base.Joiner;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
+
 import kafka.api.FetchRequest;
 import kafka.api.FetchRequestBuilder;
 import kafka.cluster.Broker;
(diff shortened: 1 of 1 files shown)


contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
```diff
@@ -203,6 +204,14 @@ public AbstractKafkaInputOperator()
    */
   protected abstract void emitTuple(Message message);
```
siyuanh, line 205: It is good to change this method signature, but can we do that?

DT-Priyanka, line 205: I haven't changed the method signature, that would create backward compatibility problems. I have introduced one more new method, and the default implementation of that method will call the existing emitTuple, so no change for existing users at all. The new method will help others to override the behavior if they need to.

siyuanh, line 205: That's why I'm asking can we do that. It does have compatibility issue but on the other hand if some one wants to emit the tuple along with partition id etc, he/she is forced to implement the emitTuple method as well. Well, we can keep it as it is now. @DT-Priyanka can you file a jira to make this change in 0.9 kafka input operator.  We can change the existing interface for that one

tweise, line 205: I don't see a backward compatibility issue, it's a new method. And yes, the same change needs to be made on the 0.9 operator also.

DT-Priyanka:
Version 0.9 doesn't have this limitation. In 0.9 version the "msg" is of type org.apache.kafka.clients.consumer.ConsumerRecord. This class gives access to Kafka topic, partition and offset along with message.

PR #325, opened 2016-06-21
APEXMALHAR-2122 Upgrading elasticsearch java api from 1.1.2 to 2.3.3. 

akshay-harale:
Upgrading elasticsearch java api from 1.1.2 to 2.3.3.  Made necessary changes in operator. Fixed build issues recognised while migration.


gauravgopi123:
@akshay-harale : I have following feedback
1. Why update the version. Jira says "Current elassticsearch version is 2.3.3 so wee need to upgrade the corresponding java api to 2.3.3.". This doesn't validate the version upgradation
2. Branch name should be APEXMALHAR-2122 instead of master

akshay-harale:
1. Use case is something like this, we were trying to dig into the data using Apex library and that data we wanted to visualise using Kibana dashboard. Kibana Dashboard uses Elastic Search to optimize and provide search on the data also let's us apply the filters. So I wanted to push data using elastic search operator available through Apex malhar and unfortunately that didn't work, was getting issues connecting to elastic search cluster. Then I tried posting the question on apex user forum, which didn't receive any response, so I went ahead and updated the elastic search version in Apex Malhar. This resulted in seamless connection and loading of data in the elastic search clusters through Apex. And that's the sole reason I feel we should update elastic search library being used in apex to latest version available.

ilganeli:
@akshay-harale Is this PR still relevant?

PR #326, opened 2016-06-21
APEXMALHAR-2116 Added FS record reader operator, module, test

yogidevendra:


yogidevendra:
@amberarrow Could you please review this?

amberarrow:
Ok, I'll take a look

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReader.java
```diff
@@ -0,0 +1,175 @@
...
+import org.apache.hadoop.fs.FSDataInputStream;
+
+import com.datatorrent.api.Context.OperatorContext;
+import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.lib.io.block.BlockMetadata;
+import com.datatorrent.lib.io.block.FSSliceReader;
+import com.datatorrent.lib.io.block.ReaderContext;
+
+/**
+ * This operator can be used for reading records/tuples from Filesystem
```
amberarrow: This line and several others have trailing whitespace; this is generally considered not a good thing -- see
for instance: http://programmers.stackexchange.com/questions/121555/why-is-trailing-whitespace-a-big-deal
Suggest removing all trailing white space.

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReader.java
```diff
@@ -0,0 +1,175 @@
...
+import com.datatorrent.api.Context.OperatorContext;
+import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.lib.io.block.BlockMetadata;
+import com.datatorrent.lib.io.block.FSSliceReader;
+import com.datatorrent.lib.io.block.ReaderContext;
+
+/**
+ * This operator can be used for reading records/tuples from Filesystem 
+ * in parallel (without ordering guarantees between tuples).
+ * Records can be delimited (e.g. newline) or fixed with records.
```
amberarrow: fixed with => fixed width

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReader.java
```diff
@@ -0,0 +1,175 @@
...
+  {
+    readerContext.initialize(stream, blockMetadata, consecutiveBlock);
+    ReaderContext.Entity entity;
+    while ((entity = readerContext.next()) != null) {
+
+      counters.getCounter(ReaderCounterKeys.BYTES).add(entity.getUsedBytes());
+
+      byte[] record = entity.getRecord();
+
+      //If the record is partial then ignore the record.
```
amberarrow: why 'partial' rather than 'null' ?

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,317 @@
...
+   * Static count of readers to read input file
+   * @param readersCount
+   */
+  public void setReadersCount(int readersCount)
+  {
+    this.readersCount = readersCount;
+  }
+
+  /**
+   * Gets is sequencial file read
```
amberarrow: sequencial => sequential (multiple places)

otterc:
@yogidevendra 
Can you please explain how the downstream operators will know which block a record belongs to?
Specifically when you converge, the information that which record belongs to which block will be lost. 

Is this being created for any specific use case? Can you please share that?

yogidevendra:
@chandnisingh 
This is mainly for the usecases where downstream operators do not care about which block the incoming record belongs to. For example, dedup, enrichment, projection, transform operator just work on POJO. They do not care about which file or which block does this tuple belong to.  

Usecases, in this context do not intend to retain the original sequence of records at sources. They can consider each tuple to be independent and just focus on how to process them at scale.  In this case, this module will be connected to "byte[] to POJO converter" (Say delimited parser) and then further to dedup or enrichment processing operator. Finally, output of dedup will be connected to some output operator.

yogidevendra:
Incorporated review comments.

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,331 @@
...
+ * The module reads data in parallel, following parameters can be configured
+ * <br/>
+ * 1. files: list of file(s)/directories to read<br/>
+ * 2. filePatternRegularExp: Files names matching given regex will be read<br/>
+ * 3. scanIntervalMillis: interval between two scans to discover new files in
+ * input directory<br/>
+ * 4. recursive: if scan recursively input directories<br/>
+ * 5. blockSize: block size used to read input blocks of file<br/>
+ * 6. readersCount: count of readers to read input file<br/>
+ * 7. sequencialFileRead: If emit file blocks in sequence?<br/>
```
amberarrow: sequencial => sequential

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,331 @@
...
+public class FSRecordReaderModule implements Module
+{
+  @NotNull
+  @Size(min = 1)
+  private String files;
+  private String filePatternRegularExp;
+  @Min(0)
+  private long scanIntervalMillis;
+  private boolean recursive = true;
+  private boolean sequencialFileRead = false;
```
amberarrow: sequencial => sequential

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,331 @@
...
+  {
+    this.readersCount = readersCount;
+  }
+
+  /**
+   * Gets is sequential file read
+   *
+   * @return sequencialFileRead
+   */
+  public boolean isSequencialFileRead()
```
amberarrow: sequencial => sequential

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  /**
+   * Sets is sequential file read
+   *
+   * @param sequencialFileRead
+   */
+  public void setSequencialFileRead(boolean sequencialFileRead)
+  {
+    this.sequencialFileRead = sequencialFileRead;
```
amberarrow: sequencial => sequential (multiple places)

amberarrow:
Still looking at this -- probably take a couple of days more.

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReader.java
```diff
@@ -0,0 +1,180 @@
...
+  {
+    return recordLength;
+  }
+
+  static {
+    /**
+     * Code for enabling BeanUtils to accept comma separated string to
+     * initialize FIELD_TYPE[]
+     */
+    class RecordReaderModeConverter extends AbstractConverter
```
amberarrow: Could you provide more detail on why this is necessary and how it is used ?

yogidevendra:
@amberarrow Removed the unnecessary code for BeanUtils convertor. But, facing some issues with travis build. Build failures seems to be non-deterministic. (Different unit tests failed at multiple attempts). 

Please let me know if you have any other suggestion regarding above code changes. I will update the PR for the same. We should keep this PR on hold till travis build is stable.

amberarrow:
@yogidevendra OK, rest looks good, though I'd like to test it with a small sample application on the cluster; do you have such a sample ? It would also be a good addition to the examples collection.

yogidevendra:
Let me know the location to add sample app. I will open separate PR for sample app in the respective repo.

amberarrow:
https://github.com/DataTorrent/examples under tutorials

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,332 @@
...
+ * This module is used for reading records/tuples from FileSystem. Records can
+ * be read in parallel using multiple partitions of record reader operator.
+ * (Ordering is not guaranteed when records are read in parallel)
+ *
+ * Input directory is scanned at specified interval to poll for new data.
+ * 
+ * The module reads data in parallel, following parameters can be configured
+ * <br/>
+ * 1. files: list of file(s)/directories to read<br/>
+ * 2. filePatternRegularExp: Files names matching given regex will be read<br/>
```
amberarrow: Files names => Files with names

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,332 @@
...
+ * 3. scanIntervalMillis: interval between two scans to discover new files in
+ * input directory<br/>
+ * 4. recursive: if scan recursively input directories<br/>
```
amberarrow: if scan recursively input directories => if true, scan input directories recursively

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,332 @@
...
+ * 5. blockSize: block size used to read input blocks of file<br/>
+ * 6. readersCount: count of readers to read input file<br/>
+ * 7. sequentialFileRead: If emit file blocks in sequence?<br/>
```
amberarrow: This description is not clear, please rewrite.

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,332 @@
...
+   * A comma separated list of directories to scan. If the path is not fully
+   * qualified the default file system is used. A fully qualified path can be
+   * provided to scan directories in other filesystems.
+   *
+   * @param files
+   *          files
+   */
+  public void setFiles(String files)
+  {
+    this.files = files;
```
amberarrow, line 156: Shouldn't the field be named "directories" if we expect it to be a list of directories ?

yogidevendra, line 156: I tried to keep it consistent with FSInputModule.java. If you feel that readability/intuitive names are more important than having consistency across operators then I am fine with changing this field name.

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,334 @@
...
+
+  /**
+   * Criteria for record split
+   */
+  private RECORD_READER_MODE mode;
+
+  /**
+   * Length for fixed width record
+   */
+  private int recordLength;
```
DT-Priyanka, line 88: We should either have default value or mark Min constraint to make sure it is set by user.

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,334 @@
...
+
+  public FileSplitterInput createFileSplitter()
```
DT-Priyanka, line 95: Please add javadoc on both create factory methods.

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,334 @@
...
+  {
+    return new FileSplitterInput();
+  }
+
+  public FSRecordReader createBlockReader()
```
DT-Priyanka: we can rename it to createRecordReader

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,334 @@
...
+      fileSplitter.getScanner().setFilePatternRegularExp(filePatternRegularExp);
+    }
+
+    recordReader.setBasePath(files);
+    if (readersCount != 0) {
+      dag.setAttribute(recordReader, Context.OperatorContext.PARTITIONER,
+          new StatelessPartitioner<FSSliceReader>(readersCount));
+    }
+    fileSplitter.setBlocksThreshold(blocksThreshold);
+    records.set(recordReader.records);
```
DT-Priyanka, line 143: can you explain what this line does?

yogidevendra, line 143: binding proxy port to the operator port.

yogidevendra:
@DT-Priyanka Incorporated changes based on your review feedback.
@amberarrow Could you please do a final round of review and then merge if there no further comments from the community?

yogidevendra:
@amberarrow Could you please merge this if there are no more comments.

yogidevendra:
@DT-Priyanka Could you please merge this if there are no more comments?

library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReaderModule.java
```diff
@@ -0,0 +1,344 @@
...
+  /**
+   * Criteria for record split
+   */
+  private RECORD_READER_MODE mode = RECORD_READER_MODE.DELIMITED_RECORD;
+
+  /**
+   * Length for fixed width record
+   */
+  @Min(1)
+  private int recordLength;
```
DT-Priyanka, line 88: Record length will be optional in case of DELIMITED_RECORD and is compulsory in case of FIXED_LENGHT records. so keeping it either optional or compulsory field both have pros and cons.

yogidevendra, line 88: Oh. Good point. 
Should we separate this into two different classes. FixedWidthRecordReader, DelimitedRecordReader? That will make configuration clean. 

@amberarrow Any thoughts?

amberarrow, line 88: Adding another class seems like overkill; how about removing the annotation and adding a check in the code to ensure that the value is positive if mode is FIXED_WIDTH_RECORD ?

yogidevendra, line 88: @amberarrow As per your suggestion, removed @ min annotation for record length.
Added validation in the code to ensure that the value is positive if FIXED_WIDTH_RECORD.
Also, added test case for validating this.

Could you have a look and merge this if it looks OK.

amberarrow:
There are a couple of minor issues:
1. The commit message is missing the JIRA (see instructions at https://apex.apache.org/contributing.html:
     "Commit messages need to reference JIRA (pull requests will be linked to ticket)"
2. Several lines (all inside comments) have whitespace at the end; please remove.

yogidevendra:
@amberarrow Updated as per the feedback.

PR #329, opened 2016-06-30
APEXMALHAR-2069:  moved creation of scanService to setup,

sanjaypujare:
@vrozov  pls review.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
@@ -80,13 +80,13 @@ public class FileSplitterInput extends AbstractFileSplitter implements InputOper
 {
   @NotNull
   private WindowDataManager windowDataManager;
-  @NotNull
-  protected final transient LinkedList<ScannedFileInfo> currentWindowRecoveryState;
+
+  protected transient LinkedList<ScannedFileInfo> currentWindowRecoveryState;
 
   @Valid
   @NotNull
   private TimeBasedDirectoryScanner scanner;
-  @NotNull
+
   private Map<String, Map<String, Long>> referenceTimes;
 
   private transient long sleepMillis;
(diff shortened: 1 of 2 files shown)


amberarrow:
Curious about what the advantage is to having it in setup versus the constructor. Since it is a transient field, it's not going to be checkpointed anyway.

Never mind, the answer is in the ticket.

vrozov:
There are more operational fields on top of scanService that can and should be moved to setup().

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -318,6 +317,9 @@ public TimeBasedDirectoryScanner()
     @Override
     public void setup(Context.OperatorContext context)
     {
+      if (scanService == null) {
```
vrozov: why is it necessary to check for null? Under what condition it will be not null? As far as I know platform calls setup/terminate once per deployment. IMO, it is safer to create new execution service each time setup is called and possibly assert or throw RuntimeException if scanService is not null to avoid misusage in unit tests.

sanjaypujare:
@vrozov  pls review

vrozov:
Please check the test failure.

sanjaypujare:
@vrozov I am looking at the failure and it's confusing. It doesn't fail in my workspace but it fails on travis. The failure is as follows:

Failed tests: 
  FileSplitterTest.testRecursive:426 window 2: files expected:<2> but was:<3>

Any suggestions about how I can troubleshoot this?

vrozov:
@sanjaypujare it may be a timing issue. Usually, travis-ci is slower compared to a development environment. Do test code review as a quick check, if you don't see anything wrong with the test, try reproducing in your environment by running the test multiple times.

sanjaypujare:
Closing after talking to vrozov (created another pull request 333 from a different branch)

PR #330, opened 2016-07-05
APEXMALHAR-2100 Implementation of Inner Join operator

chaithu14:


library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -0,0 +1,259 @@
...
+    stream1Store = new ManagedTimeStateImpl();
+    stream2Store = new ManagedTimeStateImpl();
+    stream1Store.setNumBuckets(noOfBuckets);
+    stream2Store.setNumBuckets(noOfBuckets);
+    if (bucketSpanTime != null) {
+      stream1Store.getTimeBucketAssigner().setBucketSpan(Duration.millis(bucketSpanTime));
+      stream2Store.getTimeBucketAssigner().setBucketSpan(Duration.millis(bucketSpanTime));
+    }
+
+    if (getExpiryTime() != null) {
```
tushargosavi: can it be different for each stores?

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
```diff
@@ -0,0 +1,187 @@
...
+  private boolean isStream1KeyPrimary = false;
+  private boolean isStream2KeyPrimary = false;
+
+  @NotNull
+  private String keyFieldsStr;
+  @NotNull
+  private String includeFieldStr;
+  private String timeFieldsStr;
+
+  protected void processTuple(T tuple, boolean isStream1Data)
```
tushargosavi: Instead of boolean , can you pass the store directly?

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
```diff
@@ -0,0 +1,187 @@
...
+
+  protected void joinStream(K key, T tuple, boolean isStream1Data)
+  {
+    Spillable.SpillableByteArrayListMultimap<K, T> store = isStream1Data ? stream2Data : stream1Data;
+    List<T> value = store.get(key);
+
+    // Join the input tuple with the joined tuples
+    if (value != null) {
+      for (T joinedValue : value) {
+        T result = isStream1Data ? joinTuples(Arrays.asList(tuple, joinedValue)) :
```
tushargosavi: why list as argument to joinTuples?

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -0,0 +1,259 @@
...
+
+  @Override
+  public void handleIdleTime()
+  {
+    if (waitingEvents.size() > 0) {
+      processWaitEvents();
+    } else {
+      /* nothing to do here, so sleep for a while to avoid busy loop */
+      try {
+        Thread.sleep(sleepMillis);
```
tushargosavi: is it needed, or platform adds the delay?

library/src/main/java/com/datatorrent/lib/join/managed/ManagedTimeStateMultiMap.java
```diff
@@ -0,0 +1,247 @@
...
+      if (valueSlice == null || valueSlice.length == 0 || valueSlice.buffer == null) {
+        return null;
+      }
+      value = new ArrayList<>();
+      value.add((V)streamCodec.fromByteArray(valueSlice));
+    }
+    return  value;
+  }
+
+  public Future<Slice> getAsync(@Nullable K k)
```
tushargosavi: ideally it should return Future<List<V>>. can you write a composible future to implement it.

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
chaithu14: Yes. It's possible. I added different parameters for both the streams.

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
chaithu14: No. While joining, we have to look into opposite store.

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
chaithu14: List is not needed. Will update.

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
chaithu14: Platform adds the delay.

chaithu14:
Incorporated review comments. Added unit test cases.

chaithu14:
@chandnisingh @tushargosavi @chinmaykolhatkar: Please review and merge.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * Create getters for the key and time fields and setters for the include fields.
+   */
+  protected void generateSettersAndGetters()
+  {
+    for (int i = 0; i < 2; i++) {
+      Class inputClass = inputFieldObjects[i].inputClass;
+      try {
+        Class c = ClassUtils.primitiveToWrapper(inputClass.getField(keyFields.get(i)).getType());
```
chinmaykolhatkar: keyFields is populated in setup method of the operator but this method is called from activate? activate of operator is called before setup.
Can you please check if this is giving desired result? Am I missing something?

chinmaykolhatkar: same for timeFields.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,247 @@
...
+        inputFieldObjects[i].keyGet = PojoUtils.createGetter(inputClass, keyFields.get(i), c);
+        if (timeFields != null && timeFields.size() != 0) {
+          Class tc = ClassUtils.primitiveToWrapper(inputClass.getField(timeFields.get(i)).getType());
+          inputFieldObjects[i].timeFieldGet = PojoUtils.createGetter(inputClass, timeFields.get(i), tc);
+        }
+        for (int j = 0; j < includeFields[i].length; j++) {
+          Class ic = ClassUtils.primitiveToWrapper(inputClass.getField(includeFields[i][j]).getType());
+          Class oc = ClassUtils.primitiveToWrapper(outputClass.getField(includeFields[i][j]).getType());
```
chinmaykolhatkar: Can you please add a check for whether includeFields[i] is present in both give input and output class?

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,247 @@
...
+          inputFieldObjects[i].fieldMap.put(PojoUtils.createGetter(inputClass, includeFields[i][j], ic),
```
chinmaykolhatkar: If the includeField is present in both input stream tuples and their values are different, what governs which value will be copied to output POJO?

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+    return new InnerJoinStreamCodec(getKeyFieldsStr().split(",")[1]);
+  }
+
+  private class FieldObjectMap
+  {
+    public Class<?> inputClass;
+    public PojoUtils.Getter keyGet;
+    public PojoUtils.Getter timeFieldGet;
+    public Map<PojoUtils.Getter,PojoUtils.Setter> fieldMap;
```
chinmaykolhatkar: This fieldMap seems to be outside of the scope of any specific inputClass. Can you please move this variable outside of this class and treat it as a seperate entity. Its a mapping of getter (from either stream's POJO) to setter to output POJO but there has to be a single mapping.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,247 @@
...
+ *
+ * @displayName POJO Inner Join Operator
+ * @tags join
+ */
+public class POJOInnerJoinOperator extends AbstractManagedStateInnerJoinOperator<Object,Object> implements Operator.ActivationListener<Context>
+{
+  private static final transient Logger LOG = LoggerFactory.getLogger(POJOInnerJoinOperator.class);
+  private transient long timeIncrement;
+  private transient FieldObjectMap[] inputFieldObjects = (FieldObjectMap[])Array.newInstance(FieldObjectMap.class, 2);
+  protected Class<?> outputClass;
```
chinmaykolhatkar: outputClass can be transient. Its assigned in output port's setup anyway. No need to save it.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,247 @@
...
+  private long time = System.currentTimeMillis();
```
chinmaykolhatkar: Any specific reason why time is non-transient. I believe this variable is used to extract time from the tuple. But as this is non-transient, it'll restore back from the state even after crash. Is this expected behaviour?

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -0,0 +1,265 @@
...
+      }
+    } else {
+      waitingEvents.put(new JoinEvent<>(key,tuple,isStream1Data),future);
+    }
+  }
+
+  @Override
+  public void handleIdleTime()
+  {
+    if (waitingEvents.size() > 0) {
```
chinmaykolhatkar: Can we process only one element in one handleIdleTime callback?
This is to make sure that processing inside handleIdleTime does not block windows.
handleIdleTime is called over and over again anyway.

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
chinmaykolhatkar: If platform adds delay,, can this be removed?

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -0,0 +1,265 @@
...
+   * Process the waitingEvents.
+   */
+  private void processWaitEvents()
+  {
+    Iterator<Map.Entry<JoinEvent<K,T>, Future<List>>> waitIterator = waitingEvents.entrySet().iterator();
+    while (waitIterator.hasNext()) {
+      Map.Entry<JoinEvent<K,T>, Future<List>> waitingEvent = waitIterator.next();
+      try {
+        JoinEvent<K,T> event = waitingEvent.getKey();
+        joinStream(event.value,event.isStream1Data,waitingEvent.getValue().get());
```
chinmaykolhatkar: If this Future request has not completed, the Future.get() call will hold the operator thread.
Suggesting to do following:
if (waitingEvent.getValue().isDone()) {
  // Process join
}
else {
  // To make sure idempotency is maintained join needs to happen in the same order
  return;
}

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
```diff
@@ -0,0 +1,331 @@
...
+  /**
+   * Merge the given tuple and list of values.
+   * @param tuple given tuple
+   * @param isStream1Data Specifies whether the given tuple belongs to stream1 or not.
+   * @param value list of tuples
+   */
+  protected void joinStream(T tuple, boolean isStream1Data, List<T> value)
+  {
+    // Join the input tuple with the joined tuples
+    if (value != null) {
```
chinmaykolhatkar: This condition (!= null) will take care of the fact that if opposite stream has not got matching tuple, then ignore this join call.
What happens to the tuples for which opposite stream did not receive any matching tuples within expiry time.
Should those tuples be emitted on a separate port? I understand this is the case for Outer Joins, but this is just a question in regards to data lineage. The non-matching tuples will get lost in the pipeline, IMO we should provide a way out for them.

chinmaykolhatkar:
@chaithu14 I suggest we change to includeFieldsStr, timeFields and keyFields as follows:
1. Rename includeFieldsStr to includeFields and make the type as List<String>
2. Change timeFields and keyFields to List<String>

chinmaykolhatkar:
Can you please mark all these class as Evolving?

chinmaykolhatkar:
Can you please put in javadoc comment saying whether the property is mandatory or optional and if optional what is the default value for it?

library/src/main/java/com/datatorrent/lib/join/ManagedSpillableComplexComponent.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.apex.malhar.lib.state.spillable.SpillableComplexComponent;
+import org.apache.apex.malhar.lib.utils.serde.Serde;
+
+import com.datatorrent.api.Context;
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * SpillableComplexComponent for Join Operator
+ */
+public class ManagedSpillableComplexComponent implements SpillableComplexComponent
```
chinmaykolhatkar: This is used only in AbstractManagedStateInnerJoinOperator.createStore().

In the interest of keeping the number of files less, can the content of method be directly used in createStore?
It just seem to create a new object of type ManagedTimeStateMultiValue.

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
```diff
@@ -0,0 +1,331 @@
...
+/**
+ * <p>
+ * An abstract implementation of inner join operator. Operator receives tuples from two streams,
+ * applies the join operation based on constraint and emit the joined value.
+ * Concrete classes should provide implementation to extractKey, extractTime, mergeTuples methods.
+ *
+ * <b>Properties:</b><br>
+ * <b>includeFieldStr</b>: List of comma separated fields to be added to the output tuple.
+ *                         Ex: Field1,Field2;Field3,Field4<br>
+ * <b>keyFields</b>: List of comma separated key field for both the streams. Ex: Field1,Field2<br>
```
bhupeshchawda: Does this mean Field1 is from stream1 and Field2 is from stream2? What if there are multiple key fields from each table?

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
```diff
@@ -0,0 +1,331 @@
...
+ * <b>timeFields</b>: List of comma separated time field for both the streams. Ex: Field1,Field2<br>
```
bhupeshchawda: Does this need to be of type java.util.Date?

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
```diff
@@ -0,0 +1,331 @@
...
+ * <b>expiryTime</b>: Expiry time for stored tuples<br>
```
bhupeshchawda: In seconds? ms?

bhupeshchawda: In seconds? ms?

bhupeshchawda: In seconds? ms?

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
```diff
@@ -0,0 +1,331 @@
...
+   * 3) Get the values of the key if found it in opposite store
+   * 4) Merge the given tuple and values found from step (3)
+   * @param tuple given tuple
+   * @param isStream1Data Specifies whether the given tuple belongs to stream1 or not.
+   */
+  protected void processTuple(T tuple, boolean isStream1Data)
+  {
+    Spillable.SpillableByteArrayListMultimap<K,T> store = isStream1Data ? stream1Data : stream2Data;
+    K key = extractKey(tuple,isStream1Data);
+    if (!store.put(key, tuple)) {
```
bhupeshchawda: When would this return false? Are we not allowing duplicates?

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -0,0 +1,265 @@
...
+   * Create Managed states and stores for both the streams.
+   */
+  @Override
+  public void createStores()
+  {
+    stream1Store = new ManagedTimeStateImpl();
+    stream2Store = new ManagedTimeStateImpl();
+    stream1Store.setNumBuckets(noOfBuckets);
+    stream2Store.setNumBuckets(noOfBuckets);
+    if (bucketSpanTime != null) {
```
bhupeshchawda: I will suggest making bucketSpanTime as well as the expiryTimes to be made mandatory. Reason is that if a user forgot to set one of these, there could be issues.

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
bhupeshchawda: Agree with Chinmay. In addition to that, you will need to have a flag which tells whether this is being called from handleIdleTime() or endWindow(). If it is endWindow(), you will need to call Future.get() until you have the value; else you won't be able to close the window.

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
chinmaykolhatkar: I agree with expiryTime to be mandatory but not with making bucketSpanTime mandatory. This quite advanced property and should have a default value.

Also why do we have different expiry time for separate streams?

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,247 @@
...
+  };
+
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    timeIncrement = context.getValue(Context.OperatorContext.APPLICATION_WINDOW_COUNT) *
+      context.getValue(Context.DAGContext.STREAMING_WINDOW_SIZE_MILLIS);
+    super.setup(context);
+    for (int i = 0; i < 2; i++) {
+      inputFieldObjects[i] = new FieldObjectMap();
```
bhupeshchawda: This is already done in the declaration.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,247 @@
...
+  /**
+   * Extract the time value from the given tuple
+   * @param tuple given tuple
+   * @param isStream1Data Specifies whether the given tuple belongs to stream1 or not.
+   * @return
+   */
+  @Override
+  public long extractTime(Object tuple, boolean isStream1Data)
+  {
+    return timeFields == null ? time : (long)(isStream1Data ? inputFieldObjects[0].timeFieldGet.get(tuple) :
```
bhupeshchawda: time should be System.currentTimeMillis() ?

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,247 @@
...
+   * Create getters for the key and time fields and setters for the include fields.
+   */
+  protected void generateSettersAndGetters()
+  {
+    for (int i = 0; i < 2; i++) {
+      Class inputClass = inputFieldObjects[i].inputClass;
+      try {
+        Class c = ClassUtils.primitiveToWrapper(inputClass.getField(keyFields.get(i)).getType());
+        inputFieldObjects[i].keyGet = PojoUtils.createGetter(inputClass, keyFields.get(i), c);
+        if (timeFields != null && timeFields.size() != 0) {
```
bhupeshchawda: Can you please add a check if timeFields has required number of elements. Might save a NPE.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,247 @@
...
+          Class tc = ClassUtils.primitiveToWrapper(inputClass.getField(timeFields.get(i)).getType());
```
bhupeshchawda: Please have more descriptive names for tc, ic, oc etc.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
```diff
@@ -0,0 +1,357 @@
...
+    if (streamCodec == null) {
+      streamCodec = new KryoSerializableStreamCodec();
+    }
+  }
+
+  public ManagedTimeStateMultiValue(@NotNull ManagedTimeStateImpl store, boolean isPrimaryKey)
+  {
+    this();
+    this.store = Preconditions.checkNotNull(store);
+    this.isKeyContainsMultiValue = isPrimaryKey;
```
bhupeshchawda: Intuitively, this should be `! isPrimaryKey` as only then key can contain multi values.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
```diff
@@ -0,0 +1,357 @@
...
+   */
+  @Override
+  public List<V> get(@Nullable K k)
+  {
+    List<V> value = null;
+    Slice valueSlice = store.getSync(getBucketId(k), streamCodec.toByteArray(k));
+    if (isKeyContainsMultiValue) {
+      value = (List<V>)streamCodec.fromByteArray(valueSlice);
+    }  else {
+      if (valueSlice == null || valueSlice.length == 0 || valueSlice.buffer == null) {
```
bhupeshchawda: Shouldn't this if block be outside the nested if? Perhaps, just after you populate `valueSlice` from `store.getSync()`

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
```diff
@@ -0,0 +1,357 @@
...
+  {
+    return streamCodec;
+  }
+
+  public void setStreamCodec(StreamCodec streamCodec)
+  {
+    this.streamCodec = streamCodec;
+  }
+
+  public class CompositeFuture implements Future<List>
```
bhupeshchawda, line 297: Instead of having this entire class, is it possible that you can just create a call called `getAsync()`? You can then return the slice as a `Future` and let the caller handle whether the contents of the slice is a single value or multi values.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
```diff
@@ -0,0 +1,357 @@
...
+
+    @Override
+    public List get() throws InterruptedException, ExecutionException
+    {
+      List<V> value = null;
+      Slice valueSlice = slice.get();
+      if (isKeyContainsMultiValue) {
+        value = (List<V>)streamCodec.fromByteArray(valueSlice);
+      }  else {
+        if (valueSlice == null || valueSlice.length == 0 || valueSlice.buffer == null) {
```
bhupeshchawda: Same comments as the above get() method.

chinmaykolhatkar:
@chaithu14 Any update on this PR? I think this PR can be moved to actual review mode... The title of PR still says "review only"...

chaithu14:
Thanks @chinmaykolhatkar & @bhupeshchawda for reviewing. I changed the PR title.

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
chaithu14: Yes. If there are multiple key fields from each stream then the primary key field has to specify in keyFields and the other key fields has to take care in mergeTuples().

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
chaithu14: Yes. It can be. If the timeField of the stream is not in milliseconds then the user has to override the extractTime() method and convert the field into milliseconds.

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
chaithu14: ms. Will add it into java docs.

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
chaithu14: If the tuple is already expired tuple then in this case will return false. put() will allow the duplicates.

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
chaithu14: If the tuples did not receive any matching tuples within expiry time then those will remove from store after expiry. 
No. we are not emitting on a separate port.
This is not straight forward because the tuples are storing as byte[]. I will add this feature when we will work on outer join support.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: No. In declaration, it creates the array with Map type.

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
chaithu14: Ok. Will make expiryTime as mandatory. 
I thought of different use case and this is invalid in streaming scenario. 
I will make single expiry time for both the streams.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
chaithu14, line 297: To make it consistent between the get() and getAsync(), created the compositeFuture.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: Yes. Time is initialized to  System.currentTimeMillis()  in declaration and incremented at every endWindow().

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
chaithu14: If the tuples did not receive any matching tuples within expiry time then those will remove from store after expiry. 
No. we are not emitting on a separate port.
This is not straight forward because the tuples are storing as byte[]. I will add this feature when we will work on outer join support.

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
chaithu14: Done.

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
chaithu14: Done

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
chaithu14: Done

library/src/main/java/com/datatorrent/lib/join/ManagedSpillableComplexComponent.java
chaithu14: Done

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: Yes. Made it into transient.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: Yes. It will restore back from the state when it crash and come back.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: No. activate of operator is called after operator setup.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: Done

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: Done

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: There is no api to check whether the field is present in Class. 
If the includeField not exists then it will return an exception. Is this not sufficient?

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: In this case, it copies the stream2 value.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: If it has a single mapping then it's difficult to check whether the getter is relates to which of the stream.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
chaithu14: Done

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
chaithu14: Yes. Done.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
chaithu14: Done

chaithu14:
Incorporated the review comments and updated the branch.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,248 @@
...
+    for (int i = 0; i < 2; i++) {
+      inputFieldObjects[i] = new FieldObjectMap();
+    }
+  }
+
+  /**
+   * Extract the time value from the given tuple
+   * @param tuple given tuple
+   * @param isStream1Data Specifies whether the given tuple belongs to stream1 or not.
+   * @return
```
chinmaykolhatkar: Description for what is returned.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,248 @@
...
+  public long extractTime(Object tuple, boolean isStream1Data)
+  {
+    return timeFields == null ? time : (long)(isStream1Data ? inputFieldObjects[0].timeFieldGet.get(tuple) :
+          inputFieldObjects[1].timeFieldGet.get(tuple));
+  }
+
+  /**
+   * Create getters for the key and time fields and setters for the include fields.
+   */
+  protected void generateSettersAndGetters()
```
chinmaykolhatkar: This can be made private.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,248 @@
...
+        throw new RuntimeException(e);
+      }
+    }
+  }
+
+  /**
+   * Extract the key value from the given tuple
+   * @param tuple given tuple
+   * @param isStream1Data Specifies whether the given tuple belongs to stream1 or not.
+   * @return
```
chinmaykolhatkar: description missing for return.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,248 @@
...
+  {
+    return isStream1Data ? inputFieldObjects[0].keyGet.get(tuple) :
+      inputFieldObjects[1].keyGet.get(tuple);
+  }
+
+  /**
+   * Merge the given tuples
+   * @param tuple1 tuple belongs to stream1
+   * @param tuple2 tuple belongs to stream1
+   * @return
```
chinmaykolhatkar: return description missing.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,248 @@
...
+    super.endWindow();
+    time += timeIncrement;
+  }
+
+  /**
+   * Returns the streamcodec for the streams
+   * @param isStream1data Specifies whether the codec needs for stream1 or stream2.
+   * @return the object of InnerJoinStreamCodec
+   */
+  public StreamCodec<Object> getInnerJoinStreamCodec(boolean isStream1data)
```
chinmaykolhatkar: This can be made protected. Public is not required.

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -0,0 +1,354 @@
...
+ * An abstract implementation of inner join operator over Managed state which extends from
+ * AbstractInnerJoinOperator.
+ *
+ * <b>Properties:</b><br>
+ * <b>noOfBuckets</b>: Number of buckets required for Managed state. <br>
+ * <b>bucketSpanTime</b>: Indicates the length of the time bucket. <br>
+ */
+@org.apache.hadoop.classification.InterfaceStability.Evolving
+public abstract class AbstractManagedStateInnerJoinOperator<K,T> extends AbstractInnerJoinOperator<K,T> implements
+    Operator.CheckpointNotificationListener, Operator.CheckpointListener,Operator.IdleTimeHandler
```
chinmaykolhatkar: CheckpointListener is deprecated, please use CheckpointNotificationListener instead.

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -0,0 +1,354 @@
...
+{
+  private static final transient Logger LOG = LoggerFactory.getLogger(AbstractManagedStateInnerJoinOperator.class);
```
chinmaykolhatkar: If LOG is not used, please remove it.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,248 @@
...
+/**
+ * Concrete implementation of AbstractManagedStateInnerJoinOperator and receives objects from both streams.
+ *
+ * @displayName POJO Inner Join Operator
+ * @tags join
+ */
+@org.apache.hadoop.classification.InterfaceStability.Evolving
+public class POJOInnerJoinOperator extends AbstractManagedStateInnerJoinOperator<Object,Object> implements Operator.ActivationListener<Context>
+{
+  private static final transient Logger LOG = LoggerFactory.getLogger(POJOInnerJoinOperator.class);
```
chinmaykolhatkar: LOG seems to be unsed, please remove it.

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -0,0 +1,354 @@
...
+
+    public JoinEvent(K key, T value, boolean isStream1Data)
+    {
+      this.key = key;
+      this.value = value;
+      this.isStream1Data = isStream1Data;
+    }
+  }
+
+  public class ManagedSpillableComplexComponent implements SpillableComplexComponent
```
chinmaykolhatkar: Do you need this complete class to be implemented? 
Can you not do:
new ManagedTimeStateMultiValue<K,V>(store, isPrimaryKey);
new ManagedTimeStateMultiValue<K,V>();

Wherever required?

library/src/main/java/com/datatorrent/lib/join/InnerJoinStreamCodec.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package com.datatorrent.lib.join;
+
+import com.datatorrent.lib.codec.KryoSerializableStreamCodec;
+import com.datatorrent.lib.util.PojoUtils;
+
+/**
+ * Stream codec based on keyExpression for POJO Inner Join Operator.
+ *
+ * @Since 3.5.0
```
chinmaykolhatkar: Please remove @since tag.. This will be added during release process.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
```diff
@@ -0,0 +1,363 @@
...
+  /**
+   * Insert (keySlice,valueSlice) into the store using bucketId and timeBucket.
+   * @param bucketId bucket Id
+   * @param timeBucket time bucket
+   * @param keySlice key slice
+   * @param valueSlice value slice
+   * @return true if the given (keySlice,valueSlice) is successfully inserted into the
+   *         store otherwise false.
+   */
+  public boolean insertInStore(long bucketId, long timeBucket, Slice keySlice, Slice valueSlice)
```
chinmaykolhatkar: This is not accessed anywhere else. Please make this private.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
```diff
@@ -0,0 +1,363 @@
...
+    throw new UnsupportedOperationException();
+  }
+
+  public int getBucketId(K k)
+  {
+    return k.hashCode() % store.getNumBuckets();
+  }
+
+
+  public boolean isKeyContainsMultiValue()
```
chinmaykolhatkar: isKeyContainsMultiValue  &  setKeyContainsMultiValue are unused. Please remove it.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
```diff
@@ -0,0 +1,363 @@
...
+
+    @Override
+    public boolean isDone()
+    {
+      return slice.isDone();
+    }
+
+    /**
+     * Converts the single element into the list.
+     * @return
```
chinmaykolhatkar: Missing description for return.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chinmaykolhatkar: Apologies... Yes you're right.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chinmaykolhatkar: Yes... that's sufficient.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chinmaykolhatkar: I'm not sure if that's the best behaviour... Because its really uncertain that what stream should overwrite on what. Any other approach would be possible for this?

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chinmaykolhatkar: Why do you want to relate it? Shoudn't there be a one-to-one mapping of output POJO fields to input POJO fields from either one of the stream?

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -0,0 +1,354 @@
...
+      if (future.isDone() || finalize) {
+        try {
+          JoinEvent<K,T> event = waitingEvent.getKey();
+          joinStream(event.value,event.isStream1Data,future.get());
+        } catch (InterruptedException | ExecutionException e) {
+          throw new RuntimeException("end window", e);
+        }
+        waitIterator.remove();
+        if (!finalize) {
+          break;
```
chinmaykolhatkar: You can break only when you find a future which is not done... Till then you can keep processing for all the waitingEvents which has done future.

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
chinmaykolhatkar: Agreed. I suggest we make this as assumption for this version of Inner Join. Later on when we have Outer join, we can make inner join as a special case of outer join which is when we can take care of this. Lets create a Jira for addition of this functionality in inner join when this PR gets merged.

Let me know if you think otherwise.

chinmaykolhatkar:
Can you move the join related files i.e. Abstract Operators, InnerJoinStreamCodec & POJOInnerJoinOperator to org.apache.apex.malhar.lib.join package?

chinmaykolhatkar:
Can you add following tests:
1. stream1:tuple1, next comes stream2:tuple2 and perform join
2. stream1:tuple1, next comes stream1:tuple2 which has same key as previous, next comes stream2:tuple1 and then perform join
3. stream1:tuple1, next comes stream2:tuple2, next comes stream2:tuple3... perform join.. Should emit 2 joined tuples as a result.
4. stream1 has 2 tuples which has same key... stream 2 has 2 tuples having same.. perform join with arrival orders in various combinations...

chinmaykolhatkar:
Can you add test for streamcodec that correct tuple is going to correct partition?

chinmaykolhatkar:
Can you make keyExpression similar to how its done in Deduper instead of specifying keys? This way composite keys can also be mentioned as a parameter.

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
```diff
@@ -0,0 +1,248 @@
...
+        Class c = ClassUtils.primitiveToWrapper(inputClass.getField(keyFields.get(i)).getType());
+        inputFieldObjects[i].keyGet = PojoUtils.createGetter(inputClass, keyFields.get(i), c);
+        if (timeFields != null && timeFields.size() == 2) {
+          Class timeField = ClassUtils.primitiveToWrapper(inputClass.getField(timeFields.get(i)).getType());
+          inputFieldObjects[i].timeFieldGet = PojoUtils.createGetter(inputClass, timeFields.get(i), timeField);
+        }
+        for (int j = 0; j < includeFields[i].length; j++) {
+          Class inputField = ClassUtils.primitiveToWrapper(inputClass.getField(includeFields[i][j]).getType());
+          Class outputField = ClassUtils.primitiveToWrapper(outputClass.getField(includeFields[i][j]).getType());
+          inputFieldObjects[i].fieldMap.put(PojoUtils.createGetter(inputClass, includeFields[i][j], inputField),
```
chinmaykolhatkar: Can you check whether types are also same for inputField and outputField, otherwise, they're not matching really.

library/src/main/java/com/datatorrent/lib/join/AbstractInnerJoinOperator.java
chaithu14: Ok

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
chaithu14: Done

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
chaithu14: Ok

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
chaithu14: Done

library/src/main/java/com/datatorrent/lib/join/AbstractManagedStateInnerJoinOperator.java
chaithu14: Yes. Initialized without creating component.

library/src/main/java/com/datatorrent/lib/join/InnerJoinStreamCodec.java
chaithu14: Done

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: Done

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: Done

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: Done

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: Done

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: Done

library/src/main/java/com/datatorrent/lib/join/POJOInnerJoinOperator.java
chaithu14: In MergeTuples(), it's difficult to figure out that the getter method belongs to stream1 or stream2.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
chaithu14: Yes. I made this to private.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
chaithu14: Done.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
chaithu14: Done

chaithu14:
Incorporated review comments and added unit tests as suggested.

chinmaykolhatkar:
@chaithu14 The InnerJoin Tests are failing. Can you please take a look? If this is passing in your local environment but not here then there is a possibility of timing issue. We need to have unit tests which deterministically test the underlying code.

chaithu14:
@chinmaykolhatkar: I fixed the issue. Please merge.

chinmaykolhatkar:
@chaithu14 You've added time delay between 2 tuples for calling process method which may not be the case. Calling process for 2 ports right after each other is a valid use case and synchronization for joining should happen as a part of the code. As discussed there seems to be a race condition while processing 2 tuples from 2 different stream arriving in quick succession. Lets try to fix that race condition.

siyuanh:
@chaithu14  Does this depend on WindowedOperator?

chaithu14:
@siyuanh This issue is not depend on WindowedOperator.

chaithu14:
Fixed the race condition issue and updated PR.

chinmaykolhatkar:
@chaithu14 Please check the travis build failure. Seems that some exception from POJOInnerJoinTest is being thrown again and again. Hence logs are exceeding size of 4k.

chaithu14:
@chinmaykolhatkar Those exceptions are not from POJOInnerJoinTest. I closed and open a PR then the travis build is passed.

library/src/main/java/org/apache/apex/malhar/lib/join/InnerJoinStreamCodec.java
```diff
@@ -0,0 +1,46 @@
...
+package org.apache.apex.malhar.lib.join;
+
+import com.datatorrent.lib.codec.KryoSerializableStreamCodec;
+import com.datatorrent.lib.util.PojoUtils;
+
+/**
+ * Stream codec based on keyExpression for POJO Inner Join Operator.
+ */
+@org.apache.hadoop.classification.InterfaceStability.Evolving
+public class InnerJoinStreamCodec extends KryoSerializableStreamCodec<Object>
```
bhupeshchawda: Anything specific to Inner join? Can we call it JoinStreamCodec instead?

library/src/test/java/org/apache/apex/malhar/lib/join/POJOInnerJoinOperatorTest.java
```diff
@@ -0,0 +1,378 @@
...
+
+    CountDownLatch latch = new CountDownLatch(1);
+
+    Order order = new Order(102, 1, 300);
+    oper.input2.process(order);
+
+    Order order2 = new Order(103, 3, 300);
+    oper.input2.process(order2);
+
+    latch.await(500, TimeUnit.MILLISECONDS);
```
bhupeshchawda: This seems to be equivalent to `sleep(500)`
Same with the other uses of `latch.await` in the tests

library/src/main/java/org/apache/apex/malhar/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -0,0 +1,255 @@
...
+    long timeBucket = extractTime(tuple,isStream1Data);
+    if (!((ManagedTimeStateMultiValue)store).put(key, tuple,timeBucket)) {
+      return;
+    }
+    Spillable.SpillableByteArrayListMultimap<K, T> valuestore = isStream1Data ? stream2Data : stream1Data;
+    Future<List> future = ((ManagedTimeStateMultiValue)valuestore).getAsync(key);
+    if (future.isDone()) {
+      try {
+        joinStream(tuple,isStream1Data, future.get());
+      } catch (InterruptedException e) {
```
bhupeshchawda: `catch (InterruptedException | ExecutionException e)`?

library/src/main/java/org/apache/apex/malhar/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -0,0 +1,255 @@
...
+   * @param isStream1Data Specifies whether the given tuple belongs to stream1 or not.
+   */
+  @Override
+  protected void processTuple(T tuple, boolean isStream1Data)
+  {
+    Spillable.SpillableByteArrayListMultimap<K,T> store = isStream1Data ? stream1Data : stream2Data;
+    K key = extractKey(tuple,isStream1Data);
+    long timeBucket = extractTime(tuple,isStream1Data);
+    if (!((ManagedTimeStateMultiValue)store).put(key, tuple,timeBucket)) {
+      return;
```
bhupeshchawda, line 98: Do we want to throw these tuples to some port? `expired` perhaps?

library/src/main/java/org/apache/apex/malhar/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -0,0 +1,255 @@
...
+   * @param finalize finalize Whether or not to wait for future to return
+   */
+  private void processWaitEvents(boolean finalize)
+  {
+    Iterator<Map.Entry<JoinEvent<K,T>, Future<List>>> waitIterator = waitingEvents.entrySet().iterator();
+    while (waitIterator.hasNext()) {
+      Map.Entry<JoinEvent<K,T>, Future<List>> waitingEvent = waitIterator.next();
+      Future<List> future = waitingEvent.getValue();
+      if (!finalize) {
+        break;
```
bhupeshchawda: `if (! finalize)` Check if first future is done. If yes then process it and break. If not, then break.
`if (finalize)` Then process everything irrespective of future is done or not.
In this case this is breaking out of the loop without doing anything...

library/src/main/java/org/apache/apex/malhar/lib/join/AbstractManagedStateInnerJoinOperator.java
chaithu14: No. expired port will conflict with the stored tuples which are expired.

library/src/main/java/org/apache/apex/malhar/lib/join/AbstractManagedStateInnerJoinOperator.java
chaithu14: Done.

library/src/test/java/org/apache/apex/malhar/lib/join/POJOInnerJoinOperatorTest.java
chaithu14: Yes.

library/src/main/java/org/apache/apex/malhar/lib/join/InnerJoinStreamCodec.java
chaithu14: Done.

library/src/test/java/org/apache/apex/malhar/lib/join/POJOInnerJoinOperatorTest.java
chinmaykolhatkar: There are other sleeps also which I don't think are required.

chaithu14: Yes. These are no required. I removed sleep() and updated.

chinmaykolhatkar:
Looks good to me. I'll merge this in a day if there are no further comments.

tweise:
@chinmaykolhatkar what review regarding performance and scalability was done? We need to pay attention to those aspects. Also, is there any work underway for an example app?

chinmaykolhatkar:
@tweise Performance testing needs to happen on this operator and that is because @chaithu14 has an improvise version of Join (inner+outer) impl in mind. This check-in is to bridge the gap between unavailability of join operator. The improved operator that chaitanya writes will go through complete performance testing and stability testing.

Yes chaitanya is working on example app. @chaithu14 , it would be great to have example app soon.

tweise:
Thanks, can you point me to the JIRA(s) where this was captured?

chinmaykolhatkar:
Good point. Just created the 2 APEXMALHAR jiras for chaitanya capturing above points:
APEXMALHAR-2209 & APEXMALHAR-2210.

tweise:
Thanks. In the future such discussion should happen before merging the PR and in a way that is visible to the community.

chinmaykolhatkar:
Sure. Will keep that in mind. Thanks for the direction Thomas. :+1:

PR #333, opened 2016-07-06
Fix for APEXMALHAR-2069: moved creation of scanService to setup, @vro…

sanjaypujare:
```
    : throw a RuntimeException if scanService is already set, fix unit tests to call setup only on WindowDataManager
    : move the initialization of most data members from the constructor to setup for the TimeBasedDirectoryScanner class
    : move the initialization of most data members from the ctor to setup for FileSplitterInput class
```

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
@@ -80,13 +80,13 @@ public class FileSplitterInput extends AbstractFileSplitter implements InputOper
 {
   @NotNull
   private WindowDataManager windowDataManager;
-  @NotNull
-  protected final transient LinkedList<ScannedFileInfo> currentWindowRecoveryState;
+
+  protected transient LinkedList<ScannedFileInfo> currentWindowRecoveryState;
 
   @Valid
   @NotNull
   private TimeBasedDirectoryScanner scanner;
-  @NotNull
+
   private Map<String, Map<String, Long>> referenceTimes;
 
   private transient long sleepMillis;
@@ -94,15 +94,17 @@ public class FileSplitterInput extends AbstractFileSplitter implements InputOper
   public FileSplitterInput()
   {
     super();
-    currentWindowRecoveryState = Lists.newLinkedList();
     windowDataManager = new WindowDataManager.NoopWindowDataManager();
-    referenceTimes = Maps.newHashMap();
     scanner = new TimeBasedDirectoryScanner();
   }
 
   @Override
   public void setup(Context.OperatorContext context)
   {
+    currentWindowRecoveryState = Lists.newLinkedList();
+    if (referenceTimes == null) {
+      referenceTimes = Maps.newHashMap();
+    }
     sleepMillis = context.getValue(Context.OperatorContext.SPIN_MILLIS);
     scanner.setup(context);
     windowDataManager.setup(context);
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -80,29 +80,29 @@
...
     windowDataManager = new WindowDataManager.NoopWindowDataManager();
-    referenceTimes = Maps.newHashMap();
     scanner = new TimeBasedDirectoryScanner();
   }
 
   @Override
   public void setup(Context.OperatorContext context)
   {
+    currentWindowRecoveryState = Lists.newLinkedList();
+    referenceTimes = Maps.newHashMap();
```
vrozov: `referenceTimes` is not marked as transient, so it will be stored during checkpointing and restored back from a checkpoint in case of failover. It is necessary to check referenceTimes for `null`.

vrozov:
@sanjaypujare when squashing commits, please reword commit message. The request to review and merge belongs to pull request and not the commit.

library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterInputTest.java
```diff
@@ -499,7 +499,7 @@ public void testRecoveryOfBlockMetadataIterator() throws InterruptedException
     testMeta.fileSplitterInput.getScanner().setScanIntervalMillis(500);
 
 
-    testMeta.fileSplitterInput.setup(testMeta.context);
```
otterc, line 502: The goal of writing tests is to test the operator not just setup of window data manager. 
For a trivial improvement, that is, to avoid instantiating objects in the logical instance, this change now doesn't execute setup of the operator which is a much bigger problem. 
In future if something is added to splitter setup, the test will never execute it and result in bugs.

Also, there are multiple of operators in library which instantiate field that are transient when the operator is constructed. Are you going to change all of them?

sanjaypujare, line 502: For the first point, we can (should) add another unit test just to test FileSplitterInput.setup() which in any case sounds like a better idea.

The 2nd point is not very clear. Are you saying there are many other instances of non-transient fields which should be checked for null in setup() or are you saying there are many fields which should ideally be declared transient?

otterc, line 502: It's not a point.  It's a question.  I am asking if similar changes are going to be made throughout the codebase?  
To me this change is completely  unnecessary. IMO code which was clean and intuitive is made the opposite just because a handful of unused objects will be present in stram.

vrozov, line 502: @chandnisingh
The tests are not supposed to call setup() multiple times for the same operator instance. According to the Java doc of the `Component.setup()` callback to setup() "is made exactly once during the operator lifetime." If a test relies on setup() to clean or reinitialize an operator state, such test should be modified.

It will be good to fix all operators that are frequently used. Please file new JIRA for such operators.

The code readability did not change and it helps with resource and memory utilization. There is no point to create objects that are not used and that can be safely created when a physical operator is deployed into a container.

otterc:
@vrozov 
An individual operator test is supposed to validate an operator's functionality so these tests try to mock that cycle.  Each test  manually triggers each callback of a operator.  Some of these tests validate recovery that is why the same test calls the setup again. Here the operator instance was resused between tests.  
IMO there isn't any problem with the way these tests are  written. Also I think making such changes is unnecessary.

vrozov:
Tests must mock operator lifecycle correctly. Platform guarantees that setup() will be called only once and in general it should follow setup()/teardown() pattern. Calling setup() multiple times without restoring an operator from a checkpoint does not validate operator's recovery functionality. It is simply a shortcut for operator reinitialization.

library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterInputTest.java
otterc, line 502: @vrozov 
Please take a look at the tests.  They are validating what they are supposed to. 
Well IMO cleaner and intuitive code is much more important.

vrozov, line 502: @chandnisingh I already looked at the test during PRs #329 and #333 review. The modified code is clean and readable as well as the original code. The only problem in the original operator code was a wrong time of resource allocation/initialization that is now fixed.

PR #334, opened 2016-07-07
APEXMALHAR-2136 1) Fixed the null pointer exception issue. 2) Added getter and setter for IncrementalCheckpointManager

chaithu14:


chaithu14:
@chandnisingh Please review and merge.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -203,7 +203,12 @@ public void setup(OperatorContext context)
       //delete all the wal files with windows > activationWindow.
       //All the wal files with windows <= activationWindow are loaded and kept separately as recovered data.
       try {
-        for (long recoveredWindow : checkpointManager.getWindowIds(operatorContext.getId())) {
+        long[] recoveredWindows = checkpointManager.getWindowIds(operatorContext.getId());
+        if (recoveredWindows == null) {
+          readerService = Executors.newFixedThreadPool(numReaders, new NameableThreadFactory("managedStateReaders"));
```
otterc: This is already being done at the end. why do you need to duplicate the stmt here?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -203,7 +203,12 @@ public void setup(OperatorContext context)
...
+          return;
+        }
+        for (long recoveredWindow : recoveredWindows) {
```
otterc: You can put this block within `if (recoveredWindows == null)`

chaithu14: Done

library/src/main/java/org/apache/apex/malhar/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -536,6 +538,16 @@ public void setDurationPreventingFreeingSpace(Duration durationPreventingFreeing
     this.durationPreventingFreeingSpace = durationPreventingFreeingSpace;
   }
 
+  public IncrementalCheckpointManager getCheckpointManager()
+  {
+    return checkpointManager;
+  }
+
+  public void setCheckpointManager(IncrementalCheckpointManager checkpointManager)
```
otterc: `@NotNull` to checkpointManager and please use Preconditions.checkNotNull(...)

chaithu14: Updated as per the comment.

PR #335, opened 2016-07-07
APEXMALHAR-1701 Deduper with Managed state

bhupeshchawda:
A Deduper based on managed state.
Added Pojo implementation. Added unit tests

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
@@ -20,116 +20,309 @@ package org.apache.apex.malhar.lib.dedup;
 
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 
 import javax.validation.constraints.NotNull;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import org.apache.apex.malhar.lib.state.BucketedState;
 import org.apache.apex.malhar.lib.state.managed.ManagedTimeUnifiedStateImpl;
+import org.apache.apex.malhar.lib.state.managed.TimeBucketAssigner;
+import org.apache.hadoop.classification.InterfaceStability.Evolving;
 
+import com.google.common.base.Preconditions;
 import com.google.common.collect.Maps;
+import com.google.common.util.concurrent.Futures;
 
+import com.datatorrent.api.AutoMetric;
 import com.datatorrent.api.Context;
+import com.datatorrent.api.Context.OperatorContext;
+import com.datatorrent.api.DAG;
 import com.datatorrent.api.DefaultInputPort;
 import com.datatorrent.api.DefaultOutputPort;
 import com.datatorrent.api.Operator;
+import com.datatorrent.api.Operator.ActivationListener;
 import com.datatorrent.api.annotation.OperatorAnnotation;
+import com.datatorrent.lib.fileaccess.FileAccessFSImpl;
+import com.datatorrent.lib.fileaccess.TFileImpl;
 import com.datatorrent.netlet.util.Slice;
 
 /**
- * An operator that de-dupes a stream.
+ * Abstract class which allows de-duplicating incoming tuples based on a configured key.
+ * Also supports expiry mechanism based on a configurable expiry period configured using {@link TimeBucketAssigner}
+ * in {@link ManagedTimeUnifiedStateImpl}
+ * Following steps are used in identifying the state of a particular tuple:
+ * 1. Check if the time of the tuple falls in an expired bucket. If so, the tuple is expired
+ * 2. If the tuple is a valid event, it is checked in the store whether the same key already exists in the
+ * time bucket identified by the event time. If, so, the tuple is a duplicate.
+ * 3. Otherwise the tuple is a unique tuple.
  *
  * @param <T> type of events
  */
+@Evolving
 @OperatorAnnotation(checkpointableWithinAppWindow = false)
-public abstract class AbstractDeduper<T> implements Operator, Operator.CheckpointNotificationListener,
-    Operator.IdleTimeHandler
+public abstract class AbstractDeduper<T>
+    implements Operator, Operator.IdleTimeHandler, ActivationListener<Context>, Operator.CheckpointNotificationListener
 {
-  @NotNull
-  protected final ManagedTimeUnifiedStateImpl managedState = new ManagedTimeUnifiedStateImpl();
-
-  private transient long sleepMillis;
-
-  private transient Map<T, Future<Slice>> waitingEvents = Maps.newLinkedHashMap();
-
-  public final transient DefaultOutputPort<T> output = new DefaultOutputPort<>();
-
-  public final transient DefaultOutputPort<T> duplicates = new DefaultOutputPort<>();
-
+  /**
+   * The input port on which events are received.
+   */
   public final transient DefaultInputPort<T> input = new DefaultInputPort<T>()
   {
     @Override
-    public void process(T tuple)
+    public final void process(T tuple)
     {
-      long time = getTime(tuple);
-      Future<Slice> valFuture = managedState.getAsync(time, getKey(tuple));
-      if (valFuture.isDone()) {
-        try {
-          processEvent(tuple, valFuture.get());
-        } catch (InterruptedException | ExecutionException e) {
-          throw new RuntimeException("process", e);
-        }
-      } else {
-        waitingEvents.put(tuple, valFuture);
-      }
+      processTuple(tuple);
     }
   };
 
-  protected abstract long getTime(T tuple);
+  /**
+   * The output port on which deduped events are emitted.
+   */
+  public final transient DefaultOutputPort<T> unique = new DefaultOutputPort<>();
 
-  protected abstract Slice getKey(T tuple);
+  /**
+   * The output port on which duplicate events are emitted.
+   */
+  public final transient DefaultOutputPort<T> duplicate = new DefaultOutputPort<>();
+
+  /**
+   * The output port on which expired events are emitted.
+   */
+  public final transient DefaultOutputPort<T> expired = new DefaultOutputPort<>();
+
+  /**
+   * Whether or not the order of tuples be maintained.
+   * Making this "true" might entail some cost in performance, but makes the operator idempotent.
+   */
+  private boolean preserveTupleOrder = true;
+
+  @NotNull
+  protected final ManagedTimeUnifiedStateImpl managedState = new ManagedTimeUnifiedStateImpl();
+
+  /**
+   * Map to hold the result of a tuple processing (unique, duplicate, expired or error) until previous
+   * tuples get processed. This is used only when {@link #preserveTupleOrder} is true.
+   */
+  private transient Map<T, Decision> decisions;
+  private transient Map<T, Future<Slice>> waitingEvents = Maps.newLinkedHashMap();
+  private transient Map<Slice, Long> asyncEvents = Maps.newLinkedHashMap();
+
+  // Metrics
+  @AutoMetric
+  private transient long uniqueEvents;
+  @AutoMetric
+  private transient long duplicateEvents;
+  @AutoMetric
+  private transient long expiredEvents;
 
   @Override
-  public void setup(Context.OperatorContext context)
+  public void setup(OperatorContext context)
   {
-    sleepMillis = context.getValue(Context.OperatorContext.SPIN_MILLIS);
+    FileAccessFSImpl fAccessImpl = new TFileImpl.DTFileImpl();
+    fAccessImpl.setBasePath(context.getValue(DAG.APPLICATION_PATH) + "/bucket_data");
+    managedState.setFileAccess(fAccessImpl);
     managedState.setup(context);
+
+    if (preserveTupleOrder) {
+      decisions = Maps.newLinkedHashMap();
+    }
   }
 
   @Override
-  public void beginWindow(long windowId)
+  public void beginWindow(long l)
   {
-    managedState.beginWindow(windowId);
+    // Reset Metrics
+    uniqueEvents = 0;
+    duplicateEvents = 0;
+    expiredEvents = 0;
+
+    managedState.beginWindow(l);
   }
 
+  protected abstract Slice getKey(T event);
+
+  protected abstract long getTime(T event);
+
+  /**
+   * Processes an incoming tuple
+   *
+   * @param tuple the incoming tuple
+   */
+  protected void processTuple(T tuple)
+  {
+
+    long time = getTime(tuple);
+    Future<Slice> valFuture = managedState.getAsync(time, getKey(tuple));
+
+    if (valFuture.isDone()) {
+      try {
+        processEvent(tuple, valFuture.get());
+      } catch (InterruptedException | ExecutionException e) {
+        throw new RuntimeException(e);
+      }
+    } else {
+      processWaitingEvent(tuple, valFuture);
+    }
+  }
+
+  /**
+   * Processes a looked-up event
+   *
+   * @param tuple the incoming tuple
+   * @param value the looked up key of the tuple
+   */
+  protected void processEvent(T tuple, Slice value)
+  {
+    if (value == BucketedState.EXPIRED) {
+      processInvalid(tuple);
+      return;
+    }
+    processValid(tuple, value);
+  }
+
+  /**
+   * Processes a tuple which is waiting for the lookup to return.
+   *
+   * @param tuple The tuple which needs to wait
+   * @param future The future object which will ultimately return the lookup result
+   */
+  protected void processWaitingEvent(T tuple, Future<Slice> future)
+  {
+    waitingEvents.put(tuple, future);
+    if (preserveTupleOrder) {
+      recordDecision(tuple, Decision.UNKNOWN);
+    }
+  }
+
+  /**
+   * Processes a valid (non-expired) tuple. This tuple may be a unique or a duplicate.
+   *
+   * @param tuple
+   *          The tuple to be processed
+   * @param value
+   *          Looked up key of the tuple
+   */
+  protected void processValid(T tuple, Slice value)
+  {
+    if (!preserveTupleOrder || waitingEvents.isEmpty()) {
+      if (value == null) {
+        managedState.put(getTime(tuple), getKey(tuple), new Slice(new byte[0]));
+        processUnique(tuple);
+      } else {
+        processDuplicate(tuple);
+      }
+    } else {
+      processWaitingEvent(tuple, Futures.immediateFuture(value));
+    }
+  }
+
+  /**
+   * Processes invalid tuples.
+   *
+   * @param tuple the incoming tuple
+   */
+  protected void processInvalid(T tuple)
+  {
+    if (preserveTupleOrder && !decisions.isEmpty()) {
+      recordDecision(tuple, Decision.EXPIRED);
+    } else {
+      processExpired(tuple);
+    }
+  }
+
+  /**
+   * Processes an expired tuple
+   *
+   * @param tuple the incoming tuple
+   */
+  protected void processExpired(T tuple)
+  {
+    expiredEvents++;
+    emitExpired(tuple);
+  }
+
+  /**
+   * Processes the duplicate tuple.
+   *
+   * @param tuple
+   *          The tuple which is a duplicate
+   */
+  protected void processDuplicate(T tuple)
+  {
+    if (preserveTupleOrder && !decisions.isEmpty()) {
+      recordDecision(tuple, Decision.DUPLICATE);
+    } else {
+      duplicateEvents++;
+      emitDuplicate(tuple);
+    }
+  }
+
+  /**
+   * Processes the unique tuple.
+   *
+   * @param tuple
+   *          The tuple which is a unique
+   */
+  protected void processUnique(T tuple)
+  {
+    if (preserveTupleOrder && !decisions.isEmpty()) {
+      recordDecision(tuple, Decision.UNIQUE);
+    } else {
+      uniqueEvents++;
+      emitUnique(tuple);
+    }
+  }
+
+  /**
+   * {@inheritDoc}
+   */
   @Override
   public void handleIdleTime()
+  {
+    if (preserveTupleOrder) {
+      emitProcessedTuples();
+    }
+    processAuxiliary(false);
+  }
+
+  /**
+   * Does any auxiliary processing in the idle time of the operator.
+   * Processes any tuples which are waiting for the lookup to return.
+   *
+   * @param finalize Whether or not to wait for future to return
+   */
+  protected void processAuxiliary(boolean finalize)
   {
     if (waitingEvents.size() > 0) {
       Iterator<Map.Entry<T, Future<Slice>>> waitIterator = waitingEvents.entrySet().iterator();
       while (waitIterator.hasNext()) {
         Map.Entry<T, Future<Slice>> waitingEvent = waitIterator.next();
-        if (waitingEvent.getValue().isDone()) {
+        T tuple = waitingEvent.getKey();
+        Slice tupleKey = getKey(tuple);
+        long tupleTime = getTime(tuple);
+        Future<Slice> future = waitingEvent.getValue();
+        if (future.isDone() || finalize ) {
           try {
-            processEvent(waitingEvent.getKey(), waitingEvent.getValue().get());
+            if (future.get() == null && asyncEvents.get(tupleKey) == null) {
+              managedState.put(tupleTime, tupleKey, new Slice(new byte[0]));
+              asyncEvents.put(tupleKey, tupleTime);
+              processUnique(tuple);
+            } else {
+              processDuplicate(tuple);
+            }
           } catch (InterruptedException | ExecutionException e) {
             throw new RuntimeException("handle idle time", e);
           }
           waitIterator.remove();
         }
-      }
-    } else {
-      /* nothing to do here, so sleep for a while to avoid busy loop */
-      try {
-        Thread.sleep(sleepMillis);
-      } catch (InterruptedException ie) {
-        throw new RuntimeException(ie);
-      }
-    }
-  }
-
-  protected void processEvent(T tuple, Slice value)
-  {
-    if (value == BucketedState.EXPIRED) {
-      return;
-    }
-    if (value == null) {
-      //not a duplicate event
-      output.emit(tuple);
-    } else {
-      if (duplicates.isConnected()) {
-        duplicates.emit(tuple);
+        if (!finalize) {
+          break;
+        }
       }
     }
   }
@@ -137,18 +330,58 @@ public abstract class AbstractDeduper<T> implements Operator, Operator.Checkpoin
   @Override
   public void endWindow()
   {
-    Iterator<Map.Entry<T, Future<Slice>>> waitIterator = waitingEvents.entrySet().iterator();
-    while (waitIterator.hasNext()) {
-      Map.Entry<T, Future<Slice>> waitingEvent = waitIterator.next();
-      try {
-        processEvent(waitingEvent.getKey(), waitingEvent.getValue().get());
-      } catch (InterruptedException | ExecutionException e) {
-        throw new RuntimeException("end window", e);
-      }
-      waitIterator.remove();
-
+    processAuxiliary(true);
+    if (preserveTupleOrder) {
+      emitProcessedTuples();
     }
+    Preconditions.checkArgument(waitingEvents.isEmpty());
+    asyncEvents.clear();
     managedState.endWindow();
+  }
+
+  /**
+   * Records a decision for use later. This is needed to ensure that the order of incoming tuples is maintained.
+   *
+   * @param tuple the incoming tuple
+   * @param d The decision for the tuple
+   */
+  protected void recordDecision(T tuple, Decision d)
+  {
+    decisions.put(tuple, d);
+  }
+
+  /**
+   * Processes tuples for which the decision (unique / duplicate / expired) has been made.
+   * Breaks once an undecided tuple is found, as we don't want to emit out of order
+   */
+  protected void emitProcessedTuples()
+  {
+    Iterator<Entry<T, Decision>> entries = decisions.entrySet().iterator();
+    while (entries.hasNext()) {
+      Entry<T, Decision> td = entries.next();
+      switch (td.getValue()) {
+        case UNIQUE:
+          uniqueEvents++;
+          emitUnique(td.getKey());
+          entries.remove();
+          break;
+        case DUPLICATE:
+          duplicateEvents++;
+          emitDuplicate(td.getKey());
+          entries.remove();
+          break;
+        case EXPIRED:
+          expiredEvents++;
+          emitExpired(td.getKey());
+          entries.remove();
+          break;
+        default:
+          /*
+           * Decision for this is still UNKNOWN. Tuple is still waiting for bucket to be loaded. Break.
+           */
+          break;
+      }
+    }
   }
 
   @Override
@@ -174,4 +407,49 @@ public abstract class AbstractDeduper<T> implements Operator, Operator.Checkpoin
   {
     managedState.committed(windowId);
   }
+
+  protected void emitUnique(T event)
+  {
+    unique.emit(event);
+  }
+
+  protected void emitDuplicate(T event)
+  {
+    duplicate.emit(event);
+  }
+
+  protected void emitExpired(T event)
+  {
+    expired.emit(event);
+  }
+
+  /**
+   * Checks whether output of deduper should preserve the input order
+   */
+  public boolean isOrderedOutput()
+  {
+    return preserveTupleOrder;
+  }
+
+  /**
+   * If set to true, the deduper will emit tuples in the order in which they were received. Tuples which arrived later
+   * will wait for previous tuples to get processed and emitted. If not set, the order of tuples may change as tuples
+   * may be emitted out of order as and when they get processed.
+   *
+   * @param preserveTupleOrder whether or not to preserve the order of incoming tuples
+   */
+  public void setPreserveTupleOrder(boolean preserveTupleOrder)
+  {
+    this.preserveTupleOrder = preserveTupleOrder;
+  }
+
+  /**
+   * Enum for holding all possible values for a decision for a tuple
+   */
+  protected enum Decision
+  {
+    UNIQUE, DUPLICATE, EXPIRED, UNKNOWN
+  }
+
+  private static final Logger logger = LoggerFactory.getLogger(AbstractDeduper.class);
 }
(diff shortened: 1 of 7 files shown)


bhupeshchawda:
@chandnisingh @chinmaykolhatkar Please review.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,463 @@
...
+ */
+@Evolving
+@OperatorAnnotation(checkpointableWithinAppWindow = false)
+public abstract class AbstractDeduper<T>
+    implements Operator, Operator.IdleTimeHandler, ActivationListener<Context>, Operator.CheckpointNotificationListener
+{
+  /**
+   * The input port on which events are received.
+   */
+  public final transient DefaultInputPort<T> input = new DefaultInputPort<T>()
```
chinmaykolhatkar, line 74: Can be replaced with <>

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,463 @@
...
+    public final void process(T tuple)
+    {
+      processTuple(tuple);
+    }
+  };
+
+  /**
+   * The output port on which deduped events are emitted.
+   */
+  public final transient DefaultOutputPort<T> output = new DefaultOutputPort<T>();
```
chinmaykolhatkar: Can be replaced with <>... Other places are also there where this comment is true.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,463 @@
...
+  @OutputPortFieldAnnotation(optional = true)
+  public final transient DefaultOutputPort<T> expired = new DefaultOutputPort<T>();
+
+  /**
+   * Whether or not the order of tuples be maintained
+   */
+  protected boolean orderedOutput = false;
+
+  @NotNull
+  protected final ManagedTimeUnifiedStateImpl managedState = new ManagedTimeUnifiedStateImpl();
```
chinmaykolhatkar, line 105: Here the new value is assigned and made @NotNull... I don't think this make sense.. What is the purpose of @NotNull here?

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,463 @@
...
+   * tuples get processed. This is used only when {@link #orderedOutput} is true.
+   */
+  protected transient Map<T, Decision> decisions;
+  private transient long sleepMillis;
+  private transient Map<T, Future<Slice>> waitingEvents = Maps.newLinkedHashMap();
+  private transient Map<Slice, Long> asyncEvents = Maps.newLinkedHashMap();
+
+  // Metrics
+  @AutoMetric
+  protected transient long uniqueEvents;
```
chinmaykolhatkar: uniqueEvents, duplicateEvents, expiredEvents doesn't seem to be used in subclass.. Can you make them private? Also please check if there are any other such variables which do not need access outside of this java file.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,463 @@
...
+      recordDecision(tuple, Decision.UNKNOWN);
+    }
+  }
+
+  /**
+   * Processes a valid (non-expired) tuple. This tuple may be a unique or a duplicate.
+   *
+   * @param tuple
+   *          The tuple to be processed
+   * @param bucket
```
chinmaykolhatkar: Please correct javadoc for parameters.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,463 @@
...
+/**
+ * Abstract class which allows de-duplicating incoming tuples based on a configured key.
+ * Also supports expiry mechanism based on a configurable expiry period configured using {@link TimeBucketAssigner}
+ * in {@link ManagedTimeUnifiedStateImpl}
+ *
+ * @param <T> type of events
+ */
+@Evolving
+@OperatorAnnotation(checkpointableWithinAppWindow = false)
+public abstract class AbstractDeduper<T>
```
chinmaykolhatkar, line 68: Can you please fix variable and method access in AbstractDeduper.. Most of them doesn't seem to required protected.

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperTimeBasedPOJOImpl.java
```diff
@@ -0,0 +1,161 @@
...
+import com.datatorrent.api.Context.OperatorContext;
+import com.datatorrent.api.Context.PortContext;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.Operator.ActivationListener;
+import com.datatorrent.api.annotation.InputPortFieldAnnotation;
+import com.datatorrent.lib.util.PojoUtils;
+import com.datatorrent.lib.util.PojoUtils.Getter;
+import com.datatorrent.netlet.util.Slice;
+
+public class DeduperTimeBasedPOJOImpl extends AbstractDeduper<Object> implements ActivationListener<Context>
```
chinmaykolhatkar: Please make this class Evolving.

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperTimeBasedPOJOImpl.java
```diff
@@ -0,0 +1,161 @@
...
+  public void deactivate()
+  {
+  }
+
+  public String getKeyExpression()
+  {
+    return keyExpression;
+  }
+
+  public void setKeyExpression(String keyExpression)
```
chinmaykolhatkar: Can you please add javadoc for all the setter methods? This is often the most important information that users want to see while using any operator.

Also while adding javadoc can you please add UI metatags?

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperTimeBasedPOJOImpl.java
```diff
@@ -0,0 +1,161 @@
...
+public class DeduperTimeBasedPOJOImpl extends AbstractDeduper<Object> implements ActivationListener<Context>
+{
+
+  @NotNull
+  private String keyExpression;
+
+  @NotNull
+  private String timeExpression;
+
+  private long bucketSpan;
```
chinmaykolhatkar: A minor knit: Can you categorize these properties together using javadoc sections: mandatory, optional, helper.. This way user gets to know what needs to be set.

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperTimeBasedPOJOImpl.java
```diff
@@ -0,0 +1,161 @@
...
+
+  private long expireBefore;
```
chinmaykolhatkar: If this variable is optional.. Can you explicitly assign this to 0?

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperTimeBasedPOJOImpl.java
chinmaykolhatkar: Need to add constraint for bucketSpan..

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperTimeBasedPOJOImpl.java
```diff
@@ -0,0 +1,161 @@
...
+
+  private long referenceInstant;
```
chinmaykolhatkar: Need to add constraint for referenceInstant

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,463 @@
...
+
+  /**
+   * The output port on which deduped events are emitted.
+   */
+  public final transient DefaultOutputPort<T> output = new DefaultOutputPort<T>();
+
+  /**
+   * The output port on which duplicate events are emitted.
+   */
+  @OutputPortFieldAnnotation(optional = true)
```
chinmaykolhatkar: No need t set optional=true for output ports, they're optional anyway.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
chinmaykolhatkar: Can we call this port "unique" instead of "output"... To me it makes more sense to call 3 ports of deduper as "unique", "duplicates" & "expired"

bhupeshchawda: Will change it

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda, line 74: anonymous class

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda: Done

chinmaykolhatkar:
@bhupeshchawda Can you check travis build failure? There seems to be some checkstyle violation.

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperTimeBasedPOJOImpl.java
```diff
@@ -0,0 +1,209 @@
...
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.Operator.ActivationListener;
+import com.datatorrent.api.StreamCodec;
+import com.datatorrent.api.annotation.InputPortFieldAnnotation;
+import com.datatorrent.lib.util.PojoUtils;
+import com.datatorrent.lib.util.PojoUtils.Getter;
+import com.datatorrent.netlet.util.Slice;
+
+@Evolving
+public class DeduperTimeBasedPOJOImpl extends AbstractDeduper<Object> implements ActivationListener<Context>
```
chinmaykolhatkar: Can this be renamed to "TimeBasedDeduper".. I don't think Impl or POJO is required...

chinmaykolhatkar: javadoc is missing for this. Pease describe how the deduper works.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,462 @@
...
+import com.datatorrent.api.Operator.ActivationListener;
+import com.datatorrent.api.annotation.OperatorAnnotation;
+import com.datatorrent.api.annotation.OutputPortFieldAnnotation;
+import com.datatorrent.lib.fileaccess.FileAccessFSImpl;
+import com.datatorrent.lib.fileaccess.TFileImpl;
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * Abstract class which allows de-duplicating incoming tuples based on a configured key.
+ * Also supports expiry mechanism based on a configurable expiry period configured using {@link TimeBucketAssigner}
```
chinmaykolhatkar, line 56: It would be great is deduper logic can be described here. That's important piece of information that user want to have.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,462 @@
...
+  private transient long duplicateEvents;
+  @AutoMetric
+  private transient long expiredEvents;
+
+  @Override
+  public void setup(OperatorContext context)
+  {
+    sleepMillis = context.getValue(Context.OperatorContext.SPIN_MILLIS);
+    FileAccessFSImpl fAccessImpl = new TFileImpl.DTFileImpl();
+    fAccessImpl.setBasePath(context.getValue(DAG.APPLICATION_PATH) + "/bucket_data");
```
chinmaykolhatkar, line 127: Should this be done partition-wise?

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,462 @@
...
+  {
+
+    long time = getTime(tuple);
+    Future<Slice> valFuture = managedState.getAsync(time, getKey(tuple));
+
+    if (valFuture.isDone()) {
+      try {
+        processEvent(tuple, valFuture.get());
+      } catch (InterruptedException | ExecutionException e) {
+        throw new RuntimeException("process", e);
```
chinmaykolhatkar: Better describe the exception?

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,462 @@
...
+  /**
+   * The output port on which expired events are emitted.
+   */
+  @OutputPortFieldAnnotation(optional = true)
+  public final transient DefaultOutputPort<T> expired = new DefaultOutputPort<>();
+
+  /**
+   * Whether or not the order of tuples be maintained
+   */
+  private boolean orderedOutput = false;
```
chinmaykolhatkar: I suggest we make this true by default and call it out explicitly of performance penalty of doing this.
The reason for default true is then one can be sure that default deduper is idempotent which usually a assumption while targeting end-to-end exactly once.

Choice should be given on making it performant at the cost of idempotency. Not the other way round.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,462 @@
...
+   * Processes a valid (non-expired) tuple. This tuple may be a unique or a duplicate.
+   *
+   * @param tuple
+   *          The tuple to be processed
+   * @param value
+   *          Looked up key of the tuple
+   */
+  protected void processValid(T tuple, Slice value)
+  {
+    if (!orderedOutput || waitingEvents.isEmpty()) {
```
chinmaykolhatkar: Why is waitingEvent condition required here?

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,462 @@
...
+   * @param tuple
+   *          The tuple which is a unique
+   */
+  protected void processUnique(T tuple)
+  {
+    if (orderedOutput && !decisions.isEmpty()) {
+      recordDecision(tuple, Decision.UNIQUE);
+    } else {
+      uniqueEvents++;
+      emitOutput(tuple);
```
chinmaykolhatkar: knit: emitUnique... Just to to be similar on all places.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
chinmaykolhatkar: Is it possible to rename this to maintainOrder... Ordered Output looks misleading property name to me.

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperTimeBasedPOJOImpl.java
```diff
@@ -0,0 +1,209 @@
...
+  private String timeExpression;
+
+  @NotNull
+  private long bucketSpan;
+
+  @NotNull
+  private long expireBefore;
+
+  // Optional
+  private long referenceInstant = -1;
```
chinmaykolhatkar: How is referenceInstant a property? Or is it a helper variable?
From what I understood its set to timeAssigner for calculating buckets. Do we want to expose that feature to use for time shift?

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperTimeBasedPOJOImpl.java
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  @Override
+  public void setup(OperatorContext context)
+  {
+    TimeBucketAssigner timeBucketAssigner = new TimeBucketAssigner();
+    timeBucketAssigner.setBucketSpan(Duration.standardSeconds(bucketSpan));
+    timeBucketAssigner.setExpireBefore(Duration.standardSeconds(expireBefore));
+    if (referenceInstant == -1) {
+      timeBucketAssigner.setReferenceInstant(new Instant());
```
chinmaykolhatkar: If current time is used as reference, shouldn't this be set to referenceInstant variable as well to ensure that in case of restore of operator, timeAssigner gets same base time value?

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda: Agreed. Also, will rename to `preserveTupleOrder`

chinmaykolhatkar:
I think the code is in good state to change Review Only tag.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda, line 127: This is the base path. Partitions will add buckets depending on their operator id.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda: No specific description; may be number of reasons. Will remove the description "process"

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda: We are checking whether we can directly proceed to process a tuple and emit it. If we don't need to preserve order, then we can proceeed. Also even if we need to preserve the order, and there is no other event that is waiting; then we can proceed.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,462 @@
...
+   * @param tuple
+   *          The tuple to be processed
+   * @param value
+   *          Looked up key of the tuple
+   */
+  protected void processValid(T tuple, Slice value)
+  {
+    if (!orderedOutput || waitingEvents.isEmpty()) {
+      if (value == null) {
+        managedState.put(getTime(tuple), getKey(tuple), getKey(tuple));
```
chinmaykolhatkar: I understand you don't care about value for putting in managed state. Maybe en null Object can be put... For eg..
Have a static inner class;
  private static class NullObject
  {
  }

  private static final NullObject NULL = new NullObject();

And use NULL as value while putting into managed state.

chinmaykolhatkar: Intention of above approach is following:
1. Can avoid 2 times call to getKey
2. Can avoid some space fr storage and processing at managed state.

Any other approach is also fine as long as above 2 are met.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,462 @@
...
+   * Does any auxiliary processing in the idle time of the operator.
+   * Processes any tuples which are waiting for the lookup to return.
+   *
+   * @param finalize Whether or not to wait for future to return
+   */
+  protected void processAuxiliary(boolean finalize)
+  {
+    if (waitingEvents.size() > 0) {
+      Iterator<Map.Entry<T, Future<Slice>>> waitIterator = waitingEvents.entrySet().iterator();
+      while (waitIterator.hasNext()) {
```
chinmaykolhatkar, line 303: Within a single call to handleIdleTime, can you please take care of a single waitingEvent instead of take care of all the events in a single handleIdleTime.
In endWindow, this is fine.. But in handleIdleTime, you should allow engine to control amount of time handleIdleTime is called.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,462 @@
...
+        Map.Entry<T, Future<Slice>> waitingEvent = waitIterator.next();
+        T tuple = waitingEvent.getKey();
+        Slice tupleKey = getKey(tuple);
+        long tupleTime = getTime(tuple);
+        Future<Slice> future = waitingEvent.getValue();
+        if (future.isDone() || finalize ) {
+          try {
+            if (future.get() == null && asyncEvents.get(tupleKey) == null) {
+              managedState.put(tupleTime, tupleKey, tupleKey);
+              asyncEvents.put(tupleKey, tupleTime);
```
chinmaykolhatkar, line 313: where is tuples from asyncEvent removed?

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperTimeBasedPOJOImpl.java
bhupeshchawda: Done

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperTimeBasedPOJOImpl.java
bhupeshchawda: Have set the default in the declaration. Will remove the if statement.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
chinmaykolhatkar, line 313: I don't see value present in asyncEvents being used anywhere.. Can you put NULL Object instead?

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperTimeBasedPOJOImpl.java
bhupeshchawda: It is a property for user to supply in seconds. We internally convert it to an instant.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda, line 56: done

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda, line 68: done

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda: Yes, adding value was just temporary as I did not have anything else there and null was not accepted. I'll use NullObject as suggested.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda, line 313: Used in the if condition

chinmaykolhatkar, line 313: where is asyncEvent emptied?

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperStreamCodec.java
```diff
@@ -0,0 +1,45 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.lib.dedup;
+
+import com.datatorrent.lib.codec.KryoSerializableStreamCodec;
+import com.datatorrent.lib.util.PojoUtils;
+import com.datatorrent.lib.util.PojoUtils.Getter;
+
+public class DeduperStreamCodec extends KryoSerializableStreamCodec<Object>
```
chinmaykolhatkar, line 35: Can you make this Evolving?

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperStreamCodec.java
```diff
@@ -0,0 +1,45 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.lib.dedup;
+
+import com.datatorrent.lib.codec.KryoSerializableStreamCodec;
+import com.datatorrent.lib.util.PojoUtils;
+import com.datatorrent.lib.util.PojoUtils.Getter;
+
```
chinmaykolhatkar, line 27: Please add a javadoc comment describing how this works?

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperStreamCodec.java
bhupeshchawda, line 35: done

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda, line 313: in endWindow()

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda, line 303: handling only one event in a call

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,459 @@
...
+   * may be emitted out of order as and when they get processed.
+   *
+   * @param orderedOutput
+   */
+  public void setPreserveTupleOrder(boolean preserveTupleOrder)
+  {
+    this.preserveTupleOrder = preserveTupleOrder;
+  }
+
+  public ManagedTimeUnifiedStateImpl getManagedState()
```
chinmaykolhatkar: Why is this getter required?

library/src/test/java/org/apache/apex/malhar/lib/dedup/DeduperIdempotencyTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+  @Test
+  public void testApplication() throws IOException, Exception
+  {
+    try {
+      LocalMode lma = LocalMode.newInstance();
+      Configuration conf = new Configuration(false);
+      lma.prepareDAG(new DeduperIdempotencyTestApp(), conf);
+      LocalMode.Controller lc = lma.getController();
+      lc.runAsync();
```
chinmaykolhatkar: Can you run the test in async mode of Controller using method similar to:
com.datatorrent.lib.io.fs.FileSplitterBaseTest.testSplitterInApp

This gives a deterministic way of verification the cases.

library/src/test/java/org/apache/apex/malhar/lib/dedup/DeduperPartitioningTest.java
```diff
@@ -0,0 +1,171 @@
...
+   */
+  @Test
+  public void testDeduperStreamCodec()
+  {
+    try {
+      LocalMode lma = LocalMode.newInstance();
+      Configuration conf = new Configuration(false);
+      lma.prepareDAG(new TestDedupApp(), conf);
+      LocalMode.Controller lc = lma.getController();
+      lc.run(10 * 1000); // runs for 10 seconds and quits
```
chinmaykolhatkar: Can you run the test in async mode of Controller using method similar to:
com.datatorrent.lib.io.fs.FileSplitterBaseTest.testSplitterInApp

This gives a deterministic way of verification the cases.

library/src/test/java/org/apache/apex/malhar/lib/dedup/DeduperIdempotencyTest.java
```diff
@@ -0,0 +1,132 @@
...
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.api.InputOperator;
+import com.datatorrent.api.LocalMode;
+import com.datatorrent.api.StreamingApplication;
+import com.datatorrent.common.util.BaseOperator;
+import com.datatorrent.lib.fileaccess.FileAccessFSImpl;
+import com.datatorrent.lib.fileaccess.TFileImpl;
+
+public class DeduperIdempotencyTest
```
chinmaykolhatkar: Knit: Can you rename this to DeduperOrderTest ? I don't this this test idempotency really.. :)

library/src/main/java/org/apache/apex/malhar/lib/dedup/DeduperStreamCodec.java
bhupeshchawda, line 27: done

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda: removed

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,459 @@
...
+  }
+
+  protected abstract Slice getKey(T event);
+
+  protected abstract long getTime(T event);
+
+  /**
+   * Processes an incoming tuple
+   *
+   * @param tuple
```
chinmaykolhatkar: Please add description to tuple parameter.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,459 @@
...
+      }
+    } else {
+      processWaitingEvent(tuple, valFuture);
+    }
+  }
+
+  /**
+   * Processes a looked-up event
+   *
+   * @param tuple
```
chinmaykolhatkar: Please add descriptions to parameters.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,459 @@
...
+      }
+    } else {
+      processWaitingEvent(tuple, Futures.immediateFuture(value));
+    }
+  }
+
+  /**
+   * Processes invalid tuples.
+   *
+   * @param tuple
```
chinmaykolhatkar: Please add description.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,459 @@
...
+      recordDecision(tuple, Decision.EXPIRED);
+    } else {
+      processExpired(tuple);
+    }
+  }
+
+  /**
+   * Processes an expired tuple
+   *
+   * @param tuple
```
chinmaykolhatkar: Description for tuple.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,459 @@
...
+    }
+    Preconditions.checkArgument(waitingEvents.isEmpty());
+    asyncEvents.clear();
+    managedState.endWindow();
+  }
+
+  /**
+   * Records a decision for use later. This is needed to ensure that the order of incoming tuples is maintained.
+   *
+   * @param tuple
```
chinmaykolhatkar: Missing description.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
```diff
@@ -0,0 +1,459 @@
...
+  {
+    return preserveTupleOrder;
+  }
+
+  /**
+   * If set to true, the deduper will emit tuples in the order in which they were received. Tuples which arrived later
+   * will wait for previous tuples to get processed and emitted. If not set, the order of tuples may change as tuples
+   * may be emitted out of order as and when they get processed.
+   *
+   * @param orderedOutput
```
chinmaykolhatkar: Incorrect param in javadoc. Also add description to it.

library/src/test/java/org/apache/apex/malhar/lib/dedup/DeduperIdempotencyTest.java
bhupeshchawda: done

library/src/test/java/org/apache/apex/malhar/lib/dedup/DeduperPartitioningTest.java
bhupeshchawda: done

library/src/test/java/org/apache/apex/malhar/lib/dedup/DeduperIdempotencyTest.java
bhupeshchawda: done

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda: done

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda: done

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda: done

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda: done

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda: done

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractDeduper.java
bhupeshchawda: done

chinmaykolhatkar:
Looks good to me. I'll merge it tomorrow if there are no other comments.
@bhupeshchawda There are 2 commits, is it because attribution needs to be maintained? If yes, that's fine.

bhupeshchawda:
Yes, two commits are to preserve attribution.

PR #338, opened 2016-07-11
APEXMALHAR-2139 

aayushia:
Emit only the keys for which values changed or were added in a given window.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/algo/UniqueCounter.java
+++ b/library/src/main/java/com/datatorrent/lib/algo/UniqueCounter.java
@@ -19,9 +19,7 @@
 package com.datatorrent.lib.algo;
 
 import java.util.HashMap;
-import java.util.Map;
-
-import org.apache.commons.lang.mutable.MutableInt;
+import java.util.HashSet;
 
 import com.datatorrent.api.DefaultInputPort;
 import com.datatorrent.api.DefaultOutputPort;
@@ -32,7 +30,7 @@ import com.datatorrent.lib.util.UnifierHashMapSumKeys;
 /**
  * This operator counts the number of times a tuple exists in a window.&nbsp;A map from tuples to counts is emitted at the end of each window.
  * <p>
- * Counts the number of times a key exists in a window; Count is emitted at end of window in a single HashMap.
+ * Counts the number of times a key exists or is added in that given window; Count is emitted for the modified or added keys at end of window in a single HashMap.
  * </p>
  * <p>
  * This is an end of window operator<br>
@@ -58,6 +56,7 @@ import com.datatorrent.lib.util.UnifierHashMapSumKeys;
 public class UniqueCounter<K> extends BaseUniqueKeyCounter<K>
 {
   private boolean cumulative;
+  HashSet<K> inputSet = new HashSet<>();
 
   /**
    * The input port which receives incoming tuples.
@@ -70,9 +69,9 @@ public class UniqueCounter<K> extends BaseUniqueKeyCounter<K>
     @Override
     public void process(K tuple)
     {
+      inputSet.add(tuple);
       processTuple(tuple);
     }
-
   };
 
   /**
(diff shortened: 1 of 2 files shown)


aayushia:
@devtagare please review

library/src/main/java/com/datatorrent/lib/testbench/CountAndLastTupleTestSink.java
```diff
@@ -24,7 +24,7 @@
  * @displayName Count And Last Tuple Test Sink
  * @category Test Bench
  * @tags count, sink
- * @since 0.3.2
+ * @since 0.3.2
```
tweise: Remove trailing whitespace.

library/src/main/java/com/datatorrent/lib/algo/UniqueCounter.java
```diff
@@ -90,24 +91,30 @@ public void process(K tuple)
...
+        // Key present already, add the values
+        tuple.put(e.getKey(), e.getValue().toInteger());
+        oldMap.put(e.getKey(), e.getValue().toInteger() + oldMap.get(e.getKey()));
       }
-      tuple.put(e.getKey(), e.getValue().toInteger());
-    }
-    if (tuple != null) {
-      count.emit(tuple);
-    }
-    if (!cumulative) {
```
tweise, line 108: What is the purpose of cumulative after this change? When is map cleared?

tweise:
@chinmaykolhatkar can you take a look at this PR?

chinmaykolhatkar:
@tweise Sure.. Will do.

@aayushia Can you please explain what is the purpose of this change? I see there is a "cumulative" property of this operator which can assign downstream operator to identify the change.

aayushia:
@chinmaykolhatkar The purpose is to emit only the changed keys in each window as compared to emitting the entire map in endWindow. I have made some changes to the code as per @tweise comment. Have to update the unit test. Please wait till I have made all the changes.

library/src/main/java/com/datatorrent/lib/testbench/CountAndLastTupleTestSink.java
chinmaykolhatkar: THis is not done yet.

library/src/main/java/com/datatorrent/lib/algo/UniqueCounter.java
```diff
@@ -90,24 +91,30 @@ public void process(K tuple)
...
    */
   @Override
   public void endWindow()
   {
-    HashMap<K, Integer> tuple = null;
-    for (Map.Entry<K, MutableInt> e: map.entrySet()) {
+    HashMap<K, Integer> tuple = new HashMap<>();
+
+    for (Map.Entry<K, MutableInt> e : map.entrySet()) {
       if (tuple == null) {
```
chinmaykolhatkar: Not required. Tuple will never be null.

chinmaykolhatkar: you're just initializing it on line 99.

chinmaykolhatkar:
@aayushia I am still not able to understand the reason for this PR. You said that you want to emit only the changed keys in each window. I think same can be achieved using the existing operator itself. If you set the "cumulative" to false which is the property of the operator, then the map will be cleared at every endWindow, making the "map" object a fresh one. In next endWindow, it'll have content only the ones which are received in the that window.

If you see the diff of the code, in endWindow, both if and else condition is adding the content of the "map" to the "tuple". So basically, you're adding everything of map to tuple and emitting to tuple.

Please correct if my understand is wrong here. But from the looks of it and based on your requirement, I see this change in PR is not required and your requirement can be achieved using existing operator.

aayushia:
@chinmaykolhatkar I agree with what you said, the default test case checks with the "cumulative" property as false. However, irrespective, of it being true or false, it emits the entire map it got as input for that particular window. What I am tying to achieve, is emit only the keys which got modified or newly added in that window as compared to the previous windows (and not emit every key it received in that window). I have changed the code now, if you could kindly review it.

chinmaykolhatkar:
@aayushia Thanks for the explanation. I'll review the PR.

library/src/main/java/com/datatorrent/lib/algo/UniqueCounter.java
```diff
@@ -58,6 +56,7 @@
 public class UniqueCounter<K> extends BaseUniqueKeyCounter<K>
```
chinmaykolhatkar, line 56: javadoc for this method need to be updated for logic change done.

chinmaykolhatkar (changes requested):
Please make the requested changes.
Then squash the commits and rebase with master branch.

chinmaykolhatkar (commented):
Changes looks fine. Please squash the commits and rebase with master. I can merge the changes then.

chinmaykolhatkar:
The checks passed for me locally. THe travis check has failed though. @aayushia After you squash commits and rebase with master please verify whether travis build also passes.

aayushia:
@chinmaykolhatkar Made the change you mentioned. Squashed and rebased with master, as well. Can you please review and merge?

chinmaykolhatkar:
@aayushia Your branch is not rebased with master, its very old. 
Please do "git fetch" on remote base repository and then only call "git rebase <remote>/master" on ur branch.
And then push the changes to ur remote repository.

This is important to avoid any unnecessary merge commits.

PR #343, opened 2016-07-14
[Review Only] APEXMALHAR-1701: Deduper using Windowed Operator - Initial Commit. Ad…

bhupeshchawda:
…ded Tests

Added an implementation based on Windowed Operator

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractWindowedDeduper.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractWindowedDeduper.java
@@ -1,0 +1,97 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.lib.dedup;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.apex.malhar.lib.window.Tuple;
+import org.apache.apex.malhar.lib.window.Window;
+import org.apache.apex.malhar.lib.window.impl.WindowedOperatorImpl;
+
+import com.datatorrent.api.DefaultOutputPort;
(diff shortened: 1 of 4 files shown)


library/src/test/java/org/apache/apex/malhar/lib/dedup/WindowedDeduperTest.java
```diff
@@ -0,0 +1,160 @@
...
+  public static void setup()
+  {
+    dedup = new WindowedDeduperImpl();
+    // Define accumulation
+    Accumulation dedupAccum = new DedupAccumulation<TestPojo>();
+    dedup.setAccumulation(dedupAccum);
+    // Set storage to InMemory implementation
+    dedup.setDataStorage(new InMemoryWindowedStorage<List<TestPojo>>());
+    // Set retraction storage. Not used, but is required by the abstract operator.
+    dedup.setRetractionStorage(new InMemoryWindowedStorage<TestPojo>());
```
davidyan74, line 76: Should not need this if you're not doing triggers at all here. Does it give you an error?

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractWindowedDeduper.java
```diff
@@ -0,0 +1,97 @@
...
+   */
+  @Override
+  public void accumulateTuple(Tuple.WindowedTuple<T> tuple)
+  {
+    for (Window window : tuple.getWindows()) { // There will be exactly one window
+      // process each window
+      List<T> accum = dataStorage.get(window);
+      if (accum == null) {
+        // New window?
+        accum = new ArrayList<T>();
```
davidyan74, line 81: I think it's better to call defaultAccumulatedValue() here and have defaultAccumulatedValue return a new list.

library/src/test/java/org/apache/apex/malhar/lib/dedup/WindowedDeduperTest.java
bhupeshchawda, line 76: Yes. It throws a NPE. Changes are needed in the abstract operator.

library/src/main/java/org/apache/apex/malhar/lib/dedup/AbstractWindowedDeduper.java
bhupeshchawda, line 81: agreed. Will change this.

library/src/test/java/org/apache/apex/malhar/lib/dedup/WindowedDeduperTest.java
davidyan74, line 76: I see where the problem is. I have a fix here in my newest review-only PR: #345

bhupeshchawda:
@chinmaykolhatkar @chandnisingh @sandeepdeshmukh Please review

Please also compare this with https://github.com/apache/apex-malhar/pull/335
We need to decide one implementation to go ahead with.
Thanks!

bhupeshchawda:
Closing the PR. This was some exploratory implementation to see how useful a windowed deduper would be. Here is the new JIRA: https://issues.apache.org/jira/browse/APEXMALHAR-2177

https://issues.apache.org/jira/browse/APEXMALHAR-1701 will be closed as deduper based on managed state is already merged.

PR #346, opened 2016-07-18
APEXMALHAR-2158 Fixed the duplication of messages emitted issue when the Kafka Input operator redeployed

chaithu14:


chaithu14:
@siyuanh : Please review and merge.

siyuanh:
@chaithu14 Looks good to me, but can you write a unit test to verify this（just checkpoint the offset and redeploy the operator to make sure there is duplication or data loss）? And did you verify the duplicate you see is always just one tuple?

chaithu14:
@siyuanh Sure. I will write the unit test for this fix. 
Yes I verified, the duplicate is just one tuple and this depends on number of topic partitions.

chaithu14:
@siyuanh Added unit test case and updated the PR.

PR #347, opened 2016-07-20
APEXMALHAR-2156 changes

sanjaypujare:
APEXMALHAR-2156 changes: @PramodSSImmaneni  @amberarrow @ashwinchandrap  pls review.


sanjaypujare:
Added Apache license headers to the 2 new properties files

library/sqsdevCreds.properties
```diff
@@ -0,0 +1,21 @@
+#
```
ashwinchandrap: We can provide the same information from the properties.xml file for the application using the operator. This separate properties file is not needed.

library/src/main/java/com/datatorrent/lib/io/jms/AbstractJMSInputOperator.java
```diff
@@ -514,6 +514,11 @@ public WindowDataManager getWindowDataManager()
     return this.windowDataManager;
   }
 
+  public void setTransacted(boolean value)
```
amberarrow: Please remove whitespace at end of line (they tend to cause regex searches to sometimes fail on such lines).
Multiple lines in multiple files have this issue.

library/src/main/java/com/datatorrent/lib/io/jms/AbstractJMSInputOperator.java
```diff
@@ -514,6 +514,11 @@ public WindowDataManager getWindowDataManager()
     return this.windowDataManager;
   }
 
+  public void setTransacted(boolean value) 
+  {
+    transacted = value;
+  }
+
```
amberarrow: Ditto

library/pom.xml
```diff
@@ -306,12 +306,24 @@
       <artifactId>janino</artifactId>
       <version>2.7.8</version>
       <scope>test</scope>
-    </dependency>    
-     <dependency>
+    </dependency>
+ 	<dependency>
```
amberarrow: Please remove TABs (multiple lines).

library/pom.xml
```diff
@@ -306,12 +306,24 @@
       <artifactId>janino</artifactId>
       <version>2.7.8</version>
       <scope>test</scope>
-    </dependency>    
-     <dependency>
+    </dependency>
+    <dependency>
```
pramodin, line 310: Looks like junit dependency is already available as a transitive dependency otherwise the tests wouldn't have compiled so far. Why this new addition?

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
```diff
@@ -76,18 +76,24 @@
...
   private transient Destination destination;
 
+  private ConnectionFactoryBuilder connectionFactoryBuilder;
   private String connectionFactoryClass;
   private Map<String, String> connectionFactoryProperties = Maps.newHashMap();
   private String ackMode = "CLIENT_ACKNOWLEDGE";
-  private String clientId = "TestClient";
-  private String subject = "TEST.FOO";
+  private String clientId;
+  private String subject;
```
pramodin, line 86: Is subject a parameter that user has to set? In that case add a constraint validation.

library/pom.xml
sanjaypujare, line 310: True but I was running into "NoSuchMethodError" exceptions as described at http://stackoverflow.com/questions/15833015/nosuchmethoderror-with-hamcrest-1-3-junit-4-11?rq=1 . I followed one of the workarounds suggested. If this is an issue I'll look into eliminating this explicit dependency.

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
sanjaypujare, line 86: Subject is used only in the topic mode (pub-sub mode) and only with ActiveMQ since SQS supports only the queue mode. So in the topic mode subject has to be non-null but that part of the code hasn't changed and is not even being tested (previous JMSInputOperator only had tests for the queue mode). The only thing I did was to remove the hard-coding of subject to "TEST.FOO". I'll create a JIRA to do the constraint check in the topic mode.

library/sqsdevCreds.properties
sanjaypujare: We will implement a new "credentials store" in a separate JIRA. For now I have moved the location of the properties files to under test/resources .

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
```diff
@@ -355,10 +366,13 @@ public void createConnection() throws JMSException
...
   {
-    logger.debug("class {} properties {}", connectionFactoryClass, connectionFactoryProperties);
+    logger.debug("connectionFactoryBuilder {} connectionFactoryClass {} properties {}", "" + connectionFactoryBuilder,
+        connectionFactoryClass, connectionFactoryProperties);
     ConnectionFactory cf;
     try {
-      if (connectionFactoryClass != null) {
+      if (connectionFactoryBuilder != null) {
+        cf = connectionFactoryBuilder.buildConnectionFactory();
+      } else if (connectionFactoryClass != null) {
```
pramodin: Instead of having multiple mechanisms wouldn't it be better to encapsulate the factory from class instanatiation into a default factory builder?

pramodin: Also shouldn't the connection factory properties be actually handled by the builder?

sanjaypujare: Yes, both valid points. But both the default class (ActiveMQConnectionFactory) and connection factory properties was old code. Let's see if this should be changed under a separate JIRA

sanjaypujare:
Close and reopen to restart the build

sanjaypujare:
reopen to restart the build

sanjaypujare:
try build again

sanjaypujare:
build again

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
```diff
@@ -250,12 +250,7 @@ public void setup(OperatorContext context)
     consumer.create();
     // reset the offsets to checkpointed one
     if (consumer instanceof SimpleKafkaConsumer && !offsetStats.isEmpty()) {
-      Map<KafkaPartition, Long> currentOffsets = new HashMap<>();
-      // Increment the offsets and set it to consumer
-      for (Map.Entry<KafkaPartition, Long> e: offsetStats.entrySet()) {
-        currentOffsets.put(e.getKey(), e.getValue() + 1);
-      }
-      ((SimpleKafkaConsumer)consumer).resetOffset(currentOffsets);
+      ((SimpleKafkaConsumer)consumer).resetOffset(offsetStats);
```
pramodin, line 253: Why are kafka changes in this pull request?

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
```diff
@@ -76,18 +76,52 @@
...
+    public ConnectionFactory buildConnectionFactory()
+    {
+      ConnectionFactory cf;
+      try {
+        cf = new org.apache.activemq.ActiveMQConnectionFactory();
+        BeanUtils.populate(cf, super.connectionFactoryProperties);
+        logger.debug("creation successful.");
+        return cf;
+      } catch (Exception e) {
+        throw new RuntimeException("Failed to create connection factory.", e);
```
pramodin: DTThrowabled.rethrow() or throw Throwables.propagate so you are not wrapping the exception if it is already a RuntimeException

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
```diff
@@ -76,18 +76,52 @@
...
+  }
+
+  public static class DefaultConnectionFactoryBuilder extends ConnectionFactoryBuilder
+  {
+
+    @Override
+    public ConnectionFactory buildConnectionFactory()
+    {
+      ConnectionFactory cf;
+      try {
```
pramodin, line 117: Where is the connectionFactoryClass option?

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
```diff
@@ -143,7 +172,7 @@ public void setConnectionFactoryProperties(Map<String, String> connectionFactory
   @Deprecated
   public void setUser(String user)
```
pramodin, line 196: Why do different things for connectionFactoryClass and individual properties. Looks like the set/get for connectionFactoryClass have been removed in favor of specifying it as part of the factory builder but the convenience methods for the properties are still present.

Removing connectionFactoryClass property directly in the operator is probably the right way to go but it breaks backward compatibility. Since this operator is marked evolving that is allowed. But since the usage is not completely known, I suggest you email dev and users about the change and ask folks if they are using the operator today and if so would they be ok to make changes to their operator in future. From the response you can assess if you need to also provide a convenience method for the factory class that wraps the default builder or you can get rid of the convenience wrappers all together.

Second provide a getter for the connectionFactoryBuilder. With this, I can set properties in the property file for the default factory builder like this

``` xml
<property>
   <name>dt.operator.jms.connectionFactoryBuilder.connectionFactoryClass</name>
   <value>package.myJMSFactory</value>
</property>

<property>
   <name>dt.operator.jms.connectionFactoryBuilder.connectionFactoryProperties[user]</name>
   <value>myUsername</value>
</property>
```

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
```diff
@@ -76,18 +76,52 @@
   private transient Session session;
   private transient Destination destination;
 
-  private String connectionFactoryClass;
-  private Map<String, String> connectionFactoryProperties = Maps.newHashMap();
+  private ConnectionFactoryBuilder connectionFactoryBuilder = new DefaultConnectionFactoryBuilder();
```
pramodin, line 83: Mark it with NotNull validation

stream/src/test/java/org/apache/apex/malhar/stream/FunctionOperator/FunctionOperatorTest.java
```diff
@@ -1,332 +0,0 @@
-/**
```
pramodin, line 1: Why are changes to this operator included?

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
pramodin, line 196: Another option is to provide the connectionFactoryClass helper for now and take up deprecation discussions after these changes are included.

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
sanjaypujare, line 253: These are not mine - I ended up merging changes from upstream/master (because of conflicts) which show up in the pull request for this branch.

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
sanjaypujare: Sure - will look into doing it.

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
sanjaypujare, line 117: With the new scheme of using a ConnectionFactoryBuilder we don't need it (as we discussed it last). So the caller can just provide a custom builder with his own connectionFactoryClass in that builder.

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
sanjaypujare, line 196: Good point. I will go ahead and ask dev & users about connectionFactoryClass. In the mean time will think and discuss about which option to go with (remove the convenience functions or mark them deprecated).

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
sanjaypujare, line 83: Ok

stream/src/test/java/org/apache/apex/malhar/stream/FunctionOperator/FunctionOperatorTest.java
sanjaypujare, line 1: As I said before this is due to conflicts and merging from upstream that show up in my branch. Will figure out a way to avoid this in future.

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
```diff
@@ -129,12 +173,20 @@ public void setConnectionFactoryClass(String connectionFactoryClass)
    */
   public Map<String, String> getConnectionFactoryProperties()
```
PramodSSImmaneni: Also deprecate these helper methods

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
pramodin, line 253: Try rebasing your changes on top of latest master

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
```diff
@@ -112,14 +156,14 @@ public Destination getDestination()
     return destination;
   }
 
-  public String getConnectionFactoryClass()
```
pramodin, line 115: Leave the get/set connection factory class helper methods for backwards compatibility and handle them like properties below.

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
pramodin, line 117: That means they would have to implement a builder that has similar functionality as the default builder isn't it. Why not add the connectionFactoryClass logic to the default builder

if (connectionFactoryClass != null) {
        @SuppressWarnings("unchecked")
        Class<ConnectionFactory> clazz = (Class<ConnectionFactory>)Class.forName(connectionFactoryClass);
        cf = clazz.newInstance();
      } else {
        cf = new org.apache.activemq.ActiveMQConnectionFactory();
      }

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
```diff
@@ -76,18 +79,59 @@
...
+  private String subject;
   private int batch = 10;
   private int messageSize = 255;
   private boolean durable = false;
   private boolean topic = false;
   private boolean verbose = false;
   protected boolean transacted = true;
 
+  public static interface ConnectionFactoryBuilder
+  {
```
pramodin, line 95: Looks good

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
```diff
@@ -129,12 +173,20 @@ public void setConnectionFactoryClass(String connectionFactoryClass)
    */
   public Map<String, String> getConnectionFactoryProperties()
   {
-    return connectionFactoryProperties;
+    if (connectionFactoryBuilder instanceof DefaultConnectionFactoryBuilder) {
```
pramodin, line 176: On further thought about the serialization aspects and the extra objects, it would create, suggest having the connectionFactoryBuilder null by default and creating it in the helper method. So if connectionFactoryBuilder is null or not instance of DefaultConnectionFactoryBuilder set it to a new DefaultConnectionFactoryBuilder. If someone sets a different factory builder then an extra default builder will not be used.

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
```diff
@@ -76,18 +79,59 @@
...
   protected boolean transacted = true;
 
+  public static interface ConnectionFactoryBuilder
+  {
+    public ConnectionFactory buildConnectionFactory();
+  }
+
+  public static class DefaultConnectionFactoryBuilder implements ConnectionFactoryBuilder
+  {
+    protected Map<String, String> connectionFactoryProperties = Maps.newHashMap();
```
pramodin, line 101: NotNull validation on properties as these can be set from outside

amberarrow:
I did the rebase and got conflicts in library/pom.xml. Fixed them and pushed to:
https://github.com/amberarrow/incubator-apex-malhar/tree/APEXMALHAR-2156.improvement1

You can use it to compare with your local copy if necessary. Here's what I did:
git checkout master
git pull
git checkout APEXMALHAR-2156.improvement1
git pull
git rebase master
[yields conflicts in library/pom.xml which need to be fixed by hand; thereafter:]
git add library/pom.xml
git rebase --continue

This reorders the commits so that unrelated commits by other people (such as the commits to the Kafka code) are not commingled with yours.

sanjaypujare:
these changes moved to a different branch and PR....

sanjaypujare:
Opened another PR 362 and closing this one

PR #348, opened 2016-07-20
APEXMALHAR-2157 Json formatter improvements

shubham-pathak22:


shubham-pathak22:
@devtagare @chinmaykolhatkar Please review.

chinmaykolhatkar:
@shubham-pathak22 Can you please fix JSONFormatter error seen in travis build?

shubham-pathak22:
Done

library/src/main/java/com/datatorrent/lib/formatter/Formatter.java
```diff
@@ -65,17 +73,28 @@ public void setup(PortContext context)
...
   };
 
+  @Override
+  public void beginWindow(long windowId)
+  {
+    errorTupleCount = 0;
+    emittedObjectCount = 0;
+    incomingTuplesCount = 0;
+  }
+
```
devtagare, line 99: reset counters in endWindow

chinmaykolhatkar, line 99: I think this is correct. Reset should be done in begin window only. These counts are picked between endWindow and beginWindow.. If reset is done in endWindow, then count will be 0 always.

library/src/main/java/com/datatorrent/lib/formatter/JsonFormatter.java
```diff
@@ -45,58 +38,30 @@
...
-      ObjectMapper mapper = new ObjectMapper();
-      if (dateFormat != null) {
-        mapper.setDateFormat(new SimpleDateFormat(dateFormat));
-      }
-      writer = mapper.writerWithType(clazz);
-      mapper.configure(SerializationConfig.Feature.AUTO_DETECT_FIELDS, true);
-      mapper.configure(SerializationConfig.Feature.AUTO_DETECT_GETTERS, true);
-      mapper.configure(SerializationConfig.Feature.AUTO_DETECT_IS_GETTERS, true);
-    } catch (Throwable e) {
-      throw new RuntimeException("Unable find provided class");
```
devtagare, line 64: LOG.error("Unable find provided class");

library/src/main/java/com/datatorrent/lib/formatter/JsonFormatter.java
```diff
@@ -45,58 +38,30 @@
...
   @Override
   public String convert(Object tuple)
   {
+    if (tuple == null) {
+      return null;
+    }
     try {
       return writer.writeValueAsString(tuple);
-    } catch (JsonGenerationException | JsonMappingException e) {
-      logger.debug("Error while converting tuple {} {}",tuple,e.getMessage());
```
devtagare, line 74: logger.debug("Error while converting tuple {} {}",tuple,e); instead of e.getMessage() its in debug mode

library/src/main/java/com/datatorrent/lib/formatter/JsonFormatter.java
```diff
@@ -45,58 +38,30 @@
...
-    } catch (IOException e) {
-      DTThrowable.rethrow(e);
+    } catch (JsonProcessingException e) {
+      logger.error("Error while converting tuple {} {}", tuple, e.getMessage());
     }
```
devtagare, line 59: Same as above e instead of e.getMessage()

library/src/test/java/com/datatorrent/lib/formatter/JsonFormatterTest.java
```diff
@@ -35,6 +35,8 @@
 
 import org.apache.commons.io.FileUtils;
 
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.Lists;
```
devtagare, line 44: Please update the test cases here to include additional assertions based on the autometric counters added.

library/src/main/java/com/datatorrent/lib/formatter/JsonFormatter.java
shubham-pathak22, line 64: this is removed

library/src/test/java/com/datatorrent/lib/formatter/JsonFormatterTest.java
shubham-pathak22, line 44: done

library/src/main/java/com/datatorrent/lib/formatter/Formatter.java
```diff
@@ -48,6 +49,13 @@
 {
   protected transient Class<?> clazz;
 
+  @AutoMetric
+  protected long errorTupleCount;
```
chinmaykolhatkar: Why does the AutoMetric variables needs to be protected?

library/src/test/java/com/datatorrent/lib/formatter/JsonFormatterTest.java
```diff
@@ -182,14 +173,19 @@ public void testJSONToPOJONoFieldPOJO()
...
   }
 
   public static class Test1Pojo
   {
     public int a;
     public long b;
+    @JsonProperty("c2")
     public String c;
     public List<String> d;
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy/MM/dd")
```
chinmaykolhatkar: This looks good... Can you also add few other fields and see how the values interact for them:
1. java.util.Date without annotation.. What is the expected format and does the test pass with default format?
2. java.util.Time... Is there any annotation for that? If yes, can you test out  Time field for both annotation and without annotation.... If there is no annotation, please add just a new Time field and see default format.
3. Please add tests for different patterns as well.. Just a single pattern is not enough.. Please include other time and timezone etc fields also in pattern.
4. What happens if the shape is change? What are the possible values for that?

chinmaykolhatkar:
@shubham-pathak22 
Please add ApplicationTest for this... Its important to test the serialization of operator.

Also did you get chance to launch an application using this operator on actual YARN cluster. Its important to test that because the tuples are going to be POJOs containing annotations which needs to be serializable.

library/src/main/java/com/datatorrent/lib/formatter/Formatter.java
shubham-pathak22: Those need to be private. Have made it private and added protected  methods to access the values.

library/src/test/java/com/datatorrent/lib/formatter/JsonFormatterTest.java
shubham-pathak22: 1. By default Date object is written as time in millis. Had added the test for it, it passes on local machine but getting build failures when I check in the code. Hence in the test case, I have formated the date values to limit to year, month and day
2. There is no annotation for time field. Time field in the POJO is written as "time in millis". Faced similar issue with the test case as above.
3. Have added multiple patterns. Locally I have tested with "timezone" fields as well and it works. Again, locally the test case passes but fails after checking it in.
4. http://fasterxml.github.io/jackson-annotations/javadoc/2.1.0/com/fasterxml/jackson/annotation/JsonFormat.Shape.html

shubham-pathak22:
Incorporated review comments. Added application test. Also, successfully ran a sample app on YARN .

library/src/test/java/com/datatorrent/lib/formatter/JsonFormatterTest.java
chinmaykolhatkar: The reason why local test is passing and travis is failing because, the timezones in which the Date object is generated. The string that you're matching against is based on your local time zone.
But when the Date object is created in travis, the build machine is in different time zone.

Please make the test timezone agnostic.

chinmaykolhatkar:
Looks good to me. Will merge in couple days.

PR #351, opened 2016-07-27
APEXMALHAR-2169 Removed the stuff related to dynamic partition based on load from AbstractKafkaInputOperator.

chaithu14:


chaithu14:
@siyuanh Please review the PR.

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
```diff
@@ -628,6 +630,11 @@ else if (newWaitingPartition.size() != 0) {
           }
         }
 
+        if (kPIntakeRate.size() == 0) {
```
tushargosavi: add a comment for this condition.

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
```diff
@@ -188,6 +188,8 @@
   @Min(1)
   private int initialPartitionCount = 1;
 
+  private boolean isPartitionBasedOnLoad = false;
```
tushargosavi: Add a comment about this flag.

siyuanh: I don't think we should introduce another property here. We can reuse msgRateUpperBound and/or byteRateUpperBound

tushargosavi:
other things looks fine to me. @siyuanh can you take a quick look too.

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
```diff
@@ -834,17 +840,16 @@ private boolean isPartitionRequired(int opid, List<KafkaConsumer.KafkaMeterStats
         return false;
       }
 
+      if (!isPartitionBasedOnLoad) {
```
siyuanh: Why would this line be involved. If there is new kafka partition, it would return true at line 829. And new kafka partition would be added to waiting list.

siyuanh: I do think the logic is not correct in the if block from line 599. The new kafka partition should be merged into existing partition

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
chaithu14: Is this can be achieved as follows. Please correct it, if I am wrong.
if (msgRateUpperBound != Long.MAX_VALUE) {
/// Dynamic partition based on load is enabled
} else {
/// Dynamic partition based on load is disabled
}

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
chaithu14: Consider the below scenario:
By default, repartitionInterval is 30 sec. So, operator checks whether the repartition is needed or not at 30, 60, 90, 120, etc. Suppose, Meta-data change happened at 120.
In this case, operator checks whether repartition required or not at 30, 60, 90 secs. isPartitionRequired returns true based on HardConstraint and SoftConstraint.
But, if the user wants to disable the dynamic partition based on load then isPartitionRequied() has to return false at 30, 60, 90 secs.

chaithu14:
@siyuanh Incorporated the changes as suggested in JIRA. Please review.

chaithu14:

siyuanh:
@chaithu14  Will you fix the one_to_many partition issue in another pull request?

chaithu14:
@siyuanh Yes. Will open another PR for one_to_many partition issue.

chaithu14:
@siyuanh Updated the PR with the fix of one_to_many partition issue. Please review.

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
```diff
@@ -156,9 +151,11 @@
   public PartitionStrategy strategy = PartitionStrategy.ONE_TO_ONE;
 
   // default resource is unlimited in terms of msgs per second
+  @Deprecated
```
siyuanh, line 154: Please add document here.

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
```diff
@@ -592,54 +597,9 @@ else if (newWaitingPartition.size() != 0) {
...
           newPartitions.add(createPartition(kps[i], initOffset, newManagers));
         }
-
-      }
-      else if (newWaitingPartition.size() != 0) {
-
-        logger.info("[ONE_TO_MANY]: Add operator partition for kafka partition(s): {} ", StringUtils.join(newWaitingPartition, ", "));
-        partitions.add(createPartition(Sets.newHashSet(newWaitingPartition), null, newManagers));
+        newWaitingPartition.clear();
         windowDataManager.partitioned(newManagers, deletedOperators);
```
siyuanh, line 607: Not sure if it is a existing bug but windowDataManager is always called, right? Actually if we don't delete partition, we don't have to call this method?

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
chaithu14, line 154: Added.

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
chaithu14, line 607: This method has to call whenever a new partition added or existing partition deleted.

contrib/src/main/java/com/datatorrent/contrib/kafka/AbstractKafkaInputOperator.java
```diff
@@ -592,54 +597,15 @@ else if (newWaitingPartition.size() != 0) {
...
-            KafkaConsumer.KafkaMeterStats kms = (KafkaConsumer.KafkaMeterStats) stat.counters;
-            kPIntakeRate.putAll(get_1minMovingAvgParMap(kms));
-          }
-        }
-
-        List<PartitionInfo> partitionInfos = firstFitDecreasingAlgo(kPIntakeRate);
-
         // Add the existing partition Ids to the deleted operators
-        for(Partitioner.Partition<AbstractKafkaInputOperator<K>> op : partitions)
+        for(Partition<AbstractKafkaInputOperator<K>> op : partitions)
```
siyuanh: checkstyle violation

chaithu14: Updated the code with violation fix.

tweise:
@siyuanh please resolve the JIRA.

PR #354, opened 2016-07-28
APEXMALHAR-2171 #resolve #comment Added methoed call to set maxCacheS…

ShunxinLu:
…ize when building cache.


siyuanh:
@chandnisingh  Could you please review this? Is it the only place we need to use the maxSize?

otterc:
LGTM
Typo in the commit comment.

ShunxinLu:
@chandnisingh Thank you for pointing out, will be fixed in a moment.

otterc:
There is still a typo - "buiding"

PR #358, opened 2016-07-29
APEXMALHAR-2172: Updates to JDBC Poll Input Operator

DT-Priyanka:
Changes include:
1. Using jooq query to construct sql queries
2. Update logic to query based on row numbers instead of primary key column values
3. Small fixes
4. Code refactoring to make it more readable
5. Update operator to emit pojo class constructed from table input fields
6. Update to test cases, to abstract out generic test code and add more tests


DT-Priyanka:
@bhupeshchawda  and @devtagare please review

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -27,18 +27,49 @@
...
+import org.jooq.tools.jdbc.JDBCUtils;
+import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.apex.malhar.lib.wal.FSWindowDataManager;
 import org.apache.apex.malhar.lib.wal.WindowDataManager;
 import org.apache.commons.lang3.tuple.MutablePair;
 import org.apache.hadoop.classification.InterfaceStability.Evolving;
 
+/**
```
tweise: Multiple file headers and import statements mixed.

tweise:
How does 2. work? The query needs to be repeatable, so you will still need to sort on the primary key?

tweise:
The jooq license isn't one of those listed here:

http://www.apache.org/legal/resolved.html#category-a

Where is the analysis on license compatibility?

bhupeshchawda:
Building the range query is currently based on limit and offset.
This kind of querying might create problems since the order of fetching data might change depending on various factors like presence of indexes on that column or also might depend on what plan is generated for the query. The output being a set, we cannot assume the order. In order for it to be repeatable, we have the following options:
1. Explicitly use "order by" with limit and offset. - Might be costly for larger tables. Although this is worth exploring as we can assume a unique key index on this column and improve the performance. 
2. Explicitly identify bounds for the query and change the query format to `where key >= low and key <= high`. This should work in a repeatable manner; However, the performance here is again questionable in absence of indexes. Perhaps we'll need to assume an index in this case as well for performance. 
3. Use the current query (limit and offset) using something like `RowNumber` as in oracle databases. However, need to check if such functionality is present in other databases.

DT-Priyanka:
Reading further, I guess right thing to do will be use "order by" and also avoid using "offset". 

Here are some good reads about it:
http://www.iheavy.com/2013/06/19/3-ways-to-optimize-for-paging-in-mysql/
http://www.slideshare.net/slideshow/view?login=Eweaver&preview=no&slideid=1&title=efficient-pagination-using-mysql

I will check if I can modify our query to be more efficient, also would clarify assumptions about "key" column if any.

DT-Priyanka:
Thomas, using "order by" key column in queries to ensure right results.

Would evaluate options to optimize performance by avoiding use of "offset"

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +93,48 @@
...
-  /**
-   * thread to poll database
-   */
-  private transient Thread dbPoller;
-  protected transient ArrayBlockingQueue<List<T>> emitQueue;
+  protected KeyValPair<Integer, Integer> rangeQueryPair;
+  protected Integer lower;
+  protected transient boolean isPolled;
+  protected transient Integer lastPolledBound;
+  protected transient Integer lastEmittedRecord;
```
bhupeshchawda: Can lower, lastPolledBound and lastEmittedRecord be int instead of Integer?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +93,48 @@
...
+  protected transient ScheduledExecutorService scanService;
+  protected transient LinkedBlockingDeque<T> emitQueue;
   protected transient PreparedStatement ps;
   protected WindowDataManager windowManager;
```
bhupeshchawda: It would be better if you could reorder transients / non-transients into separate categories.

library/pom.xml
```diff
@@ -335,6 +335,12 @@
       <version>7.0.6</version>
     </dependency>
     <dependency>
+      <groupId>org.jooq</groupId>
+      <artifactId>jooq</artifactId>
+      <version>3.6.4</version>
+    </dependency>
+
```
bhupeshchawda: don't add extra lines

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -51,22 +64,22 @@
...
+ * User needs to provide tableName, dbConnection, columnsExpression, look-up key<br>
+ * Optionally batchSize, pollInterval and a where clause can be given <br>
  * This operator uses static partitioning to arrive at range queries for exactly
  * once reads<br>
  * This operator will create a configured number of non-polling static
  * partitions for fetching the existing data in the table. And an additional
  * single partition for polling additive data. Assumption is that there is an
  * ordered column using which range queries can be formed<br>
- * If an emitColumnList is provided, please ensure that the keyColumn is the
+ * If an columnsExpression is provided, please ensure that the keyColumn is the
```
bhupeshchawda: If this is a list, better to call it a list, and not an expression

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +93,48 @@
...
+  private int pollInterval = DEFAULT_POLL_INTERVAL;
+  private int queueCapacity = DEFAULT_QUEUE_CAPACITY;
+  private int fetchSize = DEFAULT_FETCH_SIZE;
 
   @Min(1)
   private int partitionCount = 1;
   protected transient int operatorId;
-  protected transient boolean isReplayed;
-  protected transient boolean isPollable;
+  protected boolean isPollerPartition;
```
bhupeshchawda: If this is required, please mark it `@NotNull` or else, provide a default value

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +93,48 @@
...
   protected int batchSize;
```
bhupeshchawda: Can we set `@NotNull`?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +93,48 @@
...
-  /**
-   * size of the emit queue used to hold polled records before emit
-   */
-  private static int queueCapacity = 4 * 1024 * 1024;
   private transient volatile boolean execute;
   private transient AtomicReference<Throwable> cause;
+  private String tableName;
+  private String columnsExpression;
+  private String whereCondition = null;
+  private String key;
```
bhupeshchawda, line 114: `@NotNull` for tableName, columnsExpression and key?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +93,48 @@
...
+  private transient DSLContext create;
 
-  /**
-   * size of the emit queue used to hold polled records before emit
-   */
-  private static int queueCapacity = 4 * 1024 * 1024;
   private transient volatile boolean execute;
   private transient AtomicReference<Throwable> cause;
+  private String tableName;
+  private String columnsExpression;
```
bhupeshchawda, line 112: Can it be renamed to something like `columnsListStr` ?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +93,48 @@
...
-  protected String highestPolled;
-  private static final String user = "";
-  private static final String password = "";
-  /**
-   * thread to poll database
-   */
-  private transient Thread dbPoller;
-  protected transient ArrayBlockingQueue<List<T>> emitQueue;
+  protected KeyValPair<Integer, Integer> rangeQueryPair;
+  protected Integer lower;
```
bhupeshchawda: Better name? `lowerBound`?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +93,48 @@
...
+  protected transient boolean isPolled;
+  protected transient Integer lastPolledBound;
```
bhupeshchawda: can this actually be called `upperBound`?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
bhupeshchawda: Perhaps a lot of the above variables can be private. Can you check which ones can be made private?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -286,149 +146,119 @@ public AbstractJdbcPollInputOperator()
   public void setup(OperatorContext context)
   {
     super.setup(context);
+    intializeDSLContext();
+    if (scanService == null) {
+      scanService = Executors.newScheduledThreadPool(partitionCount);
```
bhupeshchawda: This means, if the number of partitions are set to more than one, then a threadpool with multiple threads will be created. I think we just need a single thread.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +93,48 @@
  * @tags database, sql, jdbc, partitionable,exactlyOnce
  */
 @Evolving
-public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInputOperator<T, JdbcStore>
-    implements ActivationListener<Context>, IdleTimeHandler, Partitioner<AbstractJdbcPollInputOperator<T>>
+public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInputOperator<T, JdbcStore> implements
+    ActivationListener<OperatorContext>, IdleTimeHandler, Partitioner<AbstractJdbcPollInputOperator<T>>
```
bhupeshchawda: You can remove IdleTimeHandler and handleIdleTime() method. Nothing is being done there.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +93,48 @@
...
-   * size of the emit queue used to hold polled records before emit
-   */
-  private static int queueCapacity = 4 * 1024 * 1024;
   private transient volatile boolean execute;
   private transient AtomicReference<Throwable> cause;
+  private String tableName;
+  private String columnsExpression;
+  private String whereCondition = null;
+  private String key;
   protected transient int spinMillis;
```
bhupeshchawda: Not needed. This can be removed.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -286,149 +146,119 @@ public AbstractJdbcPollInputOperator()
...
+      scanService.scheduleAtFixedRate(new DBPoller(), 0, pollInterval, TimeUnit.MILLISECONDS);
+    }
+  }
+
+  protected void initializePreparedStatement()
+  {
+    try {
+      // If its a range query pass upper and lower bounds, If its a polling query pass only the lower bound
+      if (isPollerPartition) {
+        ps = store.getConnection().prepareStatement(buildRangeQuery(rangeQueryPair.getKey(), Integer.MAX_VALUE),
```
bhupeshchawda, line 176: Are you expecting rangeQueryPair form the user? If so, mark it `@NotNull`. This is not being set anywhere..

bhupeshchawda, line 176: `rangeQueryPair` must be created some where in setup() for this particular partition

tweise:
What if the user wants to use RDBMS specific SQL?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: just wanted to be consistent with JDBCPojoInputOperator which is already there in Malhar.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: that will be calculated during "definePartition" phase.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: batchSize is optional, let me add some default value to it. What's the suggestion?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka, line 114: done.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka, line 112: should we be consistent with already present JDBCPojoInputOperator? People can use same set of properties with both operators.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: yes :)

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: Right, made it "1"

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: okay, that was already there, will remove it.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka, line 176: rangeQueryPair is calculated during "definePartition" and use is not suppose to set it. Let me try to make that part of code more readable.

DT-Priyanka:
Thomas,
In JDBCPojoInputOperator, we do allow user to directly specify raw query, but we haven't done same here, it would be little complicated considering we are firing range queries here. If user doesn't specify such range queries our operator can misbehave.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +91,45 @@
...
-  private static final String password = "";
-  /**
-   * thread to poll database
-   */
-  private transient Thread dbPoller;
-  protected transient ArrayBlockingQueue<List<T>> emitQueue;
+  private transient ScheduledExecutorService scanService;
+  protected transient boolean isPolled;
+  protected transient Integer lastPolledBound;
+  protected transient Integer lastEmittedRecord;
```
bhupeshchawda: Can you make `lowerBound`, `lastPolledBound` and `lastEmittedRecord` as `int` instead of `Integer`

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +91,45 @@
...
-public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInputOperator<T, JdbcStore>
-    implements ActivationListener<Context>, IdleTimeHandler, Partitioner<AbstractJdbcPollInputOperator<T>>
+public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInputOperator<T, JdbcStore> implements
+    ActivationListener<OperatorContext>, Partitioner<AbstractJdbcPollInputOperator<T>>
 {
-  /**
-   * poll interval in milliseconds
-   */
-  private static int pollInterval = 10000;
+  private static int DEFAULT_QUEUE_CAPACITY = 4 * 1024 * 1024;
```
bhupeshchawda: Perhaps this seems too large for the queue capacity? It means even for a 100 byte record, ~ 400MB of data can be in the queue!

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -565,67 +379,110 @@ public void emitReplayedTuples(PreparedStatement ps)
...
-
-      store.connect();
-      //The n given partitions are for range queries and n + 1 partition is for polling query
-      //The upper bound for the n+1 partition is set to null since its a pollable partition
+      AbstractJdbcPollInputOperator<T> jdbcPoller = cloneUtils.getClone();
       if (i < getPartitionCount()) {
-        jdbcPoller.setRangeQueryPair(partitionToRangeMap.get(i));
-        isPollable = false;
+        jdbcPoller.rangeQueryPair = partitionToRangeMap.get(i);
+        jdbcPoller.lastEmittedRecord = partitionToRangeMap.get(i).getKey();
```
bhupeshchawda: Set `jdbcPoller.isPollerPartition = false` here?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -286,149 +141,117 @@ public AbstractJdbcPollInputOperator()
...
-        }
-        isReplayed = false;
-        LOG.debug("Prepared statement after re-initialization - {} ", ps.toString());
-      } catch (SQLException e) {
-        // TODO Auto-generated catch block
-        throw new RuntimeException(e);
-      }
+    if (isPollerPartition) {
+      updatePollQuery();
+      isPolled = false;
```
bhupeshchawda, line 205: Shouldn't `isPolled` be false irrespective of whether it is a poller partition or not?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +91,45 @@
...
-  private static final String user = "";
-  private static final String password = "";
-  /**
-   * thread to poll database
-   */
-  private transient Thread dbPoller;
-  protected transient ArrayBlockingQueue<List<T>> emitQueue;
+  private transient ScheduledExecutorService scanService;
+  protected transient boolean isPolled;
+  protected transient Integer lastPolledBound;
```
bhupeshchawda: ```upperBound```? Seems to be more intuitive when looking at the rest of the code.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -438,119 +261,110 @@ public void endWindow()
...
-        jdbcPoller.setKey(getKey());
-        jdbcPoller.setPartitionCount(getPartitionCount());
-        jdbcPoller.setPollInterval(getPollInterval());
-        jdbcPoller.setTableName(getTableName());
-        jdbcPoller.setBatchSize(getBatchSize());
-        isPollable = false;
-
-        LOG.debug("[Window ID -" + windowId + "," + recoveredData.left + "," + recoveredData.right + "]");
+        ps = store.getConnection().prepareStatement(
+            buildRangeQuery(rangeQueryPair.getKey(), (rangeQueryPair.getValue() - rangeQueryPair.getKey())),
```
bhupeshchawda: Should this fetch data from `recoveredData.left` to `recoveredData.right`?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +91,45 @@
  * @tags database, sql, jdbc, partitionable,exactlyOnce
```
bhupeshchawda: exactlyonce => idempotent

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -438,119 +261,110 @@ public void endWindow()
...
+            Integer bound = null;
+            if (lastEmittedRecord == null) {
+              bound = rangeQueryPair.getKey();
+            } else {
+              bound = lastEmittedRecord;
+            }
+            ps = store.getConnection().prepareStatement(buildRangeQuery(bound, Integer.MAX_VALUE), TYPE_FORWARD_ONLY,
+                CONCUR_READ_ONLY);
+          }
+          scanService.scheduleAtFixedRate(new DBPoller(), 0, pollInterval, TimeUnit.MILLISECONDS);
```
bhupeshchawda, line 336: This is already being done in the `activate()` method.
I don't think there is a need for this entire if() block: `if (currentWindowId == windowManager.getLargestRecoveryWindow())`. This is already taken care by before the if() statement.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -438,119 +261,110 @@ public void endWindow()
...
+      ResultSet result = ps.executeQuery();
+      if (result.next()) {
+        do {
+          emitQueue.add(getTuple(result));
+        } while (result.next());
       }
-    } catch (InterruptedException ex) {
-      // log and ignore, ending execution anyway
-      LOG.error("exception in poller thread: ", ex);
+      isPolled = true;
```
bhupeshchawda, line 285: Does this mean that we want to poll records just once in a window?
Is this because the poll interval is sufficiently long so as to prevent multiple calls to `pollRecords()` within a single window?
Even in that case, if the poll interval is reduced, there might be multiple calls where only one call will happen in this case.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -438,119 +261,110 @@ public void endWindow()
...
+    } catch (SQLException ex) {
+      throw new RuntimeException(String.format("Error while running query"), ex);
+    } finally {
+      store.disconnect();
```
bhupeshchawda: This will be called even in case of no exceptions and create problems.
We should let the exception reach the operator thread. Otherwise this might never get lost in the thread.
Once you are ready to throw the exception from the main operator thread, we can do a `store.disconnect()` in some `finally` block

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka, line 205: non-poller partition should run only once. We make it false for poller partition because, it should try to poll again in next window to find incremental changes.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: right, would update.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka, line 336: If we don't start threads in "activate" method we want to do it in beginWindow once "reply" is done.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -438,119 +261,110 @@ public void endWindow()
     currentWindowRecoveryState = new MutablePair<>();
```
bhupeshchawda: can be removed?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -438,119 +261,110 @@ public void endWindow()
...
-  {
     try {
-      if (dbPoller != null && dbPoller.isAlive()) {
-        dbPoller.interrupt();
-        dbPoller.join();
+      ps.setFetchSize(getFetchSize());
+      ResultSet result = ps.executeQuery();
+      if (result.next()) {
+        do {
+          emitQueue.add(getTuple(result));
```
bhupeshchawda: Use `offer` instead, else we might get an `IllegalStateException`

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -286,149 +141,117 @@ public AbstractJdbcPollInputOperator()
...
-      for (Object tuple : tuples) {
-        if (lower == null) {
-          lower = tuple.toString();
-        }
-        upper = tuple.toString();
-        outputPort.emit((T)tuple);
-      }
+    int pollSize = (emitQueue.size() < batchSize) ? emitQueue.size() : batchSize;
+    while (pollSize-- > 0) {
+      T obj = emitQueue.poll();
```
bhupeshchawda, line 236: Check for `obj == null`, in case the queue becomes empty

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: changing size to 4k

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka, line 285: The thread execution is not bound to window. And even if the pollInterval is short and there are multiple calls to poll record, as it's a single thread, the subsequent calls will block till this one is done. And the isPolled will be "true" for consequent calls of non-poller threads.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: calling disconnect for non-poller threads as they are not going to connect to db again, also udpated exception handling.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: removed this and couple of other lines :)

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -80,201 +91,45 @@
  * @tags database, sql, jdbc, partitionable,exactlyOnce
  */
 @Evolving
-public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInputOperator<T, JdbcStore>
-    implements ActivationListener<Context>, IdleTimeHandler, Partitioner<AbstractJdbcPollInputOperator<T>>
+public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInputOperator<T, JdbcStore> implements
```
bhupeshchawda, line 93: Can you mark this class with `checkpointableWithinAppWindow = false`?
Better to have it in the abstract class.

DT-Priyanka:
@chandnisingh, @bhupeshchawda can you also take a look?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -286,271 +146,233 @@ public AbstractJdbcPollInputOperator()
...
+            ps = store.getConnection().prepareStatement(buildRangeQuery(bound, Integer.MAX_VALUE), TYPE_FORWARD_ONLY,
+                CONCUR_READ_ONLY);
+          }
+          scanService.scheduleAtFixedRate(new DBPoller(), 0, pollInterval, TimeUnit.MILLISECONDS);
+        } catch (SQLException e) {
+          throw new RuntimeException(e);
+        }
       }
     } catch (IOException e) {
       throw new RuntimeException("replay", e);
```
bhupeshchawda: Better description for the exception?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
bhupeshchawda, line 205: Can you explicitly make the `isPolled = false` in the declaration, so there is no confusion.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
bhupeshchawda, line 285: What about non-poller partitions? `pollRecords()` is just fetching `fetchSize` records and making `isPolled = true`. This way non-poller partitions only fetch `fetchSize` records.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
bhupeshchawda: Same as above comment. Non-poller partitions are going to connect to the db for the next query, right?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -286,271 +146,233 @@ public AbstractJdbcPollInputOperator()
...
-  {
     try {
-      if (dbPoller != null && dbPoller.isAlive()) {
-        dbPoller.interrupt();
-        dbPoller.join();
+      ps.setFetchSize(getFetchSize());
+      ResultSet result = ps.executeQuery();
+      if (result.next()) {
+        do {
+          emitQueue.offer(getTuple(result));
```
bhupeshchawda: Here, we need to check if `offer()` returns `false`. If so, keep trying (after sleeping for a while) until it returns success.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -286,271 +146,233 @@ public AbstractJdbcPollInputOperator()
...
-      for (Object tuple : tuples) {
-        if (lower == null) {
-          lower = tuple.toString();
-        }
-        upper = tuple.toString();
-        outputPort.emit((T)tuple);
+    int pollSize = (emitQueue.size() < batchSize) ? emitQueue.size() : batchSize;
+    while (pollSize-- > 0) {
+      T obj = emitQueue.poll();
+      if (obj != null) {
```
bhupeshchawda, line 237: When `obj == null`, need to break out of the loop.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -27,48 +27,62 @@
...
 import com.datatorrent.lib.util.KeyValPair;
 import com.datatorrent.lib.util.KryoCloneUtils;
 import com.datatorrent.netlet.util.DTThrowable;
 
+import static java.sql.ResultSet.CONCUR_READ_ONLY;
+import static java.sql.ResultSet.TYPE_FORWARD_ONLY;
+import static org.jooq.impl.DSL.field;
+
 /**
  * Abstract operator for for consuming data using JDBC interface<br>
```
bhupeshchawda: "for for" => for

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -27,48 +27,62 @@
...
- * This operator uses static partitioning to arrive at range queries for exactly
- * once reads<br>
+ * User needs to provide tableName, dbConnection, columnsExpression, look-up key<br>
+ * Optionally batchSize, pollInterval and a where clause can be given <br>
+ * This operator uses static partitioning to arrive at range queries for
+ * idempotent reads<br>
  * This operator will create a configured number of non-polling static
  * partitions for fetching the existing data in the table. And an additional
  * single partition for polling additive data. Assumption is that there is an
  * ordered column using which range queries can be formed<br>
```
bhupeshchawda: "ordered unique column"

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -27,48 +27,62 @@
...
- * If an emitColumnList is provided, please ensure that the keyColumn is the
+ * If an columnsExpression is provided, please ensure that the keyColumn is the
  * first column in the list<br>
```
bhupeshchawda: is "keyColumn is the first column in the list" needed anymore?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -27,48 +27,62 @@
...
- * Range queries are formed using the {@link JdbcMetaDataUtility}} Output -
- * comma separated list of the emit columns eg columnA,columnB,columnC<br>
- * Only newly added data which has increasing ids will be fetched by the polling
- * jdbc partition
+ * Only newly added data will be fetched by the polling jdbc partition,
+ * also assumption is rows won't be added or deleted in middle during scan.
  * 
  * In the next iterations this operator would support an in-clause for
```
bhupeshchawda: I don't think this can be achieved. Perhaps we should not mention this in the javadoc.

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPollInputOperator.java
```diff
@@ -18,214 +18,75 @@
...
 
 import com.datatorrent.api.Context.OperatorContext;
-import com.datatorrent.api.annotation.OperatorAnnotation;
+import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.api.annotation.OutputPortFieldAnnotation;
 
 /**
- * A concrete implementation for {@link AbstractJdbcPollInputOperator}} for
+ * A concrete implementation for {@link AbstractJdbcPollInputOperator} for
  * consuming data from MySQL using JDBC interface <br>
```
bhupeshchawda: Can we remove explicit mention for MySQL? I think it is generic as of now..

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOPollInputOperator.java
```diff
@@ -0,0 +1,321 @@
...
+import com.datatorrent.api.Context;
+import com.datatorrent.api.Context.OperatorContext;
+import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.api.annotation.OutputPortFieldAnnotation;
+import com.datatorrent.lib.db.jdbc.JdbcPOJOInputOperator.ActiveFieldInfo;
+import com.datatorrent.lib.util.FieldInfo;
+import com.datatorrent.lib.util.PojoUtils;
+
+/**
+ * A concrete implementation for {@link AbstractJdbcPollInputOperator} for
```
bhupeshchawda: Remove reference to MySQL?

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOPollInputOperator.java
```diff
@@ -0,0 +1,321 @@
...
+ * consuming data from MySQL using JDBC interface <br>
+ * User needs to provide tableName,dbConnection,setEmitColumnList,look-up key <br>
+ * Optionally batchSize,pollInterval,Look-up key and a where clause can be given <br>
+ * This operator uses static partitioning to arrive at range queries for exactly
+ * once reads<br>
+ * Assumption is that there is an ordered column using which range queries can
+ * be formed<br>
+ * If an emitColumnList is provided, please ensure that the keyColumn is the
+ * first column in the list<br>
+ * Range queries are formed using the {@link JdbcMetaDataUtility} Output - comma
```
bhupeshchawda: Please update javadoc for this operator

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPollInputOperator.java
```diff
@@ -18,214 +18,75 @@
...
-
+    StringBuilder resultTuple = new StringBuilder();
     try {
-      if (ps.isClosed()) {
-        LOG.debug("Returning due to closed ps for non-pollable partitions");
-        return;
+      for (String obj : emitColumns) {
+        resultTuple.append(rs.getObject(obj) + ",");
       }
+      resultTuple.substring(0, resultTuple.length() - 1); //remove last comma
```
bhupeshchawda: Need to assign the returned substring to `resultTuple`

library/src/test/java/com/datatorrent/lib/db/jdbc/JdbcPojoPollableOpeartorTest.java
```diff
@@ -0,0 +1,251 @@
...
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class JdbcPojoPollableOpeartorTest extends JdbcOperatorTest
+{
+  public String dir = null;
+  @Mock
+  private ScheduledExecutorService mockschedular;
```
bhupeshchawda: nit: "mockscheduler" ?

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOPollInputOperator.java
```diff
@@ -0,0 +1,321 @@
...
+ * @tags database, sql, jdbc
+ */
+@Evolving
+public class JdbcPOJOPollInputOperator extends AbstractJdbcPollInputOperator<Object>
+{
+  private final transient List<ActiveFieldInfo> columnFieldSetters = Lists.newArrayList();
+  protected List<Integer> columnDataTypes;
+  protected transient Class<?> pojoClass;
+  @NotNull
+  private List<FieldInfo> fieldInfos;
```
bhupeshchawda, line 66: For consistency with Jdbc output operator, can you please use `JdbcFieldInfo` instead?

library/src/test/java/com/datatorrent/lib/db/jdbc/JdbcPojoPollableOpeartorTest.java
```diff
@@ -0,0 +1,251 @@
...
+    JdbcPOJOPollInputOperator inputOperator = new JdbcPOJOPollInputOperator();
+    inputOperator.setStore(store);
+    inputOperator.setTableName(TABLE_POJO_NAME);
+    inputOperator.setKey("id");
+    inputOperator.setFieldInfos(fieldInfos);
+    inputOperator.setFetchSize(100);
+    inputOperator.setBatchSize(100);
+    inputOperator.setPartitionCount(2);
+
+    TestUtils.MockBatchedOperatorStats readerStats = new TestUtils.MockBatchedOperatorStats(2);
```
bhupeshchawda: Is this used?

library/src/test/java/com/datatorrent/lib/db/jdbc/JdbcPojoPollableOpeartorTest.java
```diff
@@ -0,0 +1,251 @@
...
+    DefaultPartition<AbstractJdbcPollInputOperator<Object>> apartition = new DefaultPartition<AbstractJdbcPollInputOperator<Object>>(
+        inputOperator);
+    TestUtils.MockPartition<AbstractJdbcPollInputOperator<Object>> pseudoParttion = new TestUtils.MockPartition<>(
+        apartition, readerStats);
+    List<Partitioner.Partition<AbstractJdbcPollInputOperator<Object>>> newMocks = Lists.newArrayList();
+    newMocks.add(pseudoParttion);
+
+    Collection<com.datatorrent.api.Partitioner.Partition<AbstractJdbcPollInputOperator<Object>>> newPartitions = inputOperator
+        .definePartitions(newMocks, null);
+
```
bhupeshchawda: Please cleanup this part of the code. Keep only the necessary variables for identifying the partitions.

bhupeshchawda:
@DT-Priyanka I have a few outstanding comments. Please address them.
Additionally also add documentation to data members and getters and setters along with proper doc tags. Fix documentation for refactored classes.
Also, please check a few places where store.disconnect() is not called in case of an exception.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka, line 237: As we are checking queuesize before entering the while loop, there is rare chance that "obj" will be null. I had added the check just for safety.

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOPollInputOperator.java
DT-Priyanka, line 66: to be consistent with JdbcPojoInputOpeartor, skiping it for now, we can work on both operators later to make it thoroughly consistent.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -286,271 +141,238 @@ public AbstractJdbcPollInputOperator()
...
-      if (dbPoller != null && dbPoller.isAlive()) {
-        dbPoller.interrupt();
-        dbPoller.join();
+      ps.setFetchSize(getFetchSize());
+      ResultSet result = ps.executeQuery();
+      if (result.next()) {
+        do {
+          boolean submitted = false;
+          while (!submitted) {
+            submitted = emitQueue.offer(getTuple(result));
```
bhupeshchawda: `while(!emitQueue.offer(getTuple(result))) {
 Thread.sleep(10);
}` ?

bhupeshchawda: Also 1000 sounds too much?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -638,24 +504,191 @@ public void partitioned(
...
+  @VisibleForTesting
+  protected void setScheduledExecutorService(ScheduledExecutorService service)
+  {
+    scanService = service;
+  }
+
+  /**
+   * Gets {@link WindowDataManager}
+   *
+   * @return
```
bhupeshchawda: Add `return` value
Same for other getters

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: I am doing a null check and those variables are compared and assigned to each other so won't do that.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -638,24 +504,191 @@ public void partitioned(
...
+   */
+  public WindowDataManager getWindowManager()
+  {
+    return windowManager;
+  }
+
+  /**
+   * Sets {@link WindowDataManager}
+   *
+   * @return
```
bhupeshchawda: `@param` needed instead of `@return`
Same for other setters

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOPollInputOperator.java
bhupeshchawda, line 66: Can you please create a JIRA for this?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: changing to lastPolledRow

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka, line 205: by default java boolean variables are false, so didn't do that.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka, line 285: the fetchSize, is provided to the PreparedStatement to tell it how many records to fetch at a time i..e. we can say how many records to fetch as a page. Multiple of such pages will be fetched based on how many records are present. e.g. we have 1000 records and fetchSize is 100, then 10 pages of 100 records each will be fetched.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: No, they all records should be polled once within while loop. Let's discuss this.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: Should I change it to 500, as it's our default window size?

DT-Priyanka:
Sorry had missed couple of comments before, addressing remaining comments.

bhupeshchawda:
Looks good to me. If there are no other comments, will merge by tomorrow.

DT-Priyanka:
File Splitter test is failing reopening to re-run tests.

PR #362, opened 2016-08-08
JMS Input operator changes to support SQS and ActiveMQ

sanjaypujare:
@PramodSSImmaneni these changes incorporate your comments and dont have extraneous changes. Pls review and merge...

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
+++ b/library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
@@ -25,6 +25,7 @@ import javax.jms.ConnectionFactory;
 import javax.jms.Destination;
 import javax.jms.JMSException;
 import javax.jms.Session;
+import javax.validation.constraints.NotNull;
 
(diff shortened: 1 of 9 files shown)


sanjaypujare:
@PramodSSImmaneni  changed made and commits squashed. Pls review and merge as appropriate

library/src/main/java/com/datatorrent/lib/io/jms/JMSBase.java
```diff
@@ -129,12 +281,22 @@ public void setConnectionFactoryClass(String connectionFactoryClass)
    */
   public Map<String, String> getConnectionFactoryProperties()
   {
-    return connectionFactoryProperties;
+    DefaultConnectionFactoryBuilder builder =
```
pramodin: This method should also behave like getConnectionFactoryClass, throw Unsupported exception and not create a default factory builder as it is a getter. The convenience methods below setUser etc may not be able to use this method then but their functionality is correct.

PR #363, opened 2016-08-09
APEXMALHAR-2179: Add documentation for JDBC Poll Input Operator

DT-Priyanka:
@bhupeshchawda @devtagare  and @chandnisingh  please review.


DT-Priyanka:
@bhupeshchawda  can you please take a look?

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,100 @@
+JDBC Poller Input Operator
+=============
+
+## Operator Objective
+This operator scans jdbc table in parallel fashion. Based on requrested partition count, number of non polling static partitions are created to scan table portitions in parallel. Additional single partition is poller partition which scans for new data incrementally. Assumption is that there is an ordered column using which range queries can be formed. Only newly added data will be fetched by the polling. Also out of order insertion/deletion won't be supported.
```
bhupeshchawda: Objective is to read data from a JDBC complaint database.
Explain `jdbc table` - table in jdbc complaint database. 
Typo - `requested`

bhupeshchawda: User might be new to terms like `partitions`, `non-polling` etc. Can we introduce the user before using these?
May be use `operator instances` instead of `partitions`. Also explain what polling and non-polling means.
Also explain `ordered column`
Keep this restricted to the objective. Move the `not-supported`mentions to another section like `Limitations`.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,100 @@
...
+
+JDBC Poller is **idempotent**, **fault-tolerant** and **statically partitionable**.
+
+## Operator Usecase
+1. Scan huge database tables to either copy to other DB or process it using **Apache Apex**
```
bhupeshchawda: Define DB before using it, or use `database` instead.
Please also give an example for the use case.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Operator Information
+1. Operator location: ***malhar-library***
+2. Available since: ***3.5.0***
+3. Operator state: ***Evolving***
+3. Java Packages:
+    * Operator: ***[com.datatorrent.lib.db.jdbc.AbstractJdbcPollInputOperator](https://www.datatorrent.com/docs/apidocs/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.html)***
+
+### AbstractJdbcPollInputOperator
+This is the abstract implementation that serves as base class for polling messages from jdbc store. This class doesn’t have any ports.
```
bhupeshchawda: Here, could you also expand on the functionality of the operator?

bhupeshchawda: Define jdbc store

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Properties, Attributes and Ports
+### <a name="props"></a>Properties of JDBCPoller
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *store* | JDBC Store for connection | [JDBCStore](#JDBCStore) | Yes | N/A |
+| *tableName* | table name to be scanned | String | Yes | N/A |
+| *columnsExpression* | Comma separated list of columns to select from the given table. | String | Yes | N/A |
+| *key* | Primary key column name | String | Yes | N/A |
+| *partitionCount* | Static partitions counts | int | No | 1 |
```
bhupeshchawda: Static partition count*

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,100 @@
...
+| *whereCondition* | Where condition for jdbc query | String | No | N/A |
```
bhupeshchawda: Default value will be `""`

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,100 @@
...
+| *fetchSize* | The number of rows to fetch by PreparedStatement | int | No | 20000 |
```
bhupeshchawda: Can you explain this param a bit more? What PreparedStatement does?

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,100 @@
...
+| *pollInterval* | Interval in milliseconds to poll the DB | int | No | 10000 |
```
bhupeshchawda: poll the database table*

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,100 @@
...
+| *queueCapacity* | Capacity of queue which holds DB data before emiting | int | No | 4096 |
+| *batchSize* | Number of elements  to emit in a bacth | int | No | 2000 |
```
bhupeshchawda: Typo - `batch`

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,100 @@
...
+#### <a name="JDBCStore"></a>Properties of JDBC Store (BackendStore)
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *databaseDriver* |JDBC Driver class for connection to JDBC Store. This driver should be there in classpath | String | Yes | N/A |
+| *databaseUrl* | Database url of the form jdbc:subprotocol:subname | String | Yes | N/A |
+| *connectionProps* | Comma seperated connection properties e.g. user:xyz,password:ijk | String | Yes | N/A | File format. | String | Yes | N/A |
+
+## Abstract Methods
+void emitTuple(T tuple): Abstract method that emits tuple extracted from jdbc store.
+T getTuple(ResultSet result): Abstract method to get formatted tuple.
```
bhupeshchawda: `Abstract method to extract the tuple from the JDBC ResultSet object and convert it to the required type (T)`

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,100 @@
...
+
+```xml
+  <property>
+    <name>dt.operator.{OperatorName}.prop.partitionCount</name>
+    <value>4</value>
+  </property>
+```
+
+where {OperatorName} is the name of the POJOEnricher operator.
+Above lines will partition JdbcPoller statically 4 times. Above value can be changed accordingly to change the number of static partitions.
```
bhupeshchawda: `This will create 5 operator instances in all. Four of these will fetch the data which is currently in the table. We call these static partitions. An additional partition will be created which will read any newly added data. We call such a partition as a polling partition, as it "polls" for newly added data.
Changing the value of the property will create corresponding number of static partitions.`

Also describe static and polling partitions above.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,100 @@
...
+#### Properties, Attributes and Ports
+#### <a name="props"></a>Properties of JdbcPOJOPollInputOperator
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *fieldInfos*| [FieldInfo](https://www.datatorrent.com/docs/apidocs/com/datatorrent/lib/util/FieldInfo.html) maps a store column to a pojo field name.| List | Yes | N/A |
+
+#### Platform Attributes that influences operator behavior
+| **Attribute** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *outputPort.TUPLE_CLASS* | TUPLE_CLASS attribute on output port which tells operator the class of POJO which need to be emitted | Class or FQCN | Yes |
```
bhupeshchawda: expand FQCN

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,100 @@
...
+
+#### Ports
+| **Port** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *outputPort* | Tuples that are read from jdbc store are emitted from on this port | Object (POJO) | No |
+
+### 2. JdbcPollInputOperator
```
bhupeshchawda, line 162: Does this not have any properties or attributes?

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,100 @@
...
+N/A
+
+#### Platform Attributes that influences operator behavior
+N/A
+
+#### Ports
+| **Port** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *outputPort* | Tuples that are read from jdbc store are emitted from on this port | String | No |
+
```
bhupeshchawda, line 175: Please add link to some example code.

bhupeshchawda:
@DT-Priyanka I have given my comments.
@amberarrow Could you please also take a look?

docs/operators/jdbcPollInputOperator.md
amberarrow: Strongly agree with Bhupesh's comments. We can assume reader is familiar with Java and perhaps basic Apex terms like tuple, operator, port and stream but all other terms should be explained in a way that such a reader would understand, before being used. If a term is discussed in greater detail further down in the same document, use a phrase like "see section titled ??? below".

amberarrow:
I'll take a closer look over the next few days.

DT-Priyanka:
@bhupeshchawda and @amberarrow  updated document to address most of the comments as well as incorporated review comments from #368 about writing document.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
+JDBC Poller Input Operator
+=============
+
+## Operator Objective
+This operator scans jdbc table in parallel fashion. The jdbc table is table in jdbc complaint database. To provide parallelism during read each instance of Apex input operator also known as operator partition reads one portion of jdbc table. Based on requested operator instances number of non-polling static partitions are created to scan table portitions. The non-polling partition will scan the table portion only once. Additional single operator partition is created to scan newly added data incrementally.
```
amberarrow: An important part of this operator is the use of a worker thread to asynchronously fetch data from the DB and pass it to the operator thread for further processing. So we need to mention that right up front and also explain why this is needed in the first place. This paragraph overall is very difficult to understand. Suggest rewriting to provide a broader context for the problems addressed by this operator, something like this:

```
Input operators often face a couple of problems:
1. As discussed in [Development Best Practices](https://github.com/apache/apex-core/blob/master/docs/development_best_practices.md),
    the operator callbacks such as `beginWindow()`, `beginWindow()`, `emitTuples()`, etc.
    (which are invoked by the main operator thread)
    are required to return quickly, well within the default streaming window duration of
    500ms. This requirement can be an issue when retrieving data from slow external systems
    such as databases or object stores: if the call takes too long, the platform will deem
    the operator blocked and restart it. Restarting will often run into the same issue
    causing an unbroken sequence of restarts.

2. When a large volume of data is available from a single store that allows reading from
   arbitrary locations (such as a file or a database table), reading the data sequentially
   can be throughput limiting: Having multiple readers read from non-overlapping sections
   of the store allows any downstream parallelism in the DAG to be exploited better to
   enhance throughput. For files, this approach is used by the file splitter and block
   reader operators in the Malhar library.

The current operator addresses the first with an asynchronous worker thread which
retrieves the data and adds it to an in-memory queue; the main operator thread
dequeue tuples very quickly if data is available or simply returns if not. The second is
addressed in a way that parallels the approach to files by having multiple partitions
read records from non-overlapping areas of the table. Additional details of how this is
done are described below.
```

The text about "non-polling" partitions seems like implementation detail that can be moved to a later section.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
+JDBC Poller Input Operator
+=============
+
+## Operator Objective
+This operator scans jdbc table in parallel fashion. The jdbc table is table in jdbc complaint database. To provide parallelism during read each instance of Apex input operator also known as operator partition reads one portion of jdbc table. Based on requested operator instances number of non-polling static partitions are created to scan table portitions. The non-polling partition will scan the table portion only once. Additional single operator partition is created to scan newly added data incrementally. 
+
+#### Assumption
+Assumption is that there is an ordered column using which range queries can be formed. That means database has a column which has unique constraint as well as the newly inserted record should have column value more than max value in that column.
```
amberarrow: Generally speaking, we should anticipate and answer all the questions that are likely to occur to the reader. Here
are some to come to mind in this context:
1. Why is this assumption needed ?
2. Does it have to be a single column or can it be a combination of columns ?
3. Why max value (i.e. why not min value) ?

Also, if the intent is to say that every new record is constrained in some way, "the newly inserted" should
read "every newly inserted".

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Use cases
+1. Scan huge database tables to either copy to other database or process it using **Apache Apex**. The example application to copy database contents to HDFS is available in [examples repository](https://github.com/DataTorrent/examples/tree/master/tutorials/jdbcIngest)
+
```
amberarrow, line 30: We also have a variety of operators named "AbstractJdbc*" under lib/db/jdbc/. How should the reader go about
deciding which operator to use ?

"The example application" => "An example application using this operator"
"available in" => "available in the"

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+## How to Use?
+The tuple type in the abstract class is a generic parameter. Concrete subclasses need to choose an appropriate class (such as String or java class for [POJO](https://en.wikipedia.org/wiki/Plain_Old_Java_Object) etc for the generic parameter) and also implement a couple of abstract methods: getTuple(ResultSet) to parse database row to chosen class and emitTuple(T) to emit the tuple.
```
amberarrow: Providing a link for POJO seems like overkill. Suggest removing reference to POJO and just say
"an appropriate concrete class". Also add a line stating that the concrete class should have a default (no-arg)
constructor so that it can be serialized by Kryo.
"parse database row to chosen class" => "convert database rows to objects of this concrete class"

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+
+In principle, no ports need be defined in the rare case that the operator simply writes tuples to some external sink or merely maintains aggregated statistics. But in most common scenarios, the tuples need to be sent to one or more downstream operators for additional processing such as parsing, enrichment or aggregation; in such cases, appropriate output ports are defined and the emitTuple(T) implementation dispatches tuples to the desired output ports.
+
+Couple of concrete implementations are provided in Malhar:
+
+1. [JdbcPOJOPollInputOperator](https://github.com/apache/apex-malhar/blob/master/library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOPollInputOperator.java): It uses java Object for the generic parameter. You need to set output port attribute TUPLE_CLASS to define your POJO class name to define Object type. The record fetched from database table will be parsed as object of configured POJO class. This operator defines a single output port and processes each database table record one by one as a tuple object. Details are discussed below.
```
amberarrow: Suggest moving the sentence "This operator defines a single ...." to before this one: "You need to set output..."

"set output port" => "set the output port"
"from database table" => "from the database table"
"parsed as object of configured" => "parsed, using the `getTuple` method mentioned above, as an object of the configured"

Here, since POJO is part of the operator name, it seems reasonable to provide an expansion and a link.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+```xml
+  <property>
+    <name>dt.operator.{OperatorName}.prop.partitionCount</name>
+    <value>4</value>
+  </property>
+```
+
+where {OperatorName} is the name of the JDBC Poller operator.
+
+This will create 5 operator instances in all. Four of these will fetch the data which is currently in the table. We call these static non-polling partitions. An additional partition will be created which will read any newly added data. We call such a partition as a polling partition, as it "polls" for newly added data. Changing the value of the property will create corresponding number of static partitions.
```
amberarrow: Omit sentence: "Changing the value ..." since it seems redundant. Some questions arise naturally here:
1. Will there always be one and only one polling partition ?
2. Once the 4 partitions finish reading current table contents, are they idle (since all newly added data is
    read only by the polling partition) ?

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+
+2. [JdbcPollInputOperator](https://github.com/apache/apex-malhar/blob/master/library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPollInputOperator.java): It uses String for the generic parameter. This operator defines a single port and processes each database table record one by one as String tuple. Details are discussed below.
+
+## Partitioning of JDBC Poller
+#### Static Partitioning
+Only static partitioning is supported for JDBC Poller Input Operator. Configure parameter `partitionCount` to define the desired number of initial partitions (4 in this example).
+**Note**: The additional partition will be created to poll newly added records.
```
amberarrow: Suggest rewriting this note: "An additional partition ... added records, so the total number of partitions will  always
be 1 + partitionCount.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+  <name>dt.operator.{OperatorName}.prop.store.databaseUrl</name>
+  <value>jdbc:mysql://localhost:3306/mydb</value>
+</property>
+<property>
+  <name>dt.operator.{OperatorName}.prop.store.connectionProps</name>
+  <value>user:myuser,password:mypassword</value>
+</property>
+```
+
+* If you need to filter the table records, set `whereCondition` which will be added to the generated sql query.
```
amberarrow: sql => SQL

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+
+JDBC Poller is **idempotent**, **fault-tolerant** and **statically partitionable**.
+
+## AbstractJdbcPollInputOperator
+This is the abstract implementation that serves as base class for polling messages from jdbc store. It can be extended to modify functionality or add new capabilities. This class doesn’t have any ports, so concrete subclasses will need to provide them if necessary.
+
+![AbstractJdbcPollInputOperator.png](images/jdbcinput/operatorsClassDiagram.png)
+
+###<a name="AbstractJdbcPollInputOperatorProps"></a>Properties of AbstractJdbcPollInputOperator
+Several properties are available to configure the behavior of this operator and they are summarized in the table below. Of those only store, tableName, columnsExpression and key are mandatory ones. Those properties can be set like this:
```
amberarrow: Highlight these literal names like, store, tableName, etc. by, for example, using backquotes.
"mandatory ones." => "mandatory."
Suggest moving the table here to just after the sentence ending "... table below."
The sample should include all the mandatory properties ("store" is absent).
There are references in the code to "polling query" and "range query", for example, this comment in initializePreparedStatement:

```
// If its a range query pass upper and lower bounds, If its a polling query pass only the lower bound
```

Explain how they are used and provide specific examples of the generated queries in this context so that the reader can understand in concrete terms how each query is affected by these properties.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+  <value>jdbc:mysql://localhost:3306/mydb</value>
+</property>
+<property>
+  <name>dt.operator.{OperatorName}.prop.store.connectionProps</name>
+  <value>user:myuser,password:mypassword</value>
+</property>
+```
+
+* If you need to filter the table records, set `whereCondition` which will be added to the generated sql query.
+* If you have huge database to scan set `partitionsCount` to a higher number to increase read parallelism.
```
amberarrow: "huge database" => "table with a very large number or rows"

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+* The operator uses PreparedStatement, a precompiled SQL statement to fetch records from database table. You can set `fetchSize` to hint database driver to restrict number of rows to fetch in one call. The remaining rows of resultset will be fetched in subsequent calls.
```
amberarrow: "to hint" => "as a hint to the"
Add explicit note that some DB drivers may not honor the hint.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+* The platform invokes the emitTuples() callback multiple time in each streaming window; within a single such call, if a large number of tuples are emitted, there is some risk that they may overwhelm the downstream operators especially if they are performing some compute intensive operation. For such cases, output can be throttled by reducing the value of the `batchSize` property. Conversely, if the downstream operators can handle the load, increase the value to enhance throughput.
+* If there is high frequency of incoming records in your table and you want to process it as soon they appear use lower value of `pollInterval`; if they appear rarely or if some delay in processing a new file is acceptable, increase it.
```
amberarrow: "frequency" => "rate"
"process it as soon as they appear use" => "process them as soon as they appear, use"
" a new file" => "new records"

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+* After reading the records from table they are held in memory for some time till they are emitted to next operator. The records are kept in a blocking queue. The capacity of this blocking queue can be changed using parameter `queueCapacity`.
```
amberarrow: How does the reader decide when to change it ?
"from table" => "from the table"
The reason for being "held in memory" should be clarified by describing the asynchronous nature of the fetching
by a worker thread and the limits on how much data the operator thread can dequeue and emit in one call.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *store* | JDBC Store for connection | [JDBCStore](#JDBCStore) | Yes | N/A |
+| *tableName* | table name to be scanned | String | Yes | N/A |
+| *columnsExpression* | Comma separated list of columns to select from the given table. | String | Yes | N/A |
+| *key* | Primary key column name | String | Yes | N/A |
+| *partitionCount* | Static partitions count | int | No | 1 |
+| *whereCondition* | Where condition for jdbc query | String | No | N/A |
+| *fetchSize* | The number of rows to fetch by PreparedStatement | int | No | 20000 |
```
amberarrow: Hint limiting the  number of rows to fetch in a single call. Also, seems like the default value would be driver dependent. It would be immensely useful to the reader if we can document the default value for a few commonly
used drivers. For example, it looks like the MySQL treats this value in a rather odd way:
 http://stackoverflow.com/questions/20496616/fetchsize-in-resultset-set-to-0-by-default

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+| *queueCapacity* | Capacity of queue which holds DB data before emiting | int | No | 4096 |
+| *batchSize* | Maximum number of tuples to emit in a single call to the emitTuples() callback (see explanation above). | int | No | 2000 |
+
+
+#### <a name="JDBCStore"></a>Properties of JDBC Store (BackendStore)
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *databaseDriver* |JDBC Driver class for connection to JDBC Store. This driver should be there in class path | String | Yes | N/A |
+| *databaseUrl* | Database url of the form jdbc:subprotocol:subname | String | Yes | N/A |
+| *connectionProps* | Comma separated connection properties e.g. user:xyz,password:ijk | String | Yes | N/A | File format. | String | Yes | N/A |
```
amberarrow: Stray text at end of line.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+This implementation converts jdbc store records to [POJO](https://en.wikipedia.org/wiki/Plain_Old_Java_Object) and emits POJO on output port.
+
+#### <a name="JdbcPOJOPollInputOperatorProps"></a>Properties of JdbcPOJOPollInputOperator
+This operator defines following additional properties beyond those defined in the [parent class](#AbstractJdbcPollInputOperatorProps).
+
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *fieldInfos*| [FieldInfo](https://www.datatorrent.com/docs/apidocs/com/datatorrent/lib/util/FieldInfo.html) maps a store column to a POJO field name.| List | Yes | N/A |
+
+#### Platform Attributes that influences operator behavior
```
amberarrow: influences => influence

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+| -------- | ----------- | ---- | ------------------ |
+| *outputPort.TUPLE_CLASS* | TUPLE_CLASS attribute on output port which tells operator the class of POJO which need to be emitted | Class or FQCN (Fully Qualified Class Name) | Yes |
+
+#### Ports
+| **Port** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *outputPort* | Tuples that are read from jdbc store are emitted from on this port | Object (POJO) | No |
+
+### 2. JdbcPollInputOperator
+This implementation converts jdbc store records to comma separated csv record
```
amberarrow: csv record => csv records.

Also explain the reason for why this operator exists. The reader could define a POJO for a csv record and just use
JdbcPOJOPollInputOperator. Why would the reader need this operator ?

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,155 @@
...
+
+#### <a name="props"></a>Properties of JdbcPollInputOperator
+This operator defines no additional properties beyond those defined in the [parent class](#AbstractJdbcPollInputOperatorProps).
+
+#### Ports
+| **Port** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *outputPort* | Tuples that are read from jdbc store are emitted from on this port | String | No |
```
amberarrow: from on => on

docs/operators/jdbcPollInputOperator.md
DT-Priyanka: I would add a note to check for your driver relevant properties. 20000 number is what we have set by default in our operator irrespective of which sql we are working with, so we need to mention that accordingly.

docs/operators/jdbcPollInputOperator.md
DT-Priyanka: I would prefer not to include such details. There is already a discussion to change nature of query to use lower and greater than rather than limit and offset kind of queries. Such changes in implementation needs to be in consistent with document. I would prefer to avoid such data in document which goes in depth of coding or design.

DT-Priyanka:
addressed review comments from @amberarrow

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,174 @@
...
+| *batchSize* | Maximum number of tuples to emit in a single call to the emitTuples() callback (see explanation above). | int | No | 2000 |
+
+#### <a name="JDBCStore"></a>Properties of JDBC Store (BackendStore)
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *databaseDriver* |JDBC Driver class for connection to JDBC Store. This driver should be there in class path | String | Yes | N/A |
+| *databaseUrl* | Database url of the form jdbc:subprotocol:subname | String | Yes | N/A |
+| *connectionProps* | Comma separated connection properties e.g. user:xyz,password:ijk | String | Yes | N/A |
+
+Of these only `store`, `tableName`, `columnsExpression` and `key` are mandatory. Those properties can be set like this:
```
amberarrow: "store" is not set in the example below. If it is the _properties_ of store that are mandatory (rather than the store itself), we should say that.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,174 @@
...
+</property>
+<property>
+  <name>dt.operator.{OperatorName}.prop.store.connectionProps</name>
+  <value>user:myuser,password:mypassword</value>
+</property>
+```
+
+* If you need to filter the table records, set `whereCondition` which will be added to the generated SQL query.
+* If you have table with a very large number or rows to scan set `partitionsCount` to a higher number to increase read parallelism.
+* The operator uses PreparedStatement, a precompiled SQL statement to fetch records from database table. You can set `fetchSize` as a hint to the database driver to restrict number of rows to fetch in one call. The remaining rows of resultset will be fetched in subsequent calls. Please note, some of the databases may not honor this hint. Please refer to database documentation to check recommended value for your database.
```
amberarrow: "databases may not" -- please verify if this is a limitation of the DB or the _driver_; from some docs, I got the impression that some drivers may not support it and other drivers _for the same DB_ may.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,174 @@
...
+1. Operator location: ***malhar-library***
+2. Available since: ***3.5.0***
+3. Operator state: ***Evolving***
+4. Java Packages:
+    * Operator: ***[com.datatorrent.lib.db.jdbc.AbstractJdbcPollInputOperator](https://www.datatorrent.com/docs/apidocs/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.html)***
+
+JDBC Poller is **idempotent**, **fault-tolerant** and **statically partitionable**.
+
+## AbstractJdbcPollInputOperator
+This is the abstract implementation that serves as base class for polling messages from jdbc store. It can be extended to modify functionality or add new capabilities. This class doesn’t have any ports, so concrete subclasses will need to provide them if necessary.
```
amberarrow: jdbc => JDBC

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,174 @@
...
+* The platform invokes the emitTuples() callback multiple time in each streaming window; within a single such call, if a large number of tuples are emitted, there is some risk that they may overwhelm the downstream operators especially if they are performing some compute intensive operation. For such cases, output can be throttled by reducing the value of the `batchSize` property. Conversely, if the downstream operators can handle the load, increase the value to enhance throughput.
+* If there is high rate of incoming records in your table and you want to process them as soon as they appear, use lower value of `pollInterval`; if they appear rarely or if some delay in processing new records is acceptable, increase it.
+* After reading the records from the table they are held in memory for some time till they are emitted to next operator. The records are kept in a blocking queue. The capacity of this blocking queue can be changed using parameter `queueCapacity`. You can use larger size of queue when your reader thead is very fast and you want to read more data in memory to keep it ready for emission.
```
amberarrow: thead => thread

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,174 @@
...
+
+
+### Abstract Methods
+void emitTuple(T tuple): Abstract method that emits tuple extracted from jdbc store.
+
+T getTuple(ResultSet result): Abstract method to extract the tuple from the JDBC ResultSet object and convert it to the required type (T).
+
+## Concrete Classes
+### 1. JdbcPOJOPollInputOperator
+This implementation converts jdbc store records to [POJO](https://en.wikipedia.org/wiki/Plain_Old_Java_Object) and emits POJO on output port.
```
amberarrow: jdbc => JDBC

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,174 @@
...
+### 2. JdbcPollInputOperator
+This implementation converts jdbc store records to comma separated csv records. This operator is normally used when you just want to copy the data from database to somewhere else and don't want to do much of processing.
+
+#### <a name="props"></a>Properties of JdbcPollInputOperator
+This operator defines no additional properties beyond those defined in the [parent class](#AbstractJdbcPollInputOperatorProps).
+
+#### Ports
+| **Port** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *outputPort* | Tuples that are read from jdbc store are emitted on this port | String | No |
```
amberarrow: jdbc => JDBC

docs/operators/jdbcPollInputOperator.md
DT-Priyanka: Actually store object is set in java code. Should I add that comment at the end of properties section?

docs/operators/jdbcPollInputOperator.md
DT-Priyanka: right, will update that.

docs/operators/jdbcPollInputOperator.md
amberarrow: It is set in Java code in the sample application, but please try a small experiment to see if it can be set using the properties file (we can do that for TUPLE_CLASS). If that is possible, add that stanza to the example. This is the sort of thing that users often ask about.
If not, explicitly state that the store must be set from Java and provide the code fragment from the example.

DT-Priyanka: okay will check that tomorrow.

DT-Priyanka: I couldn't find any easy to use way for setting "store" instance using properties. Also please check I have added a note to mention that the store instance needs to be created in java code.

DT-Priyanka:
@amberarrow  if there are no other comments can you please merge the PR.

DT-Priyanka:
reopening PR to build travis build again.

DT-Priyanka:
@amberarrow is this PR good to go? If yes can you merge it? I have rebased it with master again.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,175 @@
+JDBC Poller Input Operator
+=============
+
+## Operator Objective
+This operator scans JDBC database table in parallel fashion. This operator is added to address common input operator problems like,
+
+1. As discussed in [Development Best Practices](https://github.com/apache/apex-core/blob/master/docs/development_best_practices.md),
+    the operator callbacks such as `beginWindow()`, `beginWindow()`, `emitTuples()`, etc.
```
amberarrow: second beginWindow should be endWindow

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,175 @@
...
+
+#### Assumption
+Assumption is that there is an ordered column using which range queries can be formed. That means database has a column or combination of columns which has unique constraint as well as every newly inserted record should have column value more than max value in that column, as we poll only appended records.
+
+## Use cases
+1. Scan huge database tables to either copy to other database or process it using **Apache Apex**. An example application using this operator to copy database contents to HDFS is available in the [examples repository](https://github.com/DataTorrent/examples/tree/master/tutorials/jdbcIngest). Look for "PollJdbcToHDFSApp" for example of this particular operator.
+
+## How to Use?
+The tuple type in the abstract class is a generic parameter. Concrete subclasses need to choose an appropriate class (such as String or an appropriate concrete java class, having no argument constructor so that it can be serialized using kyro serializer, etc for the generic parameter). Also implement a couple of abstract methods: getTuple(ResultSet) to convert database rows to objects of concrete class and emitTuple(T) to emit the tuple.
```
amberarrow: no argument => no-argument
serialized ... generic parameter =>  serialized using Kryo
Also, highlight literal function calls like getTuple(ResultSet) and emitTuple(T) using backquotes.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,175 @@
...
+  <value>jdbc:mysql://localhost:3306/mydb</value>
+</property>
+<property>
+  <name>dt.operator.{OperatorName}.prop.store.connectionProps</name>
+  <value>user:myuser,password:mypassword</value>
+</property>
+```
+
+* If you need to filter the table records, set `whereCondition` which will be added to the generated SQL query.
+* If you have table with a very large number or rows to scan set `partitionsCount` to a higher number to increase read parallelism.
```
amberarrow: have table => have a table

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,175 @@
...
+* The operator uses PreparedStatement, a precompiled SQL statement to fetch records from database table. You can set `fetchSize` as a hint to the database driver to restrict number of rows to fetch in one call. The remaining rows of resultset will be fetched in subsequent calls. Please note, some of the database drivers may not honor this hint. Please refer to database driver documentation to know recommended value.
```
amberarrow: of resultset => omit

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,175 @@
...
+* If there is high rate of incoming records in your table and you want to process them as soon as they appear, use lower value of `pollInterval`; if they appear rarely or if some delay in processing new records is acceptable, increase it.
+* After reading the records from the table they are held in memory for some time till they are emitted to next operator. The records are kept in a blocking queue. The capacity of this blocking queue can be changed using parameter `queueCapacity`. You can use larger size of queue when your reader thread is very fast and you want to read more data in memory to keep it ready for emission.
+
+**Note**: Please set right store object instance to JDBC Poller Input Operator using your application code. It's recommended to use [JdbcStore](https://github.com/apache/apex-malhar/blob/master/library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcStore.java) for this operator.
+
+### Abstract Methods
+void emitTuple(T tuple): Abstract method that emits tuple extracted from JDBC store.
```
amberarrow: Use backquotes to highlight literal code fragments.

docs/operators/jdbcPollInputOperator.md
```diff
@@ -0,0 +1,175 @@
...
+| -------- | ----------- | ---- | ------------------ |
+| *outputPort.TUPLE_CLASS* | TUPLE_CLASS attribute on output port which tells operator the class of POJO which need to be emitted | Class or FQCN (Fully Qualified Class Name) | Yes |
+
+#### Ports
+| **Port** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *outputPort* | Tuples that are read from JDBC store are emitted from on this port | Object (POJO) | No |
+
+### 2. JdbcPollInputOperator
+This implementation converts JDBC store records to comma separated csv records. This operator is normally used when you just want to copy the data from database to somewhere else and don't want to do much of processing.
```
amberarrow: csv => (CSV)

amberarrow:
Made a few minor comments; rest looks good.

DT-Priyanka:
updated.

PR #366, opened 2016-08-10
APEXMALHAR-2129 removed moving time boundaries periodically and async…

otterc:
…hrnously based on system time


otterc:
@bhupeshchawda @ilooner 
Please review since you are familiar with this code.

ilooner:
@chandnisingh lgtm

bhupeshchawda:
@chandnisingh lgtm too.

Just a thought: `getTimeBucketFor()` also moves bucket boundaries in addition to getting the time bucket which seems like a unknown side effect. Can we perhaps rename it to something like:  `getTimeBucketAndAdjustBoundaries()`

otterc:
Addressed the review comments. Please merge.

PR #368, opened 2016-08-11
APEXMALHAR-2184: Add documentation for File Input Operator

DT-Priyanka:
@yogidevendra   @ashwinchandrap  and @amberarrow  please review.


docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,101 @@
+File Input Operator
+=============
+
+## Operator Objective
+This operator scans a directory for files. Files are then read and split into tuples, which are emitted. The default implementation scans a single directory. The operator is fault tolerant. It tracks previously read files and current offset as part of checkpoint state. In case of failure the operator will skip files that were already processed and fast forward to the offset of the current file. Supports partitioning and changes to number of partitions. The directory scanner is responsible to only accept the files that belong to a partition.
```
amberarrow: "Supports partitioning ... partitions." => Supports both static and dynamic partitioning.

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Properties, Attributes and Ports
+### <a name="props"></a>Properties of AbstractFileInputOperator
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *directory* | absolute path of directory to be scanned | String | Yes | N/A |
+| *scanIntervalMillis* | Interval in milliseconds after which directory should be scanned for new files | int | No | 5000 |
+| *emitBatchSize* | Number of tuples to emit in a batch | int | No | 1000 |
+| *partitionCount* | Desired number of partitions count | int | No | 1 |
+| *maxRetryCount* | Maximum number of times the operator will attempt to process a file | No | 5 |
```
amberarrow: Missing a column

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,101 @@
...
+  <property>
+    <name>dt.operator.{OperatorName}.prop.partitionCount</name>
+    <value>4</value>
+  </property>
+```
+
+where {OperatorName} is the name of the input operator.
+Above lines will partition operator statically 4 times. Above value can be changed accordingly to change the number of static partitions.
+
+#### Dynamic Partitioning of JDBC Poller
```
amberarrow: JDBC Poller ?

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,101 @@
...
+Implement ```getNewPartitionCount()``` method to do dymanic partitioning.
+
+## Derived Classes
+### 1. AbstractFTPInputOperator
+The class is used to read files from FTP file system.
+#### Properties, Attributes and Ports
```
amberarrow: I don't see any ports mentioned; if there are none, please add a note saying something like:
"This operator has no ports" in each case.
Also, is there any reason to mention attributes since they are not operator specific ?

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,101 @@
...
+#### <a name="props"></a>Properties of AbstractFTPInputOperator
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *host*| Hostname of ftp server.| String | Yes | N/A |
+| *port*| Port of ftp server.| int | No | 21 (default ftp port) |
+| *userName*| Username which is used for login to the server. | String | No | anonymous |
+| *password*| Password which is used for login to the server. | String | No | gues |
+
+### 2. AbstractParquetFileReader
+Reads Parquet files from input directory using GroupReadSupport. Derived classes need to implement [convertGroup(Group)](https://www.datatorrent.com/docs/apidocs/com/datatorrent/contrib/parquet/AbstractParquetFileReader.html#convertGroup(Group)) method to convert Group to other type.
```
amberarrow: Since we are being explicit about what derived classes need to do here, please verify that we do the same for all the other operators described here.

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### Properties, Attributes and Ports
+#### <a name="props"></a>Properties of AbstractParquetFileReader
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *parquetSchema*| Parquet Schema to parse record. | String | Yes | N/A |
+
+### 3. AbstractThroughputFileInputOperator
+This operator provides dynamic partitioning to AbstractFileInputOperator. The user can set the preferred number of pending files per operator as well as the max number of operators and define a repartition interval. If a physical operator runs out of files to process and an amount of time greater than or equal to the repartition interval has passed then a new number of operators are created to accommodate the remaining pending files.
```
amberarrow: This is not clear and needs to be described in greater detail. Where and how does throughput enter the picture ? Perhaps add a couple of specific scenarios and describe what happens there.

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,101 @@
...
+
+![AbstractFileInputOperator.png](images/fsInput/operatorsClassDiagram.png)
+
+## Properties, Attributes and Ports
+### <a name="props"></a>Properties of AbstractFileInputOperator
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *directory* | absolute path of directory to be scanned | String | Yes | N/A |
+| *scanIntervalMillis* | Interval in milliseconds after which directory should be scanned for new files | int | No | 5000 |
+| *emitBatchSize* | Number of tuples to emit in a batch | int | No | 1000 |
```
amberarrow: explain what a "batch" is and why the user should care about it

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Operator Information
+1. Operator location: ***malhar-library***
+2. Available since: ***1.0.2***
+3. Operator state: ***Stable***
+3. Java Packages:
+    * Operator: ***[com.datatorrent.lib.io.fs.AbstractFileInputOperator](https://www.datatorrent.com/docs/apidocs/com/datatorrent/lib/io/fs/AbstractFileInputOperator.html)***
+
+### AbstractFileInputOperator
+This is the abstract implementation that serves as base class for scanning a directory for files and read the files one by one. This class doesn’t have any ports.
```
amberarrow: Add an overview here of what the operator does here, what parts need to be implemented by concrete subclasses and show code fragments for example from the LineByLine operator. Describe what the DirectoryScanner does. Some obvious questions that will come up in the reader's mind should be answered, for example:
1. What happens if a file that has already been processed has new data appended to it ?
2. What happens if a processed file is deleted ?
3. What happens if a separate actor is in the process of writing a new file in the monitored directory ? How will this operator know when the file is ready to be read ?
4. What if the number of files in the scanned directory grows over time to be very large -- what impact, if any, does it have on this operator ?
5. Can this operator be used to monitor multiple directories ? (Point people to the fileIO-multidir example)
6. This class already implements the Partitioner interface; can a custom partitioner be set on this operator ? More generally, since we are saying this operator supports dynamic partitioning, it is useful to describe (with a code fragment if possible) how to trigger it.

amberarrow: One of the issues that users often ask about is how they can tell which file a tuple belongs to in downstream operators and, a related question, how to tell when one file has ended and new one begun. Discuss a few of ways the user can provide this info and the tradeoffs in each approach, for example (a) Attach the file name to each tuple (b) create a tuple structure that can contain either an actual data tuple or a control tuple indicating begin or end of a file (c) send control tuples on a different port. Some of the examples as well as the Top N Words tutorial have sample
implementations of some of these approaches.

amberarrow: Add a code fragment showing how to handle an input file encoded in UTF8.

DT-Priyanka: Adding information about tuple to file mapping will be difficult in the code. I would suggest instead we should have an example which does that and point it to that. If you agree to it, would create an jira for tracking it further.

amberarrow: No code is needed, but a discussion of possible approaches here will be invaluable.

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,159 @@
+File Input Operator
+=============
+
+## Operator Objective
+This operator scans a directory for files. Files are then read and split into tuples based on your logic e.g. tuple can be a line  or a block from file, which are then emitted. The operator can be used with different file systems like HDFS, S3, ftp, nfs etc provided hadoop supports that file system.
+
+The default [DirectoryScanner](#https://datatorrent.com/docs/apidocs/com/datatorrent/lib/io/fs/AbstractFileInputOperator.DirectoryScanner.html) polls a input directory for files. You can provide regular expression to scanner to select perticular files from directory. The scanner maintains a list of alredy scanned files, so only newly added files will be read on next poll of directory. The modified or deleted files will not be considered during scan.
```
amberarrow: Please run it through a spell-checker, e.g. "perticular", "accpets"

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+
+The operator is fault tolerant. It tracks previously read files and current offset as part of checkpoint state. In case of failure the operator will skip files that were already processed and fast forward to the offset of the current file. Supports both static and dynamic partitioning. For static partitioning it also accpets changes to number of partitions at run time. The directory scanner is responsible to only accept the files that belong to a partition.
+## Partitioning of File Input Operator
+#### Static Partitioning of File Input Operator
+Configure parameter "partitionCount" to define desired number of static partitions.
```
amberarrow: ... partitions. => ... partitions (4 in this example).

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+
+```xml
+  <property>
+    <name>dt.operator.{OperatorName}.prop.partitionCount</name>
+    <value>4</value>
+  </property>
+```
+
+where {OperatorName} is the name of the input operator.
+Above lines will partition operator statically 4 times. Above value can be changed accordingly to change the number of static partitions.
```
amberarrow: Remove entire line.

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### Dynamic Partitioning of File Input Operator
+Override ```getNewPartitionCount()``` method to do dymanic partitioning.
```
amberarrow: Not sure if this is the whole story: Can the property not be changed at run time to achieve dynamic partitioning ?

DT-Priyanka: The logic inside getNewPartitionCount should be such that value will be calculated based on load or other factors to achieve dynamic partitioning.
In other case "partitionCount" can be changed at runtime, but that would still be static partitioning as it doesn't undergo re-partition based on some logic as such.
Isn't this understanding right?

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+The operator read contents of a file line by line. Each line is emitted as a separate tuple in string format.
+
+#### Properties and Ports
+#### Prooperties
+This operator has no added properties. Refer [parent class](#AbstractFileInputOperatorProps) properties for more details.
+
+#### Ports
+| **Port** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *output* | Tuples that are read from file are emitted on this port | String | No |
```
amberarrow: Connecting this port is mandatory; if not connected the app fails with: An error occurred trying to launch the application. Server message: javax.validation.ValidationException: At least one output port must be connected:

docs/operators/fsInputOperator.md
amberarrow: Most people would consider any partitioning that occurs while an application is running to be  "dynamic partitioning"; "static partitioning" is partitioning that occurs once before the application starts. See for example: https://cwiki.apache.org/confluence/display/Hive/DynamicPartitions
for discussion in a related, though not exactly the same, context.

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+
+![AbstractFileInputOperator.png](images/fsInput/operatorsClassDiagram.png)
+
+## Properties and Ports
+### <a name="AbstractFileInputOperatorProps"></a>Properties of AbstractFileInputOperator
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *directory* | absolute path of directory to be scanned | String | Yes | N/A |
+| *scanIntervalMillis* | Interval in milliseconds after which directory should be scanned for new files | int | No | 5000 |
+| *emitBatchSize* | Number of tuples to emit in a batch. Batch is collection of tuples bundled together. Multiple of such batches will be emitted in a window. | int | No | 1000 |
```
amberarrow: The description of a batch is incorrect -- I see no evidence of any "bundling" of a batch and also the explanation
simply explains one undefined term ("batch") using another undefined term ("bundle"). It is also inadequate: Why is batching done here ? Why should the user care about it ? How should the user decide when to change this parameter ?

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *host*| Hostname of ftp server.| String | Yes | N/A |
+| *port*| Port of ftp server.| int | No | 21 (default ftp port) |
+| *userName*| Username which is used for login to the server. | String | No | anonymous |
+| *password*| Password which is used for login to the server. | String | No | gues |
+
+#### Ports
+This operator has no ports.
+
```
amberarrow, line 203: There is a concrete class inside this file which needs to be discussed.

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+### <a name="AbstractFileInputOperatorProps"></a>Properties of AbstractFileInputOperator
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *directory* | absolute path of directory to be scanned | String | Yes | N/A |
+| *scanIntervalMillis* | Interval in milliseconds after which directory should be scanned for new files | int | No | 5000 |
+| *emitBatchSize* | Number of tuples to emit in a batch. Batch is collection of tuples bundled together. Multiple of such batches will be emitted in a window. | int | No | 1000 |
+| *partitionCount* | Desired number of partitions count | int | No | 1 |
+| *maxRetryCount* | Maximum number of times the operator will attempt to process a file | int |No | 5 |
+| *scanner* | Scanner to scan new files in directory | [DirectoryScanner](#DirectoryScanner) | No | DirectoryScanner |
+
```
amberarrow, line 149: The window data manager property needs to be discussed and explained. Chandni had a blog about this and adding a pointer to that blog will be helpful.

amberarrow:
Please take a look at branch APEXMALHAR-2184-ram at https://github.com/amberarrow/incubator-apex-malhar

I've added more explanatory content and made various other changes. I've also made several additional comments on this PR. Please review both and merge into your PR. This guide needs more work, specifically:

The use of the WindowDataManager needs to be discussed.
The derivative abstract classes need more explanation and also need code fragments illustrating how to make concrete derived classes out of them.

DT-Priyanka:
@amberarrow  is it a good idea to discussion possible implementations for scenarios like tracking file name for tuple etc in this document? Or they should go somewhere else like discussion forum etc? Or may be a new section in this document which discuss about implementation possibilities?

DT-Priyanka:
@amberarrow  for partitioning, you have mentioned dynamic partitioning can be done by changing value of "partitionCount" parameter, but that doesn't seem to be a dynamic partitioning exactly. Provided this any way of defining our static partitioning can actually be dynamic as we allow changes of variable values on application runtime.
Also we have already implemented the "definePartitions" in Abstract class and taken care of redistribution of load to new partitions, we just expect users to override "getNewPartitionCount". Should we better abstract out the details of "definePartitions" and just expose "getNewPartitionCount" to users?

amberarrow:
In this instance, tracking the file name comes up often w.r.t this operator, so yes, I think we should provide the information in the context where it is most useful and that is here. Creating a new section in the document is fine as long as the information is present and is accurate. The new section can be titled "Common implementation issues" or even "FAQ" or something similar. Not all operators merit this treatment, but this one does.

We should not lose track of the main purpose of this document: The reader should be able to use the operator in common situations without running into any difficulties that are not discussed in the document.

Regarding dynamic partitioning -- I've explained my rationale in the earlier note. Yes we allow variable changes at runtime and it is quite reasonable and even standard to call them "dynamic" changes. Not all operators support partitioning at runtime based on such changes; it is entirely reasonable to call those that do "dynamically partitionable". I don't quite understand what this means: "Provided this any way of defining our static partitioning can actually be dynamic ".

The design question can be discussed separately.

DT-Priyanka:
@amberarrow  I have addressed most of your review comments. Can you take a look and merge PR if looks appropriate?

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,339 @@
...
+This operator is suitable for use in an environment where small to medium sized files are
+deposited in a specific directory on a regular basis. For very large files a better alternative
+is the `FileSplitter` and `BlockReader` combination since they allow such files to be processed
+by multiple partitions to achieve higher throughput. Additionally, files which are continually
+modified by other processes are not suitable for processing with this operator since they may
+yield unpredictable results.
+
+## How to Use?
+The tuple type in the abstract class is a generic parameter.
+Concrete subclasses need to choose an appropriate class (such as `String` or `byte[]` for the
```
amberarrow: missing close parenthesis after byte[]

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,339 @@
...
+  incoming parameter or any other factors, define a new desired value of
+  `partitionCount` and finally, if this value differs from the current partition
+  count, set the `repartitionRequired` flag in the response.
+
+The details of actually creating the new set of partitions can be customized by overriding
+the `definePartitions()` method. There are a couple of things to keep in mind when doing this.
+The first is that repartitioning needs some care when the operator has state (as is the
+case here): Existing state from current operator partitions needs to redistributed to the
+new partitions in a logically consistent way. The second is that some or all of the
+current set of partitions, which is an input parameter to `definePartitions()`, can by
```
amberarrow: by => be

docs/operators/fsInputOperator.md
```diff
@@ -0,0 +1,339 @@
...
+copied over to the new set; such partitions will continue running and will not be
+restarted. Any existing that are not present in the new set will be shutdown.
```
amberarrow: existing => existing partitions

amberarrow (approved):
Looks good, I'll merge in a day or two.

PR #372, opened 2016-08-16
APEXMALHAR-2195 - Fix LineReaderContext last record

yogidevendra:
1. Changes in the test app

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/block/ReaderContext.java
+++ b/library/src/main/java/com/datatorrent/lib/io/block/ReaderContext.java
@@ -152,7 +152,7 @@ public interface ReaderContext<STREAM extends InputStream & PositionedReadable>
     private final transient ByteArrayOutputStream tmpBuilder;
 
     private transient byte[] buffer;
-    private transient String strBuffer;
+    private transient String bufferStr;
     private transient int posInStr;
 
     public LineReaderContext()
@@ -190,11 +190,11 @@ public interface ReaderContext<STREAM extends InputStream & PositionedReadable>
           if (bytesRead == -1) {
             break;
           }
-          strBuffer = new String(buffer);
+          bufferStr = new String(buffer,0, bytesRead);
         }
 
-        while (posInStr < strBuffer.length()) {
-          char c = strBuffer.charAt(posInStr);
+        while (posInStr < bufferStr.length()) {
+          char c = bufferStr.charAt(posInStr);
           if (c != '\r' && c != '\n') {
             tmpBuilder.write(c);
             posInStr++;
@@ -208,8 +208,8 @@ public interface ReaderContext<STREAM extends InputStream & PositionedReadable>
         lineBuilder.write(subLine);
 
         if (foundEOL) {
-          while (posInStr < strBuffer.length()) {
-            char c = strBuffer.charAt(posInStr);
+          while (posInStr < bufferStr.length()) {
+            char c = bufferStr.charAt(posInStr);
             if (c == '\r' || c == '\n') {
               emptyBuilder.write(c);
               posInStr++;
@@ -219,6 +219,10 @@ public interface ReaderContext<STREAM extends InputStream & PositionedReadable>
           }
           usedBytes += emptyBuilder.toByteArray().length;
         } else {
+          //end of stream reached
+          if (bytesRead < bufferSize) {
+            break;
+          }
           //read more bytes from the input stream
           posInStr = 0;
         }
--- a/library/src/test/java/org/apache/apex/malhar/lib/fs/FSRecordReaderTest.java
+++ b/library/src/test/java/org/apache/apex/malhar/lib/fs/FSRecordReaderTest.java
@@ -90,14 +90,22 @@ public class FSRecordReaderTest
     LocalMode.Controller lc = lma.getController();
     lc.setHeartbeatMonitoringEnabled(true);
     lc.runAsync();
-    LOG.debug("Waiting for app to finish");
-    Thread.sleep(1000 * 1);
+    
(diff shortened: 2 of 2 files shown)


yogidevendra:
@amberarrow Could you please review this?

amberarrow:
Ok, over the weekend.

library/src/test/java/org/apache/apex/malhar/lib/fs/FSRecordReaderTest.java
```diff
@@ -89,15 +89,20 @@ public void testDelimitedRecords() throws Exception
     lma.prepareDAG(app, conf);
     LocalMode.Controller lc = lma.getController();
     lc.setHeartbeatMonitoringEnabled(true);
-    lc.runAsync();
+    lc.run(10 * 1000);
     LOG.debug("Waiting for app to finish");
     Thread.sleep(1000 * 1);
     lc.shutdown();
```
gauravgopi123: lc.shutdown is not required if you are using lc.run with time > 0

library/src/main/java/com/datatorrent/lib/io/block/ReaderContext.java
```diff
@@ -190,7 +190,7 @@ protected Entity readEntity() throws IOException
           if (bytesRead == -1) {
             break;
           }
-          strBuffer = new String(buffer);
+          strBuffer = new String(buffer,0, bytesRead);
```
gauravgopi123: strBuffer is misnomer. It refers string and not string buffer.. Can you please refactor this too

yogidevendra:
@gauravgopi123 Incorporated your review comments. Could you please have a look?

library/src/main/java/com/datatorrent/lib/io/block/ReaderContext.java
```diff
@@ -190,11 +190,11 @@ protected Entity readEntity() throws IOException
...
-          strBuffer = new String(buffer);
+          bufferStr = new String(buffer,0, bytesRead);
         }
 
-        while (posInStr < strBuffer.length()) {
-          char c = strBuffer.charAt(posInStr);
+        while (posInStr < bufferStr.length()) {
+          char c = bufferStr.charAt(posInStr);
           if (c != '\r' && c != '\n') {
             tmpBuilder.write(c);
```
gauravgopi123, line 199: This is lossy write to tmpBuilder using ByteArrayOutputStream.write(int). byte[] to String conversion (line 193) is based on default encoding. In case default encoding where multiple bytes are required to store chars, this will lead to lose of data...

After looking at the code little more, I also found that code can be simplified a lot and we can get rid of tmpBuilder and emptyBuilder

yogidevendra, line 199: @gauravgopi123 How about creating separate JIRA for addressing this issue?
We can have in-depth discussion on the issues, alternatives, possible solutions on the new JIRA.

yogidevendra:
@gauravgopi123 @amberarrow  any comments?

amberarrow:
@yogidevendra It'll probably be another week before I can look at this, swamped currently.

gauravgopi123:
@yogidevendra : Any reason to rush for this fix? If you plan to use LineReaderContext with only your fix, it will still be buggy. So I suggest that you fix the other issue and merge with this PR.

yogidevendra:
@gauravgopi123 here are some reasons:
1. We can pull this fix for malhar 3.5 release if new JIRA is created for the other problems you mentioned.
2. I have another PR in examples repo which is dependent on this fix. 
3. Splitting the JIRA would help in better tracking, discussion focus.

@amberarrow @tweise Any opinions on separate JIRA?

library/src/main/java/com/datatorrent/lib/io/block/ReaderContext.java
tweise: @yogidevendra What does it take to address the tmpBuilder issue now? Does not seem to require elaborate discussion. I would also prefer to get it fixed but if you are not able to work on it then take it up for next release.

yogidevendra: @tweise I will prefer to work on it later. Also, I have some possible alternative in my mind which I will propose in the new JIRA. 
I have created new JIRA APEXMALHAR-2198 for this.

library/src/test/java/org/apache/apex/malhar/lib/fs/FSRecordReaderTest.java
```diff
@@ -89,15 +89,18 @@ public void testDelimitedRecords() throws Exception
     lma.prepareDAG(app, conf);
     LocalMode.Controller lc = lma.getController();
     lc.setHeartbeatMonitoringEnabled(true);
-    lc.runAsync();
+    lc.run(10 * 1000);
```
tweise: Please implement proper test termination. This topic was discussed elsewhere before.

yogidevendra: @tweise updated to handle proper test termination.

tweise:
Your branch is behind by 18 commits. Please run `git rebase apache/master` to rebase.

tweise:
@yogidevendra it looks like your upstream repo is the github mirror and not the asf repo. That would explain why there is the last commit missing. In the future please pull from the asf repo as sometimes sync to the mirror is delayed (saw it happening twice since yesterday).

yogidevendra:
@tweise updated after rebase with asf. But, I noticed travis build is in pending state for long time for last few builds. Any clues?

tweise:
Yes, Travis builds also take really long. Not sure why.

PR #373, opened 2016-08-18
APEXMALHAR-2154 Update the Kafka Input Operator to use CheckpointNotificationListener

chaithu14:


sandeshh:
Should this be done for every operators in Malhar?

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaInputOperator.java
```diff
@@ -191,6 +191,12 @@ public void checkpointed(long l)
   }
 
   @Override
+  public void beforeCheckpoint(long windowId)
```
tweise, line 194: @chaithu14 what is the purpose of this change? Why add an interface with no implementation?

chaithu14: JIRA represents to update the checkpointlistner. In this operator, there is no functionality in checkpoint.

PR #377, opened 2016-08-23
*Review Only* APEXMALHAR-2192 Added some spillable method implementations

davidyan74:
This is not ready for merge yet. Please do not comment on missing unit tests, code style, debug statements, etc.

@brightchen @ilooner I added removeAll method to SpillableByteArrayListMultimapImpl but the testRemoveAll test method fails. Can you please take a look at what I did wrong?


library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableByteArrayListMultimapImpl.java
```diff
@@ -152,7 +153,15 @@ public SpillableStateStore getStore()
   @Override
   public List<V> removeAll(@Nullable Object key)
```
brightchen, line 154: The map not only exists in the Spillable Map, but could also serialized and copy to the bucket of the state management. and also could be written to one or even multiple files. So I don't think this solution could work.

davidyan74: @brightchen I know you're concerned about the deleted list ending up being garbage, I have the same concern but that does not explain why the unit test is failing.

PR #379, opened 2016-08-25
APEXMALHAR-2134 Fix the NullPointerException, if the kafka partition has no leader broker

chaithu14:


chaithu14:

contrib/src/main/java/com/datatorrent/contrib/kafka/SimpleKafkaConsumer.java
```diff
@@ -508,6 +508,9 @@ private void monitorMetadata()
               continue;
             }
             Broker b = pm.leader();
+            if (b == null) {
+              continue;
```
siyuanh, line 513: We should put a INFO log here.

chaithu14: Added a log message and updated.

tweise:
@chaithu14 can you rebase the PR so we get the unit test fix into the CI build.

PR #380, opened 2016-08-25
APEXMALHAR-2199 Added the support chroot zookeeper path in Kafka Input Operator

chaithu14:


chaithu14:

siyuanh:
@chaithu14  I think the problem is more than your change. Looks like the valid url for zookeeper client is node0:port,node1:port/chroot/your/path, not node0:port/chroot/your/path,node1:port/chroot/your/path. so unlike the kafka client connect to each node individually, the zk client always connect to "cluster". So we should change the "zookeeperMap" in kafka consumer to just a plain map from "cluster id" => "zk connection string"  then when we try to use zookeeper client we don't need to concatenate the string which would be potentially wrong.

siyuanh:
@chaithu14 I opened another PR, could you please close this one?

PR #382, opened 2016-08-26
APEXMALHAR-2200 Enabled checkstyle for demos. Fixed checkstyle errors.

ShunxinLu:


demos/frauddetect/src/main/java/com/datatorrent/demos/frauddetect/SlidingWindowSumKeyVal.java
```diff
@@ -69,31 +70,31 @@ public void processDataTuple(KeyValPair<K, V> tuple)
       }
       buffer.put(key, stateList);
     }
-
+
```
tweise: Please remove trailing whitespaces. Should be an IDE setting.

demos/pom.xml
```diff
@@ -163,9 +163,6 @@
         <plugin>
```
tweise: Remove entire plugin configuration. It is inherited from parent pom.

siyuanh: In parent, skip is set to true, shall we enforce the checkstyle check by default?

tweise: Where do you see skip set to true? Yes, it should be on by default and problematic modules blacklisted.

demos/frauddetect/src/main/java/com/datatorrent/demos/frauddetect/MerchantTransactionBucketOperator.java
```diff
@@ -102,34 +105,34 @@ private void processTuple(MerchantTransaction tuple)
...
   private void emitCreditCardKeyTuple(MerchantTransaction tuple, DefaultOutputPort<KeyValPair<MerchantKey, CreditCardData>> outputPort)
   {
     MerchantKey key = getMerchantKey(tuple);
-
+    
     CreditCardData data = new CreditCardData();
     data.fullCcNum = tuple.fullCcNum;
     data.amount = tuple.amount;
-
+
```
tweise: trailing whitespace in many places

tweise:
Great to see this cleanup happening! Please split the change into 2 commits. One for the build change and the second for the formatting changes. The second commit should be attributed to the CI user so there is no confusion when questions arise WRT the logic itself. You can do that by running:

```
git commit --amend --author "MalharJenkins <jenkins@datatorrent.com>"
```

PR #383, opened 2016-08-26
APEXMALHAR-2206: Fixed application tests

bhupeshchawda:


bhupeshchawda:
@yogidevendra Please review and merge

library/src/test/java/com/datatorrent/lib/db/jdbc/JdbcIOAppTest.java
```diff
@@ -126,6 +126,7 @@ public void testApplication() throws Exception
       Thread.sleep(3000);
 
       Assert.assertEquals("Events in store", 10, getNumOfEventsInStore());
+      lc.shutdown();
```
tweise, line 129: I think it would be better to place this before the assertion. Also above test is another example of of the anti-pattern where we don't check for presence of results while waiting for arbitrary amount of time.

tweise:
@yogidevendra the Travis build still did not complete, why was this merged in a rush?

yogidevendra:
@tweise 
I observed following problems on local build which were fixed after this fix.
1. target directory size increasing indefinitely leading to disk full on local machine.
2. local build not finishing for long time.

Regarding travis build; I thought travis build is getting blocked because of the two other builds which were not yet finished. (I assumed there could be some upper cap on no. of buillds running for a project).
Local build was finished in less than 8 mins.

library/src/test/java/com/datatorrent/lib/db/jdbc/JdbcIOAppTest.java
yogidevendra: Opened #384 for fixing placement of assertion.

PR #393, opened 2016-08-30
APEXMALHAR-2207 implemented proper assertions. Removed console outputs

shubham-pathak22:
@yogidevendra Please review.

### DIFF
--- a/library/src/test/java/com/datatorrent/lib/formatter/JsonFormatterTest.java
+++ b/library/src/test/java/com/datatorrent/lib/formatter/JsonFormatterTest.java
@@ -22,11 +22,8 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintStream;
-import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
-
-import javax.validation.ConstraintViolationException;
 
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
@@ -36,22 +33,14 @@ import org.junit.Test;
 import org.junit.runner.Description;
 
 import org.apache.commons.io.FileUtils;
-import org.apache.hadoop.conf.Configuration;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.Lists;
 
-import com.datatorrent.api.Context;
-import com.datatorrent.api.DAG;
-import com.datatorrent.api.DefaultOutputPort;
-import com.datatorrent.api.InputOperator;
-import com.datatorrent.api.LocalMode;
-import com.datatorrent.api.StreamingApplication;
-import com.datatorrent.common.util.BaseOperator;
-import com.datatorrent.lib.io.ConsoleOutputOperator;
 import com.datatorrent.lib.io.fs.AbstractFileOutputOperatorTest.FSTestWatcher;
 import com.datatorrent.lib.testbench.CollectorTestSink;
+import com.datatorrent.lib.util.KryoCloneUtils;
 import com.datatorrent.lib.util.TestUtils;
 import com.datatorrent.lib.util.TestUtils.TestInfo;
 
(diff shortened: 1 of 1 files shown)


library/src/test/java/com/datatorrent/lib/formatter/JsonFormatterTest.java
```diff
@@ -198,15 +199,25 @@ public void testJSONToPOJONoFieldPOJO()
...
-      LocalMode lma = LocalMode.newInstance();
-      Configuration conf = new Configuration(false);
-      lma.prepareDAG(new JsonFormatterApplication(), conf);
-      LocalMode.Controller lc = lma.getController();
-      lc.run(2000);// runs for 2 seconds and quits
-    } catch (ConstraintViolationException e) {
-      Assert.fail("constraint violations: " + e.getConstraintViolations());
-    }
+    LocalMode lma = LocalMode.newInstance();
+    Configuration conf = new Configuration(false);
```
tweise: What is the purpose of this code? This test case should cover JsonFormatter. It looks like it is testing generic engine behavior instead?

library/src/test/java/com/datatorrent/lib/formatter/JsonFormatterTest.java
```diff
@@ -198,15 +199,25 @@ public void testJSONToPOJONoFieldPOJO()
...
+    LogicalPlan prepareDAG = (LogicalPlan)lma.prepareDAG(new JsonFormatterApplication(), conf);
+    LogicalPlan cloneDAG = (LogicalPlan)lma.cloneDAG();
+    assertEquals("Attributes", prepareDAG.getAttributes(), cloneDAG.getAttributes());
```
sandeshh: Why can't we use 

AssertTrue( prepareDAG.equals(cloneDAG)) ?

library/src/test/java/com/datatorrent/lib/formatter/JsonFormatterTest.java
yogidevendra: @shubham-pathak22 Expectation is as follows:
1. Collect the tuples from the operator under test.
2. Compare it against expected results.
3. Assert if results are matching.
4. localmode execution should wait untill output results are available instead of some arbitrary time. 

You may refer to https://github.com/apache/apex-malhar/blob/master/library/src/test/java/org/apache/apex/malhar/lib/fs/FSRecordReaderTest.java#L90-L102 for reference. 

Another example is available here: https://github.com/apache/apex-malhar/blob/master/library/src/test/java/org/apache/apex/malhar/lib/dedup/DeduperPartitioningTest.java#L190-193

shubham-pathak22: @tweise Purpose was to check for serialization of the DAG  and to check if app runs with the operator under test. One approach was to just create and run a sample application for some time. Assertions were not included in that application test as other unit tests had covered those. But the downside was that it created a lot of console outputs.
Hence this approach was followed.

In general, do we need to add a test to verify serialization of the operator ? If yes, could you advise on correct approach ?  I will modify it accordingly.

@yogidevendra Thank you for providing the examples. However, unit tests already covers it.

shubham-pathak22: Will add a unit test to test serialization of just the operator and update.

shubham-pathak22:
@tweise Replaced the application test with a test to check just the operator serialization.

library/src/test/java/com/datatorrent/lib/formatter/JsonFormatterTest.java
```diff
@@ -196,54 +187,16 @@ public void testJSONToPOJONoFieldPOJO()
...
-      test1Pojo.sizes.add("250x350");
-      test1Pojo.sizes.add("800x600");
-      test1Pojo.startDate = new DateTime().withDate(2016, 1, 1).withHourOfDay(0).withMinuteOfHour(0)
-          .withSecondOfMinute(0).toDate();
-      test1Pojo.endDate = new DateTime().withDate(2016, 2, 1).withHourOfDay(0).withMinuteOfHour(0).withSecondOfMinute(0)
-          .withZone(DateTimeZone.UTC).toDate();
-      output.emit(test1Pojo);
-    }
+    Kryo kryo = new Kryo();
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
tweise: You can use KryoCloneUtils

shubham-pathak22: Agreed. Done.

PR #396, opened 2016-08-31
APEXMALHAR-2218-Creation of RegexSplitter operator

venkateshkottapalli:
APEXMALHAR-2218-Creation of RegexSplitter operator

@ashwinchandrap Please review.

### DIFF
--- a/contrib/src/main/java/com/datatorrent/contrib/parser/RegexParser.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/parser/RegexParser.java
@@ -1,0 +1,234 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.contrib.parser;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.util.Date;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import javax.validation.constraints.NotNull;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.commons.beanutils.BeanUtils;
+import org.apache.commons.beanutils.ConversionException;
+import org.apache.commons.beanutils.ConvertUtils;
+import org.apache.commons.beanutils.converters.DateConverter;
+import org.apache.commons.beanutils.converters.DateTimeConverter;
+import org.apache.commons.lang.StringUtils;
+import org.apache.hadoop.classification.InterfaceStability;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import com.datatorrent.api.Context;
+import com.datatorrent.lib.parser.Parser;
+import com.datatorrent.lib.util.KeyValPair;
+
+/**
+ * Operator parses tuple based on regex pattern and populates POJO matching the user defined schema <br>
+ * This operator expects the upstream operator to send every line in the file as a byte array.
+ * splitRegexPattern contains the regex pattern of lines in the file <br>
+ * Schema is specified in a json format as per {@link DelimitedSchema} that
+ * contains field information and constraints for each field.<br>
+ * Schema field names should match with the POJO variable names<br>
+ * Assumption is that each field in the delimited data should map to a simple
+ * java type.<br>
+ * <br>
+ * <b>Properties</b> <br>
+ * <b>splitRegexPattern</b>:Regex pattern as a string<br>
+ *
+ * @displayName RegexParser
+ * @category Parsers
+ * @tags pojo parser regex logs server
+ * @since 3.7.0
+ */
+@InterfaceStability.Evolving
+public class RegexParser extends Parser<byte[], KeyValPair<String, String>>
+{
+  /**
+   * Contents of the schema.Schema is specified in a json format as per
+   * {@link DelimitedSchema}
+   */
+  @NotNull
+  private String schema;
+  /**
+   * Schema is read into this object to access fields
+   */
+  private transient DelimitedSchema delimitedParserSchema;
+  /**
+   * Regex Pattern defined for the tuple
+   */
+  @NotNull
+  private String splitRegexPattern;
+  /**
+   * Pattern to store the compiled regex
+   */
+  private transient Pattern pattern;
+
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    delimitedParserSchema = new DelimitedSchema(schema);
+    pattern = Pattern.compile(splitRegexPattern);
+  }
+
+  @Override
+  public void processTuple(byte[] tuple)
+  {
+    if (tuple == null) {
+      if (err.isConnected()) {
+        err.emit(new KeyValPair<String, String>(null, "Blank/null tuple"));
+      }
+      errorTupleCount++;
+      return;
+    }
+    String incomingString = new String(tuple);
+    if (StringUtils.isBlank(incomingString)) {
+      if (err.isConnected()) {
+        err.emit(new KeyValPair<String, String>(incomingString, "Blank tuple"));
+      }
+      errorTupleCount++;
+      return;
+    }
+    try {
+      if (out.isConnected() && clazz != null) {
+        Matcher matcher = pattern.matcher(incomingString);
+        boolean patternMatched = false;
+        Constructor<?> ctor = clazz.getConstructor();
+        Object object = ctor.newInstance();
+
+        if (matcher.find()) {
+          for (int i = 0; i <= matcher.groupCount()-1; i++) {
+            if (delimitedParserSchema.getFields().get(i).getType() == DelimitedSchema.FieldType.DATE) {
+              DateTimeConverter dtConverter = new DateConverter();
+              dtConverter.setPattern((String)delimitedParserSchema.getFields().get(i).getConstraints().get(DelimitedSchema.DATE_FORMAT));
+              ConvertUtils.register(dtConverter, Date.class);
+            }
+            BeanUtils.setProperty(object, delimitedParserSchema.getFields().get(i).getName(), matcher.group(i+1));
+          }
+          patternMatched = true;
+        }
+        if (!patternMatched) {
+          throw new ConversionException("The incoming tuple do not match with the Regex pattern defined.");
+        }
+
+        out.emit(object);
+        emittedObjectCount++;
+      }
+
+    } catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException | InstantiationException | ConversionException e) {
+      if (err.isConnected()) {
+        err.emit(new KeyValPair<String, String>(incomingString, e.getMessage()));
+        logger.debug("Regex Expression : {} Incoming tuple : {}", splitRegexPattern, incomingString);
+      }
+      errorTupleCount++;
+      logger.error("Tuple could not be parsed. Reason {}", e.getMessage());
+    }
+  }
+
+  /**
+   * Set the schema that defines the format of the tuple
+   *
+   * @param schema
+   */
+  public void setSchema(String schema)
+  {
+    this.schema = schema;
+  }
+
+  /**
+   * Set the Regex Pattern expected for the incoming tuple
+   *
+   * @param splitRegexPattern
+   */
+  public void setSplitRegexPattern(String splitRegexPattern)
+  {
+    this.splitRegexPattern = splitRegexPattern;
+  }
+
+  /**
+   * Get the schema value
+   *
+   * @return schema
+   */
+  public String getSchema()
+  {
+    return schema;
+  }
+
+  /**
+   * Get the Regex Pattern value
+   *
+   * @return splitRegexPattern
+   */
+  public String getSplitRegexPattern()
+  {
+    return splitRegexPattern;
+  }
+
+  @Override
+  public Object convert(byte[] tuple)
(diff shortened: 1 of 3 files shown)


contrib/src/main/java/com/datatorrent/contrib/parser/RegexSplitter.java
```diff
@@ -0,0 +1,129 @@
...
+ * <br>
+ * <b>Properties</b> <br>
+ * <b>splitRegexPattern</b>:Regex pattern as a string<br>
+ *
+ * @displayName RegexSplitter
+ * @category Parsers
+ * @tags pojo parser regex logs server
+ * @since 3.6.0
+ */
+public class RegexSplitter extends CsvParser
```
shubham-pathak22: Any specific reason to extend from CsvParser ?  I feel RegexSplitter is not a "special case" of Csv Parser. It should extend from Parser instead of CsvParser.

contrib/src/main/java/com/datatorrent/contrib/parser/RegexSplitter.java
```diff
@@ -0,0 +1,129 @@
...
+{
+  @NotNull
+  private  String splitRegexPattern;
```
shubham-pathak22: Javadoc. In general please add JavaDocs for field members and getters and setters.

shubham-pathak22:
Would adding another port just emitting tokens after the split be useful ?  That way this operator could also be used as a tokenizer

contrib/src/test/java/com/datatorrent/contrib/parser/RegexSplitterTest.java
```diff
@@ -0,0 +1,167 @@
...
+  @Test
+  public void testNullInput() throws JSONException
+  {
+    regex.beginWindow(0);
+    regex.in.process(null);
+    regex.endWindow();
+    Assert.assertEquals(0, pojoPort.collectedTuples.size());
+    Assert.assertEquals(1, error.collectedTuples.size());
+  }
+
```
shubham-pathak22: You may add tests to verify values of counters ( autometrics ) as well.

contrib/src/main/java/com/datatorrent/contrib/parser/RegexSplitter.java
shubham-pathak22: Mark the operator as @InterfaceStability.Evolving

venkateshkottapalli: Agree. Will make the changes suggested.

contrib/src/main/java/com/datatorrent/contrib/parser/CsvParser.java
```diff
@@ -65,7 +65,7 @@
  * <b>err</b>:tuples that do not confine to schema are emitted on this port as
  * KeyValPair<String,String><br>
  * Key being the tuple and Val being the reason.
- * 
+ *
```
shubham-pathak22: Please undo these changes. Spaces have been introduced.

contrib/src/main/java/com/datatorrent/contrib/parser/RegexSplitter.java
```diff
@@ -0,0 +1,214 @@
...
+      if (err.isConnected()) {
+        err.emit(new KeyValPair<String, String>(null, "Blank/null tuple"));
+      }
+      errorTupleCount++;
+      return;
+    }
+    String incomingString = new String(tuple);
+    if (StringUtils.isBlank(incomingString)) {
+      if (err.isConnected()) {
+        err.emit(new KeyValPair<String, String>(incomingString, "Blank/header tuple"));
```
shubham-pathak22: Error message could just be "Blank tuple" as the check is for blank string

contrib/src/main/java/com/datatorrent/contrib/parser/RegexSplitter.java
```diff
@@ -0,0 +1,214 @@
...
+  /**
+   * Set the Regex Pattern expected for the incoming tuple
+   *
+   * @param splitRegexPattern
+   */
+  public void setSplitRegexPattern(String splitRegexPattern)
+  {
+    this.splitRegexPattern = splitRegexPattern;
+  }
+
```
shubham-pathak22: add getters for both schema and splitRegexPattern

contrib/src/main/java/com/datatorrent/contrib/parser/RegexSplitter.java
```diff
@@ -0,0 +1,234 @@
...
+  @Override
+  public void processTuple(byte[] tuple)
+  {
+    if (tuple == null) {
+      if (err.isConnected()) {
+        err.emit(new KeyValPair<String, String>(null, "Blank/null tuple"));
+      }
+      errorTupleCount++;
```
shubham-pathak22: Where are the counters being reset ? You might need to call super.beginWindow().
In addition to it please add a test to check whether counters are being reset for each window. You may refer to 
https://github.com/apache/apex-malhar/blob/master/contrib/src/test/java/com/datatorrent/contrib/parser/JsonParserTest.java

shubham-pathak22:
There are build failures because of missing headers. Good practice is to run this command before pushing in the changes
mvn apache-rat:check verify -Dlicense.skip=false -Dtravis=true -Dtest.log.console.threshold=INFO -Pall-modules -B -V

contrib/src/test/java/com/datatorrent/contrib/parser/RegexSplitterTest.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    public void setDate(Date date)
+    {
+      this.date = date;
+    }
+
+  }
+
+  private static final Logger logger = LoggerFactory.getLogger(RegexSplitterTest.class);
```
shubham-pathak22: You may remove the logger since it is not being used

contrib/src/main/java/com/datatorrent/contrib/parser/RegexSplitter.java
venkateshkottapalli: Counters are reset in the Parser beginWindow(). It gets called since we don't have an implementation in RegexSplitter.

I will add test cases to check that like you suggested.

contrib/src/test/java/com/datatorrent/contrib/parser/RegexSplitterTest.java
venkateshkottapalli: Sure.

shubham-pathak22:
@venkateshkottapalli  The changes look good to me. Will merge in a day if there are no further comments.

venkateshkottapalli:
Travis build failing.. Will close and reopen.

venkateshkottapalli:
APEXMALHAR-2218-Creation of RegexSplitter operator - Reopen

contrib/src/main/java/com/datatorrent/contrib/parser/RegexSplitter.java
```diff
@@ -0,0 +1,234 @@
...
+ * <b>Properties</b> <br>
+ * <b>splitRegexPattern</b>:Regex pattern as a string<br>
+ *
+ * @displayName RegexSplitter
+ * @category Parsers
+ * @tags pojo parser regex logs server
+ * @since 3.6.0
+ */
+@InterfaceStability.Evolving
+public class RegexSplitter extends Parser<byte[], KeyValPair<String, String>>
```
bhupeshchawda: Can this be called something like RegexParser?
RegexSplitter does not intuitively seem like some parser. It looks like a type of `FilterOperator`, but one that operates on raw data.

venkateshkottapalli: @bhupeshchawda Agreed, RegexParser looks more appropriate.

contrib/src/main/java/com/datatorrent/contrib/parser/RegexParser.java
```diff
@@ -0,0 +1,234 @@
...
+      return;
+    }
+    try {
+      if (out.isConnected() && clazz != null) {
+        Matcher matcher = pattern.matcher(incomingString);
+        boolean patternMatched = false;
+        Constructor<?> ctor = clazz.getConstructor();
+        Object object = ctor.newInstance();
+
+        while (matcher.find()) {
```
devtagare: Why is this not a if condition

contrib/src/main/java/com/datatorrent/contrib/parser/RegexParser.java
```diff
@@ -0,0 +1,234 @@
...
+          for (int i = 1; i <= matcher.groupCount(); i++) {
+            if (delimitedParserSchema.getFields().get(i - 1).getType() == DelimitedSchema.FieldType.DATE) {
+              DateTimeConverter dtConverter = new DateConverter();
+              dtConverter.setPattern((String)delimitedParserSchema.getFields().get(i - 1).getConstraints().get(DelimitedSchema.DATE_FORMAT));
+              ConvertUtils.register(dtConverter, Date.class);
+            }
+            BeanUtils.setProperty(object, delimitedParserSchema.getFields().get(i - 1).getName(), matcher.group(i));
```
devtagare: Use PojoUtils instead.BeanUtils uses reflection

contrib/src/main/java/com/datatorrent/contrib/parser/RegexParser.java
```diff
@@ -0,0 +1,234 @@
...
+          }
+          patternMatched = true;
+        }
+        if (!patternMatched) {
+          throw new ConversionException("The incoming tuple do not match with the Regex pattern defined.");
```
devtagare: Please dont throw an exception here emit to error port

contrib/src/main/java/com/datatorrent/contrib/parser/RegexParser.java
```diff
@@ -0,0 +1,234 @@
...
+import com.datatorrent.lib.parser.Parser;
+import com.datatorrent.lib.util.KeyValPair;
+
+/**
+ * Operator parsers the tuple based on regex pattern and populates POJO matching the user defined schema <br>
+ * This operator expects the upstream operator to send every line in the file as a byte array.
+ * splitRegexPattern contains the regex pattern of lines in the file <br>
+ * Schema is specified in a json format as per {@link DelimitedSchema} that
+ * contains field information and constraints for each field.<br>
+ * Schema field names should match with the POJO variable names<br>
```
devtagare, line 52: Why is this condition enforced.You can use https://github.com/apache/apex-malhar/blob/master/library/src/main/java/com/datatorrent/lib/util/FieldInfo.java to specify input to output mappings

contrib/src/test/java/com/datatorrent/contrib/parser/RegexParserTest.java
```diff
@@ -0,0 +1,261 @@
...
+import org.junit.runner.Description;
+
+import com.datatorrent.lib.appdata.schemas.SchemaUtils;
+import com.datatorrent.lib.testbench.CollectorTestSink;
+import com.datatorrent.lib.util.KeyValPair;
+
+public class RegexParserTest
+{
+  RegexParser regex = new RegexParser();
+  private CollectorTestSink<Object> error = new CollectorTestSink<Object>();
```
devtagare, line 39: Pleas add an application level test

contrib/src/test/resources/RegexSplitterschema.json
```diff
@@ -0,0 +1,22 @@
+{
```
devtagare, line 1: Does this operator need a CSV like schema definition ? How is this operator using a separator and quouteChar for parsing inputs ?

venkateshkottapalli: @devtagare It doesn't need the delimiters in schema definition, will remove them from the test schema.

contrib/src/main/java/com/datatorrent/contrib/parser/RegexParser.java
venkateshkottapalli: @devtagare Thanks for suggesting this and letting me know. Adding an additional mapping also makes sense but this may not be needed as the user has control over both schema and POJO.  If the schema has say 50 or more fields, it gets tedious to set these fields, we need not make it mandatory though by adding additional checks. Let me know your thoughts and did we see any issues with this type of constraint before?

contrib/src/main/java/com/datatorrent/contrib/parser/RegexParser.java
venkateshkottapalli: Both if and while serve the same purpose here, but using if is appropriate, will change it.

contrib/src/test/java/com/datatorrent/contrib/parser/RegexParserTest.java
venkateshkottapalli: @devtagare any reason for an application test in Unit tests?

sandeshh:
What is the status of this PR?

venkateshkottapalli:
I will incorporate the changes in a couple of days Shubham.. I am creating
an example to see the bottlenecks.

-Venky.

contrib/src/main/java/com/datatorrent/contrib/parser/RegexParser.java
venkateshkottapalli: PojoUtils doesn't support Date fields. Using PojoUtils for the rest and Reflection for Date can be tried, if that is feasible solution I will track it as a different issue. At present, a single instance of an operator is able to process 32K records per second with 2 strings, 1 long and 1 Date field. The test case has the String and the regex pattern example.

contrib/src/main/java/com/datatorrent/contrib/parser/RegexParser.java
venkateshkottapalli: There is no input to output mapping here as in the case of Json.  Schema field names  define the structure of the String but will not be present in data. Please look at the test case for sample data.

contrib/src/main/java/com/datatorrent/contrib/parser/RegexParser.java
shubham-pathak22: Change this to "if" as suggested by @devtagare

shubham-pathak22:
@venkateshkottapalli Looks good to me. If there are no further comments and once the changes are incorporated,  please rebase and make sure the build is successful. Will merge the PR then.

contrib/src/main/java/com/datatorrent/contrib/parser/RegexParser.java
venkateshkottapalli: @shubham-pathak22 Done with the change.

venkateshkottapalli:
Build failing.. will close and reopen!

PR #401, opened 2016-09-07
APEXMALHAR-2226 Fixed the Not supported exception while re-deploying the AbstractFileOutput Operator

chaithu14:


library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
```diff
@@ -636,7 +638,7 @@ protected FSDataOutputStream openStream(Path filepath, boolean append) throws IO
   {
     FSDataOutputStream fsOutput;
     if (append) {
-      fsOutput = fs.append(filepath);
+      return openStreamForNonAppendFS(filepath);
```
yogidevendra: How about 
fsOutput = openStreamForNonAppendFS(filepath);
to keep it consistent with the original one.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
```diff
@@ -645,6 +647,46 @@ protected FSDataOutputStream openStream(Path filepath, boolean append) throws IO
...
+   * @param filepath given file path
+   * @return output stream
+   */
+  protected FSDataOutputStream openStreamInAppendMode(Path filepath)
+  {
+    FSDataOutputStream fsOutput = null;
+    try {
+      fsOutput = fs.append(filepath);
+    } catch (IOException e) {
+      if (e.getMessage().contains("Not supported")) {
```
yogidevendra: What is output of e.getMessage() in this case? Can we make it e.getMessage().equals instead of contains?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
```diff
@@ -645,6 +647,46 @@ protected FSDataOutputStream openStream(Path filepath, boolean append) throws IO
...
+  protected FSDataOutputStream openStreamForNonAppendFS(Path filepath)
+  {
+    try {
+      Path appendTmpFile = new Path(filepath + APPEND_TMP_FILE);
+      fs.rename(filepath, appendTmpFile);
+      FSDataInputStream fsIn = fs.open(appendTmpFile);
+      FSDataOutputStream fsOut = fs.create(filepath);
+      IOUtils.copy(fsIn, fsOut);
+      flush(fsOut);
+      fs.delete(appendTmpFile);
```
yogidevendra, line 683: Do we need to close fsIn before delete?

yogidevendra (changes requested):
Added my feedback for the changes. Otherwise looks OK.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
chaithu14: Error message consists of very huge contents. I think contains() is the best option.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
chaithu14: Okay.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
chaithu14: Yes. We need to close stream before delete.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
chaithu14: Sorry, my bad. Error message contains "Not Supported" only. I changed to equals() instead of contains.

chaithu14:
@yogidevendra Incorporated review comments and Updated the branch.

yogidevendra:
Looks OK to me. Would wait for 1 day for any other comments from community else merge it.

PR #402, opened 2016-09-07
APEXMALHAR-2227 Index out of Bound Exception fix in Kafka input

otterc:
@siyuanh @tweise


siyuanh:
@chandnisingh The test fails because of FileNotFoundException which is related to your FSWindowManager change. Would you please take a look? Thanks

otterc:
@siyuanh which test fails?

otterc:
@siyuanh fixed the issue.

tweise:
Tested with 

```
cd contrib
mvn clean test -Dtest=Kafka* -DskipTests=false
```

tweise:
@siyuanh can you please check that the CI covers this for the 0.9 connector?

PR #405, opened 2016-09-08
APEXMALHAR-2130 #resolve Added a spillable map that takes two keys with support of iterating through all entries with a given first key

davidyan74:
@siyuanh @tweise @ilooner Please review. This is mostly for the scalable implementation of a WindowedKeyedStorage.

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/spillable/Spillable.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/spillable/Spillable.java
@@ -18,9 +18,12 @@
  */
 package org.apache.apex.malhar.lib.state.spillable;
 
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
+
+import org.apache.commons.lang3.tuple.Pair;
 
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Multiset;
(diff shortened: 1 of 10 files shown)


ilooner:
@davidyan74 

My concern about SpillableTwoKeyMap is that you have a SpillableByteMap where the value is a Set. If the Set is large for each key every time you do an update you will do a large write, or every time you need to check the contents of the set you need to do a large read. I think this could be avoided if we provided a SpillableSet Implementation and an implementation of SetMultiMap https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/SetMultimap.html

The implementation of SetMultiMap would be very similar to the existing SpillableArrayListMultimap

The implementation of a SpillableSet would also be straightforward. We could just use a SpillableIterableByteMap and keep a marker byte as a value for keys that exist in the set.

davidyan74:
@ilooner  Thanks for your feedback. Yes, I was also thinking about the same problem for reading and writing a large key set for each window. But I wanted to leave that for the next iteration since I could not think of a straightforward way to implement a SpillableSet. How would you implement the support of returning all elements (or an iterator over all elements) of a SpillableSet?

davidyan74:
@ilooner By the way, #403 will support iterating over entries in a TFile. However, it would not be sufficient for this use case as it only works for flushed entries. We would have to change ManagedState to use a TreeMap instead of a HashMap for the in-memory cache to support this kind of iteration. That's the reason I wanted to hold off until the next iteration.

ilooner:
I see that makes sense. Well since I just saw your comment after I wrote all this up, here is a sketch of how I would do it without those TFile feature just for fun :)

We could implement an Iterator For a SpillableSet by having a Spillable linked list

The SpillableLinked list could be made with two SpillableByteMaps

SpillableMap1 holds the forward links like so:

1 -> 2
2 -> 3
4 -> 5

SpillableMap2 holds the back links like so:

2 -> 1
3 -> 2
4 -> 5

Then you do add remove and iteration over the links like any doubly linked list.

Once you have the SpillableLinked list your SpillableSet would internally have a linked list. You could check existence of a value in the set by checking if it is a key in SpillableMap1 or SpillableMap2 and iterate over the set by iterating over the linked list.

davidyan74:
@ilooner This would work. Thanks! @siyuanh @tweise What do you guys think?

PR #407, opened 2016-09-11
APEXMALHAR-2130 WindowStorage interface changes in preparation of inc…

davidyan74:
…orporating spillable data structures

@tweise @ShunxinLu FYI


library/src/main/java/org/apache/apex/malhar/lib/window/Tuple.java
```diff
@@ -45,11 +45,12 @@
    */
   class PlainTuple<T> implements Tuple<T>
   {
-    private T value;
```
siyuanh, line 48: Why change this to final? What if I have a stateless transform in between 2 windowed operator and I want to change the value of the tuple. Then I have to recreate the Tuple instance

library/src/main/java/org/apache/apex/malhar/lib/window/Window.java
```diff
@@ -148,20 +136,49 @@ public boolean equals(Object other)
...
+      } else if (this.getDurationMillis() > o.getDurationMillis()) {
+        return 1;
+      }
+      return 0;
+    }
+
+    @Override
+    public String toString()
+    {
+      return "[TimeWindow " + getBeginTimestamp() + "(" + getDurationMillis() + ")]";
```
siyuanh, line 164: It is better to be in json formatter

library/src/main/java/org/apache/apex/malhar/lib/window/WindowedStorage.java
```diff
@@ -53,41 +50,87 @@
...
+     * @return
+     */
+    T get(Window window);
+
+    /**
+     * Returns the iterable of the entries in the storage
+     *
+     * @return
+     */
+    Iterable<Map.Entry<Window, T>> entries();
```
siyuanh, line 89: This method always return itself?

siyuanh, line 89: Actually I'd rather change the method to allWindows and remove Iterable from class signature. It's hard to understand a WindowedPlainStorage is iterable.

library/src/main/java/org/apache/apex/malhar/lib/window/WindowedStorage.java
```diff
@@ -53,41 +50,87 @@
...
+     */
+    void put(Window window, K key, V value);
+
+    /**
+     * Gets the iterator over the key/value pairs associated with the given window
+     *
+     * @param window
+     * @return
+     */
+    Iterator<Map.Entry<K, V>> iterator(Window window);
```
siyuanh: what's the difference between iterator and entries?

library/src/main/java/org/apache/apex/malhar/lib/window/WindowedStorage.java
davidyan74: @siyuanh  makes sense. It's updated.

library/src/main/java/org/apache/apex/malhar/lib/window/WindowedStorage.java
davidyan74: @siyuanh removed iterator

library/src/main/java/org/apache/apex/malhar/lib/window/Tuple.java
davidyan74: @siyuanh you have a point. @tweise @vrozov do you guys know of any value in making tuples immutable?

library/src/main/java/org/apache/apex/malhar/lib/window/Window.java
davidyan74: I agree but that's not what the convention is in the rest of apex code. This is more of a convenience for logging so it's better to make it a single line string, even though we can make json one line as well.

library/src/main/java/org/apache/apex/malhar/lib/window/Tuple.java
davidyan74: @siyuanh  I have reverted this change for now.

davidyan74:
@siyuanh please take a look at this again and see if you have any more issues before this can be merged

library/src/main/java/org/apache/apex/malhar/lib/window/Window.java
```diff
@@ -148,20 +136,49 @@ public boolean equals(Object other)
...
+    }
   }
 
   /**
    * SessionWindow is a window that represents a time slice for a key, with the time slice being variable length.
    *
    * @param <K>
    */
-  class SessionWindow<K> extends TimeWindow
+  class SessionWindow<K> extends TimeWindow<SessionWindow<K>>
```
siyuanh, line 173: javadoc for compareTo here

library/src/main/java/org/apache/apex/malhar/lib/window/Window.java
```diff
@@ -59,54 +68,33 @@ public boolean equals(Object other)
...
-  /**
-   * The singleton default comparator of windows
-   */
-  Comparator<Window> DEFAULT_COMPARATOR = new DefaultComparator();
-
-  /**
    * TimeWindow is a window that represents a time slice
    */
-  class TimeWindow implements Window
+  class TimeWindow<WINDOW extends TimeWindow> implements Window<WINDOW>
```
siyuanh: javadoc for compareTo

library/src/main/java/org/apache/apex/malhar/lib/window/Window.java
siyuanh: override in SlidingTimedWindow

library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
```diff
@@ -389,8 +397,32 @@ public void processWatermark(ControlTuple.Watermark watermark)
   @Override
   public void setup(Context.OperatorContext context)
   {
-    this.windowWidthMillis = context.getValue(Context.DAGContext.STREAMING_WINDOW_SIZE_MILLIS);
+    this.timeIncrement = context.getValue(Context.OperatorContext.APPLICATION_WINDOW_COUNT) * context.getValue(Context.DAGContext.STREAMING_WINDOW_SIZE_MILLIS);
     validate();
+    windowStateMap.setup(context);
```
siyuanh, line 402: I'm thinking if we could provide an automatic way to call these method by set annotation on the field just like those IoC frameworks do

library/src/main/java/org/apache/apex/malhar/lib/window/impl/InMemorySessionWindowedStorage.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+
+  @Override
+  public Collection<Map.Entry<Window.SessionWindow<K>, V>> getSessionEntries(K key, long timestamp, long gap)
+  {
+    List<Map.Entry<Window.SessionWindow<K>, V>> results = new ArrayList<>();
+    TreeSet<Window.SessionWindow<K>> sessionWindows = keyToWindows.get(key);
+    if (sessionWindows != null) {
+      for (Window.SessionWindow<K> window : sessionWindows) {
```
siyuanh: This might be resolved in O(logn) time

library/src/main/java/org/apache/apex/malhar/lib/window/Tuple.java
vrozov: It depends on how PlainTuple is used later. If it is used as a key, making PlainTuple immutable may be helpful. But I guess, T itself may be used as a key, instead of PlainTuple. In cases mutable or immutable use case becomes common, we may introduce (Im)MutablePlainTuple.

library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
davidyan74: not the scope of this PR :)

library/src/main/java/org/apache/apex/malhar/lib/window/Window.java
davidyan74: @siyuanh done

library/src/main/java/org/apache/apex/malhar/lib/window/impl/InMemorySessionWindowedStorage.java
davidyan74: @siyuanh fixed

library/src/main/java/org/apache/apex/malhar/lib/window/Window.java
davidyan74: WindowOption.SlidingTimeWindows generates Window.TimeWindow itself. There is no class that represents individual sliding time windows.

siyuanh:
@davidyan74 @tweise @ShunxinLu I have no more questions about the pull request. If you guys are ok with the change. I will go ahead and merge it.

demos/highlevelapi/src/main/java/org/apache/apex/malhar/stream/sample/complete/TwitterAutoComplete.java
```diff
@@ -125,7 +125,7 @@ public ComputeTopFlat(int candidatesPerPrefix, int minPrefix)
           public Tuple<KeyValPair<String, CompletionCandidate>> f(KeyValPair<String, CompletionCandidate> tuple)
           {
             // TODO: Should be removed after Auto-wrapping is supported.
-            return new Tuple.WindowedTuple<>(Window.GLOBAL_WINDOW, tuple);
+            return new Tuple.WindowedTuple<>(Window.GlobalWindow.getInstance(), tuple);
```
tweise: Why this change?

library/src/main/java/org/apache/apex/malhar/lib/window/Window.java
```diff
@@ -18,31 +18,42 @@
...
+    /**
+     * The singleton global window
+     */
+    private static GlobalWindow GLOBAL_WINDOW = new GlobalWindow();
+
     private GlobalWindow()
     {
     }
 
+    public static GlobalWindow getInstance()
```
tweise: This is not needed, rename GLOBAL_WINDOW to INSTANCE and make it public.

siyuanh: I'm also curious about this change and to me. And also the equals method can use == instead of instanceof which I believe would be slightly faster

davidyan74:
all comments addressed. @siyuanh  please merge if you see no other issues.

library/src/main/java/org/apache/apex/malhar/lib/window/Window.java
```diff
@@ -18,27 +18,33 @@
...
+   * Global window means there is only one window that spans the entire life time of the application
    */
-  class GlobalWindow implements Window
+  class GlobalWindow implements Window<GlobalWindow>
   {
+
+    /**
+     * The singleton global window
+     */
+    public static GlobalWindow INSTANCE = new GlobalWindow();
```
siyuanh: Should be final

davidyan74: @siyuanh added

tweise:
@davidyan74 LGTM please rebase.

PR #408, opened 2016-09-12
APEXMALHAR-2248 Added SpillableSet and SpillableSetMultimap implementation

davidyan74:
@tweise @siyuanh @ilooner please review

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableSetImpl.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableSetImpl.java
@@ -1,0 +1,352 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.lib.state.spillable;
+
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+
+import javax.validation.constraints.NotNull;
+
+import org.apache.apex.malhar.lib.utils.serde.Serde;
+import org.apache.commons.lang3.mutable.MutableInt;
+import org.apache.hadoop.classification.InterfaceStability;
+
+import com.esotericsoftware.kryo.DefaultSerializer;
+import com.esotericsoftware.kryo.serializers.FieldSerializer;
+import com.google.common.base.Preconditions;
+
+import com.datatorrent.api.Context;
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * A Spillable implementation of {@link List} backed by a {@link SpillableStateStore}.
+ * @param <T> The type of object stored in the {@link SpillableSetImpl}.
+ *
+ * @since 3.5.0
+ */
+@DefaultSerializer(FieldSerializer.class)
+@InterfaceStability.Evolving
+public class SpillableSetImpl<T> implements Spillable.SpillableSet<T>, Spillable.SpillableComponent
+{
+  private static class ListNode<T>
+  {
+    ListNode()
+    {
+    }
+
+    ListNode(boolean valid, T next)
+    {
+      this.valid = valid;
+      this.next = next;
+    }
+
+    boolean valid;
+    T next;
+  }
+
+  public static class SerdeListNodeSlice<T> implements Serde<ListNode<T>, Slice>
+  {
+    private Serde<T, Slice> serde;
+    private static Slice falseSlice = new Slice(new byte[]{0});
+    private static Slice trueSlice = new Slice(new byte[]{1});
+
+    public SerdeListNodeSlice(@NotNull Serde<T, Slice> serde)
+    {
+      this.serde = Preconditions.checkNotNull(serde);
+    }
+
+    @Override
+    public Slice serialize(ListNode<T> object)
+    {
+      int size = 0;
+
+      Slice slice1 = object.valid ? trueSlice : falseSlice;
+      size += 1;
+      Slice slice2 = serde.serialize(object.next);
+      size += slice2.length;
+
+      byte[] bytes = new byte[size];
+      System.arraycopy(slice1.buffer, slice1.offset, bytes, 0, slice1.length);
+      System.arraycopy(slice2.buffer, slice2.offset, bytes, slice1.length, slice2.length);
+
+      return new Slice(bytes);
+    }
+
+    @Override
+    public ListNode<T> deserialize(Slice slice, MutableInt offset)
+    {
+      ListNode<T> result = new ListNode<>();
+      result.valid = slice.buffer[offset.intValue()] != 0;
+      offset.add(1);
+      result.next = serde.deserialize(slice, offset);
+      return result;
+    }
+
+    @Override
+    public ListNode<T> deserialize(Slice object)
+    {
+      return deserialize(object, new MutableInt(0));
+    }
+  }
+
+  @NotNull
+  private SpillableStateStore store;
+  @NotNull
+  private SpillableByteMapImpl<T, ListNode<T>> map;
+
+  private T head;
+  private int size;
+
+  private SpillableSetImpl()
+  {
+    //for kryo
+  }
+
+  public SpillableStateStore getStore()
+  {
+    return store;
+  }
+
+  /**
+   * Creates a {@link SpillableSetImpl}.
+   * @param bucketId The Id of the bucket used to store this
+   * {@link SpillableSetImpl} in the provided {@link SpillableStateStore}.
+   * @param prefix The Id of this {@link SpillableSetImpl}.
+   * @param store The {@link SpillableStateStore} in which to spill to.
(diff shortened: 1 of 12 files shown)


siyuanh:
Overall looks good. But I have some objection here about the design especially the class hierarchy is apparently over-designed to me. I think the intermediate interface is useless. For example interface SpillableArrayList just extends List. It has no method, and if it is used as label, we already have Spillable and SpillableComponent.  What I'm thinking it cut off this layer, so the SpillableArrayListImpl would directly implements SpillableComponent and List.

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableSetMultimapImpl.java
```diff
@@ -0,0 +1,302 @@
...
+@DefaultSerializer(FieldSerializer.class)
+@InterfaceStability.Evolving
+public class SpillableSetMultimapImpl<K, V> implements Spillable.SpillableSetMultimap<K, V>,
+    Spillable.SpillableComponent
+{
+  public static final int DEFAULT_BATCH_SIZE = 1000;
+  public static final byte[] SIZE_KEY_SUFFIX = new byte[]{(byte)0, (byte)0, (byte)0};
+
+  private transient WindowBoundedMapCache<K, SpillableSetImpl<V>> cache = new WindowBoundedMapCache<>();
+  private transient boolean isRunning = false;
```
siyuanh: What are isRunning and isInWindow for?

library/src/main/java/org/apache/apex/malhar/lib/utils/serde/SerdeCollectionSlice.java
```diff
@@ -37,35 +37,40 @@
  * @since 3.5.0
  */
 @InterfaceStability.Evolving
-public class SerdeListSlice<T> implements Serde<List<T>, Slice>
+public class SerdeCollectionSlice<T, CollectionT extends Collection<T>> implements Serde<CollectionT, Slice>
```
siyuanh, line 40: Is it better to have Collection<T> in Serde?

library/src/main/java/org/apache/apex/malhar/lib/utils/serde/SerdeCollectionSlice.java
```diff
@@ -37,35 +37,40 @@
...
   {
     this.serde = Preconditions.checkNotNull(serde);
+    this.collectionClass = Preconditions.checkNotNull(collectionClass);
   }
 
   @Override
-  public Slice serialize(List<T> objects)
+  public Slice serialize(CollectionT objects)
   {
     Slice[] slices = new Slice[objects.size()];
```
siyuanh, line 66: Probably not in this PR. But I'm thinking about some special class like ArraySlice which holds buffer and array of offsets in one object.  But I don't know how it will fit into this method.

siyuanh, line 66: But I think this should be useful in many places

otterc:
@siyuanh dis-agree with the comment on Spillable class hierarchy. It is not over-designed. Instead it makes it easier for developer to find out all the different implementations for a specific datastructure.

We can have multiple SpillableList implementations. While writing an operator which needs a spillable list, it is easier for me to find all the existing implementations  that are of type SpillableList.

ilooner:
@siyuanh a benefit to having the interfaces that way is that we can easily add methods to a SpillableDataStructure signature. Yes there are no methods in any of those interfaces now, but there may be different methods added for different datastructures in the future. Furthermore it is useful to document the contract of each datastructure and all its implementations, which may differ from their typical in memory cousins. Ultimately I think it makes future modification and documentation easier, while preserving readability.

library/src/main/java/org/apache/apex/malhar/lib/utils/serde/SerdeCollectionSlice.java
davidyan74: The serde is for any collection, not just a List. That's why I made this change, since I need a serde for a Set and the same serde works for a Set as well.

siyuanh:
@ilooner I agree we need a interface for all/or at least a group of common Spillable data structures. It like in Java we have List/ OrderedList interface but we don't have ArrayList and TreeList interface. You can also argue they have different methods internally, but that wouldn't be in the interface level. Interface delivers another level of abstraction in my mind and my understanding of java library. To the name  SpillableArrayList and SpillableByteArrayListMultimap are already very very specific I don't see any place would have something called ArrayList interface and/or ByteArrayListMultimap interface.

siyuanh:
@chandnisingh Your point is a voting question, whether you think it's easier or I think it's easier doesn't mean anything. Let's discuss this in mailing list. And actually SpillableList is easier to understand instead of SpillableArrayList interface.

otterc:
@siyuanh well than this holds for your initial comment as well 😀.  It is just your opinion that it is "overdesigned"  which in your words doesn't mean anything.

tweise:
Perhaps the discussion about interface hierarchy can be separated from this PR. There are different perspectives. Another consideration is that you can start with something that is simple and meets the current needs and can be extended with use cases and learning from it.

davidyan74:
@ilooner I tried to implement `SpillableSetMultiMap.removeAll(K key)` and I'm having some trouble doing it. I'm making this PR review-only and will push my current implementation of removeAll later today and would love to get your feedback. Also, I am curious on your reasons behind the word "Byte" in all the class names. Should I rename SpillableSet and SpillableSetMultimap interfaces and their implementation to say SpillableByteSet, etc?

ilooner:
@davidyan74 I would recommend a lazy delete for now, by setting an empty byte array as the size of the the set for a key in SpillableSetMultimap. The data will still be in managed state, until managed state expires it. If new data is set for the same key later, it is fine, it will just overwrite the old keys. Just be sure that in SpillableSet you point the last node in your linked list to an empty byte array, and recognize an empty byte array and null as an indicator of nonexistence for a value throughout your code. 

I think this is a good compromise because it makes removals fast, and leverage's managed state's efficient purging mechanism to prevent unbounded growth of data.

davidyan74:
@ilooner I have fixed several potential bugs.
Please take a look at the last 3 commits in this PR and see whether or not you agree with them.
The recovery unit test is still failing. I'll take a deeper look at what's going on tomorrow but if you know what's going on in there, please let me know.

library/src/test/java/org/apache/apex/malhar/lib/state/spillable/SpillableSetMultimapImplTest.java
```diff
@@ -185,24 +205,8 @@ public long simpleMultiKeyTestHelper(SpillableStateStore store,
...
-
-    Assert.assertFalse(map.containsKey(key));
-
-    SpillableTestUtils.checkValue(store, 0L,
-        SliceUtils.concatenate(keyBytes, SpillableSetMultimapImpl.SIZE_KEY_SUFFIX), 0, 0, serdeInt);
-
-
     map.endWindow();
     store.endWindow();
-    store.beforeCheckpoint(nextWindowId);
```
ilooner: Why was this removed? In order to correctly simulate a checkpoint with managed state you need to beforeCheckpoint and then checkpointed. It looks like there may be a pre existing issue on line 228 because the window is incremented before checkpointed is called... My bad.

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableSetMultimapImpl.java
```diff
@@ -64,6 +64,8 @@
   private long bucket;
   private Serde<K, Slice> serdeKey;
   private Serde<V, Slice> serdeValue;
+  private int sequence = 0;
```
ilooner: I would feel better if this was a long.

ilooner:
@davidyan74 left a couple comments. overall lgtm

ilooner:
@davidyan74 just realize we need to save the current sequence for a key, if the Set for a key is requested multiple times (like on recovery) then the set may be assigned a different sequence number than the first time.

davidyan74:
@ilooner  If a new set is assigned a different sequence number, what would be the harm? And even there is harm, if the tuples are coming in in an idempotent way, the sequence number should remain the same, right?

ilooner:
@davidyan74 consider the following scenario.
1. You request key A and the corresponding set is assigned a sequence number of 20. 
2. Then lets say you request 80 more different keys, so your sequence number is now 100. 
3. Then you checkpoint
4. Then you fail and get restored to the checkpoint
5. At this point your cache of SpillableSets is empty because it is transient and your sequence number is 100
6. Now we request key A, so a new Spillable set is made with sequence number 101 for key A. But wait, the first time we requested key A the sequence number was 20! So now our new key prefix is wrong and we have lost all the data we stored for that key.

A similar thing can happen in the case when a SpillableSet gets evicted from the cache and gets brought back

davidyan74:
@ilooner I see what you're saying. That would mean a sequence number is useless in this case. I think I just need to have a marker for each spillableset that marks whether it has been removed. Or do you have other ideas?

davidyan74:
@ilooner i think what you said was the reason why the recovery test was failing.

ilooner:
@davidyan74 I think the marker would work

davidyan74:
@ilooner @chandnisingh How does Managed State decide to purge with the current Spillable data structures? The current `remove` implementation of spillable data structures seems to only do a marking (by setting the value to be EMPTY_BYTE_ARRAY) and does not do explicit remove, and Managed State has no way to know that they are actually no longer valid. Does that mean the data would just grow indefinitely?

davidyan74:
This PR depends on #416 #417 #418. Will issue a rebased PR once they are all merged.

davidyan74:

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableSetImpl.java
```diff
@@ -0,0 +1,352 @@
...
+    {
+      this.valid = valid;
+      this.next = next;
+    }
+
+    boolean valid;
+    T next;
+  }
+
+  public static class SerdeListNodeSlice<T> implements Serde<ListNode<T>, Slice>
```
siyuanh: ListNodeSliceSerde is a better name

siyuanh: Follow the convention. I see never mind

PR #404, opened 2016-09-07
APEXMALHAR-2190 #resolve #comment Use reusable buffer to serial spill…

brightchen:
…able data structure

PR #418, opened 2016-09-20
APEXMALHAR-2247 #resolve Added iteration feature in SpillableArrayListImpl and generalize SerdeListSlice to SerdeCollectionSlice

davidyan74:

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableArrayListImpl.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableArrayListImpl.java
@@ -18,6 +18,7 @@
  */
 package org.apache.apex.malhar.lib.state.spillable;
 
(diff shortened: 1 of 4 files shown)


brightchen (commented):
I generalized to support Collection in PR https://github.com/apache/apex-malhar/pull/404. Please see SerdeCollectionWithSerializeBuffer

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableArrayListImpl.java
```diff
@@ -145,7 +147,28 @@ public boolean contains(Object o)
   @Override
   public Iterator<T> iterator()
   {
-    throw new UnsupportedOperationException();
+    return new Iterator<T>()
```
siyuanh, line 150: Maybe set some modification marker so user can not modify during iteration

PR #419, opened 2016-09-20
APEXMALHAR-2250: Ignore directories while scanning.

tushargosavi:
The default DirectoryScanner defined in AbstractFileInputOperator does not handle directories correctly. If there is a directory in the configured path, it gets added as a file in pendingFile list and when operator tries to open it for reading it fails, the operator keeps retrying for configured number of time and then ignore this file.

The fix would be to not return directory name in scanned file names in the first place.


sandeshh:
Tagging @amberarrow as he has the most experience with the fileInputOperator.

vrozov:
@tushargosavi @sandeshh closed due to inactivity

PR #423, opened 2016-09-22
APEXMALHAR-2241 - The metadata kafka consumer should also pickup the properties setting on the kafka input operator

venkateshkottapalli:
Fixes two issues - 
- metadata kafka consumer not picking up properties set 
- Consumer properties set from Properties.xml are  not getting picked.
- Added test case to validate the consumer Properties are not reset.

@siyuanh Please review.

### DIFF
--- a/kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaPartitioner.java
+++ b/kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaPartitioner.java
@@ -18,6 +18,8 @@
  */
 package org.apache.apex.malhar.kafka;
 
+import java.io.IOException;
+import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaPartitioner.java
```diff
@@ -222,17 +224,36 @@ private void initMetadataClients()
...
     }
   }
 
   /**
+   * Converts the property list (key and element pairs) to String format
+   * This format is used to print to a Stream for debugging.
+   * @param prop
+   * @return String
+   */
+  private static String getPropertyAsString(Properties prop) {
```
siyuanh: Why it is static?

siyuanh: @venkateshkottapalli Can you please make the change and I will merge it

venkateshkottapalli: @siyuanh Made the changes. Please review and merge.

PR #428, opened 2016-09-23
APEXMALHAR-2237 : Changes in FSInputModule to support Dynamic partitioning.

yogidevendra:


yogidevendra:
@DT-Priyanka Could you please review this?

DT-Priyanka:
I am not aware why this change is made, can you please explain. We have seen some issues with dynamic partitioning and hence decided to go with static. Why are we changing behavior again?

yogidevendra:
This is based on operator maturity exercise. Dynamic scaling is one of the aspects considered for deciding if given operator/module is mature.

apps/filecopy/src/main/resources/META-INF/properties.xml
```diff
@@ -23,26 +23,10 @@
...
-    <value>128000000</value>
-  </property>
-  <property>
-    <name>dt.operator.HDFSInputModule.prop.scanIntervalMillis</name>
-    <value>10000</value>
-  </property>
-  <property>
-    <name>dt.operator.HDFSInputModule.prop.dedup</name>
-    <value>true</value>
+    <value>hdfs://source-namenode-service:port/path-to-input-directory</value>
```
DT-Priyanka: I will suggest we should keep default values as before.

apps/filecopy/src/main/resources/META-INF/properties.xml
```diff
@@ -23,26 +23,10 @@
 <configuration>
   <property>
     <name>dt.operator.HDFSInputModule.prop.files</name>
-    <value>hdfs://localhost:54310/user/dtadmin/input</value>
-  </property>
-  <property>
```
DT-Priyanka, line 28: why are we removing these properties?

apps/filecopy/src/site/conf/cluster-memory-conf.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
DT-Priyanka: Why there is new config file for these parameters? Do we follow this practice anywhere else?

library/src/main/java/com/datatorrent/lib/io/fs/FSInputModule.java
```diff
@@ -220,7 +243,9 @@ public void setBlockSize(long blockSize)
   /**
    * Gets readers count
    * @return readersCount
+   * @Deprecated Use getMinReaders(), getMaxReaders() instead
```
DT-Priyanka: Add better explanation, it's not clear that not having this configuration and setting others mean that we are using dynamic partitioning. Also why two parameters now, is one more question I get when I read this javadoc.

library/src/main/java/com/datatorrent/lib/io/fs/FSInputModule.java
```diff
@@ -108,8 +114,25 @@ public void populateDAG(DAG dag, Configuration conf)
     }
 
     blockReader.setBasePath(files);
-    if (readersCount != 0) {
-      dag.setAttribute(blockReader, Context.OperatorContext.PARTITIONER, new StatelessPartitioner<FSSliceReader>(readersCount));
+    
+    if (readersCount != 0 && minReaders == 0) {
+      minReaders = readersCount;
```
DT-Priyanka: remove extra line above.

library/src/main/java/com/datatorrent/lib/io/fs/FSInputModule.java
```diff
@@ -276,7 +303,61 @@ public int getBlocksThreshold()
...
+  public long getRepartitionCheckInterval()
+  {
+    return repartitionCheckInterval;
+  }
+  
+  /**
+   * Sets Interval for re-evaluating dynamic partitioning
+   * @param repartitionCheckInterval interval for re-evaluating dynamic partitioning
+   */
+  public void setRepartitionCheckInterval(long repartitionCheckInterval)
```
DT-Priyanka, line 323: can we call it as repartitionInterval ?

DT-Priyanka (commented):
In general, can we add any tests to verify these new parameters and behavior?

apps/filecopy/src/site/conf/cluster-memory-conf.xml
yogidevendra: Single configuration is not suitable for all environments. So, we are trying to capture them separately. 

properties.xml would capture mandatory properties for which there cannot be system default.
Other properties which are advanced but can be initialized with meaningful defaults are captured under apconf. These can be used using -apconf from apex cli.

library/src/main/java/com/datatorrent/lib/io/fs/FSInputModule.java
yogidevendra: I retained older parameter for maintaining backward compatibility. Other PR #439  would mark this class as evolving. Then we can remove the older parameter.

library/src/main/java/com/datatorrent/lib/io/fs/FSInputModule.java
yogidevendra: I thought so, but then I realized that interval is just to check if repartition is needed. Repartition may not always happen after this interval.

apps/filecopy/src/main/resources/META-INF/properties.xml
yogidevendra: Default values are changed to reflect the expected user input (no need to refer to documentation for specifying the value). Anyways, path given in earlier default value may or may not exists on the end user system.

apps/filecopy/src/site/conf/cluster-memory-conf.xml
yogidevendra: These are not new parameters. Existing properties in the operators are exposed with getters, setters.

yogidevendra:
@DT-Priyanka All review comments are answered. Could you please have a look.

apps/filecopy/src/main/resources/META-INF/properties.xml
DT-Priyanka: Path may not exist. But localhost and default port is likely to exist. I don't have firm opinion about this, we can go with what looks right to majority.

apps/filecopy/src/site/conf/cluster-memory-conf.xml
DT-Priyanka: So if someone wants to run this app, he will have to create new config file consolidating all these properties or there is a way to provide multiple properties file to one application?

DT-Priyanka (commented):
Rest of the changes looks okay, will merge it if there are no further comments by tomorrow.

apps/filecopy/src/site/conf/cluster-memory-conf.xml
yogidevendra: launch with -apconf option should address this.

PR #437, opened 2016-09-30
APEXMALHAR-2276 over writing the value of a key in the same time bucket

otterc:
@davidyan74 @ilooner please review and merge since you are familiar with ManagedState code

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
@@ -389,12 +389,15 @@ public interface Bucket extends ManagedStateComponent
         sizeInBytes.getAndAdd(key.length);
         sizeInBytes.getAndAdd(Long.SIZE);
       }
-      if (timeBucket > bucketedValue.getTimeBucket()) {
+      if (timeBucket >= bucketedValue.getTimeBucket()) {
 
         int inc = null == bucketedValue.getValue() ? value.length : value.length - bucketedValue.getValue().length;
         sizeInBytes.getAndAdd(inc);
         bucketedValue.setTimeBucket(timeBucket);
         bucketedValue.setValue(value);
+      } else {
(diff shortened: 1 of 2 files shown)


library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -389,7 +389,7 @@ public void put(Slice key, long timeBucket, Slice value)
         sizeInBytes.getAndAdd(key.length);
         sizeInBytes.getAndAdd(Long.SIZE);
       }
-      if (timeBucket > bucketedValue.getTimeBucket()) {
+      if (timeBucket >= bucketedValue.getTimeBucket()) {
```
davidyan74, line 392: Should we log a warning or throw an exception in the else of this if? I don't think we want to just quietly ignore a put call...

otterc: For ManagedTimeUnifiedStateImpl, `timeBucket = bucketedValue.getTimeBucket()`. 
For ManagedTimeStateImpl, receiving out of order events is common.
I think we should log a warning.

chaithu14: Sorry for the delay. I think we need to check the time instead of timeBucketId.
Consider the below case, 
ManagedTimeStateImpl mt = new ManagedTimeStateImpl();
bucketSpanTime = 0.5 secs
mt.put(1004, k1, v1);
mt.put(1002, k1, v2);
Note: Both tuples are belong to same bucket.

Now, according to the fix, mt.get(k1).equals(v2) will return true.
Is it expected behavior. Please correct it, if I am wrong.

otterc: @chaithu14 you are right. The fix for it is more complicated. I will re-open the ticket.

PR #443, opened 2016-10-05
APEXMALHAR-2276 over writing the value of key only when the time is g…

otterc:
…reater than equal to latest time of the entry

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
@@ -23,7 +23,7 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.TreeSet;
+import java.util.TreeMap;
 import java.util.concurrent.ConcurrentSkipListMap;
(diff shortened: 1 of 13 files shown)


otterc:
@ilooner @davidyan74 @chaithu14  please have a look

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -172,6 +196,36 @@ public int hashCode()
...
+      byte[] latestTimeBytes = Longs.toByteArray(bucketedValue.latestTime);
+
+      System.arraycopy(latestTimeBytes, 0, serializedVal, 0, latestTimeBytes.length);
+      System.arraycopy(bucketedValue.value.buffer, bucketedValue.value.offset, serializedVal, latestTimeBytes.length,
+          bucketedValue.value.length);
+
+      return new Slice(serializedVal);
+    }
+
+    public static BucketedValue deserialize(Slice serializedSlice, long timeBucket)
```
tweise, line 212: This is a lot of work for keeping track of the timestamp. I would like to better understand the need for this (see comment on JIRA).

PR #444, opened 2016-10-07
APEXMALHAR-2281 fixed race condition and calculation of bucket size

otterc:


ilooner (approved):
lgtm will merge if no one has any issues

library/src/main/java/org/apache/apex/malhar/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -274,6 +272,11 @@ protected void putInBucket(long bucketId, long timeBucket, @NotNull Slice key, @
     if (timeBucket != -1) {
       //time bucket is invalid data is not stored
       int bucketIdx = prepareBucket(bucketId);
+      //synchronization on a bucket isn't required for put because the event is added to flash which is
+      //a concurrent map. The assumption here is that the calls to put & get(sync/async) are being made synchronously by
```
tweise, line 276: @ilooner will this work for dimension store?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -384,20 +394,20 @@ public void put(Slice key, long timeBucket, Slice value)
...
-        LOG.warn("ignoring {} {} {}; existing {} {}", key, value, timeBucket,
-            bucketedValue.getValue(), bucketedValue.getTimeBucket());
+        if (timeBucket >= bucketedValue.getTimeBucket()) {
+
+          int inc = null == bucketedValue.getValue() ? value.length : value.length - bucketedValue.getValue().length;
+          sizeInBytes.getAndAdd(inc);
+          bucketedValue.setTimeBucket(timeBucket);
+          bucketedValue.setValue(value);
+        } else {
+          LOG.warn("ignoring {} {} {}; existing {} {}", key, value, timeBucket,
```
tweise: Shouldn't this raise an exception instead of logging if the put is not applied?

otterc: Seemed like out of order updates to an entry can be frequent. Ignoring them seemed better.
However, I am not sure after recent discussions. Maybe we should throw an exception, so the clients will have to handle it.
Copying @davidyan74 as well.

davidyan74: Can the caller check the condition before calling put()? 
IMO, a call to put() should always do what it's expected to do (creating an entry or overwriting an entry).

davidyan74: If this method was named "putIfMostRecent" or something like that, I would not have an issue of it ignoring a call quietly based on a certain condition.

otterc: Caller can do a get before put, however get returns only a Slice.
We can change it to return `BucketedValue` however that is part of the interface and the change may touch several classes using ManagedState

otterc: Actually, caller doesn't know timeBucket for the new entry as well. TimeBucket is created inside managed state so the caller can't even compare it.

otterc: Updated the pull request to throw an exception.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -384,20 +394,19 @@ public void put(Slice key, long timeBucket, Slice value)
...
-        LOG.warn("ignoring {} {} {}; existing {} {}", key, value, timeBucket,
-            bucketedValue.getValue(), bucketedValue.getTimeBucket());
+        if (timeBucket >= bucketedValue.getTimeBucket()) {
+
+          int inc = null == bucketedValue.getValue() ? value.length : value.length - bucketedValue.getValue().length;
+          sizeInBytes.getAndAdd(inc);
+          bucketedValue.setTimeBucket(timeBucket);
+          bucketedValue.setValue(value);
+        } else {
+          throw new RuntimeException("newer entry exists for " + key);
```
tweise: Nit: use AssertionError instead?

tweise:
Good to merge from my end. Please rebase.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
davidyan74: I think an exception is better since the caller may want to catch it when that happens

otterc: I added an assert statement. 
@davidyan74 IMO handling these scenarios by catching an exception is costly. Instead, the caller should cache the value of key and time if they care about it.

tweise: The problem with the assert statement is that it is turned off by default. So if we use it here, then we are back to the silently ignored put. IMO it is better to explicitly raise an error in this situation.

@davidyan74 You can catch AssertionError just as you can catch RuntimeException. Both won't provide much value for handling the situation.

otterc:
Travis fails with un-related test failures:

> POJOInnerJoinOperatorTest.testEmitMultipleTuplesFromStream2:337 Number of tuple emitted  expected:<2> but was:<4>
>   POJOInnerJoinOperatorTest.testInnerJoinOperator:184 Number of tuple emitted  expected:<1> but was:<2>
>   POJOInnerJoinOperatorTest.testMultipleValues:236 Number of tuple emitted  expected:<2> but was:<3>
>   POJOInnerJoinOperatorTest.testUpdateStream1Values:292 Number of tuple emitted  expected:<1> but was:<2>

tweise:
Yup - https://issues.apache.org/jira/browse/APEXMALHAR-2284

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateMultiValue.java
```diff
@@ -71,7 +71,6 @@ public ManagedTimeStateMultiValue(@NotNull ManagedTimeStateImpl store, boolean i
   {
     this();
     this.store = Preconditions.checkNotNull(store);
-    this.store.asyncReadSource = Bucket.ReadSource.READERS;
     this.isKeyContainsMultiValue = isKeyContainsMultiValue;
```
chaithu14, line 74: Due to these changes, Inner join operator tests are failing.

otterc: @chaithu14 please fix ManagedTimeStateMultiValue. Also once you fix failing tests, please create tests ManagedTimeStateMultiValue
There was no need of having a variable asyncReadSource in AbstractManagedState.

PR #446, opened 2016-10-07
APEXMALHAR-1852 Attempt to fix the recovery of partial file test

otterc:


otterc:
Unrelated test failure
Failed tests: 
  KafkaInputOperatorTest.testInputOperatorWithFailure:291->testInputOperator:373 TIMEOUT. testName: testtopic0; Collected data: [c1_1, c1_3, c1_5, c1_7, c1_9, c1_11, c1_13, c1_15, c1_17, c1_19, END_TUPLE, c2_2, c2_4, c2_6, c2_8, c2_10, c2_12, c2_14, c2_16, c2_18, c2_20]

otterc:
Discovered the problem. Some files have 5 blocks and some have 6. This test relies on a file having 5 blocks but any other file can be picked.

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
```diff
@@ -452,7 +447,7 @@ private void scan(Path filePath, Path rootPath, Map<String, Long> lastModifiedTi
           }
         }
       } catch (FileNotFoundException fnf) {
-        LOG.warn("Failed to list directory {}", filePath, fnf);
+        LOG.error("Failed to list directory {}", filePath, fnf);
```
tweise: If it is an error, shouldn't the exception be propagated? Otherwise, if it is an acceptable condition, why not log at INFO?

library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterTest.java
```diff
@@ -400,6 +400,11 @@ public void testRecoveryOfPartialFile() throws InterruptedException
         testMeta.blockMetadataSink.collectedTuples.get(1).getFilePath().endsWith(file2));
   }
 
+  /**
+   * Ignoring this test because {@link FileSplitter} is deprecated. In order to fix some of the race conditions in the
+   * operator, the signature of the operator changed which is why it was deprecated.
+   */
+  @Ignore
```
tweise: Is there a replacement? Should the test be removed?

tweise:
Tests pass! Will wait for your take on the two comments.

library/src/test/java/com/datatorrent/lib/io/fs/FileSplitterTest.java
otterc: FileSplitterInput which is a replacement of this class, has the same test which I fixed. We can just remove this.

tweise: +1 will wait for the updated PR

library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java
otterc: The exception should be propagated.
Don't know why this was just logged before. The scanner is provided with a list of input folders so this assumes that it is fine if the scanner fails to open some of them. It is better to highlight these issues so propagating seems better.

tweise: I agree. And if the idea was to support cases where a directory gets created after the app is already running, there should maybe be a property to control it (ignoreMissingDir).

otterc: It is not a big change to add the ignoreMissingDir property but is it okay if we create a ticket maybe ("newbie") for it and keep it separate from this?

tweise: Sounds good, will merge this.

PR #448, opened 2016-10-10
APEXMALHAR-2254-2269-2270 Fix bugs on replay tuple skipping and idemp…

mattqzhang:
@PramodSSImmaneni please review it.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
@@ -592,16 +592,26 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
                 pendingFiles.remove(recoveryEntry.file);
               }
               inputStream = retryFailedFile(new FailedFile(recoveryEntry.file, recoveryEntry.startOffset));
+
+              while (--skipCount >= 0) {
+                readEntity();
+              }
               while (offset < recoveryEntry.endOffset) {
                 T line = readEntity();
                 offset++;
                 emit(line);
+              }
+              if (recoveryEntry.fileClosed) {
+                closeFile(inputStream);
               }
             } else {
               while (offset < recoveryEntry.endOffset) {
                 T line = readEntity();
                 offset++;
                 emit(line);
+              }
+              if (recoveryEntry.fileClosed) {
+                closeFile(inputStream);
               }
             }
           }
@@ -654,6 +664,7 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
     if (inputStream != null) {
       long startOffset = offset;
       String file  = currentFile; //current file is reset to null when closed.
+      boolean fileClosed = false;
 
       try {
         int counterForTuple = 0;
@@ -662,6 +673,7 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
           if (line == null) {
             LOG.info("done reading file ({} entries).", offset);
             closeFile(inputStream);
+            fileClosed = true;
             break;
           }
 
@@ -679,9 +691,9 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
       } catch (IOException e) {
         failureHandling(e);
       }
-      //Only when something was emitted from the file then we record it for entry.
-      if (offset > startOffset) {
-        currentWindowRecoveryState.add(new RecoveryEntry(file, startOffset, offset));
+      //Only when something was emitted from the file, or we have a closeFile(), then we record it for entry.
+      if (offset >= startOffset) {
+        currentWindowRecoveryState.add(new RecoveryEntry(file, startOffset, offset, fileClosed));
       }
     }
   }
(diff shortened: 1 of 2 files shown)


library/src/test/java/com/datatorrent/lib/io/fs/AbstractFileInputOperatorTest.java
```diff
@@ -816,6 +819,137 @@ public void testStateWithIdempotency() throws Exception
...
+    oper.endWindow();
+    Assert.assertEquals("lines", beforeRecovery4, queryResults.collectedTuples);
+
+    Assert.assertEquals("number tuples", 7, queryResults.collectedTuples.size());
+
+    oper.teardown();
+  }
+
+  /**
+   * This method checkpoints the given writer.
```
amberarrow: "writer" is a bit confusing since it is actually a reader reading from an external source. How about "operator" ?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -591,23 +591,33 @@ protected void replay(long windowId)
                 pendingFiles.remove(recoveryEntry.file);
               }
               inputStream = retryFailedFile(new FailedFile(recoveryEntry.file, recoveryEntry.startOffset));
+
+              while (--skipCount >= 0) {
+                readEntity();
```
amberarrow, line 597: One thing to keep in mind is that for a large file, while we are spinning in this loop, enough time can elapse that STRAM deems the operator hung and kills it. We'll need to increase TIMEOUT_WINDOW_COUNT if that happens (not much we can do here).

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -591,23 +591,33 @@ protected void replay(long windowId)
                 pendingFiles.remove(recoveryEntry.file);
               }
               inputStream = retryFailedFile(new FailedFile(recoveryEntry.file, recoveryEntry.startOffset));
+
```
amberarrow, line 595: There is similar code to skip tuples in emitTuples(), but I believe it will never be executed since there is a check at the top of that function that simply returns if recovery is under way. Additionally, I think this check will _never_ be true:
{code}
if (inputStream == null) {
      try {
        if (currentFile != null && offset > 0) {
        ....
{code}

This logic should be carefully examined and all dead code removed from emitTuples().

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -661,6 +672,7 @@ public void emitTuples()
           if (line == null) {
             LOG.info("done reading file ({} entries).", offset);
             closeFile(inputStream);
+            fileClosed = true;
```
amberarrow, line 676: Check what happens with an empty file: Looks like on replay, both openFile() and closeFile() calls are skipped since there are no recovery entries for that file.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
amberarrow: Actually, I take the first comment back: looks like the skipping in emitTuples might be needed for failed files (but not for replay) but please double check.

library/src/test/java/com/datatorrent/lib/io/fs/AbstractFileInputOperatorTest.java
mattqzhang: fixed

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: Yes I think you're right. The skipCount in emitTuples() are for failed files. Skipping in the replay is in the replay() function.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: Thank you for pointing it out. It does help me identify a bug when the current window finishes reading all the data, and closeFile() is called in next window by itself only. I've updated the fix, and it also covers the case for empty file. I've added both unit test cases.

pramodin:
looking at it

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -592,23 +592,33 @@ protected void replay(long windowId)
...
+              if (recoveryEntry.fileClosed) {
+                closeFile(inputStream);
+              }
             }
           }
         }
       }
     } catch (IOException e) {
-      throw new RuntimeException("replay", e);
+      failureHandling(e);
```
pramodin: Looks like, with failure in reading from a file it skips all other recovery entries in the window and continues with next window (if within max failure count). If there are other files to be read within this window, they would be skipped.

pramodin: I would revert this change as handling failure here will result in non-idempotent behavior. When there is an exception it will try recovery of the operator again.

PR #451, opened 2016-10-12
Renamed MergeWindowedOperator to WindowedMergeOperator.

ShunxinLu:
@davidyan74 @tweise Please review.

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/window/WindowedMergeOperator.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/window/WindowedMergeOperator.java
@@ -21,7 +21,7 @@ package org.apache.apex.malhar.lib.window;
 /**
  * Interface for Merge Windowed Operator.
  */
-public interface MergeWindowedOperator<InputT1, InputT2>
+public interface WindowedMergeOperator<InputT1, InputT2>
     extends WindowedOperator<InputT1>
 {
   /**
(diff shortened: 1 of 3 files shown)


library/src/test/java/org/apache/apex/malhar/lib/window/sample/Merge/KeyedWindowedMergeOperatorTestApplication.java
```diff
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.apex.malhar.lib.window.impl;
+package org.apache.apex.malhar.lib.window.sample.Merge;
```
tweise: package names should be lowercase

tweise (changes requested):

tweise:
@davidyan74 shouldn't these "tests" that look more like examples be moved under demos and fixed to actually run in the CI. Then we could also point to them as examples in the documentation.

davidyan74:
@tweise I don't have preferences either way. But should that be separate from this PR?

davidyan74:
@ShunxinLu please just change the package name. The location of "tests" should be separate and I can just work on that after this is merged

ShunxinLu:
@davidyan74 @tweise The name of package is renamed to lowercases. Please take a look.

tweise:
Please rebase to HEAD of master.

Please also remove the renames for the sample apps and create a JIRA to provide them as examples under demos.

```
rename library/src/test/java/org/apache/apex/malhar/lib/window/{impl => sample/merge}/KeyedWindowedMergeOperatorTestApplication.java (95%)
 rename library/src/test/java/org/apache/apex/malhar/lib/window/{impl => sample/merge}/WindowedMergeOperatorTestApplication.java (96%)
```

ShunxinLu:
@tweise Does it mean that I only change the name of the interface in this PR? And move the sample apps under demos in another PR?

tweise:
Yes, that's my suggestion. It would be great if you could pick up the demos work!

PR #466, opened 2016-10-23
APEXMALHAR-2181 Added Cassandra Upsert Operator with PreparedStatemen…

ghost:
@PramodSSImmaneni / @ashwinchandrap / @sanjaypujare / @DT-Priyanka : Please review 

@tweise  / @PramodSSImmaneni  - This pull request bumps the guava libraries to higher versions ( from 14.x to 16.x ).This was required to support new functionality in the Cassandra upsert operator (because cassandra driver needed an update ). 

Brief description of the new functionalities being supported by this operator are here : https://issues.apache.org/jira/browse/APEXMALHAR-2181

### DIFF
--- a/contrib/pom.xml
+++ b/contrib/pom.xml
@@ -487,13 +487,13 @@
     <dependency>
       <groupId>com.datastax.cassandra</groupId>
       <artifactId>cassandra-driver-core</artifactId>
-      <version>2.1.8</version>
+      <version>3.1.0</version>
       <optional>true</optional>
     </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>14.0.1</version>
+      <version>16.0.1</version>
       <scope>provided</scope>
       <optional>true</optional>
     </dependency>
@@ -639,13 +639,13 @@
            <artifactId>*</artifactId>
          </exclusion>
        </exclusions>
-    </dependency>  
+    </dependency>
     <dependency>
       <groupId>com.googlecode.json-simple</groupId>
       <artifactId>json-simple</artifactId>
       <version>1.1.1</version>
       <optional>true</optional>
-    </dependency> 
+    </dependency>
     <dependency>
       <groupId>com.github.fge</groupId>
       <artifactId>json-schema-validator</artifactId>
--- a/contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
@@ -1,0 +1,1012 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.contrib.cassandra;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.lang.reflect.Field;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.apex.malhar.lib.wal.FSWindowDataManager;
+import org.apache.apex.malhar.lib.wal.WindowDataManager;
+import org.apache.hadoop.classification.InterfaceStability;
+
+import com.codahale.metrics.Timer;
+
+import com.datastax.driver.core.BoundStatement;
+import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.CodecRegistry;
+import com.datastax.driver.core.ColumnMetadata;
+import com.datastax.driver.core.ConsistencyLevel;
+import com.datastax.driver.core.DataType;
+import com.datastax.driver.core.Host;
+import com.datastax.driver.core.KeyspaceMetadata;
+import com.datastax.driver.core.Metrics;
+import com.datastax.driver.core.PreparedStatement;
+import com.datastax.driver.core.ResultSet;
+import com.datastax.driver.core.Session;
+import com.datastax.driver.core.TableMetadata;
+import com.datastax.driver.core.TypeCodec;
+
+
+
+import com.datatorrent.api.AutoMetric;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.Operator;
+import com.datatorrent.api.annotation.InputPortFieldAnnotation;
+import com.datatorrent.api.annotation.Stateless;
+import com.datatorrent.common.util.BaseOperator;
+
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * An abstract operator that is used to mutate cassandra rows using PreparedStatements for faster executions
+ * and accommodates EXACTLY_ONCE Semantics if concrete implementations choose to implement an abstract method with
+ * meaningful implementation (as Cassandra is not a pure transactional database , the burden is on the concrete
+ * implementation of the operator ONLY during the reconciliation window (and not for any other windows).
+ *
+ * The typical implementation model is as follows :
+ *  1. Create a concrete implementation of this class by extending this class and implement a few methods.
+ *  2. Define the payload that is the POJO that represents a Cassandra Row is part of this execution context
+ *     {@link UpsertExecutionContext}. The payload is a template Parameter of this class
+ *  3. The Upstream operator that wants to write to Cassandra does the following
+ *      a. Create an instance of {@link UpsertExecutionContext}
+ *      b. Set the payload ( an instance of the POJO created as step two above )
+ *      c. Set additional execution context parameters like CollectionHandling style, List placement Styles
+ *         overriding TTLs, Update only if Primary keys exist and Consistency Levels etc.
+ *  4. The concrete implementation would then execute this context as a cassandra row mutation
+ *
+ * This operator supports the following features
+ * 1. Highly customizable Connection policies. This is achieved by specifying the ConnectionStateManager.
+ *    There are quite a few connection management aspects that can be
+ *    controlled via {@link ConnectionStateManager} like consistency, load balancing, connection retries,
+ *    table to use, keyspace to use etc. Please refer javadoc of {@link ConnectionStateManager}
+ * 2. Support for Collections : Map, List and Sets are supported
+ *    User Defined types as part of collections is also supported.
+ * 3. Support exists for both adding to an existing collection or removing entries from an existing collection.
+ *    The POJO field that represents a collection is used to represent the collection that is added or removed.
+ *    Thus this can be used to avoid a pattern of read and then write the final value into the cassandra column
+ *    which can be used for low latency / high write pattern applications as we can avoid a read in the process.
+ * 4. Supports List Placements : The execution context can be used to specify where the new incoming list
+ *    is to be added ( in case there is an existing list in the current column of the current row being mutated.
+ *    Supported options are APPEND or PREPEND to an existing list
+ * 5. Support for User Defined Types. A pojo can have fields that represent the Cassandra Columns that are custom
+ *    user defined types. Concrete implementations of the operator provide a mapping of the cassandra column name
+ *    to the TypeCodec that is to be used for that field inside cassandra. Please refer javadocs of
+ *    {@link this.getCodecsForUserDefinedTypes() } for more details
+ * 6. Support for custom mapping of POJO payload field names to that of cassandra columns. Practically speaking,
+ *    POJO field names might not always match with Cassandra Column names and hence this support. This will also avoid
+ *    writing a POJO just for the cassandra operator and thus an existing POJO can be passed around to this operator.
+ *    Please refer javadoc {@link this.getPojoFieldNameToCassandraColumnNameOverride()} for an example
+ * 7. TTL support - A default TTL can be set for the Connection ( via {@link ConnectionStateManager} and then used
+ *    for all mutations. This TTL can further be overridden at a tuple execution level to accomodate use cases of
+ *    setting custom column expirations typically useful in wide row implementations.
+ * 8. Support for Counter Column tables. Counter tables are also supported with the values inside the incoming
+ *    POJO added/subtracted from the counter column accordingly. Please note that the value is not absolute set but
+ *    rather representing the value that needs to be added to or subtracted from the current counter.
+ * 9. Support for Composite Primary Keys is also supported. All the POJO fields that map to the composite
+ *    primary key are used to resolve the primary key in case of a Composite Primary key table
+ * 10. Support for conditional updates : This operator can be used as an Update Only operator as opposed to an
+ *     Upsert operator. i.e. Update only IF EXISTS . This is achieved by setting the appropriate boolean in the
+ *     {@link UpsertExecutionContext} tuple that is passed from the upstream operator.
+ * 11. Lenient mapping of POJO fields to Cassandra column names. By default the POJO field names are case insensitive
+ *     to cassandra column names. This can be further enhanced by over-riding mappings. Please refer feature 6 above.
+ * 12. Defaults can be overridden at at tuple execution level for TTL & Consistency Policies
+ * 13. Support for handling Nulls i.e. whether null values in the POJO are to be persisted as is or to be ignored so
+ *     that the application need not perform a read to populate a POJO field if it is not available in the context
+ * 14. A few autometrics are provided for monitoring the latency aspects of the cassandra cluster
+ */
+
+@InterfaceStability.Evolving
+public abstract class AbstractUpsertOutputOperator extends BaseOperator implements
+    Operator.ActivationListener<Context.OperatorContext>, Operator.CheckpointNotificationListener
+{
+
+  protected ConnectionStateManager connectionStateManager;
+
+  private WindowDataManager windowDataManager;
+
+  private long currentWindowId;
+
+  private transient boolean isInSafeMode;
+
+  private transient long reconcilingWindowId;
+
+  private transient boolean isInReconcilingMode;
+
+  protected transient Session session;
+
+  protected transient Cluster cluster;
+
+  transient Map<String, TypeCodec> complexTypeCodecs;
+
+  transient Map<String, Class> userDefinedTypesClass;
+
+  transient Map<String, DataType> columnDefinitions;
+
+  transient Map<String, String> colNamesMap;
+
+  transient Set<String> pkColumnNames;
+
+  transient Set<String> counterColumns;
+
+  transient Set<String> collectionColumns;
+
+  transient Set<String> listColumns;
+
+  transient Set<String> mapColumns;
+
+  transient Set<String> setColumns;
+
+  transient Set<String> userDefinedTypeColumns;
+
+  transient Set<String> regularColumns;
+
+  protected Map<String, Object> getters;
+
+  protected Map<String, TypeCodec> codecsForCassandraColumnNames;
+
+  CassandraPreparedStatementGenerator cassandraPreparedStatementGenerationUtil;
+
+  transient Map<Long, PreparedStatement> preparedStatementTypes;
+  transient Class<?> tuplePayloadClass;
+
+  private static final transient Logger LOG = LoggerFactory.getLogger(AbstractUpsertOutputOperator.class);
+
+  private static final String CASSANDRA_CONNECTION_PROPS_FILENAME = "CassandraOutputOperator.properties";
+
+  // Metrics
+
+  @AutoMetric
+  transient long ignoredRequestsDuetoIfExistsCheck = 0;
+
+  @AutoMetric
+  transient long successfullWrites = 0;
+
+  @AutoMetric
+  long totalWriteRequestsAttempted = 0;
+
+  @AutoMetric
+  transient int numberOfHostsWrittenTo = 0;
+
+  @AutoMetric
+  transient double fifteenMinuteWriteRateLatency = 0.0;
+
+  @AutoMetric
+  transient double fiveMinuteWriteRateLatency = 0.0;
+
+  @AutoMetric
+  transient double oneMinuteWriteRateLatency = 0.0;
+
+  @AutoMetric
+  transient double meanWriteRateLatency = 0.0;
+
+  @AutoMetric
+  transient long totalIgnoresInThisWindow = 0;
+
+  @AutoMetric
+  long totalIgnoresSinceStart = 0;
+
+  @AutoMetric
+  transient long totalWriteTimeoutsInThisWindow = 0;
+
+  @AutoMetric
+  long totalWriteTimeoutsSinceStart = 0;
+
+  @AutoMetric
+  transient long totalWriteRetriesInThisWindow =  0;
+
+  @AutoMetric
+  long totalWriteRetriesSinceStart = 0;
+
+  @AutoMetric
+  transient long writesWithConsistencyOne = 0;
+
+  @AutoMetric
+  transient long writesWithConsistencyTwo = 0;
+
+  @AutoMetric
+  transient long writesWithConsistencyThree = 0;
+
+  @AutoMetric
+  transient long writesWithConsistencyAll = 0;
+
+  @AutoMetric
+  transient long writesWithConsistencyLocalOne = 0;
+
+  @AutoMetric
+  transient long writesWithConsistencyQuorum = 0;
+
+  @AutoMetric
+  transient long writesWithConsistencyLocalQuorum = 0;
+
+  @AutoMetric
+  transient long writeWithConsistencyLocalSerial = 0;
+
+  @AutoMetric
+  transient long writesWithConsistencyEachQuorum = 0;
+
+  @AutoMetric
+  transient long writesWithConsistencySerial = 0;
+
+  @AutoMetric
+  transient long writesWithConsistencyAny = 0;
+
+  transient Set<Host> uniqueHostsWrittenToInCurrentWindow;
+
+
+  @InputPortFieldAnnotation
+  public final transient DefaultInputPort<UpsertExecutionContext> input = new DefaultInputPort<UpsertExecutionContext>()
+  {
+    @Override
+    public void process(final UpsertExecutionContext tuple)
+    {
+      if (!isEligbleForPassivation(tuple)) {
+        return;
+      }
+      BoundStatement stmnt = setDefaultsAndPrepareBoundStatement(tuple);
+      ResultSet result = session.execute(stmnt);
+      updatePerRowMetric(result);
+    }
+  }; // end of input port implementation
+
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    super.setup(context);
+    windowDataManager = getWindowDataManager();
+    if ( windowDataManager == null) {
+      windowDataManager = new FSWindowDataManager();
+    }
+    windowDataManager.setup(context);
+  }
+
+  @Override
+  public void teardown()
+  {
+    super.teardown();
+    if (null != windowDataManager) {
+      windowDataManager.teardown();
+    }
+  }
+
+
+  /**
+   * Primarily resets the per window counter metrics.
+   * @param windowId The windowid as provided by the apex framework
+   */
+  @Override
+  public void beginWindow(long windowId)
+  {
+    super.beginWindow(windowId);
+    totalIgnoresInThisWindow = 0;
+    totalWriteTimeoutsInThisWindow = 0;
+    totalWriteRetriesInThisWindow =  0;
+    uniqueHostsWrittenToInCurrentWindow.clear();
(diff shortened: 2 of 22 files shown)


tweise:
Hadoop depends on guava 11.0.2 and I seem to remember some earlier discussion that there is a conflict between that version and the one required for Cassandra. Can you confirm that this works on a Hadoop cluster and which version is being used? @DT-Priyanka we should probably confirm this by updating the Cassandra examples from https://github.com/DataTorrent/examples/tree/master/tutorials and make them part of Malhar demos as a follow-up work.

DT-Priyanka:
@tweise  I will try it in example app after review cycle to confirm about guava dependencies. We can move apps to malhar after that.

contrib/pom.xml
```diff
@@ -487,13 +487,13 @@
     <dependency>
       <groupId>com.datastax.cassandra</groupId>
       <artifactId>cassandra-driver-core</artifactId>
-      <version>2.1.8</version>
+      <version>3.1.0</version>
```
DT-Priyanka, line 490: does this version strictly require java 8 or works with java 7 as well?

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+@InterfaceStability.Evolving
+public abstract class AbstractUpsertOutputOperator extends BaseOperator implements
+    Operator.ActivationListener<Context.OperatorContext>, Operator.CheckpointNotificationListener
+{
+
+  public static final String TTL_PARAM_NAME = "ttl";
+
+  ConnectionStateManager connectionStateManager;
+
+  transient WindowDataManager windowDataManager;
```
DT-Priyanka: this should not be transient, also can we make it private?

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+ */
+
+@InterfaceStability.Evolving
+public abstract class AbstractUpsertOutputOperator extends BaseOperator implements
+    Operator.ActivationListener<Context.OperatorContext>, Operator.CheckpointNotificationListener
+{
+
+  public static final String TTL_PARAM_NAME = "ttl";
+
+  ConnectionStateManager connectionStateManager;
```
DT-Priyanka: can we make it private?

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+
+  private transient long reconcilingWindowId;
+
+  private transient boolean isInReconcilingMode;
+
+  transient Session session;
+
+  transient Cluster cluster;
+
+  transient String keyspaceName;
```
DT-Priyanka: I think we can make most of these variables private. Also why have you marked all of them as transient?

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+  }
+
+  private BoundStatement setDefaultsAndPrepareBoundStatement(UpsertExecutionContext tuple)
+  {
+    UpsertExecutionContext.NullHandlingMutationStyle nullHandlingMutationStyle = tuple.getNullHandlingMutationStyle();
+    if (UpsertExecutionContext.NullHandlingMutationStyle.UNDEFINED == nullHandlingMutationStyle) {
+      nullHandlingMutationStyle = UpsertExecutionContext.NullHandlingMutationStyle.SET_NULL_COLUMNS;
+    }
+    boolean setNulls = true;
+    if (nullHandlingMutationStyle != UpsertExecutionContext.NullHandlingMutationStyle.SET_NULL_COLUMNS) {
```
DT-Priyanka, line 504: so default strategy is also set null?

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+      setNulls = false;
+    }
+    UpsertExecutionContext.CollectionMutationStyle collectionMutationStyle = tuple.getCollectionMutationStyle();
+    if (collectionMutationStyle != null) {
```
DT-Priyanka: can we say,
if (collectionMutationStyle == null || collectionMutationStyle == UpsertExecutionContext.CollectionMutationStyle.UNDEFINED) {
      tuple.setCollectionMutationStyle(UpsertExecutionContext.CollectionMutationStyle.ADD_TO_EXISTING_COLLECTION);
    }
for simplicity of code? same for lists

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+    for (String colNameEntry : columnDefinitions.keySet()) {
+      if (pkColumnNames.contains(colNameEntry)) {
+        continue;
+      }
+      if (count > 0) {
+        queryExpression.append(",");
+      }
+      count += 1;
+      if (counterColumns.contains(colNameEntry)) {
+        queryExpression.append(" " + colNameEntry + " = " + colNameEntry + " + :" + colNameEntry);
```
DT-Priyanka: can you please explain this syntax? I am not aware of it.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+  {
+    List<ColumnMetadata> primaryKeyColumns = tableMetadata.getPrimaryKey();
+    for (ColumnMetadata primaryColumn : primaryKeyColumns) {
+      columnDefinitions.put(primaryColumn.getName(), primaryColumn.getType());
+      pkColumnNames.add(primaryColumn.getName());
+      parseForSpecialDataType(primaryColumn);
+    }
+  }
+
+  private void generatePreparedStatements()
```
DT-Priyanka, line 776: Can we think of using some query DSL for simplicity of constructing prepared statements?

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+    queryExpTTLSetCollRemove.append(ttlSetString);
+    buildNonPKColumnsExpression(queryExpTTLSetCollRemove,
+        UpsertExecutionContext.ListPlacementStyle.APPEND_TO_EXISTING_LIST, // Just in case user sets it
+        UpsertExecutionContext.CollectionMutationStyle.REMOVE_FROM_EXISTING_COLLECTION);
+    queryStrings.put(getSlotIndexForMutationContextPreparedStatement(EnumSet.of(
+        OperationContext.TTL_SET, OperationContext.COLLECTIONS_REMOVE, OperationContext.LIST_APPEND
+    )), queryExpTTLSetCollRemove.toString());
+
+    // TTL Not set , Collections Append , List prepend
+    StringBuilder queryExpTTLNotSetCollAppendListPrepend = new StringBuilder(updateQueryRoot.toString());
```
DT-Priyanka: can we move each of these query type build to a function and give the functions relevant name. That would make code more readable.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+  {
+    return new HashMap<>();
+  }
+
+  /***
+   * Allows the concrete implementations to override the preferred sink for the WindowDataManager. This might
+   * come in handy when apex moves to a loosely coupled system from HDFS stores perhaps in the future.
+   * @return
+   */
+  protected WindowDataManager getWindowDataManagerImpl()
```
DT-Priyanka: Should we have default instance in setup and have getter/setter for overriding purposes? If we use this way, there is possibility the instance will be null and we don't do null check everywhere we use it in code.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+
+  private long currentWindowId;
+
+  private transient boolean isInSafeMode;
+
+  private transient long reconcilingWindowId;
+
+  private transient boolean isInReconcilingMode;
+
+  transient Session session;
```
DT-Priyanka: why we need these instance variables session, cluster, keyspacename, tableName when they can be accessed using connectionStateManager?

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+  {
+    ConnectionStateManager.ConnectionBuilder connectionBuilder = withConnectionBuilder();
+    checkNotNull(connectionBuilder, " Connection Builder cannot be null.");
+    connectionStateManager = connectionBuilder.initialize();
+    cluster = connectionStateManager.getCluster();
+    session = connectionStateManager.getSession();
+    keyspaceName = connectionStateManager.getKeyspaceName();
+    tableName = connectionStateManager.getTableName();
+    checkNotNull(session, "Cassandra session cannot be null");
+    tuplePayloadClass = getPayloadPojoClass();
```
DT-Priyanka: should we make it a configurable parameter so that user can set class FQDN from a config file. We have already done it at many other places e.g. check CassandraPOJOInputOperator

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+    columnDefinitions = new HashMap<>();
```
DT-Priyanka: columnDefinitions, dataTypeGroupedColumns and pojoFieldNamesToCassandraColumns are referred anywhere.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+  public final transient DefaultInputPort<UpsertExecutionContext> input = new DefaultInputPort<UpsertExecutionContext>()
+  {
+    @Override
+    public void process(final UpsertExecutionContext tuple)
+    {
+      if (!isEligbleForPassivation(tuple)) {
+        return;
+      }
+      BoundStatement stmnt = setDefaultsAndPrepareBoundStatement(tuple);
+      ResultSet result = session.execute(stmnt);
```
DT-Priyanka, line 277: Are we updating one row at a time? I understand using batches and also provide a guarantee of exactly once processing is not easy. But hitting db for each query can also be expensive. Any thoughts on this?

DT-Priyanka (commented):
First thing, it's really good to see the depth you have covered the features in this operator.
I am still not done with review. But these are few comments I have so far. Other than these comments I wanted to check can we refactor this code a little bit to make it more readable. Some suggestions, we can be,  move code to construct queries etc to a different class, add more methods with proper names to do small small tasks etc.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1401 @@
...
+  transient Map<String, String> pojoFieldNamesToCassandraColumns;
+
+  Map<String, Object> getters;
+
+  Map<String, TypeCodec> codecsForCassandraColumnNames;
+
+  transient Map<Long, PreparedStatement> preparedStatementTypes;
+  transient Class<?> tuplePayloadClass;
+
+  private transient Logger LOG = LoggerFactory.getLogger(AbstractUpsertOutputOperator.class);
```
DT-Priyanka: static final ?

contrib/src/main/java/com/datatorrent/contrib/cassandra/ConnectionStateManager.java
```diff
@@ -0,0 +1,471 @@
...
+ * - Defaults to be used in case the execution context tuple {@link UpsertExecutionContext} does not want to
+ * set one explicitly.
+ * Note that the {@link ConnectionBuilder} is used to build an instance of the ConnectionStateManager.
+ * An instance of this class is typically instantiated by implementing the following inside
+ * {@link AbstractUpsertOutputOperator} withConnectionBuilder() method.
+ * {@code
+      @Override
+      public ConnectionStateManager.ConnectionBuilder withConnectionBuilder()
+      {
+        return ConnectionStateManager.withNewBuilder()
```
DT-Priyanka, line 65: is there anyway to configure these parameters from properties file (like we do normally set other operator parameters from outside.) I think we can accept defaults from properties file and those can be overridden at tuple level.

contrib/pom.xml
ghost: https://docs.datastax.com/en/cassandra/3.0/cassandra/features.html states that the cassandra server needs to be upgraded to java 8 for sure. There is not much clarity on the client side requirements. Will add comments after digging into this a bit more.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: Since the class itself is an abstract, perhaps we should let all extending classes access this ( if at all someone needs to extend some logic in the future ).. Perhaps will make this protected ?

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: Will do.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: The thinking for this was that the session and cluster objects have a lot more information about the current state of the session and cluster ( metrics , alive nodes etc ) and hence represent a different set of information.  Of course session object can be obtained from the connection statemanager everytime it needs to be accessed but that seems to be more typing for every access as we access this pretty frequently for metrics etc. Session and Cluster properties themselves change over time because of the connection policies ( Reconnections etc ) ...

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: will make the keyspacename and table name as private. Marked them transient only to avoid unnecessary state passivation .. My understanding was that activate() is called for all situations i.e. when an operator comes back alive after restarts/crashes. I was hoping to make the ConnectionStatemanager passivated state so that the all aspects like reconnection times etc which are not exposed as variables inside the operator itself can be extracted from the connection state manager and a new session established ...  ( and this also makes the session object inside the clustermanager a transient aspect of the passivation state )

Please let me know if my understanding of activate is wrong here and will modify the code aspects accordingly.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: Yes. It has been purposefully designed to update one row at a time. The main reason is to support the following use cases:
1. This operator is letting the end users specify an operational context and not just setting the values. For example TTL, consistency etc, handle nulls etc 
2. This can be done at the singular level. Setting this at a batch level does not look to be a simplistic view of the world. 
3. Also some aspects like TTL cannot be set at a record level if it is a batch statement. This means the feature of the operator needs to be compromised. Since cassandra TTL is at a cell level and a given row can have varying TTLs across its columns, I believe giving a per row flexibility would be more beneficial for the end user. 
3. Regarding the question on performance, sure there is a little overhead ( for the network call from the operator JVM instance node to the Cassandra cluster coordinator node ) but it would be interesting to see what are the overheads with respect to a batch ( Because all statements in a batch become one transaction , it looks like the coordinator node does a lot more tricks to manage its transaction state like a multi-phase handshake across all participating cassandra nodes for a given batch statement) 
5. Perhaps prepared statements mitigate this to a good extent. 

I was able to reach 5000 QPS on a 10 node cassandra cluster with 10 instances of Cassandra operator ( different implementation but taking the same approach of a single statement as opposed to a batch ) and all this without any stress on both the cassandra cluster and the Apex containers. ( I understand that the numbers here are a bit non-descriptive with respect to the hardware used etc but just wanted to point out that single statement executions are not that bad perhaps ) .

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: There is an object mapper equivalent from datastax itself but there are aspects like 
1. If exists clauses
2. TTLs

are typically not part of the DSL language and there are other messy issues of : 
1. Not all of the above constructs are occurring in a sequence of strings after the query is formed in a DSL generated statement and hence cant simply append etc. 
2. Not all of them give control of list placement styles etc. 

Hence had to handcraft the permutations of all prepared statement strings rather than using any DSL.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: That is correct.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: will do.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: I am not sure that I understood your comment correctly. Here is some more context regarding this issue. Let me know if the below does not make sense. 

The upsert operator does not need the user to specify the CQL statement that needs to be executed.The aim is to keep it as simple as possible for the user of this operator. Forcing a query based approach might have its own overheads: 
1. Cant be flexible in the types of queries to run ( an in fact can only support one query for the entire operator if we want to keep the implementation simple ) 
2. Query parsing will become non trivial when there are composite primary keys etc
3. We need a lot of metadata book keeping to be done ( details below ) 

It takes the approach of : if the definition of the POJO class is known at constructor/activate time, then it can use reflection to map the pojo field names to the cassandra column names. It then performs metadata lookups on cassandra schema definition to see which of it is a counter, which of them are contributing to primary key definition etc. Subsequently it then makes this mapping case sensitive as POJO will have camel case typically and this feature allows us to use  POJOs that are java coding standards compatible. Finally it lets the user override this mapping for situations wherein cassandra column name cant be represented as a POJO field name ( cassandra column names can have underscores etc). 

Since there are a couple of methods that are already marked abstract that the concrete implementation needs to implement, I was thinking that the getPayloadPojoClass() definition also can be abstract to let the user define the pojo payload class definition. Please note that the it is not the POJO class that is coming in as the tuple in the input port. It is the upsertexecutioncontext instance which embeds the pojo class that represents a tuple for this operator. Since Java does not allow for a neat constraint based syntax for this situation and since it cant be templated ( as I need to use reflection to build metadata ) , the method has been made abstract.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: will remove dataTypeGroupedColumns and pojoFieldNamesToCassandraColumns.  

columnDefinitions represents the DataType mapping to a name. This is used in every tuple execution wherein we extract the correct getter for every pojo field that is processed. It is also used during the phase of metadata generation.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: I am not sure if you are referring to the named parameters or the "+" part of it. Hence detailing both.

The aim is to not pin down the operator to just one type of query. Hence we do not force a query to be defined in the operator itself. As tuples flow in, we needed to choose the right prepared statement to execute ( as some tuples choose to set nulls , others not , some append to lists, others prepend to lists etc). For us to achieve these features, it would be impractical to build all permutations and combinations for all pojo fields that can be used to generate the entire possibility of prepared statements. 

In this context, instead of using a positional parameter based "setter" for a given prepared statement, I chose to take the approach of naming the parameters. This way an incoming pojo field name can be used to set the value ( by calling the getter on the pojo ) in the prepared statement itself. Hence we build all of our prepared statements using named parameters rather than positional parameters. 

Regarding the presence of "+" , we are using to imply the meaning of append to an existing value. This would prevent a read on the row and would thus highly benefit write heavy workload patterns. Of course , we can also use it in the pattern of read and set and can be done upstream if required.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: will do.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: will do. I too felt that the method is too "long" initially. Although splitting it meant passing in a lot of context for each split, it is perhaps not a bad idea  as it is only invoked once )

ghost:
@DT-Priyanka Thanks for the review. 
I have addressed some of the review comments and commented on rest of the review comments. I have pushed the changes to this pull request that address the review comments. Could you please give it a look ?

Could you please let me know if anything more needs to be done to get this merged ? 

Also if it is merged, will it be part of 3.6 ? Asking this in case we need to let others in the mail thread for 3.6 discussion be notified. 

Thanks for your time on this thread.

contrib/src/test/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperatorCodecsTest.java
```diff
@@ -0,0 +1,488 @@
...
+    Attribute.AttributeMap.DefaultAttributeMap portAttributes = new Attribute.AttributeMap.DefaultAttributeMap();
+    portAttributes.put(Context.PortContext.TUPLE_CLASS, User.class);
+    testPortContextForUserUpserts = new TestPortContext(portAttributes);
+
+    userUpsertOperator.setup(contextForUserUpsertOperator);
+    userUpsertOperator.activate(contextForUserUpsertOperator);
+    userUpsertOperator.input.setup(testPortContextForUserUpserts);
+  }
+
+  @Ignore
```
DT-Priyanka: can you give a quick try for using embedded Cassandra server for testing e.g. https://github.com/jsevellec/cassandra-unit/wiki/How-to-use-it-in-your-code

DT-Priyanka: In any case don't need to mark these tests as Ignore, as many of contrib tests depends on external systems they are not run by default.

DT-Priyanka:
Sounds good, I have few more comments, I wan't totally done last time, adding them now.
Also I would still request try to refactor code to make it more readable.

contrib/src/test/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperatorCodecsTest.java
ghost: Did not know that contrib tests are not run. I see some tests running as part of every build and hence the annotations. Will remove them ( although I am not sure whether it is the right thing to do since enabling contrib tests at some point in the future would be a number of test failures ) .

ghost: I did try cassandra-unit initially but it does not seem to work great. Also when I tried a couple of months back, there were driver protocol version differences than what we wanted for some feature support and the Embedded server versions. NoSQLUnit was also evaluated as part of that but it did not do great either. Hence had to take the current implementation approach.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
ghost: will do.

contrib/src/main/java/com/datatorrent/contrib/cassandra/ConnectionStateManager.java
ghost: I have added functionality to read from a properties file. However here are my thoughts on this: 

There are pros and cons with a properties file approach.

Pros : 
- Easier to "configure" by writing the properties file. However the concrete implementation of the operator still needs to implement abstract methods and hence as well might need some code to be in place. Hence the gain is not that much. 

Cons:
- Severely restricts that there is only one operator in the container that can output to cassandra. Since this operator is severely tied down to a single table and keyspace at instantiation time and these properties are getting loaded from a standard file name, we will be restricting that only one logical Cassandra Upsert operator can be used  at any given point in time. 

However the implementation allows the end user to use the builder pattern to override the properties file based approach.

ghost:
@DT-Priyanka  Incorporated the final set of review comments and also added comments for some of the other review comments. 

Could you please give it a look ? 

Thanks for your time on this.

contrib/pom.xml
DT-Priyanka: did you confirm this? Have you tried running this operator in any app for testing?

DT-Priyanka (commented):
Overall looks good, have you tried running this operator in any app?
Can we move code for query building etc in some other helper class? And if possible we should create more functions for logical code blocks. Also rearrange some code like having setup(), activate(), methods on top followed by other methods (consider logical flow). The code needs to be more readable.

ghost:
@DT-Priyanka  : 

Regarding moving code related to query building, I did exactly as you were hinting in the previous pull request. Not sure if you saw this. Is there any thing else you are referring to as code regarding query building. Please refer class "CassandraPreparedStatementGenerator". 

I can rearrange the code as suggested related to setup() etc methods. Is there any best practice doc for this ? For now, I can put the operator methods first and other methods later. Perhaps we need a clear doc suggesting this to avoid multiple iterations of code review. 

Regarding the comment , "The code needs to be more readable." , is there anything specific you are pointing to or the entire code itself ? I did refactor the code a few times in this thread to meet this objective. 

Regarding 3.1.0 driver version change, I did post a question to Cassandra channel and the response was that it should work with java 7 as well. Datastax documentation does not declare an explicit requirement of java 8 on the client side JVM. 

Is there a place I can put a end to end cassandra demo app code with this new operator ?

DT-Priyanka:
Hi Ananth,

Sorry for high level comment. When I say it needs to be more readable what I expect is,
1. Smaller functions
2. Small functions taking care of small responsibilities having meaningful name so while reading code one can actually decide to ignore details just looking at function name.
3. Arrangement of code 
4. Not very big classes and classes carrying single responsibility.

In general it took me more time to go through code though I am not new to java or cassandra so I though there is a scope to make the code more readable. After all you have written really good and in depth features for cassandra operator, my concern was people should not fear to use it if they don't understand it.

I have no issue with technical aspect of code so I am good to merge it when you rearrange the functions.

DT-Priyanka:
This is one example app: https://github.com/DataTorrent/examples/tree/master/tutorials/cassandraOutput
you can replace cassandra output operator with your operator to test your operator.

DT-Priyanka:
@ananthc  let me know if there is testing/refactoring pending, otherwise we can go ahead and merge the code changes.

ghost:
@DT-Priyanka  - Apologies for the delay on this thread .. I could give it one more try if you wish ... I did go through the code once more and there seems to be very less that can be done .. A few things that come to my mind are to move Autometrics to a different Util class ...Most of the functions seem to be of moderate sizes.. Perhaps I can try to split the operator into two types .. One is Exactly Once Operator and the other which does not provide that guarantee .. Thoughts ? .. The exactly once operator will only implement one more extra method over the Base Upsert operator .. Both of them would be abstract though ... Unfortunately my computer is broken at this point in time and I can resume the refactoring perhaps a week or two from now ( as the replacement part is yet to arrive for my computer ) ... 

If you reckon it is ok to merge please do so.

DT-Priyanka:
@ananthc  no need of splitting operator in two. For other suggested changes it's nice to have but I am okay to go ahead without those changes as refactoring will take time.
But you have to rebase the code once with latest changes so that I would be able to merge it.

ghost:
@DT-Priyanka  - I just pushed the rebased code. Please let me know if anything more needs to be done. Thank you for your time on this thread.

contrib/src/main/java/com/datatorrent/contrib/cassandra/AbstractUpsertOutputOperator.java
```diff
@@ -0,0 +1,1012 @@
...
+  {
+    return windowDataManager;
+  }
+
+  public void setWindowDataManager(WindowDataManager windowDataManager)
+  {
+    this.windowDataManager = windowDataManager;
+  }
+
+  /***
```
tweise, line 925: This is invalid javadoc. I noticed this while creating the release. Would someone volunteer to change the Travis CI build to incorporate javadoc checking @amberarrow

DT-Priyanka: I will do that.

PR #469, opened 2016-10-25
APEXMALHAR-2315 Ignore tests in sql which uses POJOInnerJoinOperator 

chinmaykolhatkar:
Ignore tests in sql which uses POJOInnerJoinOperator for SQL JOIN functionality. This should be reverted once POJOInnerJoinOperator is replaced with Windowed variant of it.

@tweise Please review and merge.


chinmaykolhatkar:
@bhupeshchawda Can you please review and merge?

PR #471, opened 2016-10-28
Added S3 Record Reader module

ajaygit158-zz:
@yogidevendra @chaithu14 Kindly Review


ajaygit158-zz:
travis build is failing due to java version issues. Code is probably using laer version of Java. Will recompile and raise PR again with correct java version

PR #472, opened 2016-10-28
APEXMALHAR-2304 SQL Support Examples

chinmaykolhatkar:
This is example for newly added SQL support in apex-malhar.


chinmaykolhatkar:
@tweise Please review and merge.

tweise:
I wonder if it is possible to reduce repetitive boilerplate populateDAG code so that it looks more like run(sql)? Otherwise this looks rather complicated to a user.

demos/sql/pom.xml
```diff
@@ -0,0 +1,223 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.apache.apex</groupId>
+      <artifactId>malhar-sql</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <!-- Dependency for malhar-sql... Figure out how can this be avoided -->
+    <dependency>
+      <groupId>org.apache.apex</groupId>
```
tweise: This shouldn't be necessary when malhar-sql is properly setup, those should already be transitive dependencies. Except for those cases where they were marked provided.

chinmaykolhatkar:
@tweise The examples mentioned in this are for using the existing APIs of malhar-sql.
Its very much possible to reduce boilerplate code... That's part of APEXMALHAR-2293... This is being worked on now.
Once this is done, one could use only SQL to execute the DAG in following 2 ways:
1. 

public void populateDAG(DAG dag, Configuration conf)
{
  runSQL(dag, conf, <List of sql statements to execute>);
}
1. Input can be a file where only SQL statements can be mentioned to a command line utility and then the cmdline utility will launch the DAG.

demos/sql/pom.xml
chinmaykolhatkar: Done.

chinmaykolhatkar:
@tweise Can you review and merge?

PR #473, opened 2016-10-28
APEXMALHAR-2289 #resolve #comment Add resource monitor

brightchen:


brightchen:
@davidyan74 Could you review. This PR add resource monitor to monitor the system resource such as garbage collection and CPU usage to identify the benchmark issue.

benchmark/src/main/java/com/datatorrent/benchmark/monitor/ResourceMonitorService.java
```diff
@@ -0,0 +1,96 @@
...
+import org.slf4j.LoggerFactory;
+
+public class ResourceMonitorService
+{
+  public static ResourceMonitorService create(long period)
+  {
+    return new ResourceMonitorService(period);
+  }
+
+  protected long period;
```
davidyan74: Change this to private, and all other protected members and methods need to be looked at. In most cases, private should be used instead of protected unless there are subclasses or there will be subclasses that need them

benchmark/src/main/java/com/datatorrent/benchmark/monitor/ResourceMonitorService.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  public static class LogSystemResourceTask implements Runnable
+  {
+    private static final Logger logger = LoggerFactory.getLogger(LogSystemResourceTask.class);
+
+    @Override
+    public void run()
+    {
+      logger.info("System Resource: {}",
```
davidyan74: The create method call should be separate from the logger call.

benchmark/src/main/java/com/datatorrent/benchmark/monitor/SystemResourceMonitor.java
```diff
@@ -0,0 +1,170 @@
...
+  public Map<SystemResource, Number> getSystemResourceUsage()
+  {
+    systemResourceUsage.clear();
+    for (SystemResource sr : systemResources) {
+      Method getterMethod = getGetterMethod(sr);
+      getterMethod.setAccessible(true);
+      try {
+        systemResourceUsage.put(sr, (Number)getterMethod.invoke(operatingSystemMXBean));
+      } catch (Exception e) {
+        throw new RuntimeException("Can't invoke method: " + getterMethod.getName(), e);
```
davidyan74: use Throwables.propagate

benchmark/src/main/java/com/datatorrent/benchmark/monitor/SystemResourceMonitor.java
```diff
@@ -0,0 +1,170 @@
...
+  {
+    sb.append("\n\t").append(entry.getKey()).append(": ").append(NumberFormat.getInstance().format(entry.getValue()));
+  }
+
+  protected Method getGetterMethod(SystemResource sr)
+  {
+    try {
+      return operatingSystemMXBean.getClass().getMethod("get" + sr.name(), null);
+    } catch (Exception e) {
+      throw new RuntimeException("Invalid SystemResource: " + sr);
```
davidyan74: use Throwables.propagate and please also look at other places that does a new RuntimeException

benchmark/src/main/java/com/datatorrent/benchmark/state/StoreOperator.java
```diff
@@ -46,14 +47,16 @@
   {
     INSERT,
     UPDATESYNC,
```
davidyan74: Let's put underscores in these variables names.

davidyan74 (changes requested):

brightchen:
@davidyan74 Applied comments, please review

davidyan74:
Can you check all protected usages?

davidyan74:
Also, RuntimeException has not been changed to Thowables.propagate

brightchen:
@davidyan74 I changed all class fields into private. But still keep some constructors and methods as protected

davidyan74:
@brightchen  I would say make all of them private for now unless there is a subclass that needs them.

brightchen:
@davidyan74 changed all protected to private as your commented.

benchmark/src/main/java/com/datatorrent/benchmark/monitor/SystemResourceMonitor.java
```diff
@@ -0,0 +1,171 @@
...
+  public Map<SystemResource, Number> getSystemResourceUsage()
+  {
+    systemResourceUsage.clear();
+    for (SystemResource sr : systemResources) {
+      Method getterMethod = getGetterMethod(sr);
+      getterMethod.setAccessible(true);
+      try {
+        systemResourceUsage.put(sr, (Number)getterMethod.invoke(operatingSystemMXBean));
+      } catch (Exception e) {
+        Throwables.propagate(e);
```
davidyan74: should be

``` java
 throw Throwables.propagate(e);
```

so that both IDE and the compiler know the execution flow stops here

benchmark/src/main/java/com/datatorrent/benchmark/state/StoreOperator.java
```diff
@@ -42,18 +43,20 @@
...
-    UPDATESYNC,
-    UPDATEASYNC
+    UPDATE_SYNC,
+    UPDATE_ASYNC,
+    GET_SYNC,
+    DO_NOTHING
   }
 
-  protected static final int numOfWindowPerStatistics = 10;
+  protected static final int numOfWindowPerStatistics = 120;
```
davidyan74: please change all members of this class to private

benchmark/src/main/java/com/datatorrent/benchmark/monitor/SystemResourceMonitor.java
```diff
@@ -0,0 +1,171 @@
...
+      SystemResource.SystemCpuLoad,
+      SystemResource.ProcessCpuLoad};
+
+  public static SystemResource[] fileResources = new SystemResource[] {
+      SystemResource.OpenFileDescriptorCount,
+      SystemResource.MaxFileDescriptorCount};
+
+  public static SystemResource[] allResources = SystemResource.values();
+
+  private static Map<Set<SystemResource>, SystemResourceMonitor> resourcesToMonitors = Maps.newHashMap();
```
davidyan74, line 72: why do we need this?

davidyan74 (changes requested):

benchmark/src/main/java/com/datatorrent/benchmark/monitor/SystemResourceMonitor.java
brightchen: My intension is to prevent create two instances monitoring same group of resource

PR #475, opened 2016-10-31
APEXMALHAR-2324 Megh flume

pramodin:
Flume support from megh project. Preserved attribution for individual commits, filtered flume related commits only from the megh project. Applied formatting and licensing changes compatible with the Apex project as author Apex Dev <dev@apex.apache.org>. @tweise @amolhkekre @gauravgopi123 @243826 @chandnisingh please see

megh: https://github.com/datatorrent/megh


tweise:
@PramodSSImmaneni this PR contains a lot of noise when you look at the commit history. A lot of this can be compressed with rebase.

pramodin:
@tweise some of the newer build commits are easier to compress as they are clumped together, however, for others where there are consecutive commits from the same author it is possible to rebase them but it is a tedious process as the different author contributions are mixed together. How important is it to get it fully compressed. Would it be ok to clear out some of the recent build commit noise.

tweise:
First pass should be interactive rebase that squashes subsequent commits of same author. That should trim it quite a bit. Recent commits that have to do with releases should all be squashed. Maybe we look at it again once these low hanging fruits are taken care of?

amberarrow:
In a historic case like this, how important is it to preserve attribution ? Phrased differently: Have we ever had a case where this kind of distant attribution provided critically needed information ?

pramodin:
@tweise tried to squash successive commits from the same author but running into quite a few problems, since the commits are the result of multiple git filter-branch passes and there are resultant merge commits in there that throw rebase off. How about letting the commits be, many of them are old and if somebody is looking at all the commits in a log tool these will not add a lot of noise till they get to a while back. Also, the number of commits is around 350, while these are not a small number would only amount to 5% of total commits and will not add too much delay to fetch times.

tweise:
@PramodSSImmaneni After looking at it a bit more I don't think that this commit history belongs into the Apex repository. I suggest that we squash and either attribute all files to Apex Dev or divide them between the main contributors. Most of the main code was written by Chetan, storage package by Gaurav, so it would be straightforward to commit the files like that.

I would also want to see the packages renamed and some testing done so that what goes into the repository is known to be working. What about the documentation?

pramodin:
A separate PR has been submitted with the requested changes
https://github.com/apache/apex-malhar/pull/598

PR #476, opened 2016-10-31
APEXMALHAR-2326 Use elasticMQ as a mock SQS server instead of using creds

sanjaypujare:
@tweise  this is to fix the test failures in SQSStringInputOperatorTest. Pls review and merge

### DIFF
--- a/library/pom.xml
+++ b/library/pom.xml
@@ -378,6 +378,12 @@
         </exclusion>
       </exclusions> 
     </dependency>
+    <dependency>
+      <groupId>org.elasticmq</groupId>
+      <artifactId>elasticmq-rest-sqs_2.11</artifactId>
+      <version>0.10.0</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <profiles>
--- a/library/src/test/java/com/datatorrent/lib/io/jms/SQSStringInputOperatorTest.java
+++ b/library/src/test/java/com/datatorrent/lib/io/jms/SQSStringInputOperatorTest.java
@@ -22,7 +22,11 @@ import java.io.File;
 
 import javax.jms.ConnectionFactory;
 
+import org.elasticmq.rest.sqs.SQSRestServer;
+import org.elasticmq.rest.sqs.SQSRestServerBuilder;
+import org.junit.AfterClass;
 import org.junit.Assert;
+import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.internal.AssumptionViolatedException;
@@ -30,13 +34,15 @@ import org.junit.rules.TestWatcher;
 import org.junit.runner.Description;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 import org.apache.commons.io.FileUtils;
 
 import com.amazon.sqs.javamessaging.SQSConnectionFactory;
-import com.amazonaws.auth.PropertiesFileCredentialsProvider;
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.auth.BasicAWSCredentials;
 import com.amazonaws.regions.Region;
 import com.amazonaws.regions.Regions;
+import com.amazonaws.services.sqs.AmazonSQSClient;
 import com.datatorrent.api.Attribute;
 import com.datatorrent.api.Context;
 import com.datatorrent.lib.helper.OperatorContextTestHelper;
@@ -71,7 +77,17 @@ public class SQSStringInputOperatorTest
     JMSStringInputOperator operator;
     CollectorTestSink<Object> sink;
     Context.OperatorContext context;
+    String accessKeyId = "";
+    String secretKey = "";
     SQSTestBase testBase;
+
+    private void createElasticMQMockServer()
+    {
+      // just use x and y as access and secret keys respectively: not checked by the mock server
+      testBase.sqs = new AmazonSQSClient(new BasicAWSCredentials("x", "y"));
+      // have it talk to the embedded server
+      testBase.sqs.setEndpoint("http://localhost:9324");
+    }
 
     @Override
     protected void starting(Description description)
@@ -80,10 +96,8 @@ public class SQSStringInputOperatorTest
       final String className = description.getClassName();
 
       testBase = new SQSTestBase();
-      if (testBase.validateTestCreds() == false) {
-        return;
-      }
       testBase.generateCurrentQueueName(methodName);
+      createElasticMQMockServer();
       try {
         testBase.beforTest();
       } catch (AssumptionViolatedException ave) {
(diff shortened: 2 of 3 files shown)


sanjaypujare:
@amberarrow @vrozov  could one of you review this and merge?

amberarrow:
Is there a ticket associated with this ?

sanjaypujare:
@amberarrow  just created one https://issues.apache.org/jira/browse/APEXMALHAR-2326

sanjaypujare:
@tweise @amberarrow  any comments? Can this be merged?

library/src/test/java/com/datatorrent/lib/io/jms/SQSStringInputOperatorTest.java
```diff
@@ -71,47 +77,85 @@
...
+    String accessKeyId = "";
+    String secretKey = "";
     SQSTestBase testBase;
 
+    private void createElasticMQMockServer()
+    {
+      // just use x and y as access and secret keys respectively: not checked by the mock server
+      testBase.sqs = new AmazonSQSClient(new BasicAWSCredentials("x", "y"));
+      // have it talk to the embedded server
+      testBase.sqs.setEndpoint("http://localhost:9324");
```
tweise, line 89: Should find available port instead?

library/pom.xml
```diff
@@ -378,6 +378,12 @@
         </exclusion>
       </exclusions> 
     </dependency>
+    <dependency>
```
tweise, line 381: Shouldn't there be also dependencies removed as they are replaced by the mock server?

tweise (changes requested):

tweise:
+1 for the idea in general, see comments.

tweise:
I seem to remember some comment regarding required JDK for this?

sanjaypujare:
@tweise  yes the following was in an email from me (should have put it in the PR here)

There is a Travis build failure because of dependency on Java 8 (https://travis-ci.org/apache/apex-malhar/builds/172145189)

Tests in error:
  SQSStringInputOperatorTest.createMockServer:203 » UnsupportedClassVersion akka...

This article explains it: http://stackoverflow.com/questions/32959187/akka-hello-world-problems-unsupported-major-minor-version-52-0

So:
-          Can we use Java8 for building/testing on the Travis machine?
-          Note that akka-actor is used only in test-mode (test scope for elasticmq) so for the non-test code we are still okay (i.e. there is no dependency on Java 8)

Otherwise I’ll have to see if I can make it work with older akka or older elasticMQ

vrozov:
We use JDK 1.8 in Jenkins builds, so my preference is to continue to use default JDK 1.7 in Travis. This allows covering both 1.7 and 1.8 in CI builds.

sanjaypujare:
@vrozov  in that case I will have to suppress the unit test when the current JAVA version is < 1.8. Is that okay?

vrozov:
Does 0.8.x require JDK 1.8 as well? Will downgrading the dependency help?

library/pom.xml
sanjaypujare: Nothing is replaced by the mock server for unit tests. For unit tests we used the actual SQS service and all those jar files are still needed in the regular mode.

library/src/test/java/com/datatorrent/lib/io/jms/SQSTestBase.java
```diff
@@ -173,7 +127,6 @@ public void validateAssumption()
   @Before
   public void beforTest() throws Exception
```
amberarrow, line 128: beforTest => beforeTest

library/src/test/java/com/datatorrent/lib/io/jms/SQSTestBase.java
```diff
@@ -182,6 +135,6 @@ public void beforTest() throws Exception
   @After
   public void afterTest() throws Exception
```
amberarrow, line 136: Can this function be deleted since it does nothing ?

library/src/test/java/com/datatorrent/lib/io/jms/SQSStringInputOperatorTest.java
```diff
@@ -71,47 +77,85 @@
...
+
+          };
           // Create the connection factory using the environment variable credential provider.
           // Connections this factory creates can talk to the queues in us-east-1 region.
           SQSConnectionFactory connectionFactory =
               SQSConnectionFactory.builder()
               .withRegion(Region.getRegion(Regions.US_EAST_1))
-              .withAWSCredentialsProvider(new PropertiesFileCredentialsProvider(testBase.getDevCredsFilePath()))
+              .withAWSCredentialsProvider(provider)
+              .withEndpoint("http://localhost:9324")
```
amberarrow, line 158: Suggest making that URL a variable since it is used in multiple places

library/src/test/java/com/datatorrent/lib/io/jms/SQSStringInputOperatorTest.java
```diff
@@ -154,7 +194,20 @@ protected void finished(Description description)
...
+
+  @BeforeClass
+  public static void createMockServer()
+  {
+    // By default the urls will use http://localhost:9324 as the base URL
+    mockServer = SQSRestServerBuilder.start();
+  }
 
+  @AfterClass
+  public static void shutDownMockServer() throws InterruptedException
```
amberarrow, line 207: Is this function necessary ?

library/src/test/java/com/datatorrent/lib/io/jms/SQSStringInputOperatorTest.java
```diff
@@ -71,47 +77,85 @@
     JMSStringInputOperator operator;
     CollectorTestSink<Object> sink;
     Context.OperatorContext context;
+    String accessKeyId = "";
+    String secretKey = "";
     SQSTestBase testBase;
 
+    private void createElasticMQMockServer()
+    {
+      // just use x and y as access and secret keys respectively: not checked by the mock server
```
amberarrow, line 86: Suggest a similar explanatory comment on accessKeyId and secretKey which are initialized to empty strings above.

library/src/test/java/com/datatorrent/lib/io/jms/SQSStringInputOperatorTest.java
```diff
@@ -22,21 +22,27 @@
...
-import com.amazonaws.auth.PropertiesFileCredentialsProvider;
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.auth.BasicAWSCredentials;
 import com.amazonaws.regions.Region;
 import com.amazonaws.regions.Regions;
+import com.amazonaws.services.sqs.AmazonSQSClient;
 import com.datatorrent.api.Attribute;
 import com.datatorrent.api.Context;
 import com.datatorrent.lib.helper.OperatorContextTestHelper;
```
amberarrow, line 48: Some extra leading whitespace at lines 58-59

vrozov:
@sanjaypujare @tweise @amberarrow closed due to inactivity

PR #477, opened 2016-11-01
APEXMALHAR-2320 #resolve #comment use SerializationBuffer implement F…

brightchen:
…SWindowDataManager.toSlice()

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
@@ -18,7 +18,6 @@
  */
 package org.apache.apex.malhar.lib.wal;
 
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -38,6 +37,8 @@ import org.slf4j.LoggerFactory;
 
 import org.apache.apex.malhar.lib.state.managed.IncrementalCheckpointManager;
 import org.apache.apex.malhar.lib.utils.FileContextUtils;
+import org.apache.apex.malhar.lib.utils.serde.SerializationBuffer;
+import org.apache.apex.malhar.lib.utils.serde.WindowedBlockStream;
 import org.apache.hadoop.fs.FileContext;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.Path;
@@ -45,7 +46,6 @@ import org.apache.hadoop.fs.RemoteIterator;
 
 import com.esotericsoftware.kryo.Kryo;
 import com.esotericsoftware.kryo.io.Input;
-import com.esotericsoftware.kryo.io.Output;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSet;
@@ -149,6 +149,8 @@ public class FSWindowDataManager implements WindowDataManager
 
   private transient FileContext fileContext;
 
+  private transient SerializationBuffer serializationBuffer;
+
   public FSWindowDataManager()
   {
     kryo.setClassLoader(Thread.currentThread().getContextClassLoader());
@@ -157,6 +159,7 @@ public class FSWindowDataManager implements WindowDataManager
   @Override
   public void setup(Context.OperatorContext context)
   {
+    serializationBuffer = new SerializationBuffer(new WindowedBlockStream());
     operatorId = context.getId();
 
     if (isStatePathRelativeToAppPath) {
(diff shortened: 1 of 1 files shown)


brightchen:
@davidyan74 Please review

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
```diff
@@ -416,7 +418,11 @@ public void save(Object object, long windowId) throws IOException
 
     byte[] windowIdBytes = Longs.toByteArray(windowId);
     writer.append(new Slice(windowIdBytes));
+
+    //this call will copy all data to the file input stream, so the data is useless any more after call return
```
davidyan74: please fix this comment

davidyan74 (changes requested):
don't you also need to call beginWindow and endWindow on the serializationBuffer?

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
brightchen: The stream memory was reset all instead of reset to certain window, so beginWindow(), endWindow() don't need to be called

brightchen:
@davidyan74 Please review

davidyan74:
LGTM. @chandnisingh can you also review? I'll merge if this looks good to you.

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
```diff
@@ -149,6 +149,8 @@
 
   private transient FileContext fileContext;
 
+  private SerializationBuffer serializationBuffer = new SerializationBuffer(new WindowedBlockStream());
```
otterc: Does this need to be checkpointed?

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
```diff
@@ -416,7 +418,17 @@ public void save(Object object, long windowId) throws IOException
...
     byte[] windowIdBytes = Longs.toByteArray(windowId);
     writer.append(new Slice(windowIdBytes));
+
+    /**
+     * writer.append() will copy the data to the file output stream.
+     * So the data in the buffer is not needed any more, and it is safe to reset the serializationBuffer.
+     *
+     * And the stream memory was reset all instead of reset to certain window,
+     * so beginWindow() and endWindow() don't need to be called
+     */
```
otterc, line 429: please re-write the comment. Hard to understand.

otterc (changes requested):

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
davidyan74: Good point @chandnisingh . 
@brightchen Also, since this is not windowed, should this be using BlockStream?

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
```diff
@@ -594,13 +606,10 @@ public void committed(long committedWindowId) throws IOException
 
   private Slice toSlice(Object object)
   {
-    ByteArrayOutputStream baos = new ByteArrayOutputStream();
-    Output output = new Output(baos);
-    kryo.writeClassAndObject(output, object);
-    output.close();
-    byte[] bytes = baos.toByteArray();
+    kryo.writeClassAndObject(serializationBuffer, object);
+    serializationBuffer.close();
```
davidyan74: Why close?

library/src/main/java/org/apache/apex/malhar/lib/wal/FSWindowDataManager.java
brightchen: SerializationBuffer in fact is something like kyro Output, it writes data to WindowedBlockStream. The intention is to reuse allocated memory and avoid memory copy. It is not related to checkpoint

brightchen: BlockStream is ok, SerializationBuffer constructor only take parameter with type WindowedBlockStream

brightchen: I meant SerializationBuffer can't constructed by BlockStream, it need WindowedBlockStream

otterc: since it is not transient, it will be checkpointed. My question was whether you want that?

brightchen: thank you. It should be transient.

brightchen:
z

brightchen:
@chandnisingh @davidyan74 Applied comments. Please review

davidyan74:
I'm ok with the change. @chandnisingh any other concerns?

otterc (approved):
LGTM

davidyan74:
@brightchen  can you quickly rebase this with the current master? I'll merge it.

brightchen:
@davidyan74 Rebased

PR #479, opened 2016-11-02
APEXMALHAR-2325 1) Set the file system default block size to the reader. 2) Set the block size to the reader context

chaithu14:


DT-Priyanka:
can you add some tests using mocks ?

library/src/main/java/com/datatorrent/lib/io/block/FSSliceReader.java
```diff
@@ -39,6 +42,15 @@ public FSSliceReader()
   }
 
   @Override
+  public void setup(Context.OperatorContext context)
+  {
+    super.setup(context);
+    if (basePath != null) {
+      ((ReaderContext.FixedBytesReaderContext)this.readerContext).setLength((int)fs.getDefaultBlockSize(new Path(basePath)));
```
DT-Priyanka, line 49: can you add instanceof check?

library/src/test/java/com/datatorrent/lib/io/block/FSSliceReaderTest.java
```diff
@@ -135,4 +141,32 @@ public void testBytesReceived() throws IOException
...
+    {
+      return fileSystem;
+    }
+  }
+
+  @Test
+  public void testBlockSize() throws IOException
+  {
+    long blockSize = 1000;
+    MockitoAnnotations.initMocks(this);
```
DT-Priyanka: can you move initMocks to setup method (annotate setup to @BeforeTest) so that every added test case need not do it.

PR #482, opened 2016-11-03
APEXMALHAR-2327 #resolve #comment BucketsFileSystem.writeBucketData()…

brightchen:
… call Slice.toByteArray() cause allocate unnecessary memory

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/fileaccess/FileAccess.java
+++ b/library/src/main/java/com/datatorrent/lib/fileaccess/FileAccess.java
@@ -147,11 +147,21 @@ public interface FileAccess extends Closeable
   {
     /**
      * Appends key/value pair to the underlying file.
+     * @deprecated use {@link FileWriter#append(Slice, Slice)} instead.
      * @param key
      * @param value
      * @throws IOException
      */
+    @Deprecated
     void append(byte[] key, byte[] value) throws IOException;
+
+    /**
+     * Appends key/value pair to the underlying file.
+     * @param key
+     * @param value
+     * @throws IOException
+     */
+    void append(Slice key, Slice value) throws IOException;
 
     /**
      * Returns number of bytes written to the underlying stream.
--- a/library/src/main/java/com/datatorrent/lib/fileaccess/TFileWriter.java
+++ b/library/src/main/java/com/datatorrent/lib/fileaccess/TFileWriter.java
@@ -25,6 +25,8 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.io.file.tfile.TFile.Writer;
 
+import com.datatorrent.netlet.util.Slice;
+
 /**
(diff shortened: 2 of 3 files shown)


brightchen:
@davidyan74 @chandnisingh Please review

library/src/main/java/com/datatorrent/lib/fileaccess/FileAccess.java
```diff
@@ -151,7 +151,7 @@
      * @param value
      * @throws IOException
      */
-    void append(byte[] key, byte[] value) throws IOException;
+    void append(Slice key, Slice value) throws IOException;
```
otterc, line 164: @brightchen I understand that this interface is evolving so making backward incompatible change with next version is okay. However, this interface exists for a long time. Is it better here to deprecate old methods and add new one?
In some cases, we have broken backward compatibility but the reason there was that all the api methods changed, not just one or two, and the change was quite drastic that existing implementation couldn't be re-used. Deprecating old ones would have just made the code messier.
On the other hand, if this is just being used by utilities in Malhar, we can break backward compatibility.  I think it will be good to get @tweise  input on this.

davidyan74: @chandnisingh 
Adding a method to an interface would also break backward compatibility for existing subclasses so declaring the old method deprecated and adding a new one actually does not help much except for the existing callers of the old method. The only caller of the old method in the entire Malhar is BucketsFileSystem. Granted there may be some code outside of Apex that will be affected by this change but I think it's not worth keeping around that method just because of that possibility.

otterc: @davidyan74 
Adding a method to an interface is a binary compatible change. 
https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html

davidyan74: Yes, but it would make existing subclasses not compilable. 
Anyway, my point is that this change does not actually affect anybody in reality.

otterc: We will modify existing implementations/utilities in Malhar to use the new ones anyways, but we don't break anything outside of Apex by deprecating old ones.

My point is that this is an old interface, and deprecating just one method and adding a new one, doesn't seem to cause much disruption. If HDHT moves to Malhar some day, having the old deprecated method may make the move easy. I don't have any strong opinion on this. However, it will be helpful to seek guidance from @tweise on this.

tweise: Keeping the old method and deprecating it seems a good choice when it does not cause extra complication and helps the downstream consumer.

davidyan74: OK, let's deprecate the old method then. @brightchen can you make the change?

PR #484, opened 2016-11-07
APEXMALHAR-2316 Moved suitable code to activate() for initializing tu…

KapoorHitesh:
…ple class attribute. @bhupeshchawda @chinmaykolhatkar please review.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
+++ b/library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
@@ -18,6 +18,7 @@
  */
 package com.datatorrent.lib.parser;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 
 import javax.xml.XMLConstants;
@@ -35,6 +36,7 @@ import javax.xml.validation.Validator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.w3c.dom.Document;
+import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 import org.apache.hadoop.classification.InterfaceStability;
 import org.apache.hadoop.conf.Configuration;
@@ -42,7 +44,9 @@ import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 
+import com.datatorrent.api.Context;
 import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.api.Operator;
 import com.datatorrent.lib.util.ReusableStringReader;
 import com.datatorrent.netlet.util.DTThrowable;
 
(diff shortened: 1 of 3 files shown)


devtagare (commented):
Changes seem fine.
Could you please check why the build failed ?

KapoorHitesh:
@devtagare I analyzed the build failure and it appears to be failing because DocumentBuilder is using String argument instead of InputSource and hence the test application is throwing the same exception multiple times and the log limit is getting exceeded. 
Should I remove the test application and add it when I make the code changes for the above scenario?

KapoorHitesh:
@devtagare have made the appropriate code changes please review

library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
```diff
@@ -35,14 +36,18 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.w3c.dom.Document;
+import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
+
```
bhupeshchawda: remove unnecessary changes

library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
```diff
@@ -94,9 +100,8 @@ public void processTuple(String inputTuple)
         DocumentBuilder builder;
         try {
           builder = factory.newDocumentBuilder();
-          Document doc = builder.parse(inputTuple);
+          Document doc = builder.parse(new InputSource(new ByteArrayInputStream(inputTuple.getBytes("utf-8"))));
           parsedOutput.emit(doc);
-
```
bhupeshchawda, line 99: unnecessary

library/src/main/java/com/datatorrent/lib/parser/XmlParser.java
```diff
@@ -94,9 +100,8 @@ public void processTuple(String inputTuple)
         DocumentBuilder builder;
         try {
           builder = factory.newDocumentBuilder();
-          Document doc = builder.parse(inputTuple);
+          Document doc = builder.parse(new InputSource(new ByteArrayInputStream(inputTuple.getBytes("utf-8"))));
```
bhupeshchawda: Can you make it capital "UTF-8" ?

library/src/test/java/com/datatorrent/lib/parser/XmlParserApplicationTest.java
```diff
@@ -0,0 +1,93 @@
...
+{
+  @Test
+  public void testApplication() throws IOException, Exception
+  {
+    try {
+      LocalMode lma = LocalMode.newInstance();
+      Configuration conf = new Configuration(false);
+      lma.prepareDAG(new XmlParserTest(), conf);
+      LocalMode.Controller lc = lma.getController();
+      lc.run(10000);// runs for 10 seconds and quits
```
bhupeshchawda, line 67: Not the right way to run a test. Need to validate the results after the test is run.
See https://github.com/apache/apex-malhar/blob/master/stream/src/test/java/org/apache/apex/malhar/stream/FunctionOperator/FunctionOperatorTest.java for an example.

library/src/test/java/com/datatorrent/lib/parser/XmlParserApplicationTest.java
```diff
@@ -0,0 +1,93 @@
...
+      public static boolean emitTuple = true;
+
+      public final transient DefaultOutputPort<String> output = new DefaultOutputPort<String>();
+
+      @Override
+      public void emitTuples()
+      {
+        if (emitTuple) {
+          output.emit(xmlSample);
+          emitTuple = false;
```
bhupeshchawda: Why emit just one tuple?

bhupeshchawda (changes requested):

KapoorHitesh:
@bhupeshchawda have made the suggested code changes. Please review.

PR #345, opened 2016-07-15
APEXMALHAR-2130 REVIEW ONLY (WindowedOperator): incorporating Spillable data structures

davidyan74:

PR #485, opened 2016-11-08
APEXMALHAR-2284 *Review Only* 1) Implemented SpillableMap, SpillableArrayList, SpillableArrayListMultiMap over TimeSlicedBucketState 2) Integrated Spillable data structure into Inner Join operator

chaithu14:
1) Implemented SpillableMap, SpillableArrayList, SpillableArrayListMultiMap over TimeSlicedBucketState 2) Integrated Spillable data structure into Inner Join operator

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/join/AbstractManagedStateInnerJoinOperator.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/join/AbstractManagedStateInnerJoinOperator.java
@@ -18,18 +18,42 @@
  */
 package org.apache.apex.malhar.lib.join;
 
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import javax.annotation.Nullable;
+import javax.validation.constraints.NotNull;
 
 import org.joda.time.Duration;
-import org.apache.apex.malhar.lib.state.managed.ManagedTimeStateImpl;
-import org.apache.apex.malhar.lib.state.managed.ManagedTimeStateMultiValue;
+
+import org.apache.apex.malhar.lib.state.managed.KeyBucketExtractor;
+import org.apache.apex.malhar.lib.state.managed.TimeExtractor;
+import org.apache.apex.malhar.lib.state.spillable.SequentialSpillableIdentifierGenerator;
 import org.apache.apex.malhar.lib.state.spillable.Spillable;
+import org.apache.apex.malhar.lib.state.spillable.SpillableComplexComponent;
+import org.apache.apex.malhar.lib.state.spillable.SpillableIdentifierGenerator;
+import org.apache.apex.malhar.lib.state.spillable.SpillableTimeSlicedArrayListMultiMapImpl;
+import org.apache.apex.malhar.lib.state.spillable.SpillableTimeSlicedMapImpl;
+import org.apache.apex.malhar.lib.state.spillable.SpillableTimeStateStore;
+import org.apache.apex.malhar.lib.state.spillable.managed.ManagedTimeStateSpillableStore;
+import org.apache.apex.malhar.lib.utils.serde.Serde;
 import org.apache.hadoop.fs.Path;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Multiset;
+import com.google.common.collect.Sets;
+
 import com.datatorrent.api.Context;
 import com.datatorrent.api.DAG;
 import com.datatorrent.api.Operator;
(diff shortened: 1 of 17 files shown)


chaithu14:
@chandnisingh @bhupeshchawda @tweise Please review the PR.

library/src/main/java/org/apache/apex/malhar/lib/join/POJOInnerJoinOperator.java
```diff
@@ -245,4 +274,46 @@ public FieldObjectMap()
...
+      if (timeFieldGet == null) {
+        Class timeField = null;
+        try {
+          timeField = ClassUtils.primitiveToWrapper(o.getClass().getDeclaredField(timeFieldExpr).getType());
+        } catch (NoSuchFieldException e) {
+          throw new RuntimeException(e);
+        }
+        timeFieldGet = PojoUtils.createGetter(o.getClass(), timeFieldExpr, timeField);
+      }
+      return (long)timeFieldGet.get(o);
```
bhupeshchawda: Typecasting an object to long might not be a good idea.

library/src/main/java/org/apache/apex/malhar/lib/join/POJOInnerJoinOperator.java
```diff
@@ -245,4 +274,46 @@ public FieldObjectMap()
...
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public long getTime(Object o)
+    {
+      if (timeFieldGet == null) {
+        Class timeField = null;
+        try {
+          timeField = ClassUtils.primitiveToWrapper(o.getClass().getDeclaredField(timeFieldExpr).getType());
```
bhupeshchawda: I'll suggest to have a default implementation for for the time field as `Date` and create a getter for that.
Do not have a generic getter. We don't know what type the configured date field has.

library/src/main/java/org/apache/apex/malhar/lib/join/POJOInnerJoinOperator.java
```diff
@@ -245,4 +274,46 @@ public FieldObjectMap()
...
+    public long getBucket(Object o)
+    {
+      return o.hashCode() % getNoOfBuckets();
+    }
+  }
+
+  /**
+   * Extract the time value from Object
+   */
+  public class POJOExtractTime implements TimeExtractor<Object>
```
bhupeshchawda: Rename to `POJOTimeExtractor`?

library/src/main/java/org/apache/apex/malhar/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -90,17 +113,36 @@ public void createStores()
...
+  @SuppressWarnings("unchecked")
   @Override
   protected void processTuple(T tuple, boolean isStream1Data)
   {
     Spillable.SpillableListMultimap<K,T> store = isStream1Data ? stream1Data : stream2Data;
     K key = extractKey(tuple,isStream1Data);
-    long timeBucket = extractTime(tuple,isStream1Data);
-    if (!((ManagedTimeStateMultiValue)store).put(key, tuple,timeBucket)) {
+    //long timeBucket = extractTime(tuple,isStream1Data);
+    if (!store.put(key, tuple)) {
```
bhupeshchawda: Please mention why `put` would return false? Seems like all implementations return true by default.

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableTimeStateStore.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.apache.apex.malhar.lib.state.TimeSlicedBucketedState;
+import org.apache.apex.malhar.lib.state.managed.BucketProvider;
+import org.apache.hadoop.classification.InterfaceStability;
+
+import com.datatorrent.api.Component;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.Operator;
+
+@InterfaceStability.Evolving
```
bhupeshchawda, line 32: Please add java docs explaining use of this interface.

bhupeshchawda (commented):
I have some very initial comments. I haven't yet looked at all the changes.
Will add more comments soon.

bhupeshchawda:
@chaithu14 Can you update the description of the PR?

library/src/main/java/org/apache/apex/malhar/lib/join/AbstractManagedStateInnerJoinOperator.java
chaithu14: Yes you are right. Before inserting, need to check whether the tuple is expired or not. For this check, I added isExpiredTuple() API.

chaithu14:
Incorporated review comments. Added Unit test cases for new spillable structures and updated PR.

tweise:
Why do we need new spillable DS for this? Very similar work was done for WindowOperator recently, why can't it be reused? It may also be useful to call out the difference for the join functionality itself compared to what WindowOperator already supports. @davidyan74 any thoughts?

chaithu14:
@tweise Innerjoin operator uses the ManagedTimeStateImpl store which has the key bucket and time bucket as we need to put(keyBucket, timeBucket, key, value) call.
Existing spillable structures uses the BucketedState which has the below calls
             - put(bucketId, key, value) 
             - getSync(bucketId, key)  
             - getAsync(bucketId, key) where bucketId is either key bucket or time bucket. 
So, existing spillable structures can't be reused for InnerJoin operator.

davidyan74:
@chaithu14: Work is in progress to use ManagedTimeStateImpl for the spillable data structures. The work is currently being done by @siyuanh and it's recorded in this umbrella ticket https://issues.apache.org/jira/browse/APEXMALHAR-2130

davidyan74:
@tweise Yes, this PR looks like most of it we are already working on so I feel a bit uneasy about this PR because of duplication of similar work. 

About your comment about the join functionality between this and WindowedOperator, the current innerjoin operator does not require the other concepts required by the WindowedOperator (watermark, triggering etc) so that may be a reason to continue development on the inner join operator. Whether it's a big enough reason or not is debatable. 

My take is that we should halt further development of the join operator and evaluate the WindowedMergeOperator and see if there is anything missing from the current use cases of the join operator.

tweise:
@davidyan74 this is blocking the release. But APEXMALHAR-2130 was also supposed to be finished for 3.6.0. What is that status, can it be completed and used to make this join operator functional until we decide about longer term plan?

davidyan74:
OK, I'm asking @siyuanh to update the status of the affected tickets.

chaithu14:
@davidyan74 Are you talking about this PR (#345)? If yes, then the duplication is very minimal and I have seen only the interface (SpillableTimeStateStore) is common. 
Spillable structures in this PR are over the TimeSlicedBucketedState and timeExtractor be the time extract from value.

ilganeli:
@chaithu14 Is this work still relevant or can this be closed?

PR #486, opened 2016-11-08
APEXMALHAR-2278 Implement KuduNonTransactional Output Operator

ghost:
@tweise  / @PramodSSImmaneni  : Could you please review or suggest the right person who can complete this review for me?

### DIFF
--- a/contrib/pom.xml
+++ b/contrib/pom.xml
@@ -678,5 +678,11 @@
       <artifactId>jackson-databind</artifactId>
       <version>2.7.0</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.kudu</groupId>
+      <artifactId>kudu-client</artifactId>
+      <version>1.3.0</version>
+      <optional>true</optional>
+    </dependency>
   </dependencies>
 </project>
--- a/contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
+++ b/contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
@@ -1,0 +1,660 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.contrib.kudu;
+
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.nio.ByteBuffer;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import javax.validation.constraints.NotNull;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.apache.apex.malhar.lib.wal.FSWindowDataManager;
+import org.apache.apex.malhar.lib.wal.WindowDataManager;
+import org.apache.hadoop.classification.InterfaceStability;
+import org.apache.kudu.ColumnSchema;
+import org.apache.kudu.Schema;
+import org.apache.kudu.Type;
+import org.apache.kudu.client.Delete;
+import org.apache.kudu.client.Insert;
+import org.apache.kudu.client.KuduClient;
+import org.apache.kudu.client.KuduException;
+import org.apache.kudu.client.KuduSession;
+import org.apache.kudu.client.KuduTable;
+import org.apache.kudu.client.Operation;
+import org.apache.kudu.client.PartialRow;
+import org.apache.kudu.client.Statistics;
+import org.apache.kudu.client.Update;
+import org.apache.kudu.client.Upsert;
+import com.datatorrent.api.AutoMetric;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.DefaultInputPort;
+import com.datatorrent.api.Operator;
+import com.datatorrent.api.annotation.Stateless;
+import com.datatorrent.common.util.BaseOperator;
+import com.datatorrent.lib.util.PojoUtils;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * An Abstract operator that would allow concrete implementations to write a POJO value into a given Kudu table.
+ * <p>
+ * To use this operator, the following needs to be done by the implementor
+ * <ol>
+ * <li>Create a concrete implementation of this operator and implement the method to define the connection
+ * properties to Kudu. The connection properties are set using the {@link ApexKuduConnection} using a builder pattern.
+ * </li>
+ * <li>Implement the logic how tuples need to ne handled in the event of a reconciliation phase ( i.e. when an
+ * operator is resuming back from failure ). See javadoc of the method for more details.</li>
+ * <li>Define the payload class</li>
+ * </ol>
+ * </p>
+ * <p>
+ * Note that the tuple that is getting processed is not the POJO class. The tuple that is getting processed is the
+ * {@link KuduExecutionContext} instance. This is because the operator supports mutation types as a higher level
+ * construct than simply writing a POJO into a Kudu table row.
+ * </p>
+ * <p>
+ * Supported mutations are:
+ * <ol>
+ *      <li>INSERT</li>
+ *      <li>UPDATE</li>
+ *      <li>UPSERT</li>
+ *      <li>DELETE</li>
+ * </ol>
+ * </p>
+ * <p>
+ * Please note that the Update mutation allows to change a subset of columns only even though there might be columns
+ * that were defined to be non-nullable. This is because the original mutation of type insert would have written the
+ * mandatory columns. In such scenarios, the method setDoNotWriteColumns() in {@link KuduExecutionContext} can be
+ * used to specify only those columns that need an update. This ways a read and Update pattern can be merged to a
+ * simple update pattern thus avoiding a read if required.</p>
+ * */
+@InterfaceStability.Evolving
+public abstract class AbstractKuduOutputOperator extends BaseOperator
+    implements Operator.ActivationListener<Context.OperatorContext>,Operator.CheckpointNotificationListener
+{
+
+  private transient ApexKuduConnection apexKuduConnection;
+
+  private transient  KuduTable kuduTable;
+
+  private transient  KuduSession kuduSession;
+
+  private transient KuduClient kuduClientHandle;
+
+  private transient Map<String,ColumnSchema> allColumnDefs;
+
+  private transient Map<String,Object> kuduColumnBasedGetters;
+
+  private Set<String> primaryKeyColumnNames;
+
+  private static final transient Logger LOG = LoggerFactory.getLogger(AbstractKuduOutputOperator.class);
+
+  @NotNull
+  private WindowDataManager windowDataManager;
+
+  private transient long currentWindowId;
+
+  private transient boolean isInReplayMode;
+
+  private transient boolean isInReconcilingMode;
+
+  private transient long reconcilingWindowId;
+
+  @AutoMetric
+  transient long numInserts;
+
+  @AutoMetric
+  transient long numUpserts;
+
+  @AutoMetric
+  transient long numDeletes;
+
+  @AutoMetric
+  transient long numUpdates;
+
+  @AutoMetric
+  transient long numOpsErrors;
+
+  @AutoMetric
+  transient long numBytesWritten;
+
+  @AutoMetric
+  transient long numRpcErrors;
+
+  @AutoMetric
+  transient long numWriteOps;
+
+  @AutoMetric
+  transient long numWriteRPCs;
+
+  @AutoMetric
+  long totalOpsErrors = 0;
+
+  @AutoMetric
+  long totalBytesWritten = 0;
+
+  @AutoMetric
+  long totalRpcErrors = 0;
+
+  @AutoMetric
+  long totalWriteOps = 0;
+
+  @AutoMetric
+  long totalWriteRPCs = 0;
+
+  @AutoMetric
+  long totalInsertsSinceStart;
+
+  @AutoMetric
+  long totalUpsertsSinceStart;
+
+  @AutoMetric
+  long totalDeletesSinceStart;
+
+  @AutoMetric
+  long totalUpdatesSinceStart;
+
+  public final transient DefaultInputPort<KuduExecutionContext> input = new DefaultInputPort<KuduExecutionContext>()
+  {
+    @Override
+    public void process(KuduExecutionContext kuduExecutionContext)
+    {
+      processTuple(kuduExecutionContext);
+    }
+  }; // end input port implementation
+
+  public void processTuple(KuduExecutionContext kuduExecutionContext)
+  {
+    if ( isInReconcilingMode || isInReplayMode) {
+      if ( !isEligibleForPassivationInReconcilingWindow(kuduExecutionContext, currentWindowId)) {
+        return;
+      }
+    }
+    KuduMutationType mutationType = kuduExecutionContext.getMutationType();
+    switch (mutationType) {
+      case DELETE:
+        processForDelete(kuduExecutionContext);
+        numDeletes += 1;
+        totalDeletesSinceStart += 1;
+        break;
+      case INSERT:
+        processForInsert(kuduExecutionContext);
+        numInserts += 1;
+        totalInsertsSinceStart += 1;
+        break;
+      case UPDATE:
+        processForUpdate(kuduExecutionContext);
+        numUpdates += 1;
+        totalUpdatesSinceStart += 1;
+        break;
+      case UPSERT:
+        processForUpsert(kuduExecutionContext);
+        numUpserts += 1;
+        totalUpsertsSinceStart += 1;
+        break;
+      default:
+        break;
+    }
+  }
+
+  /***
+   * Sets the values from the Pojo into the Kudu mutation object.
+   * @param currentOperation The operation instance that represents the current mutation. This will be applied to the
+   *                         current session
+   * @param kuduExecutionContext The tuple that contains the payload as well as other information like mutation type etc
+   */
+  @SuppressWarnings(value = "unchecked")
+  private void performCommonProcessing(Operation currentOperation, KuduExecutionContext kuduExecutionContext)
+  {
+    currentOperation.setExternalConsistencyMode(kuduExecutionContext.getExternalConsistencyMode());
+    currentOperation.setPropagatedTimestamp(kuduExecutionContext.getPropagatedTimestamp());
+    PartialRow partialRow = currentOperation.getRow();
+    Object payload = kuduExecutionContext.getPayload();
+    Set<String> doNotWriteColumns = kuduExecutionContext.getDoNotWriteColumns();
+    if (doNotWriteColumns == null) {
+      doNotWriteColumns = new HashSet<>();
+    }
+    for (String columnName: kuduColumnBasedGetters.keySet()) {
+      if ( doNotWriteColumns.contains(columnName)) {
+        continue;
+      }
+      ColumnSchema columnSchema = allColumnDefs.get(columnName);
+      Type dataType = columnSchema.getType();
+      try {
+        switch (dataType) {
+          case STRING:
+            PojoUtils.Getter<Object, String> stringGetter = ((PojoUtils.Getter<Object, String>)kuduColumnBasedGetters
+                .get(columnName));
+            if (stringGetter != null) {
+              final String stringValue = stringGetter.get(payload);
+              if (stringValue != null) {
+                partialRow.addString(columnName, stringValue);
+              }
+            }
+            break;
+          case BINARY:
+            PojoUtils.Getter<Object, ByteBuffer> byteBufferGetter = ((PojoUtils.Getter<Object, ByteBuffer>)
+                kuduColumnBasedGetters.get(columnName));
+            if (byteBufferGetter != null) {
+              final ByteBuffer byteBufferValue = byteBufferGetter.get(payload);
+              if (byteBufferValue != null) {
+                partialRow.addBinary(columnName, byteBufferValue);
+              }
+            }
+            break;
+          case BOOL:
+            PojoUtils.GetterBoolean<Object> boolGetter = ((PojoUtils.GetterBoolean<Object>)kuduColumnBasedGetters.get(
+                columnName));
+            if (boolGetter != null) {
+              final boolean boolValue = boolGetter.get(payload);
+              partialRow.addBoolean(columnName, boolValue);
+            }
+            break;
+          case DOUBLE:
+            PojoUtils.GetterDouble<Object> doubleGetter = ((PojoUtils.GetterDouble<Object>)kuduColumnBasedGetters.get(
+                columnName));
+            if (doubleGetter != null) {
+              final double doubleValue = doubleGetter.get(payload);
+              partialRow.addDouble(columnName, doubleValue);
+            }
+            break;
+          case FLOAT:
+            PojoUtils.GetterFloat<Object> floatGetter = ((PojoUtils.GetterFloat<Object>)kuduColumnBasedGetters.get(
+                columnName));
+            if (floatGetter != null) {
+              final float floatValue = floatGetter.get(payload);
+              partialRow.addFloat(columnName, floatValue);
+            }
+            break;
+          case INT8:
+            PojoUtils.GetterByte<Object> byteGetter = ((PojoUtils.GetterByte<Object>)kuduColumnBasedGetters.get(
+                columnName));
+            if (byteGetter != null) {
+              final byte byteValue = byteGetter.get(payload);
+              partialRow.addByte(columnName, byteValue);
+            }
+            break;
+          case INT16:
+            PojoUtils.GetterShort<Object> shortGetter = ((PojoUtils.GetterShort<Object>)kuduColumnBasedGetters.get(
+                columnName));
+            if (shortGetter != null) {
+              final short shortValue = shortGetter.get(payload);
+              partialRow.addShort(columnName, shortValue);
+            }
+            break;
+          case INT32:
+            PojoUtils.GetterInt<Object> intGetter = ((PojoUtils.GetterInt<Object>)
+                kuduColumnBasedGetters.get(columnName));
+            if (intGetter != null) {
+              final int intValue = intGetter.get(payload);
+              partialRow.addInt(columnName, intValue);
+            }
+            break;
+          case INT64:
+          case UNIXTIME_MICROS:
+            PojoUtils.GetterLong<Object> longGetter = ((PojoUtils.GetterLong<Object>)kuduColumnBasedGetters.get(
+                columnName));
+            if (longGetter != null) {
+              final long longValue = longGetter.get(payload);
+              partialRow.addLong(columnName, longValue);
+            }
+            break;
+          default:
+            LOG.error(columnName + " is not of the supported data type");
+            throw new UnsupportedOperationException("Kudu does not support data type for column " + columnName);
+        }
+      } catch ( Exception ex ) {
+        LOG.error(" Exception while fetching the value of " + columnName + " because " + ex.getMessage());
+        partialRow.setNull(columnName);
+      }
+    }
+    try {
+      kuduSession.apply(currentOperation);
+    } catch (KuduException e) {
+      LOG.error("Could not execute operation because " + e.getMessage(), e);
+      throw new RuntimeException(e.getMessage());
+    }
+  }
+
+  protected void processForUpdate(KuduExecutionContext kuduExecutionContext)
+  {
+    Update thisUpdate = kuduTable.newUpdate();
+    performCommonProcessing(thisUpdate,kuduExecutionContext);
+  }
+
+
+  protected void processForUpsert(KuduExecutionContext kuduExecutionContext)
+  {
+    Upsert thisUpsert = kuduTable.newUpsert();
+    performCommonProcessing(thisUpsert,kuduExecutionContext);
+  }
+
+
+
+  protected void processForDelete(KuduExecutionContext kuduExecutionContext)
+  {
+    Delete thisDelete = kuduTable.newDelete();
+    // Kudu does not allow column values to be set in case of a delete mutation
+    Set<String> doNotWriteCols = kuduExecutionContext.getDoNotWriteColumns();
+    if ( doNotWriteCols == null) {
+      doNotWriteCols = new HashSet<>();
+    }
+    doNotWriteCols.clear();
+    for (String columnName : allColumnDefs.keySet()) {
+      if ( !(primaryKeyColumnNames.contains(columnName))) {
+        doNotWriteCols.add(columnName);
+      }
+    }
+    kuduExecutionContext.setDoNotWriteColumns(doNotWriteCols);
+    performCommonProcessing(thisDelete,kuduExecutionContext);
+  }
+
+
+  protected void processForInsert(KuduExecutionContext kuduExecutionContext)
+  {
+    Insert thisInsert = kuduTable.newInsert();
+    performCommonProcessing(thisInsert,kuduExecutionContext);
+  }
+
+  @Override
+  public void activate(Context.OperatorContext context)
+  {
+    ApexKuduConnection.ApexKuduConnectionBuilder apexKuduConnectionBuilder = getKuduConnectionConfig();
+    apexKuduConnection = apexKuduConnectionBuilder.build();
+    checkNotNull(apexKuduConnection,"Kudu connection cannot be null");
+    kuduTable = apexKuduConnection.getKuduTable();
+    kuduSession = apexKuduConnection.getKuduSession();
+    kuduClientHandle = apexKuduConnection.getKuduClient();
+    checkNotNull(kuduTable,"Kudu Table cannot be null");
+    checkNotNull(kuduSession, "Kudu session cannot be null");
+    allColumnDefs = new HashMap();
+    primaryKeyColumnNames = new HashSet<>();
+    kuduColumnBasedGetters = new HashMap();
+    buildGettersForPojoPayload();
+    reconcilingWindowId = Stateless.WINDOW_ID;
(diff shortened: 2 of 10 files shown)


tweise:
Great to see this addition!

Package for this should probably be: org.apache.apex.malhar.contrib.kudu

I would suggest to send a separate email to dev@ summarizing the contribution and asking for reviewers.

ghost:
@tweise - I shall refactor to the above package structure, test it with the latest release and then send a mail to dev for feedback and subsequently review

tweise:
@tushargosavi @bhupeshchawda can you help reviewing this PR?

ghost:
@tweise  : Apologies I got side tracked due to other commitments. 
Regarding the movement of the code to contrib/org/apache/apex package structure instead of the original contrib/com/datatorrrent/....., I see that this will be the only package sitting in that structure. I was wondering if org/apache/apex is the package structure going forward for all new additions ? 

@tushargosavi / @bhupeshchawda  : I will send a summary and description of the implementation of this PR today to the dev mailing list.

tweise:
@ananthc yes, that is the correct package structure. All new code needs to be added under org.apache.apex packages.

ghost:
@tushargosavi / @bhupeshchawda : sent a description to the dev mailing list regarding the design. Please let me know if any changes are required for the design and I will resubmit the PR based on the inputs.

tushargosavi:
@ananthc I will start reviewing it.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+import com.datatorrent.api.Operator;
+
+
+import com.datatorrent.api.annotation.InputPortFieldAnnotation;
+import com.datatorrent.api.annotation.Stateless;
+import com.datatorrent.common.util.BaseOperator;
+import com.datatorrent.lib.util.PojoUtils;
+
+
+
```
bhupeshchawda: Why empty lines?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+      default:
+        break;
+    }
+    if (getter != null) {
+      kuduColumnBasedGetters.put(columnSchema.getName(),getter);
+    }
+  }
+
+
+  public static ApexKuduConnection.ApexKuduConnectionBuilder usingKuduConnectionBuilder()
```
bhupeshchawda, line 610: where is this being used?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+ * mandatory columns. In such scenarios, the method setDoNotWriteColumns() in {@link KuduExecutionContext} can be
+ * used to specify only those columns that need an update. This ways a read and Update pattern can be merged to a
+ * simple update pattern thus avoiding a read if required.</p>
+ * */
+@InterfaceStability.Evolving
+public abstract class AbstractKuduOutputOperator extends BaseOperator
+    implements Operator.ActivationListener<Context.OperatorContext>,Operator.CheckpointNotificationListener
+{
+
+  private ApexKuduConnection apexKuduConnection;
```
bhupeshchawda: transient?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+  private transient long currentWindowId;
+
+  private transient boolean isInSafeMode;
+
+  private transient boolean isInReconcilingMode;
+
+  private transient long reconcilingWindowId;
+
+  @AutoMetric
+  transient long numInsertsInThisWindow;
```
bhupeshchawda: `private transient long`?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+  transient long rpcErrorsInThisWindow;
+
+  @AutoMetric
+  transient long writeOpsInThisWindow;
+
+  @AutoMetric
+  transient long writeRPCsInThisWindow;
+
+  @AutoMetric
+  long totalOpsErrors = 0;
```
bhupeshchawda: `private transient long`?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+  @AutoMetric
+  long totalUpsertsSinceStart;
+
+  @AutoMetric
+  long totalDeletesSinceStart;
+
+  @AutoMetric
+  long totalUpdatesSinceStart;
+
+  @InputPortFieldAnnotation
```
bhupeshchawda: Why empty annotation?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+  public final transient DefaultInputPort<KuduExecutionContext> input = new DefaultInputPort<KuduExecutionContext>()
+  {
+    @Override
+    public void process(KuduExecutionContext kuduExecutionContext)
```
bhupeshchawda: Can we create a new method `processTuple(KuduExecutionContext kuduExecutionContext)` and call it within `process`? Helps when extending functionality.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+
+  private transient Map<String,ColumnSchema> allColumnDefs;
+
+  private transient Map<String,Object> kuduColumnBasedGetters;
+
+  private Set<String> primaryKeyColumnNames;
+
+  private static final transient Logger LOG = LoggerFactory.getLogger(AbstractKuduOutputOperator.class);
+
+  private WindowDataManager windowDataManager;
```
bhupeshchawda: `@NotNull`?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+    }
+  }
+
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    super.setup(context);
+    windowDataManager = getWindowDataManager();
+    if ( windowDataManager == null) {
+      windowDataManager = new FSWindowDataManager();
```
bhupeshchawda, line 454: Should we use `WindowDataManager.NoopWindowDataManager()` as the default window data manager? If the user needs idempotency, then `FSWindowDataManager` must be set explicitly.

contrib/pom.xml
```diff
@@ -679,5 +679,10 @@
       <artifactId>jackson-databind</artifactId>
       <version>2.7.0</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.kudu</groupId>
+      <artifactId>kudu-client</artifactId>
+      <version>1.3.0</version>
```
tushargosavi, line 684: you can mark the dependency as optional.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+    kuduSession = apexKuduConnection.getKuduSession();
+    kuduClientHandle = apexKuduConnection.getKuduClient();
+    checkNotNull(kuduTable,"Kudu Table cannot be null");
+    checkNotNull(kuduSession, "Kudu session cannot be null");
+    allColumnDefs = new HashMap();
+    primaryKeyColumnNames = new HashSet<>();
+    kuduColumnBasedGetters = new HashMap();
+    buildGettersForPojoPayload();
+    isInSafeMode = false;
+    isInReconcilingMode = false;
```
bhupeshchawda: Redundant initializations for `isInSafeMode` and `isInReconcilingMode`

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+    reconcilingWindowId = Stateless.WINDOW_ID;
+    if ( (context.getValue(Context.OperatorContext.ACTIVATION_WINDOW_ID) != Stateless.WINDOW_ID) &&
+        context.getValue(Context.OperatorContext.ACTIVATION_WINDOW_ID) <
+        windowDataManager.getLargestCompletedWindow()) {
+      isInSafeMode = true;
+      reconcilingWindowId = windowDataManager.getLargestCompletedWindow() + 1;
+      isInReconcilingMode = false;
```
bhupeshchawda: redundant

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+    super.teardown();
+    windowDataManager.teardown();
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
+    super.beginWindow(windowId);
+    currentWindowId = windowId;
+    if ( currentWindowId != Stateless.WINDOW_ID) {
```
bhupeshchawda: This condition will always be true. unnecessary..

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+  /***
+   * This is used to give control to the concrete implementation of this operator how to resolve whether to write a
+   * mutation into a given kudu table in the event of a failure and the operator subsequently resumes. This window
+   * is marked as a reconciling window. It is only for this reconciling window that we need to give control to the
+   * concrete operator implementor how to actually resolve if the entry needs to be excuted as a mutation in Kudu.
+   * @param executionContext The tuple which represents the execution context along with the payload
+   * @param reconcilingWindowId The window Id of the reconciling window
+   * @return true if we would like the entry to result in a mutation in the Kudu table.
+   */
+  protected abstract boolean isEligibleForPassivationInReconcilingWindow(KuduExecutionContext executionContext,
```
bhupeshchawda, line 657: Is this supposed to have some kind of side effect on the kudu table? If so, can we name it differently?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+
+  private transient boolean isInReconcilingMode;
+
+  private transient long reconcilingWindowId;
+
+  @AutoMetric
+  transient long numInsertsInThisWindow;
+
+  @AutoMetric
+  transient long numUpsertsInThisWindow;
```
bhupeshchawda: can we remove `InThisWindow` from the metric names?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+  @AutoMetric
+  long totalUpdatesSinceStart;
+
+  @InputPortFieldAnnotation
+  public final transient DefaultInputPort<KuduExecutionContext> input = new DefaultInputPort<KuduExecutionContext>()
+  {
+    @Override
+    public void process(KuduExecutionContext kuduExecutionContext)
+    {
+      if (kuduExecutionContext == null) {
```
bhupeshchawda: Is this possible? Can we log this event?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+        return;
+      }
+      if ( isInSafeMode) {
+        return;
+      }
+      if ( isInReconcilingMode) {
+        if ( !isEligibleForPassivationInReconcilingWindow(kuduExecutionContext, currentWindowId)) {
+          return;
```
bhupeshchawda: The current window is a new window - one that has not been processed yet. Do we want to decide whether or not to process this record depending just on `isEligibleForPassivationInReconcilingWindow()`?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+  transient long writeOpsInThisWindow;
+
+  @AutoMetric
+  transient long writeRPCsInThisWindow;
+
+  @AutoMetric
+  long totalOpsErrors = 0;
+
+  @AutoMetric
+  long totalBytesWritten = 0;
```
bhupeshchawda, line 158: I think we can remove the `total*` metrics. AFAIK the total metrics could be derived from the per window stats. @tushargosavi  comments?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+          break;
+        case UPSERT:
+          processForUpsert(kuduExecutionContext);
+          numUpsertsInThisWindow += 1;
+          totalUpsertsSinceStart += 1;
+          break;
+        default:
+          break;
+      }
+      kuduSession.getPendingErrors().getRowErrors();
```
tushargosavi: The return value is not used.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+    totalOpsErrors = kuduClientHandle.getStatistics().getClientStatistic(Statistics.Statistic.OPS_ERRORS);
+    totalBytesWritten = kuduClientHandle.getStatistics().getClientStatistic(Statistics.Statistic.BYTES_WRITTEN);
+    totalRpcErrors = kuduClientHandle.getStatistics().getClientStatistic(Statistics.Statistic.RPC_ERRORS);
+    totalWriteOps = kuduClientHandle.getStatistics().getClientStatistic(Statistics.Statistic.WRITE_OPS);
+    totalWriteRPCs = kuduClientHandle.getStatistics().getClientStatistic(Statistics.Statistic.WRITE_RPCS);
+  }
+
+  private void buildGettersForPojoPayload()
+  {
+    Class payloadClass = getTuplePayloadClass();
```
bhupeshchawda, line 525: how about using a parametrized constructor to initialize the payload class?

bhupeshchawda (commented):
Made a few comments. Will add more.
Can we also add a concrete implementation?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+
+  @Override
+  public void endWindow()
+  {
+    super.endWindow();
+    try {
+      kuduSession.flush();
+    } catch (KuduException e) {
+      LOG.error("Could not flush kudu session on an end window boundary " + e.getMessage(), e);
+      throw new RuntimeException(e.getMessage());
```
tushargosavi: you can use entire exception instead of just e.getMessage while creating RuntimeException.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,666 @@
...
+    for ( Field aFieldDef : classFields) {
+      String currentFieldName = aFieldDef.getName().toLowerCase();
+      if (allKuduTableColumnNames.contains(currentFieldName)) {
+        extractGetterForColumn(normalizedColumns.get(currentFieldName),aFieldDef);
+      } else {
+        if (columnNameOverrides.containsKey(aFieldDef.getName())) {
+          extractGetterForColumn(normalizedColumns.get(columnNameOverrides.get(aFieldDef.getName()).toLowerCase()),
+              aFieldDef);
+        }
+        if (columnNameOverrides.containsKey(aFieldDef.getName().toLowerCase())) {
```
tushargosavi: else if ??
If you have aFieldDef.getName() == aFieldDef.getName().toLowerCase() then getter will be created twice.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: This metric is a different one. The totals of course are simply summed up but that is only applicable since the application start ( and those metrics are transient by nature). The ones above are named total because they represent the total metrics since the applications first launch ( and hence not marked transient for the same reason )

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: Since the default value of optional suffices ? ( defaults to optional = false )

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: Good idea. Refactored the code accordingly.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: I think FSWindowManager makes a sense default for the following reason:

We are actually using the values stored in the WindowManager to implement logic to identify states related
1. A set of tuples are being replayed from the upstream buffer server
2. Identifying a reconcile window ( The first window that is resuming since the last exit of the window ) 

In the reconciling window phase, we are transferring the onus of "EXACTLY_ONCE" semantics to the end user ( please see below comments for this aspect ) so that the business logic can be used to decide whether we want to mutate the kudu store given the above two situations.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: Resolved this implementing a Default implementation for the Abstract Kudu output operator. This default implementation would accept a properties file ( along with a no arg constructor to use a default property file name ). The property file would help to specify the above value as well as a few others. The java docs were updated to reflect this accordingly.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: This will be used in all concrete implementations of the AbstractKuduOutputOperator. This is allowing for a builder pattern to define the kudu table name, connection parameters and behaviour. The newly added BaseKuduOutputOperator which extends this abstract class uses this pattern to instantiate the underlying Async client to the Kudu cluster.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: This is helping in achieving correct semantics for "EXACTLY_ONCE". In scenarios when the tuples are being replayed or the operator is recovering after an abrupt crash ( termed as a reconciling window ) , we call this method to let the user decide whether s/he would like to use this incoming tuple to be used in mutating the kudu store. 

An example implementation would be to check ( perform a read in kudu store ) and then decide whether this tuple is already processed etc. Since this is a business level check, we let the user take control of deciding whether this tuple needs to be mutated into the Kudu store. Note that the Default Kudu output operator is working in a "ATLEAST_ONCE" semantics as the default implementation is true for this method.

contrib/pom.xml
ghost: done

ghost:
@tushargosavi / @bhupeshchawda  - Worked on the review comments and addressed them accordingly.

tushargosavi:
LGTM , I will wait for @bhupeshchawda review before merge.

vrozov:
Can it be moved out of contrib to a separate module similar to kafka?

ghost:
@vrozov : sure . @tweise suggested we have a different package structure under contrib (under the org/apache path).. Why was Kafka taken out of contrib and made separate ? ( My understanding was that we had multiple versions of the Kafka drivers that need to be colocated and hence moved out of contrib..) .. I can refactor if you think it should be out of contrib  as the dependencies do not seem to be that big at all to impact other modules...

tweise:
@vrozov I would like such restructuring discussed and taken up separately. This PR has been open for a long time and it appears the original reviewers are quite happy with it.

I have my own thoughts on what should be a separate module, and there maturity of the operator, tests and documentation play in. That discussion goes beyond the Kudu operator.

vrozov:
@tweise I'd suggest that you initiate a discussion on dev@apex on the policy for new operators in the malhar main and contrib libraries. For this particular PR my concern is introducing a brand new operator with "org.apache.apex.malhar.*contrib*.kudu" package name. Having contrib in the name means that once an operator matures and is moved into a separate module, it's name will change affecting applications. Why not to mature an operator in a separate module?

ghost:
@bhupeshchawda  : Are there any further code review comments that you would like to see in this operator ?

bhupeshchawda:
@ananthc Sorry for the delay. Will have a look today..

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
bhupeshchawda: As long as the implementation does not do a write on the kudu table, it should be fine..

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
bhupeshchawda: I think you can remove this. It will be non-optional anyway.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,662 @@
...
+          case UNIXTIME_MICROS:
+            PojoUtils.GetterLong<Object> longGetter = ((PojoUtils.GetterLong<Object>)kuduColumnBasedGetters.get(
+                columnName));
+            if (longGetter != null) {
+              final long longValue = longGetter.get(payload);
+              partialRow.addLong(columnName, longValue);
+            }
+            break;
+          default:
+            break;
```
bhupeshchawda: throw unsupported operation exception here?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,662 @@
...
+        }
+      } catch ( Exception ex ) {
+        LOG.debug(" Exception while fetching the value of " + columnName + " because " + ex.getMessage());
```
bhupeshchawda: `LOG.error`?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,662 @@
...
+    if ( windowDataManager == null) {
+      windowDataManager = new FSWindowDataManager();
+    }
+    windowDataManager.setup(context);
+  }
+
+  @Override
+  public void teardown()
+  {
+    super.teardown();
```
bhupeshchawda: unnecessary

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,662 @@
...
+    windowDataManager.teardown();
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
+    super.beginWindow(windowId);
```
bhupeshchawda, line 468: unnecessary

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
```diff
@@ -0,0 +1,662 @@
...
+        break;
+      case INT32:
+        getter = PojoUtils.createGetterInt(pojoClass, fieldDefinition.getName());
+        break;
+      case INT64:
+      case UNIXTIME_MICROS:
+        getter = PojoUtils.createGetterLong(pojoClass, fieldDefinition.getName());
+        break;
+      default:
+        break;
```
bhupeshchawda: unsupported exception?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/ApexKuduConnection.java
```diff
@@ -0,0 +1,223 @@
...
+{
+  private static final long serialVersionUID = 4720185362997969198L;
+
+  private transient  KuduSession kuduSession;
+
+  private transient  KuduTable kuduTable;
+
+  private transient  KuduClient kuduClient;
+
+  Logger LOG = LoggerFactory.getLogger(ApexKuduConnection.class);
```
bhupeshchawda: public static final?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/ApexKuduConnection.java
```diff
@@ -0,0 +1,223 @@
...
+
+    int workerThreads = 2 *  Runtime.getRuntime().availableProcessors();
+
+    boolean isWorkerThreadsCountSet = false;
+
+    long socketReadTimeOutMs = 10000;
+
+    boolean isSocketReadTimeOutSet = false;
+
+    long operationTimeOutMs = 30000;
```
bhupeshchawda, line 150: can the constant values be declared as public static final fields and then used?

contrib/src/test/java/org/apache/apex/malhar/contrib/kudu/KuduCreateUpdateDeleteOutputOperatorTest.java
```diff
@@ -0,0 +1,336 @@
...
+    RowResultIterator rowResultItr = scanner.nextRows();
+    while (rowResultItr.hasNext()) {
+      RowResult thisRow = rowResultItr.next();
+      keyInfo.setFloatdata(thisRow.getFloat("floatdata"));
+      keyInfo.setBooldata(thisRow.getBoolean("booldata"));
+      keyInfo.setBinarydata(thisRow.getBinary("binarydata"));
+      keyInfo.setLongdata(thisRow.getLong("longdata"));
+      keyInfo.setTimestampdata(thisRow.getLong("timestampdata"));
+      keyInfo.setStringdata("stringdata");
+      break;
```
bhupeshchawda, line 196: break??

contrib/src/test/java/org/apache/apex/malhar/contrib/kudu/KuduCreateUpdateDeleteOutputOperatorTest.java
```diff
@@ -0,0 +1,336 @@
+/**
```
bhupeshchawda, line 1: Does the test expect a Kudu server running on localhost?

bhupeshchawda (commented):
@ananthc added a few minor comments. Please see.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: I can do that .. A quick question though .. Do we use the annotation for any special processing inside the apex core ? i.e. get all annotated fields that have this annotation and do any extra processing ? What is the annotation used for if it is something that can be ignored in the first place ?

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: I am sorry .. I did not understand what you mean by the above comment .

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/ApexKuduConnection.java
ghost: Well this is part of the builder pattern and hence:
1. It cannot be final
2. It is ideal this not be static as well as there might be use cases where one would like to use multiple instances of the kudu output operator in a single application each one having its own connection properties that would be built by the builder. 
3. It is not public because it is part of the builder pattern.

contrib/src/test/java/org/apache/apex/malhar/contrib/kudu/KuduCreateUpdateDeleteOutputOperatorTest.java
ghost: no. I was told that the tests are explicitly disabled in the contrib packages and hence this should not effect ? The default implementation looks for a property file name to use the host names that it will use to connect to. The java doc of the operator gives the file name as well as the property key names.

contrib/src/test/java/org/apache/apex/malhar/contrib/kudu/KuduCreateUpdateDeleteOutputOperatorTest.java
ghost: yes. The method is a "utility" method that is used across other tests in the same test class. All of these would need the first row that matches the lookup criteria. The loop is broken after resolving the first record that matches the criteria and hence the break in case there are multiple entries.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: done

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/ApexKuduConnection.java
ghost: done

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: My thinking is that it is a good practice to call the super methods especially when you are extending the base operator ( as the word base meant that it might have common logic and hence the call to super)

Removing it though as the base does not have any useful implementation as it stands today.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: My thinking is that it is a good practice to call the super methods especially when you are extending the base operator ( as the word base meant that it might have common logic and hence the call to super)

Removing this call as the base implementation does not have anything useful.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: done

ghost:
@bhupeshchawda : addressed the second set of review comments as well. Let me know if you think there is any more review comments that you think might need to be addressed.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
ghost: done

contrib/src/test/java/org/apache/apex/malhar/contrib/kudu/KuduCreateUpdateDeleteOutputOperatorTest.java
bhupeshchawda: It seems the `while` is just populating keyinfo. `while` could be replaced by an `if`.

contrib/src/test/java/org/apache/apex/malhar/contrib/kudu/KuduCreateUpdateDeleteOutputOperatorTest.java
bhupeshchawda: Yes, you are right. Forgot it was contrib.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/ApexKuduConnection.java
bhupeshchawda: What I meant was for the "constants" to be public static final default values, not the variables.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/AbstractKuduOutputOperator.java
bhupeshchawda: You are right, should be okay; however sometimes it implies something is happening in super :-)

bhupeshchawda (approved):
Have a few nitpicks. LGTM too.

ghost:
@tweise : Could you please merge if you think it is ok ?

bhupeshchawda:
@ananthc Will wait for a couple of days before merging, to allow any other feedback from the community.

ghost:
Wondering if this can be merged now since it has been 7 days since approved ?

tweise:
@ananthc yes, please rebase and I will merge it.

PR #489, opened 2016-11-09
APEXMALHAR-2330 JdbcPOJOPollInputOperator fails with NullPointerExcep…

deepak-narkhede:
**Problem:** 
JdbcPOJOPollInputOperator fails with NullPointerException using PostgreSQL driver.

**Problem Description:**
1) When JdbcPOJOPollInputOperator tries to populateColumnDataTypes, column names retrieved from resultmetadata from database ( this case : Postgres) are all lowercase.
2) Whereas columnDatatypes specified in fieldinfos might be in same case.
3) Internally hashmap ( nameToType) is used which mismatches if column name and fieldinfo are not in same case. Hence columnDataTypes is empty which causes null exception in activate call.

**Solution:**
Using similar case for hashmap and column names irrespective of any database used for JdbcPOJOPollInputOperator.

**Testing:**
Tested with Database-to-hdfs app also with instrumented logs.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOPollInputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOPollInputOperator.java
@@ -195,11 +195,11 @@ public class JdbcPOJOPollInputOperator extends AbstractJdbcPollInputOperator<Obj
         int type = rsMetaData.getColumnType(i);
         String name = rsMetaData.getColumnName(i);
         LOG.debug("column name {} type {}", name, type);
-        nameToType.put(name, type);
+        nameToType.put(name.toUpperCase(), type);
       }
 
       for (FieldInfo fieldInfo : fieldInfos) {
-        columnDataTypes.add(nameToType.get(fieldInfo.getColumnName()));
+        columnDataTypes.add(nameToType.get(fieldInfo.getColumnName().toUpperCase()));
       }
     }
   }


devtagare (changes requested):
@deepak-narkhede  Can you please update the test cases in JdbcPojoPollableOpeartorTest to accommodate this  change ?

library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOPollInputOperator.java
```diff
@@ -195,11 +195,11 @@ protected void populateColumnDataTypes() throws SQLException
         int type = rsMetaData.getColumnType(i);
         String name = rsMetaData.getColumnName(i);
         LOG.debug("column name {} type {}", name, type);
-        nameToType.put(name, type);
+        nameToType.put(name.toUpperCase(), type);
```
DT-Priyanka: can we write a unit test for this?

deepak-narkhede:
@devtagare and @DT-Priyanka , As I could't find the embedded PostgreSQL driver. Hence cannot add on to same test cases. Will try Priyanka's suggestion using Mock.

deepak-narkhede:
APEXMALHAR-2349 is opened to track the Mocking entire PostgreSQL database connection. As embedded drivers so far are unavailable.

DT-Priyanka:
I am okay to go ahead with merge as you have created new JIRA to create tests.

DT-Priyanka:
can you please rebase so that I can merge it?

deepak-narkhede:
Done with rebase.

PR #490, opened 2016-11-10
APEXMALHAR-2274 Handle large number of files for AbstractFileInputOpe…

mattqzhang:
@PramodSSImmaneni Please review

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
@@ -35,6 +35,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -53,6 +59,7 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 
 import com.google.common.base.Preconditions;
+import com.google.common.base.Throwables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 
@@ -100,9 +107,12 @@ import com.datatorrent.lib.util.KryoCloneUtils;
  */
 @org.apache.hadoop.classification.InterfaceStability.Evolving
 public abstract class AbstractFileInputOperator<T> implements InputOperator, Partitioner<AbstractFileInputOperator<T>>, StatsListener,
-    Operator.CheckpointListener, Operator.CheckpointNotificationListener
+    Operator.CheckpointListener, Operator.CheckpointNotificationListener, Operator.ActivationListener<OperatorContext>
 {
   private static final Logger LOG = LoggerFactory.getLogger(AbstractFileInputOperator.class);
+
+  private static final int DEFAULT_MAX_FILES_PER_WINDOW = 1024;
+  private static final int DEFAULT_QUEUE_CAPACITY = 4 * 1024;
 
   @NotNull
   protected String directory;
@@ -110,9 +120,11 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
   protected DirectoryScanner scanner = new DirectoryScanner();
   @Min(0)
   protected int scanIntervalMillis = 5000;
+  private int queueCapacity = DEFAULT_QUEUE_CAPACITY;
+  private int maxFilesPerWindow = DEFAULT_MAX_FILES_PER_WINDOW;
   protected long offset;
   protected String currentFile;
-  protected Set<String> processedFiles = new HashSet<String>();
+  protected ConcurrentHashSet<String> processedFiles = new ConcurrentHashSet<String>();
   @Min(1)
   protected int emitBatchSize = 1000;
   protected int currentPartitions = 1;
@@ -121,7 +133,9 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
   @Min(0)
   private int maxRetryCount = 5;
   protected transient long skipCount = 0;
-  private transient OperatorContext context;
+  protected transient OperatorContext context;
+  protected transient ScheduledExecutorService scanDirService;
+  protected transient ArrayBlockingQueue<Path> fileQueue;
 
   private final BasicCounters<MutableLong> fileCounters = new BasicCounters<MutableLong>(MutableLong.class);
   protected MutableLong globalNumberOfFailures = new MutableLong();
@@ -137,6 +151,8 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
   protected transient long currentWindowId;
   protected final transient LinkedList<RecoveryEntry> currentWindowRecoveryState = Lists.newLinkedList();
   protected int operatorId; //needed in partitioning
+  protected int fileCount;
+  protected transient AtomicReference<Throwable> atomicThrowable = new AtomicReference<>();
 
   /**
    * Class representing failed file, When read fails on a file in middle, then the file is
@@ -378,6 +394,46 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
   }
 
   /**
+   * Returns the capacity of the emit queue
+   *
+   * @return queueCapacity
+   */
+  public int getQueueCapacity()
+  {
+    return queueCapacity;
+  }
+
+  /**
+   * Sets the capacity of the emit queue
+   *
+   * @param queueCapacity
+   */
+  public void setQueueCapacity(int queueCapacity)
+  {
+    this.queueCapacity = queueCapacity;
+  }
+
+  /**
+   * Returns the max file per window
+   *
+   * @return queueCapacity
+   */
+  public int getMaxFilesPerWindow()
+  {
+    return maxFilesPerWindow;
+  }
+
+  /**
+   * Sets the max file per window
+   *
+   * @param maxFilesPerWindow
+   */
+  public void setMaxFilesPerWindow(int maxFilesPerWindow)
+  {
+    this.maxFilesPerWindow = maxFilesPerWindow;
+  }
+
+  /**
    * Returns the number of tuples emitted in a batch.
    * @return The number of tuples emitted in a batch.
    */
@@ -474,6 +530,9 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
       currentFile = null;
       offset = 0;
     }
+
+    fileQueue = new ArrayBlockingQueue<Path>(queueCapacity);
+    scanDirService = Executors.newSingleThreadScheduledExecutor();
   }
 
   /**
@@ -528,9 +587,22 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
   }
 
   @Override
+  public void activate(OperatorContext context)
+  {
+    scanDirService.scheduleWithFixedDelay(new DirectoryScannerTask(), 0, scanIntervalMillis, TimeUnit.MILLISECONDS);
+  }
+
+  @Override
+  public void deactivate()
+  {
+    scanDirService.shutdownNow();
+  }
+
+  @Override
   public void beginWindow(long windowId)
   {
     currentWindowId = windowId;
+    fileCount = 0;
     if (windowId <= windowDataManager.getLargestCompletedWindow()) {
       replay(windowId);
     }
@@ -581,6 +653,7 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
                 closeFile(inputStream);
               }
               processedFiles.add(recoveryEntry.file);
+
               //removing the file from failed and unfinished queues and pending set
               Iterator<FailedFile> failedFileIterator = failedFiles.iterator();
               while (failedFileIterator.hasNext()) {
@@ -641,6 +714,11 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
       return;
     }
 
+    Throwable throwable;
+    if ((throwable = atomicThrowable.get()) != null) {
+      Throwables.propagate(throwable);
+    }
+
     if (inputStream == null) {
       try {
         if (currentFile != null && offset > 0) {
@@ -666,7 +744,15 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
         } else if (!failedFiles.isEmpty()) {
           retryFailedFile(failedFiles.poll());
         } else {
-          scanDirectory();
+          while ((!fileQueue.isEmpty()) && (fileCount < getMaxFilesPerWindow())) {
+            Path newPath = fileQueue.peek();
+            String newPathString = newPath.toString();
+            pendingFiles.add(newPathString);
+            processedFiles.add(newPathString);
+            localProcessedFileCount.increment();
+            fileQueue.remove();
+            fileCount++;
+          }
         }
       } catch (IOException ex) {
         failureHandling(ex);
@@ -706,25 +792,6 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
       if (offset >= startOffset) {
         currentWindowRecoveryState.add(new RecoveryEntry(file, startOffset, offset, fileClosed));
       }
-    }
-  }
-
-  /**
-   * Scans the directory for new files.
-   */
-  protected void scanDirectory()
-  {
-    if (System.currentTimeMillis() - scanIntervalMillis >= lastScanMillis) {
-      Set<Path> newPaths = scanner.scan(fs, filePath, processedFiles);
-
-      for (Path newPath : newPaths) {
-        String newPathString = newPath.toString();
-        pendingFiles.add(newPathString);
-        processedFiles.add(newPathString);
-        localProcessedFileCount.increment();
-      }
-
-      lastScanMillis = System.currentTimeMillis();
     }
   }
 
@@ -873,6 +940,7 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
 
       // Do state transfer for processed files.
       oper.processedFiles.addAll(totalProcessedFiles);
+
       oper.globalNumberOfFailures = tempGlobalNumberOfRetries;
       oper.localNumberOfFailures.setValue(0);
       oper.globalNumberOfRetries = tempGlobalNumberOfFailures;
@@ -1009,6 +1077,31 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
       throw new IllegalArgumentException("maxRetryCount should be greater than or equal to 0.");
     }
     this.maxRetryCount = maxRetryCount;
+  }
+
+  /**
+   * This class scans a directory and put the list of files in a blocking queue.
+   */
+  public class DirectoryScannerTask implements Runnable
+  {
+    @Override
+    public void run()
+    {
+      try {
+        if (pendingFiles.isEmpty() && fileQueue.isEmpty() && (System.currentTimeMillis() - scanIntervalMillis >= lastScanMillis)) {
+          LinkedHashSet<Path> newFileSet = scanner.scan(fs, filePath, processedFiles);
+          for (Path newFile : newFileSet) {
+            fileQueue.put(newFile);
+          }
+
+          lastScanMillis = System.currentTimeMillis();
+        }
+      } catch (Throwable throwable) {
+        LOG.error("service", throwable);
+        atomicThrowable.set(throwable);
+        Throwables.propagate(throwable);
+      }
+    }
   }
 
   /**
@@ -1168,7 +1261,7 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
     /**
      * True if recursive; false otherwise.
      *
-     * @param recursive true if recursive; false otherwise.
+     * @return true if recursive; false otherwise.
      */
     public boolean isRecursive()
     {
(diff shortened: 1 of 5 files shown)


library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -98,15 +102,19 @@
  */
 @org.apache.hadoop.classification.InterfaceStability.Evolving
 public abstract class AbstractFileInputOperator<T> implements InputOperator, Partitioner<AbstractFileInputOperator<T>>, StatsListener,
-    Operator.CheckpointListener, Operator.CheckpointNotificationListener
+    Operator.CheckpointListener, Operator.CheckpointNotificationListener, Operator.ActivationListener<OperatorContext>
 {
   private static final Logger LOG = LoggerFactory.getLogger(AbstractFileInputOperator.class);
 
+  protected static int MAX_FILE_PER_WINDOW = 1024;
+  private static int DEFAULT_QUEUE_CAPACITY = 4 * 1024;
```
tushargosavi: final

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -517,6 +552,18 @@ public void teardown()
   }
 
   @Override
+  public void activate(OperatorContext context)
+  {
+    scanDirService.scheduleAtFixedRate(new DirectoryScannerThread(), 0, scanIntervalMillis, TimeUnit.MILLISECONDS);
```
tushargosavi: This is using scheduleAtFixedRate but task DirectoryScannerThread never terminates.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -569,7 +616,9 @@ protected void replay(long windowId)
               if (inputStream != null) {
                 closeFile(inputStream);
               }
-              processedFiles.add(recoveryEntry.file);
+              synchronized (processedFiles.getClass()) {
```
tushargosavi: why not use concurrent set? ConcurrentSkipListSet, ConcurrentHashMap.newKeySet()

library/src/main/java/com/datatorrent/lib/io/fs/AbstractThroughputFileInputOperator.java
```diff
@@ -122,8 +123,18 @@ public void setPartitionCount(int requiredPartitions)
   @Override
   public void emitTuples()
   {
-    scanDirectory();
-
+    int fileCt = 0;
+    while ((!fileQueue.isEmpty()) && (fileCt < MAX_FILE_PER_WINDOW)) {
+      Path newPath = fileQueue.peek();
+      String newPathString = newPath.toString();
+      pendingFiles.add(newPathString);
```
tushargosavi: I think pending file is not required as your fileQueue becomes list of pending files.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -998,6 +1044,34 @@ public void setMaxRetryCount(int maxRetryCount)
...
+        while (execute) {
+          if (pendingFiles.isEmpty() && fileQueue.isEmpty() && (System.currentTimeMillis() - scanIntervalMillis >= lastScanMillis)) {
+            LinkedHashSet<Path> newFileSet = scanner.scan(fs, filePath, processedFiles);
+            for (Path newFile : newFileSet) {
+              fileQueue.put(newFile);
+            }
+
+            lastScanMillis = System.currentTimeMillis();
+          } else {
+            // Thread.sleep(DEFAULT_SLEEP_TIME);
```
tushargosavi: remove comment

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -998,6 +1044,34 @@ public void setMaxRetryCount(int maxRetryCount)
...
+            Thread.sleep(scanIntervalMillis);
+          }
+        }
+      } catch (Exception e) {
+        execute = false;
```
tushargosavi: In case of Exception, we should notify the main thread and take appropriate action there. Take a look at 
https://github.com/apache/apex-malhar/blob/master/library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java#L163 and
https://github.com/apache/apex-malhar/blob/master/library/src/main/java/com/datatorrent/lib/io/fs/FileSplitterInput.java#L383

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -998,6 +1044,34 @@ public void setMaxRetryCount(int maxRetryCount)
...
+  {
+    @Override
+    public void run()
+    {
+      try {
+        while (execute) {
+          if (pendingFiles.isEmpty() && fileQueue.isEmpty() && (System.currentTimeMillis() - scanIntervalMillis >= lastScanMillis)) {
+            LinkedHashSet<Path> newFileSet = scanner.scan(fs, filePath, processedFiles);
+            for (Path newFile : newFileSet) {
+              fileQueue.put(newFile);
```
tushargosavi: does not help to put limit on the size of fileQueue, as newFileSet already has all the file name, And emitTuple is scanning  only max 1024 entries per second.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractThroughputFileInputOperator.java
```diff
@@ -122,8 +123,18 @@ public void setPartitionCount(int requiredPartitions)
   @Override
   public void emitTuples()
   {
-    scanDirectory();
-
+    int fileCt = 0;
```
tushargosavi: resetting fileCt to zero in emitTuple does not help, It needs to be cleared in beginWindow. as emitTuple can get called multipleTimes in a window.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -655,7 +707,18 @@ public void emitTuples()
         } else if (!failedFiles.isEmpty()) {
           retryFailedFile(failedFiles.poll());
         } else {
-          scanDirectory();
+          int fileCt = 0;
+          while ((!fileQueue.isEmpty()) && (fileCt < MAX_FILE_PER_WINDOW)) {
```
tushargosavi: can MAX_FILE_PER_WINDOW be a configuration? and why is it needed?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -98,15 +102,19 @@
  */
 @org.apache.hadoop.classification.InterfaceStability.Evolving
 public abstract class AbstractFileInputOperator<T> implements InputOperator, Partitioner<AbstractFileInputOperator<T>>, StatsListener,
-    Operator.CheckpointListener, Operator.CheckpointNotificationListener
+    Operator.CheckpointListener, Operator.CheckpointNotificationListener, Operator.ActivationListener<OperatorContext>
 {
   private static final Logger LOG = LoggerFactory.getLogger(AbstractFileInputOperator.class);
 
+  protected static int MAX_FILE_PER_WINDOW = 1024;
```
tushargosavi: What is the purpose of MAX_FILE_PER_WINDOW? any explanation for 1024 and can it be configured by user?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: I want to keep the data structure simple. Classes like ConcurrentSkipListSet is too complex for the job, and its running time of get, put is log(n) vs constant in HashMap.  I have also tried a similar one using Collections.synchronizedSet(new HashSet<String>()). Many others have similar problems as described below:

When I ran the tests I see several errors like:
Class cannot be created (missing no-arg constructor): java.util.Collections SynchronizedSet
When the code is trying to do a Kryo copy:
LineByLineFileInputOperator initialState = new Kryo().copy(oper);

The reason is the Collections.synchronizedSet() doesn't have a no-arg constructor. Since I cannot modify this class, I have to specifically lock on each access to this processedFiles and do something like:
synchronized (processedFiles.getClass())

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: I followed the same design settings according to AbstractJdbcPollInputOperator. But yes I agree in this case maybe we should implement it as configurable so that users can adjust it according to their file size. I'll make the changes.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: I'll update this.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: I follow similar design as AbstractJdbcPollInputOperator, and make it periodic to reduce the I/O cost of continuous scan. But I do see the problem with the while() loop you mentioned. I'll remove the while loop in the DirectoryScannerThread(). Looks like the DBPoll() in AbstractJdbcPollInputOperator has the same error :)

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: As mentioned above, I'll make it configurable. The purpose of it is to limit the number of files processed in each window when we have large number of files.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: done

library/src/main/java/com/datatorrent/lib/io/fs/AbstractThroughputFileInputOperator.java
mattqzhang: pendingFiles is serializable and used for recovery and idempotency, while fileQueue is not serialized and need to be recomputed upon failure and replay.  Moreover currently pending files is protected. It may cause regression issue if we change the whole design.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractThroughputFileInputOperator.java
mattqzhang: Good catch! I'll change it.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: I'll implement the similar throwable for exception handling here.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: fileQueue is of type ArrayBlockingQueue. It will block if it reaches the predefined size.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -50,7 +55,9 @@
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 
+import com.esotericsoftware.minlog.Log;
```
amberarrow: Looks like this was inserted by some IDE ?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -998,6 +1073,33 @@ public void setMaxRetryCount(int maxRetryCount)
...
+          for (Path newFile : newFileSet) {
+            fileQueue.put(newFile);
+          }
+
+          lastScanMillis = System.currentTimeMillis();
+        } else {
+          Thread.sleep(scanIntervalMillis);
+        }
+      } catch (Throwable throwable) {
+        Log.error("service", throwable);
```
amberarrow: Log => LOG

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -98,15 +105,20 @@
  */
 @org.apache.hadoop.classification.InterfaceStability.Evolving
 public abstract class AbstractFileInputOperator<T> implements InputOperator, Partitioner<AbstractFileInputOperator<T>>, StatsListener,
-    Operator.CheckpointListener, Operator.CheckpointNotificationListener
+    Operator.CheckpointListener, Operator.CheckpointNotificationListener, Operator.ActivationListener<OperatorContext>
 {
   private static final Logger LOG = LoggerFactory.getLogger(AbstractFileInputOperator.class);
 
+  private static final int MAX_FILE_PER_WINDOW = 1024;
```
amberarrow: MAX_FILE => DEFAULT_MAX_FILES
{the DEFAULT prefix stays consistent with DEFAULT_QUEUE_CAPACITY}

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -98,15 +105,20 @@
...
+  private static final int DEFAULT_QUEUE_CAPACITY = 4 * 1024;
+
   @NotNull
   protected String directory;
   @NotNull
   protected DirectoryScanner scanner = new DirectoryScanner();
   protected int scanIntervalMillis = 5000;
+  private int queueCapacity = DEFAULT_QUEUE_CAPACITY;
+  private int maxFilePerWindow = MAX_FILE_PER_WINDOW;
```
amberarrow: maxFile => maxFiles

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -132,6 +146,8 @@
   protected transient long currentWindowId;
   protected final transient LinkedList<RecoveryEntry> currentWindowRecoveryState = Lists.newLinkedList();
   protected int operatorId; //needed in partitioning
+  protected int fileCt;
```
amberarrow: Suggest fileCount instead of fileCt (abbreviated or cryptic name are OK if used as local variables within a small context of a few lines; for variables with a larger scope or lifetime, longer more descriptive names are preferable).

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -617,6 +691,9 @@ protected void replay(long windowId)
           }
         }
       }
+      if (windowId == windowDataManager.getLargestCompletedWindow()) {
+        scanDirService.scheduleAtFixedRate(new DirectoryScannerThread(), 0, scanIntervalMillis, TimeUnit.MILLISECONDS);
```
amberarrow: Since we are doing this in activate(), why do we need to do it again here ?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -998,6 +1073,33 @@ public void setMaxRetryCount(int maxRetryCount)
   }
 
   /**
+   * This class scans a directory and put the list of files in a blocking queue.
+   */
+  public class DirectoryScannerThread implements Runnable
```
amberarrow: Suggest renaming this DirectoryScannerRunnable, DirectoryScannerTask, or something similar since it does not create a new thread for each instance -- the same thread in the Executor runs all the instances.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
amberarrow: Consider the following scenario: Each DirectoryScannerThread takes 10s and scanIntervalMillis has the default value of 5000. The queue of tasks will continue to grow in an unbounded way. Suggest using `ScheduledThreadPoolExecutor.scheduleWithFixedDelay()` instead.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -463,6 +519,9 @@ public void setup(OperatorContext context)
       currentFile = null;
       offset = 0;
     }
+
+    fileQueue = new ArrayBlockingQueue<Path>(queueCapacity);
+    scanDirService = Executors.newScheduledThreadPool(1);
```
amberarrow: Why not `Executors.newSingleThreadExecutor()` ?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
amberarrow: Why synchronize on the class rather than the instance ?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -861,7 +934,9 @@ protected void closeFile(InputStream is) throws IOException
       oper.setScanner(scn);
 
       // Do state transfer for processed files.
-      oper.processedFiles.addAll(totalProcessedFiles);
+      synchronized (processedFiles.getClass()) {
```
amberarrow: definePartitions() executes on Stram, so do we really need to synchronize ?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
amberarrow: I'm leaning towards @tushargosavi 's position: Explicit synchronization is error prone and is better avoided if possible. `ConcurrentHashMap` has a default constructor and can simply be used as a set by using a single fixed value (e.g. `Boolean.TRUE`) for keys that are present in the set. Or you could even explicitly define a `ConcurrentHashSet` class as shown by response 4 at:
http://stackoverflow.com/questions/6992608/why-there-is-no-concurrenthashset-against-concurrenthashmap

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: @amberarrow Currently the scanDirService is defined as  Executors.newScheduledThreadPool(1), which returns type ScheduledThreadPoolExecutor. I've updated the method to scheduleWithFixedDelay() to avoid building the queue.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: I've added the ConcurrentHashSet class to the implementation.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: Was added by IDE. Fixed.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: fixed

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: updated

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: updated

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: updated

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: updated

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: Changing processedFiles to ConcurrentSet automatically resolves this case.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
mattqzhang: I followed the design in AbstractJdbcPollInputOperator. I had thought of using Executors.newSingleThreadScheduledExecutor() as we need to specify delay in the scheduling. The implementation of that one is using ScheduledThreadPoolExecutor(1). So essentially the same.

ilganeli:
@mattqzhang Is this patch still relevant given recent updates to the AbstractFileInputOperator?

amberarrow:
The content of this patch is now subsumed by https://github.com/apache/apex-malhar/pull/597

When that is merged, we can close this PR.

tushargosavi:
@amberarrow can you fix following check style errors. 

[INFO] There are 2 errors reported by Checkstyle 6.15 with apex_checks.xml ruleset.
[ERROR] src/main/java/com/datatorrent/lib/io/fs/AbstractThroughputFileInputOperator.java:[27,8] 
(imports) UnusedImports: Unused import - org.apache.hadoop.fs.Path.
[ERROR] src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java:[790,38] (blocks) LeftCurly: '{' at column 38 should be on a new line.

PR #493, opened 2016-11-10
APEXMALHAR-2335 APEXMALHAR-2333 APEXMALHAR-2334 #resolve #comment Pro…

brightchen:
…blems on StateTracker

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
@@ -435,17 +435,51 @@ public interface Bucket extends ManagedStateComponent, KeyValueByteStreamProvide
      * Free memory up to the given windowId
      * This method will be called by another thread. Adding concurrency control to Stream would impact the performance.
      * This method only calculates the size of the memory that could be released and then sends free memory request to the operator thread
+     *
+     * We intend to manage memory by keyStream and valueStream. But the we can't avoid caller use other mechanism to manage memory.
+     * It is required to cleanup maps in this case.
      */
     @Override
     public long freeMemory(long windowId) throws IOException
     {
-      // calculate the size first and then send the release memory request. It could reduce the chance of conflict and increase the performance.
-      long size = keyStream.dataSizeUpToWindow(windowId) + valueStream.dataSizeUpToWindow(windowId);
+      long memoryFreed = 0;
+      Iterator<Map.Entry<Long, Map<Slice, BucketedValue>>> entryIter = committedData.entrySet().iterator();
+      while (entryIter.hasNext()) {
+        Map.Entry<Long, Map<Slice, BucketedValue>> bucketEntry = entryIter.next();
+        if (bucketEntry.getKey() > windowId) {
+          break;
+        }
+
+        Map<Slice, BucketedValue> windowData = bucketEntry.getValue();
+        entryIter.remove();
+
+        for (Map.Entry<Slice, BucketedValue> entry : windowData.entrySet()) {
+          memoryFreed += entry.getKey().length + entry.getValue().getSize();
+        }
+      }
+
+      fileCache.clear();
+      if (cachedBucketMetas != null) {
+
+        for (BucketsFileSystem.TimeBucketMeta tbm : cachedBucketMetas.values()) {
+          FileAccess.FileReader reader = readers.remove(tbm.getTimeBucketId());
+          if (reader != null) {
+            memoryFreed += tbm.getSizeInBytes();
+            reader.close();
+          }
+        }
+      }
+
+      sizeInBytes.getAndAdd(-memoryFreed);
+
+      //add the windowId to the queue to let operator thread release memory from keyStream and valueStream
       windowsForFreeMemory.add(windowId);
-      return size;
+
+      return memoryFreed;
     }
 
     /**
+     * Release the memory managed by keyStream and valueStream.
      * This operation must be called from operator thread. It won't do anything if no memory to be freed
      */
     protected long releaseMemory()
@@ -459,10 +493,10 @@ public interface Bucket extends ManagedStateComponent, KeyValueByteStreamProvide
         memoryFreed += originSize - (keyStream.size() + valueStream.size());
       }
 
-      if (memoryFreed > 0) {
-        LOG.debug("Total freed memory size: {}", memoryFreed);
-        sizeInBytes.getAndAdd(-memoryFreed);
-      }
+      //release the free memory immediately
+      keyStream.releaseAllFreeMemory();
+      valueStream.releaseAllFreeMemory();
+
       return memoryFreed;
     }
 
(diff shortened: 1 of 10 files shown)


brightchen:
@davidyan74 Please review

brightchen:
@davidyan74 please review

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -435,17 +435,46 @@ public void put(Slice key, long timeBucket, Slice value)
...
     @Override
     public long freeMemory(long windowId) throws IOException
     {
-      // calculate the size first and then send the release memory request. It could reduce the chance of conflict and increase the performance.
-      long size = keyStream.dataSizeUpToWindow(windowId) + valueStream.dataSizeUpToWindow(windowId);
+
+      long memoryFreed = 0;
+      Long clearWindowId;
+
+      while ((clearWindowId = committedData.floorKey(windowId)) != null) {
```
davidyan74: each floorKey and remove call is O(log n) so the loop is going to be O(n log n). Why don't we just have a normal ordered iteration to make it O(n) instead?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/StateTracker.java
```diff
@@ -32,56 +35,55 @@
...
 
   protected transient AbstractManagedStateImpl managedStateImpl;
 
+  private transient long lastUpdateAccessTime = 0;
+  private transient Set<Long> accessedBucketIds = Sets.newHashSet();
+  private transient TreeSet<BucketIdTimeWrapper> bucketHeapForAccess = Sets.newTreeSet(new BucketIdTimeWrapper.BucketTimeComparator());
+  private transient TreeSet<BucketIdTimeWrapper> bucketHeapForRelease = Sets.newTreeSet(new BucketIdTimeWrapper.BucketTimeComparator());
+  private transient ReadWriteLock bucketAcessLock = new ReentrantReadWriteLock();
+  private int updateAccessTimeInterval = 500;
+
```
davidyan74, line 54: Can we make some of these members final?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/StateTracker.java
```diff
@@ -148,6 +179,19 @@ void teardown()
    */
   private static class BucketIdTimeWrapper
   {
+    public static class BucketTimeComparator implements Comparator<BucketIdTimeWrapper>
+    {
+      @Override
+      public int compare(BucketIdTimeWrapper o1, BucketIdTimeWrapper o2)
+      {
+        //same bucket are equals.
+        if (o1.bucketId == o2.bucketId) {
```
davidyan74: why do we need this if statement?

davidyan74 (changes requested):

library/src/main/java/org/apache/apex/malhar/lib/state/managed/StateTracker.java
```diff
@@ -32,56 +35,55 @@
...
-
   private final transient Timer memoryFreeService = new Timer();
 
   protected transient AbstractManagedStateImpl managedStateImpl;
 
+  private transient long lastUpdateAccessTime = 0;
+  private transient Set<Long> accessedBucketIds = Sets.newHashSet();
+  private transient TreeSet<BucketIdTimeWrapper> bucketHeapForAccess = Sets.newTreeSet(new BucketIdTimeWrapper.BucketTimeComparator());
+  private transient TreeSet<BucketIdTimeWrapper> bucketHeapForRelease = Sets.newTreeSet(new BucketIdTimeWrapper.BucketTimeComparator());
+  private transient ReadWriteLock bucketAcessLock = new ReentrantReadWriteLock();
```
davidyan74: spelling

library/src/main/java/org/apache/apex/malhar/lib/state/managed/StateTracker.java
```diff
@@ -105,12 +107,30 @@ public void run()
...
+          for (BucketIdTimeWrapper releaseItem : bucketHeapForRelease) {
+            bucketHeapForAccess.add(releaseItem);
+          }
+          bucketHeapForRelease.clear();
+
+          //switch the access and release list
+          TreeSet<BucketIdTimeWrapper> tmp = bucketHeapForAccess;
+          bucketHeapForAccess = bucketHeapForRelease;
+          bucketHeapForRelease = tmp;
+        } finally {
```
davidyan74: Not sure what this code is doing. Is it the same as putting all items in bucketHeapForAccess to bucketHeapForRelease?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/StateTracker.java
```diff
@@ -32,56 +32,53 @@
...
   }
 
   void bucketAccessed(long bucketId)
   {
-    BucketIdTimeWrapper idTimeWrapper = bucketAccessTimes.get(bucketId);
-    if (idTimeWrapper != null) {
-      bucketHeap.remove(idTimeWrapper);
-    }  else {
-      idTimeWrapper = new BucketIdTimeWrapper(bucketId);
+    accessedBucketIds.add(bucketId);
```
davidyan74: accessBucketIds is not a ConcurrentSet. Won't this create a problem when multiple threads call this method?

brightchen: accessedBucketIds all used by operator main thread

library/src/main/java/org/apache/apex/malhar/lib/state/managed/StateTracker.java
```diff
@@ -32,56 +32,53 @@
...
-
   private final transient Timer memoryFreeService = new Timer();
 
   protected transient AbstractManagedStateImpl managedStateImpl;
 
+  private transient long lastUpdateAccessTime = 0;
+  private final transient Set<Long> accessedBucketIds = Sets.newHashSet();
+  private transient TreeSet<BucketIdTimeWrapper> bucketHeapForAccess = Sets.newTreeSet();
+  private transient TreeSet<BucketIdTimeWrapper> bucketHeapForRelease = Sets.newTreeSet();
+
```
davidyan74, line 52: I think these two TreeSets and the BucketIdTimeWrapper can be replaced by a LinkedHashMap with accessOrder = true

brightchen: it is possible to it by a map from time to a list of bucket. but seems not simplify the code. The reason to use two lists in order two threads can run without concurrent control. the concurrent control only need to happen when merge/swap

davidyan74: Please take a look at this and let me know what you think:
https://github.com/apache/apex-malhar/commit/eadf39b11f495e84c1d1e6ffbb83c20b8dee404b

brightchen: it is possible that last update time is different. for example: in first 500ms, updated b1, b2; in second 500ms, updated b2, b3, so the last update of b1 is different from last update of b2 and b3

davidyan74: This is my latest FYI: https://github.com/davidyan74/apex-malhar/tree/bucketLastAccess

brightchen:
@davidyan74 Merged your update, please review

PR #499, opened 2016-11-16
APEXMALHAR-2343 #resolve #comment Count Accumulation should only incr…

brightchen:
…ease one for each tuple


brightchen:
@davidyan74 Please review

davidyan74:
@siyuanh and @chinmaykolhatkar please also take a look.

PR #500, opened 2016-11-16
APEXMALHAR-2343 #resolve #comment Count Accumulation should only incr…

brightchen:
…ease one for each tuple

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/Count.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/Count.java
@@ -26,7 +26,7 @@ import org.apache.commons.lang3.mutable.MutableLong;
  *
  * @since 3.5.0
  */
-public class Count implements Accumulation<Long, MutableLong, Long>
+public class Count<T> implements Accumulation<T, MutableLong, Long>
 {
 
   @Override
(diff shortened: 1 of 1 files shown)


brightchen:
@davidyan74 Please review

davidyan74:
LGTM @siyuanh @chinmaykolhatkar please also review and make sure this does not break existing examples that make use of the Count accumulation.

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/Count.java
```diff
@@ -26,7 +26,7 @@
  *
  * @since 3.5.0
  */
-public class Count implements Accumulation<Long, MutableLong, Long>
+public class Count<T> implements Accumulation<T, MutableLong, Long>
```
chinmaykolhatkar, line 29: It doesn't make sense to parameterize Count class as the type of input really doesn't matter.

davidyan74: That's true. Without this parameterization, the accumulation input type parameter needs to be a supertype of the input tuple type. I will issue a PR to reflect that.

davidyan74: The PR is #517

davidyan74:
This PR is superseded by #517. Please refer to that PR instead. This PR can be closed.

PR #502, opened 2016-11-17
Windowed Operator benchmark

brightchen:


brightchen:
@davidyan74 Please review

brightchen:
@davidyan74 Please review

brightchen:
@davidyan74 Could you please review and merge this PR when got time.

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/Count.java
```diff
@@ -26,7 +26,7 @@
  *
  * @since 3.5.0
  */
-public class Count implements Accumulation<Long, MutableLong, Long>
```
davidyan74, line 29: This change should not be part of this PR, since it's covered in another PR.

benchmark/src/main/java/com/datatorrent/benchmark/window/KeyedWindowedOperatorBenchmarkApp.java
```diff
@@ -0,0 +1,143 @@
...
+    public void processTuple(Tuple tuple)
+    {
+      super.processTuple(tuple);
+      ++tupleCount;
+    }
+  }
+
+  protected static class KeyedWindowedGenerator extends AbstractGenerator
+  {
+    public final transient DefaultOutputPort<Tuple.TimestampedTuple<KeyValPair<String, Long>>> data = new DefaultOutputPort<Tuple.TimestampedTuple<KeyValPair<String, Long>>>();
```
davidyan74: Currently, the AbstractGenerator actually does not have anything that has to do with "generating". I think we should have the AbstractGenerator class take the output type parameter and declare the data port there, and have the emitTuples be inside the abstract class. The concrete class only provide the tuple. Like this:
```java
public abstract class AbstractGenerator<T> extends BaseOperator implements InputOperator
{
  ...
  public final transient DefaultOutputPort<T> data = new DefaultOutputPort<>(); 
  ...
  @Override
  public void emitTuples()
  {
    for (int i = 0; i < emitBatchSize && emitCount < rate; i++) {
      data.emit(generateTuple());
      emitCount++;
    }
  }
  public abstract T generateTuple();
}
```
The rest should be obvious.

davidyan74 (commented):
The rest looks good to me. I will merge this after #517 has been merged and you have rebased after the merge.

brightchen:
@davidyan74 Applied comments

PR #514, opened 2016-11-28
APEXMALHAR-2358 #resolve #comment Optimise GenericSerde to use specif…

brightchen:
…ic serde to improve the performance

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/utils/serde/GenericSerde.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/utils/serde/GenericSerde.java
@@ -18,11 +18,16 @@
  */
 package org.apache.apex.malhar.lib.utils.serde;
 
+import java.util.Map;
+
+import org.apache.apex.malhar.lib.window.Window;
+import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.hadoop.classification.InterfaceStability;
 
 import com.esotericsoftware.kryo.Kryo;
 import com.esotericsoftware.kryo.io.Input;
 import com.esotericsoftware.kryo.io.Output;
+import com.google.common.collect.Maps;
 
 /**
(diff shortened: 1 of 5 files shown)


brightchen:
@davidyan74 Please review

library/src/main/java/org/apache/apex/malhar/lib/utils/serde/GenericSerde.java
```diff
@@ -34,31 +39,60 @@
 @InterfaceStability.Evolving
 public class GenericSerde<T> implements Serde<T>
 {
-  private transient ThreadLocal<Kryo> kryos = new ThreadLocal<Kryo>()
```
davidyan74, line 39: https://github.com/EsotericSoftware/kryo says kryo is not thread safe. Are we sure that GenericSerde is always only used by one thread?

brightchen: @davidyan74 Yes, in our current use cases, only the operator main thread will use it

brightchen:
@davidyan74 fixed checkstyle

library/src/main/java/org/apache/apex/malhar/lib/utils/serde/GenericSerde.java
```diff
@@ -36,31 +41,60 @@
...
+  {
+    return typeToSerde.get(type);
   }
 
   @Override
   public void serialize(T object, Output output)
   {
-    Kryo kryo = kryos.get();
+    Class type = object.getClass();
+    if (clazz == null) {
```
davidyan74: This is going to be a problem because the clazz may not be known when you deserialize. The clazz member variable, if not null assumes that the class is known during deserialization.

brightchen:
@davidyan74 I applied comments, please review.

PR #515, opened 2016-11-29
APEXMALHAR-2350 #resolve #comment The key and value stream should mat…

brightchen:
…ch with the bucket

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeUnifiedStateImpl.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeUnifiedStateImpl.java
@@ -216,7 +216,7 @@ public class ManagedTimeUnifiedStateImpl extends AbstractManagedStateImpl implem
     @Override
     protected void addBucketName(long bucketId)
     {
-      long operatorId = (long)managedStateContext.getOperatorContext().getId();
+      long operatorId = managedStateContext.getOperatorContext().getId();
       if (!bucketNamesOnFS.contains(operatorId)) {
         bucketNamesOnFS.add(operatorId);
       }
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableMapImpl.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableMapImpl.java
@@ -26,6 +26,7 @@ import java.util.Set;
 import javax.validation.constraints.NotNull;
 
 import org.apache.apex.malhar.lib.state.BucketedState;
+import org.apache.apex.malhar.lib.state.managed.ManagedTimeUnifiedStateImpl;
 import org.apache.apex.malhar.lib.state.managed.TimeExtractor;
 import org.apache.apex.malhar.lib.utils.serde.AffixKeyValueSerdeManager;
 import org.apache.apex.malhar.lib.utils.serde.BufferSlice;
@@ -150,7 +151,7 @@ public class SpillableMapImpl<K, V> implements Spillable.SpillableMap<K, V>, Spi
       return val;
     }
 
-    Slice valSlice = store.getSync(getBucket(key), keyValueSerdeManager.serializeDataKey(key, false));
+    Slice valSlice = store.getSync(getBucketTimeOrId(key), keyValueSerdeManager.serializeDataKey(key, false));
 
     if (valSlice == null || valSlice == BucketedState.EXPIRED || valSlice.length == 0) {
       return null;
@@ -236,13 +237,29 @@ public class SpillableMapImpl<K, V> implements Spillable.SpillableMap<K, V>, Spi
   @Override
   public void endWindow()
   {
+    boolean isTimeUnifiedStore = (store instanceof ManagedTimeUnifiedStateImpl);
     for (K key: cache.getChangedKeys()) {
-      store.put(getBucket(key), keyValueSerdeManager.serializeDataKey(key, true),
+      //the getBucket() returned in fact is time, the bucket assign then assigned the bucketId
+      long timeOrBucketId = bucket;
+      long bucketId = timeOrBucketId;
+      if (isTimeUnifiedStore) {
+        timeOrBucketId = getBucketTimeOrId(key);
+        bucketId = ((ManagedTimeUnifiedStateImpl)store).getTimeBucketAssigner().getTimeBucket(timeOrBucketId);
+      }
+      keyValueSerdeManager.updateBuffersForBucketChange(bucketId);
+      store.put(timeOrBucketId, keyValueSerdeManager.serializeDataKey(key, true),
           keyValueSerdeManager.serializeValue(cache.get(key)));
     }
 
     for (K key: cache.getRemovedKeys()) {
-      store.put(getBucket(key), keyValueSerdeManager.serializeDataKey(key, true), BufferSlice.EMPTY_SLICE);
+      long timeOrBucketId = bucket;
+      long bucketId = timeOrBucketId;
+      if (isTimeUnifiedStore) {
+        timeOrBucketId = getBucketTimeOrId(key);
+        bucketId = ((ManagedTimeUnifiedStateImpl)store).getTimeBucketAssigner().getTimeBucket(timeOrBucketId);
+      }
+      keyValueSerdeManager.updateBuffersForBucketChange(bucketId);
+      store.put(timeOrBucketId, keyValueSerdeManager.serializeDataKey(key, true), BufferSlice.EMPTY_SLICE);
     }
     cache.endWindow();
     keyValueSerdeManager.resetReadBuffer();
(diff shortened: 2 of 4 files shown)


davidyan74:
@siyuanh Can you please take a look? This seems to be related to the work you did for time bucket assignment for the windowed operator.

brightchen:
@siyuanh Could you have a look when you got time

library/src/main/java/org/apache/apex/malhar/lib/utils/serde/KeyValueSerdeManager.java
```diff
@@ -47,13 +50,28 @@ public KeyValueSerdeManager(Serde<K> keySerde, Serde<V> valueSerde)
...
-    this.valueBuffer = new SerializationBuffer(bucketInst.getValueStream());
+    bucketProvider = bp;
+    currentBucketId(bucketId);
+  }
 
+  /**
+   * The bucket can be changed. The write buffer should also be changed if bucket changed.
+   * @param bucketId
+   */
+  public void currentBucketId(long bucketId)
```
bhupeshchawda: rename to: `updateBuffersForBucket` ?

bhupeshchawda (commented):
Can you also add a unit tests for your changes?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeUnifiedStateImpl.java
```diff
@@ -64,6 +64,11 @@ public int getNumBuckets()
     return timeBucketAssigner.getNumBuckets();
   }
 
+  public long getTimeBucketId(long time)
+  {
+    return timeBucketAssigner.getTimeBucketAndAdjustBoundaries(time);
```
bhupeshchawda: Also, this method call has changed to `getTimeBucket()`. Please change.

brightchen:
@bhupeshchawda What I did is to improve the performance and try to avoid copy bytes and garbage collection. So the test is in the range of benchmark which we already have benchmark application. On the other side, this change should not impact the function which can be monitored by all existing unit tests. 
I implemented a mechanism to monitor all the memory used by all SerializationBuffer instances, but it was not accepted. So it's hard to monitor all SerializationBuffer instances.

bhupeshchawda:
@brightchen I understand that most of these changes are for performance enhancement.
However, as the JIRA indicates, this is a bug. I think it arises when ManagedTimeUnifiedStateImpl is used as the store, which has time buckets instead of key buckets. It seems you have also made changes to solve this issue. The test I was requesting was to make sure that this bug is resolved.

brightchen:
@bhupeshchawda Added unit test, please review

bhupeshchawda (approved):
Looks good. Will merge it soon if there are no other comments from the community.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeUnifiedStateImpl.java
```diff
@@ -65,6 +65,11 @@ public long getNumBuckets()
     return timeBucketAssigner.getNumBuckets();
   }
 
+  public long getTimeBucketId(long time)
+  {
```
chaithu14: Same can be achieve through the below call:
getTimeBucketAssigner().getTimeBucket(time). I would suggest remove this API.

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableMapImpl.java
```diff
@@ -236,13 +237,27 @@ public void beginWindow(long windowId)
   @Override
   public void endWindow()
   {
+    boolean isTimeUnifiedStore = (store instanceof ManagedTimeUnifiedStateImpl);
```
chaithu14, line 240: Move this out of endWindow(). I would suggest make this as class variable.

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableMapImpl.java
```diff
@@ -236,13 +237,27 @@ public void beginWindow(long windowId)
...
     for (K key: cache.getChangedKeys()) {
-      store.put(getBucket(key), keyValueSerdeManager.serializeDataKey(key, true),
+      //the getBucket() returned in fact is time, the bucket assign then assigned the bucketId
+      long timeOrBucketId = getBucket(key);
+      long bucketId = timeOrBucketId;
+      if (isTimeUnifiedStore) {
+        bucketId = ((ManagedTimeUnifiedStateImpl)store).getTimeBucketId(timeOrBucketId);
+      }
```
chaithu14: Can be done as follows: 
((ManagedTimeUnifiedStateImpl)store).getTimeBucketAssigner().getTimeBucket(..)

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableMapImpl.java
```diff
@@ -236,13 +237,27 @@ public void beginWindow(long windowId)
...
 
     for (K key: cache.getRemovedKeys()) {
-      store.put(getBucket(key), keyValueSerdeManager.serializeDataKey(key, true), BufferSlice.EMPTY_SLICE);
+      long timeOrBucketId = getBucket(key);
+      long bucketId = timeOrBucketId;
+      if (isTimeUnifiedStore) {
+        bucketId = ((ManagedTimeUnifiedStateImpl)store).getTimeBucketId(timeOrBucketId);
+      }
+      keyValueSerdeManager.updateBuffersForBucketChange(bucketId);
+      store.put(timeOrBucketId, keyValueSerdeManager.serializeDataKey(key, true), BufferSlice.EMPTY_SLICE);
```
chaithu14: If the type of managed state is ManagedTimeUnifiedStateImpl then the first argument for put() is time not time bucket.

I would suggest replace the above changes as below:
```java
long timeOrBucketId = bucket;
long bucketId = timeOrBucketId;
if (isTimeUnifiedStore) {
      timeOrBucketId = getBucket(key);
      bucketId = ((ManagedTimeUnifiedStateImpl)store).getTimeBucketAssigner().getTimeBucket(timeOrBucketId);
}
keyValueSerdeManager.updateBuffersForBucketChange(bucketId);
store.put(timeOrBucketId, keyValueSerdeManager.serializeDataKey(key, true), BufferSlice.EMPTY_SLICE);	
```

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableMapImpl.java
brightchen: make a class variable can slightly improve the performance. But this flag only apply in endWindow() and it's not really a class variable. I prefer put here

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableMapImpl.java
brightchen: I not really get the intension or benefit or this section of code. Could you explain more detail

brightchen: I kind got what your want. I don't have too much prefer in this case

brightchen:
@bhupeshchawda @chaithu14 applied comments. please review

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableMapImpl.java
chaithu14: Why do we need to recompute it in every end window? I think you can make it a class variable. It can be transient and initialized in the setup method.

chaithu14 (commented):
Rest looks good to me.

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableMapImpl.java
brightchen: The only reason to add it as class variable is due to performance, right? The benefit is increase very very little performance. It's happened every endWindow()  instead of each tuple, and the operation instanceof is in fact pretty light. But the problem is increase the complexity. I don't think buy it

brightchen:
@bhupeshchawda @chaithu14 Could you please merge if no further comments

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableMapImpl.java
bhupeshchawda: How would this increase the complexity? I think it would simplify the logic as this would need to be done just once in setup().

brightchen: As it added one member variable, So it increase the complexity. I don't think it simplify the logic, as the variable expand it's scoop.

chaithu14:
Looks good to me.

library/src/main/java/org/apache/apex/malhar/lib/state/spillable/SpillableMapImpl.java
brightchen: Here the complex I meant logic complex instead of runtime complex

PR #516, opened 2016-11-29
APEXMALHAR-2345 Purging time buckets in WindowedStorage

davidyan74:
@siyuanh Please review. This is based on the other PR you have #503. I'll add unit test for purging if this looks good.

### DIFF
--- a/benchmark/src/test/resources/log4j.properties
+++ b/benchmark/src/test/resources/log4j.properties
@@ -41,5 +41,5 @@ log4j.logger.org=info
 #log4j.logger.org.apache.commons.beanutils=warn
 log4j.logger.com.datatorrent=debug
 log4j.logger.org.apache.apex=debug
-log4j.logger.org.apache.apex.malhar.lib.state.managed=info
+log4j.logger.org.apache.apex.malhar.lib.state.managed=debug
 log4j.logger.com.datatorrent.common.util.FSStorageAgent=info
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/managed/AbstractManagedStateImpl.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/managed/AbstractManagedStateImpl.java
@@ -220,6 +220,8 @@ public abstract class AbstractManagedStateImpl
               buckets.get(bucketIdx).recoveredData(stateEntry.getKey(), bucketEntry.getValue());
             }
           }
+          // Skip write to WAL during recovery during replay from WAL.
+          // Data only needs to be transferred to bucket data files.
           checkpointManager.save(state, stateEntry.getKey(), true /*skipWritingToWindowFile*/);
         }
       } catch (IOException e) {
@@ -369,6 +371,7 @@ public abstract class AbstractManagedStateImpl
     }
     if (!flashData.isEmpty()) {
       try {
+        // write incremental state to WAL (skipWrite=false) before the checkpoint
         checkpointManager.save(flashData, windowId, false);
       } catch (IOException e) {
         throw new RuntimeException(e);
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
@@ -72,6 +72,7 @@ public class IncrementalCheckpointManager extends FSWindowDataManager
   protected transient ManagedStateContext managedStateContext;
 
   private final transient AtomicLong latestExpiredTimeBucket = new AtomicLong(-1);
+  private long latestPurgedTimeBucket = -1;
 
   private transient int waitMillis;
   private volatile long lastTransferredWindow = Stateless.WINDOW_ID;
@@ -109,8 +110,8 @@ public class IncrementalCheckpointManager extends FSWindowDataManager
           transferWindowFiles();
           if (latestExpiredTimeBucket.get() > -1) {
             try {
-              managedStateContext.getBucketsFileSystem().deleteTimeBucketsLessThanEqualTo(
-                  latestExpiredTimeBucket.getAndSet(-1));
+              latestPurgedTimeBucket = latestExpiredTimeBucket.getAndSet(-1);
+              managedStateContext.getBucketsFileSystem().deleteTimeBucketsLessThanEqualTo(latestPurgedTimeBucket);
             } catch (IOException e) {
               throwable.set(e);
               LOG.debug("delete files", e);
@@ -133,8 +134,10 @@ public class IncrementalCheckpointManager extends FSWindowDataManager
           Map<Long, Map<Slice, Bucket.BucketedValue>> buckets = savedWindows.remove(windowId);
 
           for (Map.Entry<Long, Map<Slice, Bucket.BucketedValue>> singleBucket : buckets.entrySet()) {
-            managedStateContext.getBucketsFileSystem().writeBucketData(windowId, singleBucket.getKey(),
-                singleBucket.getValue());
+            long bucketId = singleBucket.getKey();
+            if (bucketId > latestPurgedTimeBucket) {
+              managedStateContext.getBucketsFileSystem().writeBucketData(windowId, bucketId, singleBucket.getValue());
+            }
           }
           committed(windowId);
         } catch (Throwable t) {
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
@@ -111,6 +111,7 @@ public class IncrementalCheckpointManager extends FSWindowDataManager
           if (latestExpiredTimeBucket.get() > -1) {
             try {
               latestPurgedTimeBucket = latestExpiredTimeBucket.getAndSet(-1);
+              //LOG.debug("latestPurgedTimeBucket {}", latestPurgedTimeBucket);
               managedStateContext.getBucketsFileSystem().deleteTimeBucketsLessThanEqualTo(latestPurgedTimeBucket);
             } catch (IOException e) {
               throwable.set(e);
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeUnifiedStateImpl.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeUnifiedStateImpl.java
@@ -142,7 +142,7 @@ public class ManagedTimeUnifiedStateImpl extends AbstractManagedStateImpl implem
   @Override
   public void setup(Context.OperatorContext context)
   {
-    // set UnBoundedTimeBucketAssigner to this managed state impl
+    // set UnboundedTimeBucketAssigner to this managed state impl
     if (timeBucketAssigner == null) {
       UnboundedTimeBucketAssigner unboundedTimeBucketAssigner = new UnboundedTimeBucketAssigner();
       setTimeBucketAssigner(unboundedTimeBucketAssigner);
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/managed/MovingBoundaryTimeBucketAssigner.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/managed/MovingBoundaryTimeBucketAssigner.java
@@ -67,7 +67,7 @@ public class MovingBoundaryTimeBucketAssigner extends TimeBucketAssigner
   private int numBuckets;
   private transient long fixedStart;
   private transient boolean triggerPurge;
-  private transient long lowestPurgeableTimeBucket;
+  private transient long lowestPurgeableTimeBucket = -1;
 
 
   @Override
@@ -125,8 +125,9 @@ public class MovingBoundaryTimeBucketAssigner extends TimeBucketAssigner
       long move = (diffInBuckets + 1) * bucketSpanMillis;
       start += move;
       end += move;
-      triggerPurge = true;
       lowestPurgeableTimeBucket += diffInBuckets;
+      // trigger purge when lower bound changes
+      triggerPurge = (diffInBuckets > 0);
     }
     return key;
 
--- a/library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
@@ -25,6 +25,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 
 import javax.validation.ValidationException;
 
@@ -89,6 +90,8 @@ public abstract class AbstractWindowedOperator<InputT, OutputT, DataStorageT ext
   private long currentDerivedTimestamp = -1;
   private long timeIncrement;
   protected long fixedWatermarkMillis = -1;
+  private transient long streamingWindowId;
+  private transient TreeMap<Long, Long> streamingWindowToLatenessHorizon = new TreeMap<>();
 
   private Map<String, Component<Context.OperatorContext>> components = new HashMap<>();
 
@@ -407,7 +410,6 @@ public abstract class AbstractWindowedOperator<InputT, OutputT, DataStorageT ext
   public void dropTuple(Tuple input)
   {
     // do nothing
-    LOG.debug("Dropping late tuple {}", input);
   }
 
   @Override
@@ -465,6 +467,7 @@ public abstract class AbstractWindowedOperator<InputT, OutputT, DataStorageT ext
     } else {
       currentDerivedTimestamp += timeIncrement;
     }
+    streamingWindowId = windowId;
   }
 
   /**
@@ -517,6 +520,7 @@ public abstract class AbstractWindowedOperator<InputT, OutputT, DataStorageT ext
           }
         }
       }
+      streamingWindowToLatenessHorizon.put(streamingWindowId, horizon);
       controlOutput.emit(new WatermarkImpl(nextWatermark));
       this.currentWatermark = nextWatermark;
     }
@@ -622,6 +626,16 @@ public abstract class AbstractWindowedOperator<InputT, OutputT, DataStorageT ext
       if (component instanceof CheckpointNotificationListener) {
         ((CheckpointNotificationListener)component).committed(windowId);
       }
+    }
+    Long floorWindowId = streamingWindowToLatenessHorizon.floorKey(windowId);
+    if (floorWindowId != null) {
+      long horizon = streamingWindowToLatenessHorizon.get(floorWindowId);
+      windowStateMap.purge(horizon);
+      dataStorage.purge(horizon);
+      if (retractionStorage != null) {
+        retractionStorage.purge(horizon);
+      }
+      streamingWindowToLatenessHorizon.headMap(windowId, true).clear();
     }
   }
 }
--- a/library/src/main/java/org/apache/apex/malhar/lib/window/impl/SpillableWindowedPlainStorage.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/window/impl/SpillableWindowedPlainStorage.java
@@ -22,8 +22,13 @@ import java.util.Map;
 
 import javax.validation.constraints.NotNull;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.apex.malhar.lib.state.managed.ManagedTimeUnifiedStateImpl;
 import org.apache.apex.malhar.lib.state.spillable.Spillable;
 import org.apache.apex.malhar.lib.state.spillable.SpillableComplexComponent;
+import org.apache.apex.malhar.lib.state.spillable.SpillableStateStore;
 import org.apache.apex.malhar.lib.utils.serde.GenericSerde;
 import org.apache.apex.malhar.lib.utils.serde.Serde;
 import org.apache.apex.malhar.lib.window.Window;
@@ -45,6 +50,7 @@ public class SpillableWindowedPlainStorage<T> implements WindowedStorage.Windowe
   private long bucket;
   private Serde<Window> windowSerde;
   private Serde<T> valueSerde;
+  private static final Logger LOG = LoggerFactory.getLogger(SpillableWindowedPlainStorage.class);
 
   protected Spillable.SpillableMap<Window, T> windowToDataMap;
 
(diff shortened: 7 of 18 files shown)


davidyan74:
@tweise please also review

brightchen:
@davidyan74 I merged this PR and did test Spillable Benchmark. But I didn't see removal of any bucket folder. At the beginning, there are folder 345585, 345586, 345587 etc. And the end, these folders still there.

davidyan74:
@brightchen I have made some changes to fix the purging. I am seeing removal of the time bucket files now when I run your benchmark app. Please confirm.

@chandnisingh I made some small changes to IncrementalCheckpointManager.java. I am skipping the timebuckets that are already purged when transferring window files. Please review that part of this PR. Thanks.

@siyuanh and @tweise: Please review this PR.

davidyan74:
@chandnisingh The ManageStateImplTest.testFreeWindowTransferRaceCondition fails and I think it's because I'm skipping the timebuckets that are supposed to be purged. It looks like the test is checking the transfer of time bucket 0 even after time bucket 0 is purged. Please let me know if you have any idea how to fix it.

library/src/main/java/org/apache/apex/malhar/lib/join/AbstractManagedStateInnerJoinOperator.java
```diff
@@ -72,8 +73,12 @@ public void createStores()
       stream1Store.getTimeBucketAssigner().setBucketSpan(Duration.millis(bucketSpanTime));
       stream2Store.getTimeBucketAssigner().setBucketSpan(Duration.millis(bucketSpanTime));
     }
-    stream1Store.getTimeBucketAssigner().setExpireBefore(Duration.millis(getExpiryTime()));
-    stream2Store.getTimeBucketAssigner().setExpireBefore(Duration.millis(getExpiryTime()));
+    if (stream1Store.getTimeBucketAssigner() instanceof MovingBoundaryTimeBucketAssigner) {
```
tweise, line 76: Does it make sense to have different time bucket assigners for the stores? The stores anyways would need to be backed by a single state backend to ensure consistency, is that guaranteed?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -176,6 +177,12 @@ public void setup(OperatorContext context)
     operatorContext = context;
     fileAccess.init();
 
+    if ( timeBucketAssigner == null ) {
+      // set default time bucket assigner
+      MovingBoundaryTimeBucketAssigner movingBoundaryTimeBucketAssigner = new MovingBoundaryTimeBucketAssigner();
+      movingBoundaryTimeBucketAssigner.setPurgeListener(this);
```
tweise: Can the listener not be supplied through the constructor? Or is it needed at all since it is set below anyways?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -186,9 +193,9 @@ public void setup(OperatorContext context)
     if (buckets == null) {
       //create buckets array only once at start when it is not created.
```
tweise: update comment, maybe say why we need a map now

library/src/main/java/org/apache/apex/malhar/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -474,7 +481,7 @@ public FileAccess getFileAccess()
   /**
    * Sets the time bucket assigner. This can be used for plugging any custom time bucket assigner.
    *
-   * @param timeBucketAssigner a {@link TimeBucketAssigner}
+   * @param timeBucketAssigner a {@link MovingBoundaryTimeBucketAssigner}
```
tweise: Fix link

library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
```diff
@@ -48,7 +48,7 @@
 
 /**
  * Manages state which is written to files by windows. The state from the window files are then transferred to bucket
- * data files. This class listens to time expiry events issued by {@link TimeBucketAssigner}.
+ * data files. This class listens to time expiry events issued by {@link MovingBoundaryTimeBucketAssigner}.
```
tweise: why this change?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
```diff
@@ -72,6 +72,7 @@
   protected transient ManagedStateContext managedStateContext;
 
   private final transient AtomicLong latestExpiredTimeBucket = new AtomicLong(-1);
+  private transient long latestPurgedTimeBucket = -1;
```
tweise: why transient?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
```diff
@@ -109,8 +110,8 @@ public void run()
           transferWindowFiles();
           if (latestExpiredTimeBucket.get() > -1) {
             try {
-              managedStateContext.getBucketsFileSystem().deleteTimeBucketsLessThanEqualTo(
-                  latestExpiredTimeBucket.getAndSet(-1));
+              latestPurgedTimeBucket = latestExpiredTimeBucket.getAndSet(-1);
```
tweise, line 113: Is it correct to change the value asynchronously when it is used in transferWindowFiles?

tweise (changes requested):

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedStateImpl.java
```diff
@@ -88,7 +88,7 @@ public void endWindow()
 
   @Min(1)
   @Override
-  public int getNumBuckets()
+  public long getNumBuckets()
```
tweise, line 91: Why is this needed for number of buckets? Or is it something else?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeUnifiedStateImpl.java
```diff
@@ -121,14 +122,14 @@ public void endWindow()
   }
 
   @Override
-  protected void handleBucketConflict(int bucketIdx, long newBucketId)
+  protected void handleBucketConflict(long bucketIdx, long newBucketId)
```
tweise: bucketIdx seems a misnomer now, it is used like an ID.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeUnifiedStateImpl.java
```diff
@@ -59,22 +60,22 @@ public ManagedTimeUnifiedStateImpl()
...
 
   @Override
-  public int getNumBuckets()
+  public long getNumBuckets()
   {
     return timeBucketAssigner.getNumBuckets();
   }
 
   @Override
   public void put(long time, @NotNull Slice key, @NotNull Slice value)
```
tweise, line 69: Can you please call out in the class level javadoc how this class differs from ManagedTimeStateImpl?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/MovingBoundaryTimeBucketAssigner.java
```diff
@@ -0,0 +1,184 @@
...
+ * These properties once configured shouldn't be changed because that will result in different time-buckets
+ * for the same (key,time) pair after a failure.
+ * <p/>
+ *
+ * The time boundaries- start and end, move by multiples of time-bucket span. Any event with time < start
+ * is considered expired. The boundaries slide by {@link #getTimeBucket(long)}. The time which is passed as an
+ * argument to this method can be ahead of <code>end</code>. This means that the corresponding event is a future event
+ * (wrt TimeBucketAssigner) and cannot be ignored. Therefore it is accounted by sliding boundaries further.
+ *
+ * @since 3.4.0
```
tweise: remove since tag

library/src/main/java/org/apache/apex/malhar/lib/state/managed/MovingBoundaryTimeBucketAssigner.java
```diff
@@ -0,0 +1,184 @@
...
+      end = start + (numBuckets * bucketSpanMillis);
+
+      setInitialized(true);
+    }
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
+
```
tweise: extra linebreak

library/src/main/java/org/apache/apex/malhar/lib/state/managed/TimeExtractor.java
```diff
@@ -26,4 +26,27 @@
 public interface TimeExtractor<T>
 {
   long getTime(T t);
+
+  class FixedTimeExtractor<V> implements TimeExtractor<V>
```
tweise, line 30: Would BucketTimeMapper be a more appropriate name?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/UnboundedTimeBucketAssigner.java
```diff
@@ -0,0 +1,72 @@
...
+ *
+ *
+ */
+public class UnboundedTimeBucketAssigner extends TimeBucketAssigner
+{
+  @Override
+  public long getTimeBucket(long time)
+  {
+    if (time < 0) {
+      throw new IllegalArgumentException("Time: " + time + " is illegal");
```
tweise: use PreConditions.checkArgument(...)

library/src/main/java/org/apache/apex/malhar/lib/state/managed/UnboundedTimeBucketAssigner.java
```diff
@@ -0,0 +1,72 @@
...
+  public void setup(@NotNull ManagedStateContext managedStateContext)
+  {
+    super.setup(managedStateContext);
+    setInitialized(true);
+  }
+
+  @Override
+  public void teardown()
+  {
+
```
tweise: remove extra lines

library/src/main/java/org/apache/apex/malhar/lib/window/impl/AbstractWindowedOperator.java
```diff
@@ -407,7 +410,7 @@ public boolean isTooLate(long timestamp)
   public void dropTuple(Tuple input)
   {
     // do nothing
-    LOG.debug("Dropping late tuple {}", input);
+    //LOG.debug("Dropping late tuple {}", input);
```
tweise: remove this?

tweise (changes requested):

library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
davidyan74: transferWindowFiles and this assignment are executed by the same thread. This shouldn't be a problem. @chandnisingh any thoughts?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
davidyan74: removed transient

library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
tweise: OK. However it is still not clear how with the asynchronous transfer call we achieve idempotent behavior WRT latestPurgedTimeBucket?

davidyan74: I see what you're saying. However, deleting time buckets on a separate thread can also be a problem with idempotency, right? The intention of the change here is to record the latest purged time bucket so that the transferWindowFIles call will not try to transfer the deleted time buckets.

tweise: Asynchronous delete should only be done for data of committed windows, then there is no question of idempotency.

davidyan74: In the case of windowed operator, this is for for event-time windows that are beyond the lateness horizon at the time when the streaming window is committed. In that case, it's not possible that the data deleted here may be needed in the future. This may not be true for other usages, and maybe that's why the unit test is failing. I need suggestion on how to do this the correct way.

benchmark/src/test/resources/log4j.properties
```diff
@@ -23,7 +23,7 @@ log4j.appender.CONSOLE=org.apache.log4j.ConsoleAppender
 log4j.appender.CONSOLE.layout=org.apache.log4j.PatternLayout
 log4j.appender.CONSOLE.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
 log4j.appender.CONSOLE.threshold=${test.log.console.threshold}
-test.log.console.threshold=INFO
+test.log.console.threshold=DEBUG
```
siyuanh: Will this make the test log too much and kill the CI?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/AbstractManagedStateImpl.java
```diff
@@ -128,14 +129,14 @@
 {
   private long maxMemorySize;
 
-  protected int numBuckets;
+  protected long numBuckets;
```
siyuanh, line 132: Can we merge my change and do another PR?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/TimeExtractor.java
siyuanh: The class is supposed to assign a fixed time to all events.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/AbstractManagedStateImpl.java
davidyan74: @siyuanh  Sorry there is something wrong with my github email notification so I didn't know you had changed #503. I have replied to your comments about time buckets for session windows there.

davidyan74: The plan is to have your PR merged and then I would do a rebase for this PR

siyuanh:
@davidyan74  Hey David, I made some change on my branch, can you please refresh? Thanks!

siyuanh:
@davidyan74 Hold on, there is a bug

library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
tweise: If you record latest purged after the time buckets were deleted and the operator fails before the next checkpoint, what happens?

davidyan74: Each streaming window records its own lateness horizon. The purge happens only if the time bucket is beyond the lateness horizon at the committed streaming window, so we never need the time bucket that is purged.

davidyan74: Oh wait, I see that you're referring to the recorded latest purged time bucket. In that case, the latest purged time bucket will be reverted to the previous value but the time bucket was actually purged forever before the checkpoint. However, that should not matter since we will never need the purged time bucket anyway.

tweise:
@chandnisingh I'm looking at IncrementalCheckpointManager. It puts the data into a transient map and in the background writes it to the store. What happens when the operator fails and some data wasn't written? Shouldn't data be flushed from beforeCheckpoint (since there is an option to skip write in save(.., skipWriteToWindowFile))?

davidyan74:
rebased after merging #503

davidyan74:
The issue with IncrementalCheckpointManager and the failure of the unit test ManageStateImplTest.testFreeWindowTransferRaceCondition are still pending.

otterc:
@tweise 
lines 366 in AbstractManagedStateImpl, triggers the CheckpointManager to write the incremental state for all buckets in wal. skipWrite flag is set to  `false` here so this data is saved to wal. If there is a failure here, the operator is recovered from previous checkpoint  (wal is truncated as well) and tuples are going to be replayed. 

skipWrite flag is true only during recovery of ManagedState where it reads from WAL. Since the data  is already in wal, it doesn't need to be written to WAL again. However this data needs to be transferred to bucket file system (data files)

benchmark/src/test/resources/log4j.properties
tweise: @davidyan74 is this intended to remain in the PR?

tweise:
@chandnisingh thanks, this makes sense now. I might submit a PR to add this to the javadoc. Can you please also look at the question from @davidyan74 regarding the unit test failure.

My understanding is that the change in IncrementalCheckpointManager.transferWindowFiles should not be necessary since the state will be purged via deleteTimeBucketsLessThanEqualTo?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
```diff
@@ -133,8 +134,10 @@ protected void transferWindowFiles()
           Map<Long, Map<Slice, Bucket.BucketedValue>> buckets = savedWindows.remove(windowId);
 
           for (Map.Entry<Long, Map<Slice, Bucket.BucketedValue>> singleBucket : buckets.entrySet()) {
-            managedStateContext.getBucketsFileSystem().writeBucketData(windowId, singleBucket.getKey(),
-                singleBucket.getValue());
+            long bucketId = singleBucket.getKey();
+            if (bucketId > latestPurgedTimeBucket) {
```
tweise, line 139: Is there a valid scenario where expired data is written at this point?

davidyan74: Yes, this happened when I tried to run the benchmark app written by @brightchen

davidyan74: The purged time bucket should probably be removed from the meta data per window within managed state upon the purge, if it's not already doing it. I'll take a look. @chandnisingh please comment on this when you have a chance.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
otterc: `latestPurgedTimeBucket` is modified by another thread, and checkpointing is done by operator thread so what value is persisted as part of checkpoint? I mean what is saved cannot be dependent on as it may not be consistent.

library/src/main/java/org/apache/apex/malhar/lib/window/impl/SpillableWindowedPlainStorage.java
```diff
@@ -143,4 +149,16 @@ public void setup(Context.OperatorContext context)
   public void teardown()
   {
   }
+
+  @Override
+  public void purge(long horizonMillis)
+  {
+    SpillableStateStore store = scc.getStore();
+    if (store instanceof ManagedTimeUnifiedStateImpl) {
```
tweise, line 157: Why is this conditional? Is it possible to use a different store and if so how will data be purged?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
otterc: What @tweise mentioned is right that the state will be purged via deleteTimeBucketsLessThanEqualTo. So by having this condition are we trying to avoid data which is not transferred yet for a bucket but is already ready for purging? In that case it makes sense. However adding a test to validate that will be good.
Also, `latestPurgedTimeBucket` seems to be a wrong name for this variable. In here it is the main bucket so it should be`lasterPurgedBucketId`. Also I think it should be transient because I mentioned in my earlier comment that it is being modified by another thread and checkpointed by another.

davidyan74: The intention here is to avoid transferring buckets that are **already** purged. Without this condition, my test run throws an exception complaining that the bucket it tries to transfer does not exist. I will add a unit test for this, but this looks like it's causing another unit test failure in ManageStateImplTest.testFreeWindowTransferRaceCondition so I wanted to make sure what I'm doing is correct first.

Did you mean `latestPurgedBucketId`? If this variable is not transient, the value will not be valid after recovery, and it would be possible that it would still try to transfer purged buckets after recovery. I'm not exactly familiar with the Managed State code so correct me if I'm wrong.

library/src/main/java/org/apache/apex/malhar/lib/window/impl/SpillableWindowedPlainStorage.java
davidyan74: It will not be purged. I think I should make SpillableStateStore extend ManagedTimeUnifiedStateImpl.

benchmark/src/test/resources/log4j.properties
davidyan74: No, I will change it back when this PR is ready to be merged.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/IncrementalCheckpointManager.java
tweise: I found an issue in MovingBoundaryTimeBucketAssigner, where the lower bound was initialized with 0 instead of -1, causing purging to be triggered with 0, which is a valid bucket ID. I also made another change to avoid trigger purging unless necessary. @chandnisingh I also see that `StateTrackerTest` takes a long time to run (3 minutes), is that expected?

davidyan74:
@tweise Your fix looks good and it has fixed the unit test failure. We still need to write unit tests for purge verification and failure recovery verification. They can be done separately though. If you are okay with this, please merge this PR.

library/src/test/java/org/apache/apex/malhar/lib/state/managed/MovingBoundaryTimeBucketAssignerTest.java
```diff
@@ -96,20 +98,43 @@ public void testTimeBucketKey()
...
     long time1 = Duration.standardSeconds(9).getMillis() + referenceTime;
     Assert.assertEquals("time bucket", 10, testMeta.timeBucketAssigner.getTimeBucket(time1) );
+    testMeta.timeBucketAssigner.endWindow();
+    Assert.assertEquals("purgeLessThanEqualTo", 7, purgeLessThanEqualTo.longValue());
+    purgeLessThanEqualTo.setValue(-2);
 
     long time2 = Duration.standardSeconds(10).getMillis()  + referenceTime;
     Assert.assertEquals("time bucket", 11, testMeta.timeBucketAssigner.getTimeBucket(time2) );
+    testMeta.timeBucketAssigner.endWindow();
+// TODO: why is purgeLessThanEqualTo not moving to 8 here?
```
tweise, line 131: @siyuanh please have a look at this test and the purge callbacks. This does not look right. When I comment out time0 then the previous value changes from 7 to 8.. This PR is merged, so any further changes should be a new PR.

PR #517, opened 2016-11-30
APEXMALHAR-2343 Count fix and accumulation input parameter type change 

davidyan74:
@brightchen @chinmaykolhatkar @siyuanh  Please have a look

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/Count.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/Count.java
@@ -26,7 +26,7 @@ import org.apache.commons.lang3.mutable.MutableLong;
  *
  * @since 3.5.0
  */
-public class Count<T> implements Accumulation<T, MutableLong, Long>
+public class Count implements Accumulation<Object, MutableLong, Long>
 {
 
   @Override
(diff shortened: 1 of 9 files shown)


chinmaykolhatkar:
LGTM.

chinmaykolhatkar:
@davidyan74 The changes looks good to me. Can you rebase and squash?
The test failure doesn't seem to be related to these code changes, hence I think the PR can be merged, unless anyone sees any issues.

siyuanh:
@davidyan74 @brightchen  In ApexWindowedStreamImpl line 88. If you always increase the count by 1 you don't need that map function

davidyan74:
@siyuanh I have changed ApexWindowedStreamImpl to use SumLong instead of Count. I don't want to remove the map function because it actually does something other than mapping to 1. It actually adds the current timestamp if the timestamp does not already exist. And by the way, I think the automatic adding of the timestamp may break idempotency. In any case, it's not the scope of this PR.

davidyan74:
@siyuanh If this looks fine with you, please merge. The Count accumulation is no longer used anywhere within Malhar.

davidyan74:
@siyuanh please review and merge when you have a chance. This is blocking some other PRs. Thanks.

siyuanh:
@davidyan74 Just reviewed and rebased yesterday, will merge it today

PR #521, opened 2016-12-09
APEXMALHAR-2366 #resolve #comment Apply BloomFilter to Bucket, use in…

brightchen:
…ternal BloomFilter

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
@@ -232,6 +232,12 @@ public interface Bucket extends ManagedStateComponent, KeyValueByteStreamProvide
 
     protected ConcurrentLinkedQueue<Long> windowsForFreeMemory = new ConcurrentLinkedQueue<>();
 
+    private static boolean disableBloomFilterByDefault = false;
+    private boolean disableBloomFilter = disableBloomFilterByDefault;
+    private static int bloomFilterDefaultBitSize = 1000000;
+    private SliceBloomFilter bloomFilter = null;
+    private int bloomFilterBitSize = bloomFilterDefaultBitSize;
+
     private DefaultBucket()
     {
       //for kryo
@@ -247,6 +253,9 @@ public interface Bucket extends ManagedStateComponent, KeyValueByteStreamProvide
     public void setup(@NotNull ManagedStateContext managedStateContext)
     {
       this.managedStateContext = Preconditions.checkNotNull(managedStateContext, "managed state context");
+      if (!disableBloomFilter && bloomFilter == null) {
+        bloomFilter = new SliceBloomFilter(bloomFilterBitSize, 0.99);
+      }
     }
 
     @Override
@@ -353,6 +362,33 @@ public interface Bucket extends ManagedStateComponent, KeyValueByteStreamProvide
       }
     }
 
+
+    private int filteredCount = 0;
+    private int unfilteredCount = 0;
+    private static final int STATISTICS_BARRIER_NUM = 10000;
+    /**
+     * Test the result of using bloom filter and remove it if it not helpful for improving the performance.
+     *
+     * @param mightContain true if collection might contain the value
+     */
+    private void verifyBloomFilter(boolean mightContain)
+    {
+      if (!mightContain) {
+        filteredCount++;
+      } else {
+        unfilteredCount++;
+      }
+
+      if (unfilteredCount + filteredCount > STATISTICS_BARRIER_NUM && unfilteredCount > filteredCount * 4) {
+        unloadBloomFilter();
+      }
+    }
+
+    private void unloadBloomFilter()
+    {
+      bloomFilter = null;
+    }
+
     /**
      * Returns the value for the key from a valid time-bucket reader. Here, valid means the time bucket which is not purgeable.
      * If the timebucketAssigner is of type MovingBoundaryTimeBucketAssigner and the time bucket is purgeable, then return null.
@@ -362,11 +398,21 @@ public interface Bucket extends ManagedStateComponent, KeyValueByteStreamProvide
      */
     private BucketedValue getValueFromTimeBucketReader(Slice key, long timeBucket)
     {
-
       if (managedStateContext.getTimeBucketAssigner() instanceof MovingBoundaryTimeBucketAssigner &&
           timeBucket <= ((MovingBoundaryTimeBucketAssigner)managedStateContext.getTimeBucketAssigner()).getLowestPurgeableTimeBucket()) {
         return null;
       }
+
+      if (bloomFilter != null) {
+        boolean mightContain = bloomFilter.mightContain(key);
+
+        verifyBloomFilter(mightContain);
+
+        if (!mightContain) {
+          return null;
+        }
+      }
+
       FileAccess.FileReader fileReader = readers.get(timeBucket);
       if (fileReader != null) {
         return readValue(fileReader, key, timeBucket);
@@ -417,6 +463,7 @@ public interface Bucket extends ManagedStateComponent, KeyValueByteStreamProvide
     {
       // This call is lightweight
       releaseMemory();
+
       key = SliceUtils.toBufferSlice(key);
       value = SliceUtils.toBufferSlice(value);
 
@@ -460,6 +507,14 @@ public interface Bucket extends ManagedStateComponent, KeyValueByteStreamProvide
         entryIter.remove();
 
         for (Map.Entry<Slice, BucketedValue> entry : windowData.entrySet()) {
+          /**
+           * The data still in memory and reachable before the memory released
+           * So put key into bloom filter here
+           */
+          if (bloomFilter != null) {
+            bloomFilter.put(entry.getKey());
+          }
+
           memoryFreed += entry.getKey().length + entry.getValue().getSize();
         }
       }
@@ -511,6 +566,7 @@ public interface Bucket extends ManagedStateComponent, KeyValueByteStreamProvide
     public Map<Slice, BucketedValue> checkpoint(long windowId)
     {
       releaseMemory();
+
       try {
         //transferring the data from flash to check-pointed state in finally block and re-initializing the flash.
         return flash;
@@ -634,6 +690,48 @@ public interface Bucket extends ManagedStateComponent, KeyValueByteStreamProvide
     {
       return valueStream;
     }
+
+    public static int getBloomFilterDefaultBitSize()
+    {
+      return bloomFilterDefaultBitSize;
+    }
+
+    public static void setBloomFilterDefaultBitSize(int bloomFilterDefaultBitSize)
+    {
+      DefaultBucket.bloomFilterDefaultBitSize = bloomFilterDefaultBitSize;
+    }
+
+    public int getBloomFilterBitSize()
+    {
+      return bloomFilterBitSize;
+    }
+
+    public void setBloomFilterBitSize(int bloomFilterBitSize)
+    {
+      this.bloomFilterBitSize = bloomFilterBitSize;
+    }
+
+    public static boolean isDisableBloomFilterByDefault()
+    {
+      return disableBloomFilterByDefault;
+    }
+
+    public static void setDisableBloomFilterByDefault(boolean disableBloomFilterByDefault)
+    {
+      DefaultBucket.disableBloomFilterByDefault = disableBloomFilterByDefault;
+    }
+
+    public boolean isDisableBloomFilter()
+    {
+      return disableBloomFilter;
+    }
+
+    public void setDisableBloomFilter(boolean disableBloomFilter)
+    {
+      this.disableBloomFilter = disableBloomFilter;
+      this.unloadBloomFilter();
+    }
+
 
     private static final Logger LOG = LoggerFactory.getLogger(DefaultBucket.class);
   }
--- a/library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
@@ -1,0 +1,459 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.lib.state.managed;
+
+import java.util.BitSet;
+
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * This class implemented BloomFilter algorithm, the key is Slice
+ *
+ */
+public class SliceBloomFilter
+{
+  private BitSet bitset;
+  private int bitSetSize;
+  private int expectedNumberOfFilterElements; // expected (maximum) number of elements to be added
+  private int numberOfAddedElements; // number of elements actually added to the Bloom filter
+  private int numberOfHashes; // number of hash functions
+  protected transient HashFunction hasher = new HashFunction();
+
+  /**
+   * Set the attributes to the empty Bloom filter. The total length of the Bloom
+   * filter will be bitsPerElement*expectedNumberOfFilterElements.
+   *
+   * @param bitsPerElement
+   *          is the number of bits used per element.
+   * @param expectedNumberOfFilterElements
+   *          is the expected number of elements the filter will contain.
+   * @param numberOfHashes
+   *          is the number of hash functions used.
+   */
+  private void SetAttributes(double bitsPerElement, int expectedNumberOfFilterElements, int numberOfHashes)
+  {
+    this.expectedNumberOfFilterElements = expectedNumberOfFilterElements;
+    this.numberOfHashes = numberOfHashes;
+    this.bitSetSize = (int)Math.ceil(bitsPerElement * expectedNumberOfFilterElements);
+    numberOfAddedElements = 0;
+    this.bitset = new BitSet(bitSetSize);
+  }
+
+  private SliceBloomFilter()
+  {
+    //for kyro
+  }
+
+  /**
+   * Constructs an empty Bloom filter with a given false positive probability.
+   *
+   * @param expectedNumberOfElements
+   *          is the expected number of elements the filter will contain.
+   * @param falsePositiveProbability
+   *          is the desired false positive probability.
+   */
+  public SliceBloomFilter(int expectedNumberOfElements, double falsePositiveProbability)
+  {
+    if (this.bitset == null) {
+      SetAttributes(Math.ceil(-(Math.log(falsePositiveProbability) / Math.log(2))) / Math.log(2), // c = k / ln(2)
+          expectedNumberOfElements, (int)Math.ceil(-(Math.log(falsePositiveProbability) / Math.log(2))));
+    }
+  }
+
+  /**
+   * Generate integer array based on the hash function till the number of
+   * hashes.
+   *
+   * @param slice
+   *          specifies input slice.
+   * @return array of int-sized hashes
+   */
+  private int[] createHashes(Slice slice)
+  {
+    int[] result = new int[numberOfHashes];
+    long hash64 = hasher.hash(slice);
+    // apply the less hashing technique
+    int hash1 = (int)hash64;
+    int hash2 = (int)(hash64 >>> 32);
+    for (int i = 1; i <= numberOfHashes; i++) {
+      int nextHash = hash1 + i * hash2;
+      if (nextHash < 0) {
+        nextHash = ~nextHash;
+      }
+      result[i - 1] = nextHash;
+    }
+    return result;
+  }
+
+  /**
+   * Calculates the expected probability of false positives based on the number
+   * of expected filter elements and the size of the Bloom filter.
+   *
+   * @return expected probability of false positives.
+   */
+  public double expectedFalsePositiveProbability()
+  {
+    return getFalsePositiveProbability(expectedNumberOfFilterElements);
+  }
+
+  /**
+   * Calculate the probability of a false positive given the specified number of
+   * inserted elements.
+   *
+   * @param numberOfElements
+   *          number of inserted elements.
+   * @return probability of a false positive.
+   */
+  public double getFalsePositiveProbability(double numberOfElements)
+  {
+    // (1 - e^(-k * n / m)) ^ k
+    return Math.pow((1 - Math.exp(-numberOfHashes * numberOfElements / bitSetSize)), numberOfHashes);
+  }
+
+  /**
+   * Get the current probability of a false positive. The probability is
+   * calculated from the size of the Bloom filter and the current number of
+   * elements added to it.
+   *
+   * @return probability of false positives.
+   */
+  public double getFalsePositiveProbability()
+  {
+    return getFalsePositiveProbability(numberOfAddedElements);
+  }
+
+  /**
+   * Returns the value chosen for numberOfHashes.<br />
+   * <br />
+   * numberOfHashes is the optimal number of hash functions based on the size of
+   * the Bloom filter and the expected number of inserted elements.
+   *
+   * @return optimal numberOfHashes.
+   */
+  public int getNumberOfHashes()
+  {
+    return numberOfHashes;
+  }
+
+  /**
+   * Sets all bits to false in the Bloom filter.
+   */
+  public void clear()
+  {
+    bitset.clear();
+    numberOfAddedElements = 0;
+  }
+
+  /**
+   * Adds a slice of byte array to the Bloom filter.
+   *
+   * @param slice
+   *          slice of byte array to add to the Bloom filter.
+   */
+  public void put(Slice slice)
+  {
+    int[] hashes = createHashes(slice);
+    for (int hash : hashes) {
+      bitset.set(Math.abs(hash % bitSetSize), true);
+    }
+    numberOfAddedElements++;
+  }
+
+  /**
+   * Returns true if the slice of byte array could have been inserted into the Bloom
+   * filter. Use getFalsePositiveProbability() to calculate the probability of
+   * this being correct.
+   *
+   * @param slice
+   *          slice of byte array to check.
+   * @return true if the array could have been inserted into the Bloom filter.
+   */
+  public boolean mightContain(Slice slice)
+  {
+    int[] hashes = createHashes(slice);
+    for (int hash : hashes) {
+      if (!bitset.get(Math.abs(hash % bitSetSize))) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
+   * Read a single bit from the Bloom filter.
+   *
+   * @param bit
+   *          the bit to read.
+   * @return true if the bit is set, false if it is not.
+   */
+  public boolean getBit(int bit)
+  {
+    return bitset.get(bit);
+  }
+
+  /**
+   * Set a single bit in the Bloom filter.
+   *
+   * @param bit
+   *          is the bit to set.
+   * @param value
+   *          If true, the bit is set. If false, the bit is cleared.
+   */
+  public void setBit(int bit, boolean value)
+  {
+    bitset.set(bit, value);
+  }
+
+  /**
+   * Return the bit set used to store the Bloom filter.
+   *
+   * @return bit set representing the Bloom filter.
+   */
+  public BitSet getBitSet()
+  {
+    return bitset;
+  }
+
+  public void setBitSet(BitSet bitset)
+  {
+    this.bitset = bitset;
+  }
+
+  /**
+   * Returns the number of bits in the Bloom filter. Use count() to retrieve the
+   * number of inserted elements.
+   *
+   * @return the size of the bitset used by the Bloom filter.
+   */
+  public int size()
+  {
+    return this.bitSetSize;
+  }
+
+  /**
+   * Returns the number of elements added to the Bloom filter after it was
+   * constructed or after clear() was called.
+   *
+   * @return number of elements added to the Bloom filter.
+   */
+  public int count()
+  {
+    return this.numberOfAddedElements;
+  }
+
+  /**
+   * Returns the expected number of elements to be inserted into the filter.
+   * This value is the same value as the one passed to the constructor.
+   *
+   * @return expected number of elements.
+   */
+  public int getExpectedNumberOfElements()
+  {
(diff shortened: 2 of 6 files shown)


library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -353,6 +362,41 @@ public Slice get(Slice key, long timeBucket, ReadSource readSource)
       }
     }
 
+
+    private int filtedCount = 0;
```
davidyan74: Check spelling

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -353,6 +362,41 @@ public Slice get(Slice key, long timeBucket, ReadSource readSource)
...
+    private int unfiltedCount = 0;
```
davidyan74: Check spelling

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -353,6 +362,41 @@ public Slice get(Slice key, long timeBucket, ReadSource readSource)
...
+     * @param mightContain true if collection might contain the value
+     */
+    private void verifyBloomFilter(boolean mightContain)
+    {
+      if (!mightContain) {
+        filtedCount++;
+      } else {
+        unfiltedCount++;
+      }
+      long now = System.currentTimeMillis();
```
davidyan74: Would this introduce problems with idempotency?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
```diff
@@ -0,0 +1,459 @@
...
+
+import java.util.BitSet;
+
+import com.datatorrent.netlet.util.Slice;
+
+/**
+ * This class implemented BloomFilter algorithm, the key is Slice
+ *
+ */
+public class SliceBloomFilter
```
davidyan74, line 29: Is the code of this class copied from somewhere? If so, is there a reason why we don't use it as a library?

davidyan74 (changes requested):

library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
brightchen: @davidyan74 
Yes, I copied from org.apache.hadoop.util.bloom.BloomFilter and modified. The reason of copy/modify instead of using are
- The original BloomFilter can't be serialized.
- The original BloomFilter use temporary memory, this one use SerializationBuffer

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: @davidyan74 this change only impact performance but don't impact functional. So I think it does not break the idempotency.

brightchen:
@davidyan74 applied comments, please review

library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
davidyan74: It would be good if you include that information in the javadoc of this class. Let me spend some time later to look at this more. Since this has only to do with performance and does not affect the functionality of the Windowed Operator, this can wait, correct?

davidyan74: @brightchen is there a reason why we are not using Guava's Bloom Filter implementation?

bhupeshchawda:
@brightchen Added some comments in the JIRA

brightchen:
@bhupeshchawda commented in the jira

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -232,6 +232,12 @@ public int hashCode()
 
     protected ConcurrentLinkedQueue<Long> windowsForFreeMemory = new ConcurrentLinkedQueue<>();
 
+    private static boolean disableBloomFilterByDefault = false;
```
bhupeshchawda, line 235: Do we need this constant? Can't we directly use `TRUE`/ `FALSE` instead?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -232,6 +232,12 @@ public int hashCode()
...
+    private boolean disableBloomFilter = disableBloomFilterByDefault;
```
bhupeshchawda, line 236: Rename to `enableBloomFilter`? Also can we enable it by default? `enableBloomFilter = TRUE`

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -232,6 +232,12 @@ public int hashCode()
...
+    private static int bloomFilterDefaultBitSize = 1000000;
```
bhupeshchawda, line 237: If this is supposed to be a constant, can we make it something like: `DEFAULT_BLOOM_FILTER_SIZE`?
Also make all constants final?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -353,6 +362,41 @@ public Slice get(Slice key, long timeBucket, ReadSource readSource)
       }
     }
 
+
+    private int filteredCount = 0;
+    private int unfilteredCount = 0;
+    private long verifyBloomFilterBeginTime = 0;
+    private long verifyBloomFilterInterval = 5000;
```
bhupeshchawda: Can we move these declarations to the top?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -353,6 +362,41 @@ public Slice get(Slice key, long timeBucket, ReadSource readSource)
...
+    /**
+     * Test the result of using bloom filter and remove it if it not helpful for improving the performance.
+     *
+     * @param mightContain true if collection might contain the value
+     */
+    private void verifyBloomFilter(boolean mightContain)
```
bhupeshchawda, line 374: I think this might not be the right way to see if the bloom filter is performing well. Since the number of elements added to the bloom filter will affect the false positive prob., can you try to use `bloomFilter.expectedFalsePositiveProbability()` to see when we can stop using the bloom filter?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -411,6 +465,7 @@ public void put(Slice key, long timeBucket, Slice value)
     {
       // This call is lightweight
       releaseMemory();
+
```
bhupeshchawda, line 466: unnecessary change

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -361,6 +405,16 @@ public Slice get(Slice key, long timeBucket, ReadSource readSource)
      */
     private BucketedValue getValueFromTimeBucketReader(Slice key, long timeBucket)
     {
+      if (bloomFilter != null) {
```
bhupeshchawda: I think we should use the bloom filter at the first `get` call.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -454,6 +509,14 @@ public long freeMemory(long windowId) throws IOException
         entryIter.remove();
 
         for (Map.Entry<Slice, BucketedValue> entry : windowData.entrySet()) {
+          /**
+           * The data still in memory and reachable before the memory released
+           * So put key into bloom filter here
+           */
+          if (bloomFilter != null) {
```
bhupeshchawda, line 514: Adding an element to the bloom filter is best done when the element is fetched. This should also be in the first `get` call.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -506,6 +569,7 @@ protected long releaseMemory()
     public Map<Slice, BucketedValue> checkpoint(long windowId)
     {
       releaseMemory();
+
```
bhupeshchawda, line 569: unnecessary

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -630,6 +694,49 @@ public WindowedBlockStream getValueStream()
...
+    public void setDisableBloomFilter(boolean disableBloomFilter)
+    {
+      this.disableBloomFilter = disableBloomFilter;
+      this.unloadBloomFilter();
+    }
+
+
     private static final Logger LOG = LoggerFactory.getLogger(DefaultBucket.class);
+
```
bhupeshchawda: avoid unnecessary changes

library/src/test/java/org/apache/apex/malhar/lib/state/managed/DefaultBucketTest.java
```diff
@@ -52,6 +53,9 @@
     @Override
     protected void starting(Description description)
     {
+      //lots of test case get around the normal workflow and directly write to file. So should disable bloom filter
+      DefaultBucket.setDisableBloomFilterByDefault(true);
```
bhupeshchawda, line 57: Don't understand the reason why the bloom filter is disabled here?

library/src/test/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateImplTest.java
```diff
@@ -81,6 +83,7 @@ public void testAsyncGetFromReaders() throws IOException, ExecutionException, In
     Slice zero = ManagedStateTestUtils.getSliceFor("0");
     long time = System.currentTimeMillis();
 
+    DefaultBucket.setDisableBloomFilterByDefault(true);
```
bhupeshchawda, line 88: Same here, why disable?

library/src/test/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeUnifiedStateImplTest.java
```diff
@@ -116,6 +118,8 @@ public void testAsyncGet() throws ExecutionException, InterruptedException
   @Test
   public void testSyncGetFromFiles() throws IOException, ExecutionException, InterruptedException
   {
+    DefaultBucket.setDisableBloomFilterByDefault(true);
```
bhupeshchawda, line 121: Same comment.

bhupeshchawda (changes requested):

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: My intension is giving the client code a choice to use or disable the bloom filter for all bucket. As we don't have a central place to manage the instances. It's hard to get the instance for app developer. I'll change it into public

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: I treat enable bloom filter as the default. that's why use disableBloomFilter for special case

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: I give a chance for the application developer to change the default bit size in class level instead of instance level.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: The reason I put them here is all of them in fact only used by method verifyBloomFilter()

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: I'll look into it.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: First get is get from memory, which don't take too much time. And if we take check for the first get, which mean we need to put the flag at the first put(), which probably decrease the accurate very quickly

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: The flag will be put when read data from files.
Let assume everything is in memory, do we still need bloom filter? probably not as it's simple a map. So I think filter only applies to decrease the chance of unnecessary reading from files

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: I already provides setter/getter, so private is ok

library/src/test/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateImplTest.java
brightchen: the bloom filter works only go through put() and get() method of bucket. But this unit test case write data direct data from file instead of go through put(), so the bloom filter didn't keep the flags and get() return null due to filtered by Bloom Filter.

library/src/test/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeUnifiedStateImplTest.java
brightchen: same as previous one.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: just separate the logic section

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: just separate the logic section

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
bhupeshchawda: If a centralized place is needed, wouldn't AbstractManagedStateImpl be a better place to enable / disable managed state?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
bhupeshchawda: I still think the boolean should be named `enableBloomFilter`. You can set it to true and move it to the AbstractManagedStateImpl instead of Bucket.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
bhupeshchawda: I meant the constant value set to 1000000 should be a static final field. In fact all defaults must be public static final constants and set in the AbstractManagedStateImpl, instead of Bucket.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
bhupeshchawda: How about the `getFromReaders` call then? That can be a place for using the bloom filter. Also once the value is fetched, it should immediately be inserted into the bloom filter. I don't understand how that would decrease the accuracy of the filter. Can you clarify the second part of your comment?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
bhupeshchawda: I agree. Can we use `getFromReaders` in that case to insert into bloom filter? I don't think `freeMemory` is the right place to put values in the bloom filter.

library/src/test/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateImplTest.java
bhupeshchawda: I understand. In that case, I think we should have a unit test which actually executes the code changes done for supporting bloom filter. Probably a test which inserts a fixed number of keys into the store and uses bloom filter to see how much can be fetched without accessing the disk.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
bhupeshchawda: @davidyan74 are you okay with the code for bloom filter that @brightchen has taken from  org.apache.hadoop.util.bloom.BloomFilter? We'll definitely add this part into the javadoc of the class. Is there anything else that you wanted to look into?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: What I meant about instance management is something like Spring which can easily configure the property of instances. As we don't have this mechanism, so the client code very difficult to get the instance(for this case is Bucket) to configure the property. That's why provide a static default value, client code can't easily change the default value. The unit test ManagedTimeStateImplTest.testAsyncGetFromReaders() is one example

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: Different developer has different preference. It's not really impact anything like performance or complexity. 
My logic is:
if(special case)  // in this case is disabled bloom filter as treat bloom filter as default behavior
{
  do something
}

instead of 
if(not special case) 
{
  do something
}

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: But the bloomFilterDefaultBitSize is changeable.
And why put it into AbstractManagedStateImpl? this variable is totally depends on the implementation. Some implementation probably don't have conception at all

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: Why the value should be inserted into the bloom filter after fetched? Do you try to avoid update with same value? Could you explain more here?

What I usage is different here. What I try to avoid to is unnecessary get as there are much more getting than setting. 

The reason why decrease accurate is due to there are much more put() in all than put() to the file. and each put add flags to the bloom filter,

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: I guess this is due to misunderstanding the usage of bloom filter.
freeMemory() is only place which we know data going to write to files

library/src/test/java/org/apache/apex/malhar/lib/state/managed/ManagedTimeStateImplTest.java
brightchen: Agree.
I'll add unit test case

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: expectedFalsePositiveProbability() depends on expectedNumberOfFilterElements, which we in fact have no idea at all. At it is a fix value instead of dynamic value

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
bhupeshchawda: Agreed. Makes sense.
Can we move the usage of bloom filter to the `getFromReaders()` call? That is the first place where we know that files might be accessed.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
bhupeshchawda: Sorry, I meant using `getFalsePositiveProbability()`. We can keep track of the number of keys inserted into the bloom filter, right?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
bhupeshchawda: agreed

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
bhupeshchawda: I think the properties which allow to configure a bloom filter should not be per bucket. I am talking about variables like: `disableBloomFilter` and `bloomFilterBitSize`. I think these should be kept configurable. Would the user ever want to configure these differently for different buckets? How would a user of managed state be able to configure the bloom filter size?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
bhupeshchawda: Similar comment as below.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
bhupeshchawda: @brightchen Can you please also add the source of the code in javadoc?

library/src/test/java/org/apache/apex/malhar/lib/state/managed/DefaultBucketTest.java
```diff
@@ -227,4 +233,33 @@ public void testFreeMemory() throws IOException
 
     testMeta.defaultBucket.teardown();
   }
+
+  @Test
+  public void testBloomFilter() throws IOException
```
bhupeshchawda, line 238: This test would succeed whether or not bloom filter is enabled. Can we do some sort of validation whether the bloom filter in the bucket is consistent with the results?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -247,6 +253,9 @@ public DefaultBucket(long bucketId)
     public void setup(@NotNull ManagedStateContext managedStateContext)
     {
       this.managedStateContext = Preconditions.checkNotNull(managedStateContext, "managed state context");
+      if (!disableBloomFilter && bloomFilter == null) {
+        bloomFilter = new SliceBloomFilter(bloomFilterBitSize, 0.99);
+      }
```
chaithu14, line 258: I think "bloomFilter" has to move to time bucket because bloomFilter under DefaultBucket will work only if the managedState is of type ManagedTimeUnifiedStateImpl. Here, DefaultBucket behaves as time Bucket.
If the managedState is of type ManagedStateImpl/ManagedTimeStateImpl then DefaultBucket behaves as key bucket.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: In most of case, disableBloomFilter and bloomFilterBitSize should not be per bucket. It's kind depend the type of bucket. If this is the case, the static attributes can be used.
What I concern about put them into state manager is the it totally depends on Bucket implementation and not related to state manager at all. Let's think about we have another implementation, such as MemoryBucket, these attributes are non-sense to this implementation at all.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: I did very simple test, getFalsePositiveProbability() in fact is not very light. it cost 142ms for 1M operation.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: Yes, for ManagedStateImpl, bloom filter would be useless for a long run(But I still useful for certain period). ManagedTimeStateImpl should be useful for most of time. 
That's why I add the simple mechanism to check the efficiency of bloomfilter to dynamic unload it if it is useless.

library/src/test/java/org/apache/apex/malhar/lib/state/managed/DefaultBucketTest.java
brightchen: Yes, bloom filter should only helpful to increase the performance which should be tested in Benchmark application as it need run for a long time . I think what we should test in unit test is apply bloom filter will not break the current result.

brightchen:
@bhupeshchawda @chaithu14 I think I applied all comments or comment back. Please let me know if any leftover or new comments

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
```diff
@@ -353,6 +362,32 @@ public Slice get(Slice key, long timeBucket, ReadSource readSource)
...
+    private void verifyBloomFilter(boolean mightContain)
+    {
+      if (!mightContain) {
+        filteredCount++;
+      } else {
+        unfilteredCount++;
+      }
+
+      if (unfilteredCount > filteredCount * 4) {
+        unloadBloomFilter();
```
chaithu14: I think this looks like not a feasible way of disabling bloomFilter. Consider the case where the first search element(which would be on disk) found in state then the if condition becomes true and then the bloomFilter is disabled.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
```diff
@@ -0,0 +1,459 @@
...
+{
+  private BitSet bitset;
```
chaithu14, line 31: BitSet is not Kryo serializable. Could you please bind it under JavaSerializer?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
```diff
@@ -0,0 +1,459 @@
...
+      SetAttributes(Math.ceil(-(Math.log(falsePositiveProbability) / Math.log(2))) / Math.log(2), // c = k / ln(2)
+          expectedNumberOfElements, (int)Math.ceil(-(Math.log(falsePositiveProbability) / Math.log(2))));
+    }
+  }
+
+  /**
+   * Generate integer array based on the hash function till the number of
+   * hashes.
+   *
+   * @param data
```
chaithu14: Mismatching with the argument. Could you please correct it.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
```diff
@@ -0,0 +1,459 @@
...
+  public void clear()
+  {
+    bitset.clear();
+    numberOfAddedElements = 0;
+  }
+
+  /**
+   * Adds an array of bytes to the Bloom filter.
+   *
+   * @param bytes
```
chaithu14: Same here, update the comment.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
```diff
@@ -0,0 +1,459 @@
...
+    }
+    numberOfAddedElements++;
+  }
+
+  /**
+   * Returns true if the array of bytes could have been inserted into the Bloom
+   * filter. Use getFalsePositiveProbability() to calculate the probability of
+   * this being correct.
+   *
+   * @param bytes
```
chaithu14: Same here

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
bhupeshchawda: Then we should probably remove the setter methods for disableBloomFilter and bloomFilterBitSize, right?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
bhupeshchawda: I think that should be okay. We don't have to check this every time. We can do it once every n elements inserted, right?

library/src/main/java/org/apache/apex/malhar/lib/state/managed/Bucket.java
brightchen: agree. This logic is based on statistics, so the base number should be relative big. I'll add logic to check base number.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
brightchen: I believe I saw an implementation which was annotated by JavaSerializer and I then removed it. I can't recalled the reason now. But I check the BitSet class, it seems serializable by kryo. Could you please check the version of BitSet? I'll add a unit test case for building a simple application to test this. The unit test application should throw exception if any non-transient object is not serializable.

brightchen: I added unit test and no exception thrown. So it should be serializable

brightchen:
@bhupeshchawda @chaithu14 Could you please merge if no further comments

library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
chaithu14: :+1:

brightchen:
@PramodSSImmaneni @vrozov I talked with @bhupeshchawda He want one of you take look at this PR

library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
pramodin: I will add the reference

pramodin:
It looks like this is close to merge and needs a couple of small things. Since @brightchen is not looking at this actively, I will make those changes on top of a merge commit containing his commit and create a new PR for same JIRA. This way the final merge will also close this PR.

library/src/main/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilter.java
pramodin: I notice that the original hadoop bloom filter code includes a second license from European Commission on top of the regular apache license. We may have to do the same. Will discuss this on the dev list.

PR #522, opened 2016-12-09
SPOI-8533 Missing property tag in Iteration Demo's properties.xml

mattqzhang:


bhupeshchawda:
@mattqzhang can you please fix the commit message as well as the branch name. Please use APEXMALHAR jira id instead of SPOI.

PR #523, opened 2016-12-14
APEXMALHAR-2368 code changes to optimize memory usage in JdbcPoll ope…

KapoorHitesh:
…rator @DT-Priyanka @tushargosavi @bhupeshchawda please review

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
@@ -100,9 +100,14 @@ public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInpu
   private static int DEFAULT_FETCH_SIZE = 20000;
   private static int DEFAULT_BATCH_SIZE = 2000;
   private static int DEFAULT_SLEEP_TIME = 100;
+  private static int DEFAULT_RESULT_LIMIT = 100000;
   private int pollInterval = DEFAULT_POLL_INTERVAL; //in miliseconds
   private int queueCapacity = DEFAULT_QUEUE_CAPACITY;
   private int fetchSize = DEFAULT_FETCH_SIZE;
+  /**
+   * Parameter to limit the number of results to fetch in one query by the Poller partition.
+   */
+  private int resultLimit = DEFAULT_RESULT_LIMIT;
 
   @Min(1)
   private int partitionCount = 1;
@@ -127,12 +132,13 @@ public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInpu
   private transient ScheduledExecutorService scanService;
   private transient AtomicReference<Throwable> threadException;
   protected transient boolean isPolled;
-  protected transient Integer lastPolledRow;
   protected transient LinkedBlockingDeque<T> emitQueue;
   protected transient PreparedStatement ps;
   protected boolean isPollerPartition;
 
   protected transient MutablePair<Integer, Integer> currentWindowRecoveryState;
+
+  private transient int lastOffset;
 
   public AbstractJdbcPollInputOperator()
   {
@@ -175,8 +181,7 @@ public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInpu
     try {
       // If its a range query pass upper and lower bounds, If its a polling query pass only the lower bound
       if (isPollerPartition) {
-        ps = store.getConnection().prepareStatement(buildRangeQuery(rangeQueryPair.getKey(), Integer.MAX_VALUE),
-            TYPE_FORWARD_ONLY, CONCUR_READ_ONLY);
+        lastOffset = rangeQueryPair.getKey();
       } else {
         ps = store.getConnection().prepareStatement(
             buildRangeQuery(rangeQueryPair.getKey(), (rangeQueryPair.getValue() - rangeQueryPair.getKey())),
@@ -203,28 +208,9 @@ public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInpu
       }
     }
     if (isPollerPartition) {
-      updatePollQuery();
       isPolled = false;
     }
     lowerBound = lastEmittedRow;
-  }
-
-  private void updatePollQuery()
-  {
-    if ((lastPolledRow != lastEmittedRow)) {
-      if (lastEmittedRow == null) {
-        lastPolledRow = rangeQueryPair.getKey();
-      } else {
-        lastPolledRow = lastEmittedRow;
-      }
-      try {
-        ps = store.getConnection().prepareStatement(buildRangeQuery(lastPolledRow, Integer.MAX_VALUE),
-            TYPE_FORWARD_ONLY, CONCUR_READ_ONLY);
-      } catch (SQLException e) {
-        throw new RuntimeException(e);
-      }
-    }
-
   }
 
   @Override
@@ -269,20 +255,42 @@ public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInpu
     store.disconnect();
   }
 
+  /**
+   * Function to insert results of a query in emit Queue
+   * @param preparedStatement PreparedStatement to execute the query and store the results in emit Queue.
+   */
+  protected void insertDbDataInQueue(PreparedStatement preparedStatement) throws SQLException, InterruptedException
+  {
+    preparedStatement.setFetchSize(getFetchSize());
+    ResultSet result = preparedStatement.executeQuery();
+    if (result.next()) {
+      do {
+        while (!emitQueue.offer(getTuple(result))) {
+          Thread.sleep(DEFAULT_SLEEP_TIME);
+        }
+      } while (result.next());
+      result.close();
+    }
+    preparedStatement.close();
+  }
+
   protected void pollRecords()
   {
     if (isPolled) {
       return;
     }
     try {
-      ps.setFetchSize(getFetchSize());
-      ResultSet result = ps.executeQuery();
-      if (result.next()) {
-        do {
-          while (!emitQueue.offer(getTuple(result))) {
-            Thread.sleep(DEFAULT_SLEEP_TIME);
-          }
-        } while (result.next());
+      if (isPollerPartition) {
+        int nextOffset = getRecordsCount();
+        while (lastOffset < nextOffset) {
+          PreparedStatement preparedStatement = store.getConnection().prepareStatement(buildRangeQuery(lastOffset, resultLimit),
+              TYPE_FORWARD_ONLY, CONCUR_READ_ONLY);
+          insertDbDataInQueue(preparedStatement);
+          lastOffset = lastOffset + resultLimit;
+        }
+        lastOffset = nextOffset;
+      } else {
+        insertDbDataInQueue(ps);
       }
       isPolled = true;
     } catch (SQLException ex) {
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -127,13 +129,15 @@
...
   protected transient boolean isPolled;
-  protected transient Integer lastPolledRow;
   protected transient LinkedBlockingDeque<T> emitQueue;
   protected transient PreparedStatement ps;
   protected boolean isPollerPartition;
 
   protected transient MutablePair<Integer, Integer> currentWindowRecoveryState;
 
+  private int lastOffset;
+  private int nextOffset;
```
DT-Priyanka: should these variables be transient?

DT-Priyanka: can we name it as lower offset and higher offset for simplicity?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -269,20 +253,38 @@ public void deactivate()
     store.disconnect();
   }
 
+  protected void insertResultsInQueue(PreparedStatement preparedStatement) throws SQLException, InterruptedException
```
DT-Priyanka: The function name is misleading as you are also running the DB query in this function.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -269,20 +253,38 @@ public void deactivate()
...
-      if (result.next()) {
-        do {
-          while (!emitQueue.offer(getTuple(result))) {
-            Thread.sleep(DEFAULT_SLEEP_TIME);
-          }
-        } while (result.next());
+      if (isPollerPartition) {
+        nextOffset = getRecordsCount();
+        while (lastOffset < nextOffset) {
+          PreparedStatement preparedStatement = store.getConnection().prepareStatement(buildRangeQuery(lastOffset, rangeSize),
```
DT-Priyanka: shouldn't we use (lastoffset, (nextOffset > rangeSize) ? rangeSize : nextOffset)

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -269,20 +253,38 @@ public void deactivate()
...
+              TYPE_FORWARD_ONLY, CONCUR_READ_ONLY);
+          insertResultsInQueue(preparedStatement);
+          lastOffset = lastOffset + rangeSize;
```
DT-Priyanka: if nextOffset i.e. number of records in table are lesser than rangeSize, this logic won't work, we might end up using wrong lower offset for next poll call.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -717,6 +723,24 @@ public void setKey(String key)
     this.key = key;
   }
 
+  /**
+   * gets the Range Size
```
DT-Priyanka: Please explain what is range? I am not sure if it is a generic term for sql.

DT-Priyanka (changes requested):
Can we add a test which sets small value of rangeSize, and check if poll works properly?
Also please provide the reason these changes are required in Jira. Put your finding it it, so that we can justify the changes.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
KapoorHitesh: I think they should be non-transient to ensure exactly once.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
KapoorHitesh: Please suggest a better name. The only contribution of this function is to insert rows/results in emit queue.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
KapoorHitesh: This logic is working because I am setting lastOffset = nextOffset; after the while loop is finished.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
KapoorHitesh: I will add a JavaDoc for it. It is not  a generic term in sql.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
KapoorHitesh: Basically the loop should continue until all the records currently in the table have been emitted.

KapoorHitesh:
The unit test cases are already available to check if polling is working correctly.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: how does it help in exactly once processing?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: "insertDbDataInQueue"??

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
DT-Priyanka: okay, can we instead set lastOffset = lastOffset + (greater of (rangeSize, nextOffset) and remove "lastOffset = nextOffset", just for clarity of code?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
KapoorHitesh: I was thinking on the lines if we poller fails while adding the records then we may need to restore the value , but we are assuring the same via variable 'bound'. So will change it.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
KapoorHitesh: Done on your suggestion

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
KapoorHitesh: lastOffset = nextOffset signifies that we have processed rows till nextOffset.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
KapoorHitesh: Also changed the name.

KapoorHitesh:
@bhupeshchawda  @DT-Priyanka made the changes, please review. 
Please review.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -269,20 +255,42 @@ public void deactivate()
...
-      if (result.next()) {
-        do {
-          while (!emitQueue.offer(getTuple(result))) {
-            Thread.sleep(DEFAULT_SLEEP_TIME);
-          }
-        } while (result.next());
+      if (isPollerPartition) {
+        int nextOffset = getRecordsCount();
+        while (lastOffset < nextOffset) {
+          PreparedStatement preparedStatement = store.getConnection().prepareStatement(buildRangeQuery(lastOffset, resultLimit),
```
DT-Priyanka, line 286: do we need new PreparedStatement instance everytime? or we can reuse "ps" here?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -175,8 +181,7 @@ protected void initializePreparedStatement()
     try {
       // If its a range query pass upper and lower bounds, If its a polling query pass only the lower bound
       if (isPollerPartition) {
-        ps = store.getConnection().prepareStatement(buildRangeQuery(rangeQueryPair.getKey(), Integer.MAX_VALUE),
-            TYPE_FORWARD_ONLY, CONCUR_READ_ONLY);
+        lastOffset = rangeQueryPair.getKey();
       } else {
         ps = store.getConnection().prepareStatement(
```
DT-Priyanka, line 186: can we move this part to poll records as we did for pollar partition? if we do that we should change name of this function.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -717,6 +729,24 @@ public void setKey(String key)
     this.key = key;
   }
 
+  /**
+   * gets the Result Limit size, parameter to limit the number of results
+   * to fetch in one query by the Poller partition.
+   */
+  public int getResultLimit()
+  {
+    return resultLimit;
```
DT-Priyanka, line 738: can we call it as querySizeLimit ?

KapoorHitesh: querySizeLimit is not sounding correct to me because we cannot have a limit on the size of the query we can only limit the result set.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
KapoorHitesh: We need it every time. We are closing it after the operation.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
KapoorHitesh: I don't think it will help to fix this JIRA in any way. We can open a separate discussion on that.

DT-Priyanka: As you have moved  part to initialize prepared statement from "if" and not from "else" the code doesn't look clean. Anyway it was just a suggestion.

DT-Priyanka:
can you rebase it?

PR #530, opened 2017-01-02
APEXMALHAR-2379 Validation of attributes for AbstractFileOutputOperat…

francisf:
…or, regex on name instead of path

@bhupeshchawda please review

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
@@ -19,6 +19,7 @@
 package com.datatorrent.lib.io.fs;
 
 import java.io.BufferedReader;
+import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
@@ -37,6 +38,7 @@ import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
 
 import org.slf4j.Logger;
@@ -106,14 +108,17 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
   protected String directory;
   @NotNull
   protected DirectoryScanner scanner = new DirectoryScanner();
+  @Min(0)
   protected int scanIntervalMillis = 5000;
   protected long offset;
   protected String currentFile;
   protected Set<String> processedFiles = new HashSet<String>();
+  @Min(1)
   protected int emitBatchSize = 1000;
   protected int currentPartitions = 1;
   protected int partitionCount = 1;
   private int retryCount = 0;
+  @Min(0)
   private int maxRetryCount = 5;
   protected transient long skipCount = 0;
   private transient OperatorContext context;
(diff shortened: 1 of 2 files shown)


tweise:
Why is this not done using the javax.validation constraint annotations ?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -366,6 +367,9 @@ public int getScanIntervalMillis()
    */
   public void setScanIntervalMillis(int scanIntervalMillis)
   {
+    if (scanIntervalMillis < 0) {
+      throw new IllegalArgumentException("scanIntervalMillis should be greater than or equal to 0.");
+    }
```
chaithu14, line 376: Can we add the validation annotation to scanIntervalMillis, instead fixing it in setter method ?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -384,6 +388,9 @@ public int getEmitBatchSize()
    */
   public void setEmitBatchSize(int emitBatchSize)
   {
+    if (emitBatchSize <= 0) {
+      throw new IllegalArgumentException("emitBatchSize should be greater than 0.");
+    }
```
chaithu14, line 397: Can we add the validation annotation to emitBatchSize, instead fixing it in setter method ?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -994,6 +1001,9 @@ public int getMaxRetryCount()
    */
   public void setMaxRetryCount(int maxRetryCount)
   {
+    if (maxRetryCount < 0) {
+      throw new IllegalArgumentException("maxRetryCount should be greater than or equal to 0.");
+    }
```
chaithu14, line 1010: Same here.

library/src/test/java/com/datatorrent/lib/io/fs/AbstractFileInputOperatorTest.java
```diff
@@ -117,7 +117,7 @@ public void testSinglePartiton() throws Exception
     oper.output.setSink(sink);
 
     oper.setDirectory(testMeta.dir);
-    oper.getScanner().setFilePatternRegexp(".*file[\\d]");
+    oper.getScanner().setFilePatternRegexp("((?!target).)*file[\\d]");
```
chaithu14, line 133: Can you add some tests?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
francisf: @tweise @chaithu14
I took the change in https://github.com/apache/apex-malhar/blob/master/library/src/main/java/org/apache/apex/malhar/lib/fs/FSRecordReader.java#L158

I had tried the \@Min constraint annotation, which throws a constraint violation exception if the values are invalid in the properties.xml (bean validation). However, I do not see it throwing exceptions during the setting of the property of the operator when the application is running. Please let me know if I have misunderstood something here.

chaithu14: Are talking about dynamic property changes ? If yes, I think we need to fix it up in core.

francisf: @chaithu14 agreed. Will raise a ticket for the same on APEXCORE. Can the existing changes be merged? I've added the @ Min constraint annotation so that it can be picked up.

francisf: @chaithu14  Created [APECORE-600](https://issues.apache.org/jira/browse/APEXCORE-600)

chaithu14:
Looks good to me.

PR #533, opened 2017-01-16
APEXMALHAR-2384 adding documentation for fixed width parser

KapoorHitesh:
@amberarrow please review.


mkdocs.yml
```diff
@@ -19,4 +19,5 @@ pages:
     - Json Formatter: operators/jsonFormatter.md
     - Transform Operator: operators/transform.md
     - Xml Parser: operators/xmlParserOperator.md
+    - Fixed Width Parser: operators/fixedWidthParserOperator.md
```
trusli: We should sort this alphabetically rather than at the end of the list.

KapoorHitesh:
@amberarrow please review.

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,170 @@
+Fixed Width Parser Operator
+=============
+
+## Operator Objective
+This operator is designed to parse fixed width records and construct a map or concrete java class also known as ["POJO"](https://en.wikipedia.org/wiki/Plain_Old_Java_Object) out of it. User need to provide the schema to describe the fixed width data. Based on schema definition the operator will parse the incoming record to object map and POJO.
```
amberarrow: Based on schema .... => The incoming records will be parsed according to the schema and either a map or a POJO (or both) emitted.

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,170 @@
...
+Valid records will be emitted as POJO / map while invalid ones are emitted on error port with error message.
```
amberarrow: Revise: Invalid records will be emitted on the error port along with an error message.

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,170 @@
...
+
+**Note**: field names in schema must match field names of POJO and in the same order as it appears in the incoming data.
```
amberarrow: in schema => in the schema
of POJO => of the POJO
and in the same order as it appears => and must be in the same order as they appear

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,170 @@
...
+2. Available since:**_3.8.0_**
+3. Operator state:**_Evolving_**
+4. Java Package:[com.datatorrent.contrib.parser.FixedWidthParser](https://github.com/apache/apex-malhar/blob/master/contrib/src/main/java/com/datatorrent/contrib/parser/FixedWidthParser.java)
+
+
+## <a name="props"></a>Properties of FixedWidthParser
+User need to set the schema which describes fixed width data.
+e.g.
+
+```xml
```
amberarrow: This is not XML

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Operator Information
+1. Operator location:**_malhar-contrib_**
+2. Available since:**_3.8.0_**
+3. Operator state:**_Evolving_**
+4. Java Package:[com.datatorrent.contrib.parser.FixedWidthParser](https://github.com/apache/apex-malhar/blob/master/contrib/src/main/java/com/datatorrent/contrib/parser/FixedWidthParser.java)
+
+
+## <a name="props"></a>Properties of FixedWidthParser
+User need to set the schema which describes fixed width data.
```
amberarrow: Providing the actual XML fragment will be more useful than just the JSON, e.g.

The user needs to set the `jsonSchema` property to an appropriate JSON string that describes the fixed width data. An example of how to set it in the `properties.xml` file is shown below:
```xml
<property>
    <name>jsonSchema</name>
    <value> .... </value>
</property>
```

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,170 @@
...
+        },
+        {
+            "name": "valid",
+            "type": "String",
+            "length": "10",
+            "padding":" "
+        }
+    ]
+}
+
```
amberarrow: When setting this in code, should the value be a string or a JSONObject ? Show an example of doing that.

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,170 @@
...
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *jsonSchema* | [Schema](https://github.com/apache/apex-malhar/blob/master/contrib/src/main/java/com/datatorrent/contrib/parser/FixedWidthSchema.java)  describing fixed width data. Based on schema definition the operator will parse the incoming record to object map and POJO.
+Valid records will be emitted as POJO / map while invalid ones are emitted on error port with error message. | String | YES | N/A |
+
+## Platform Attributes that influences operator behavior
+
+| **Attribute** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *out.TUPLE_CLASS* | TUPLE_CLASS attribute on output port which tells operator the class of POJO which need to be emitted | Class| Yes |
+
```
amberarrow: Show an example of setting this in both XML and in code.

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,170 @@
...
+            "length": "1",
+            "trueValue": "y",
+            "falseValue": "n"
+        },
+        {
+            "name": "optimized",
+            "type": "Boolean",
+            "length": "1",
+            "trueValue": "y",
+            "falseValue": "n"
```
amberarrow: Do these have to be single characters ? What if the user wants "true" and "false" ? They are different lengths, so what should the user use for the length value ? Are they case sensitive ? If so how does the user request case insensitive matching ?

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,170 @@
...
+  * Float
+
+
+## Ports
+
+| **Port** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *in* | Tuples that needs to be parsed are recieved on this port | byte[] | Yes |
+| *out* | Valid Tuples that are emitted as pojo | Object (POJO) | No |
+| *parsedOutput* | Valid Tuples that are emitted as map | Map | No |
```
amberarrow: map => maps
So looks like emitting only maps and no POJOs is not an option ? If so, that should be clarified in the first paragraph at the top.

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,170 @@
...
+| *err* | Invalid Tuples are emitted with error message | KeyValPair <String, String\> | No |
+
+## Partitioning
+Fixed Width Parser is both statically and dynamically partitionable.
+### Static Partitioning
+This can be achieved in 2 ways
+
+1. Specifying the partitioner and number of partitions in the populateDAG() method
```
amberarrow: highlight literals like populateDAG() in some way; use the same convention consistently throughout

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Thanks Done.

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Thanks Done.

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Thanks Done.

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Thanks Done.

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Thanks Done.

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Thanks Done.

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Thanks Done.

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Modified the example to clarify this

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Thanks Done.

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: I am sorry I am unable to understand the confusion here. If the user will not connect the port corresponding to POJO, POJO will not be emitted similarly if the user will not connect the port for map, map will not be emitted.

KapoorHitesh:
@amberarrow have made the suggested changes. Please review and merge

docs/operators/fixedWidthParserOperator.md
amberarrow: Ok, makes sense (for some reason I thought 'out' had to be connected).

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,231 @@
...
+      "length": "1"
+      },
+      {
+      "name": "valid",
+      "type": "String",
+      "length": "10",
+      "padding":" "
+      }
+      ]
+      }
```
amberarrow: Duplicating the entire JSON string seems pointless; either remove the earlier copy or change the text here so something like this:
<value>  </value>

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,231 @@
...
+    </value>
+</property>
+  ```
```
amberarrow: The leading space here causes the rest of the document to not render properly.

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,231 @@
...
+     <name>dt.operator.{OperatorName}.attr.PARTITIONER</name>
+     <value>com.datatorrent.common.partitioner.StatelessPartitioner:2</value>
+   </property>
+  ```
+
+where {OperatorName} is the name of the FixedWidthParser operator.
+ Above lines will partition FixedWidthParser statically 2 times. Above value can be changed accordingly to change the number of static partitions.
+
+
+### Dynamic Paritioning
```
amberarrow: typo

docs/operators/fixedWidthParserOperator.md
amberarrow: I don't see the code example

docs/operators/fixedWidthParserOperator.md
amberarrow: I don't see this change

KapoorHitesh: The value just above this 
{
      "name": "active",
      "type": "Boolean",
      "length": "4",
      "trueValue": "true",
      "falseValue": "false"
      },

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Done

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Thanks Done

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Added the code example. Sorry missed it earlier.

KapoorHitesh:
@amberarrow have done the required changes please merge.

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,205 @@
...
+      "name": "bidPrice",
+      "type": "Double",
+      "length": "3"
+      },
+      {
+      "name": "startDate",
+      "type": "Date",
+      "format": "yyyy-MM-dd HH:mm:ss",
+      "length": "19"
+
```
amberarrow: Omit empty line

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,205 @@
...
+      "name": "securityCode",
+      "type": "Long",
+      "length": "5"
+      },
+      {
+      "name": "active",
+      "type": "Boolean",
+      "length": "4",
+      "trueValue": "true",
+      "falseValue": "false"
```
amberarrow: As mentioned in my earlier comment, please explain using English text either above or below this example, whether these values are case sensitive: Using this JSON, if the input value is "True" will it work ? Also, I'm puzzled by why the length value is 4 when the 2 strings are of different lengths -- this was also part of my original comment.

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,205 @@
...
+| **Attribute** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *out.TUPLE_CLASS* | TUPLE_CLASS attribute on output port which tells operator the class of POJO which need to be emitted | Class| Yes |
+
+User can set this property in properties.xml in the following way
+
+```xml
+<property>
+    <name>dt.application.{ApplicationName}.operator.{OperatorName}.port.out.attr.TUPLE_CLASS</name>
+    <value>fixedwidthparser.Ad</value>
```
amberarrow: Suggest changing this to something like: com.example.fixedwidthparser.Ad
to emphasize that it should be the fully qualified class name (note that this is what is done in the example).

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,205 @@
...
+</property>
+```
+Where {OperatorName} is the name of the Operator and {ApplicationName} is the name of the application and "fixedwidthparser.Ad" is the Tuple class.
```
amberarrow: Tuple class => fully qualified name of the Tuple class

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,205 @@
...
+
+Following code can be added to `populateDAG()` method of application to set Tuple Class:
+```java
+dag.setOutputPortAttribute({OperatorName}.out, Context.PortContext.TUPLE_CLASS, fixedwidthparser.Ad)
```
amberarrow: Please confirm that you've verified this code fragment in an actual working application.

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,205 @@
...
+      {
+      "name": "weatherTargeted",
+      "type": "Character",
+      "length": "1"
+      },
+      {
+      "name": "valid",
+      "type": "String",
+      "length": "10",
+      "padding":" "
```
amberarrow: Please explain what "padding" is -- is it expected before or after ? Does it have to be exactly one space or will multiple spaces work ? Will TAB characters work ? What if the user has a file that has TAB characters, what value should be specified here ? Please test these out to make sure of behavior before updating the document.

KapoorHitesh: I have verified that for TAB user will have to give '\t' as the padding. Padding is a single character

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Good catch. This is an incorrect example the length should be 5 in this case and then for true the actual data should be "true_" (assuming "_" as the padding character)

KapoorHitesh: I have verified the above comment

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: verified and modified it.

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Done

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Done

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Done

KapoorHitesh:
@amberarrow please review

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,205 @@
+Fixed Width Parser Operator
+=============
+
+## Operator Objective
+This operator is designed to parse fixed width records and construct a map or concrete java class also known as ["POJO"](https://en.wikipedia.org/wiki/Plain_Old_Java_Object) out of it. User need to provide the schema to describe the fixed width data. The incoming records will be parsed according to the schema and either a map or a POJO (or both) is emitted.
```
amberarrow: User need => User needs

docs/operators/fixedWidthParserOperator.md
amberarrow: User need => User needs

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,205 @@
...
+
+
+## <a name="props"></a>Properties of FixedWidthParser
+User need to set the schema which describes fixed width data.
+
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *jsonSchema* | [Schema](https://github.com/apache/apex-malhar/blob/master/contrib/src/main/java/com/datatorrent/contrib/parser/FixedWidthSchema.java)  describing fixed width data. Based on schema definition the operator will parse the incoming record to object map and POJO. Valid records will be emitted as POJO / map while invalid ones are emitted on error port with error message. | String | YES | N/A |
+
+User can set this property in properties.xml in the following way
```
amberarrow: properties.xml => `properties.xml`
way => way:

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,205 @@
...
+
+```xml
+<property>
+    <name>dt.application.{ApplicationName}.operator.{OperatorName}.prop.jsonSchema</name>
+    <value>{
```
amberarrow: Please fix indentation

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,205 @@
...
+      "type": "String",
+      "length": "10",
+      "padding":" "
+      }
+      ]
+      }
+    </value>
+</property>
+```
+Where {OperatorName} is the name of the Operator and {ApplicationName} is the name of the application. In this property padding is a character used to specify padding used in the incoming records. If there is a single padding for the entire record then there is no need to specify separate padding for separate fields.
```
amberarrow: This explanation is inadequate.

Looks like this operator uses the `FixedWidthSchema` class in a crucial way, so that class needs to be discussed in some detail here.

Put yourself in the position of a user who has a file of records with each line having a single record containing first name, last name, and id separated by one or more (variable number) blanks; how does the user determine if this file is parseable ? Since each line has variable length, is this considered "fixed length" or not ? How does the user construct a schema for this situation ?

Some keys in the example above seem to have specific meanings; those meanings should be explained with suitable examples. A lot more explanation and examples are needed here.

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Done

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Done

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Done.

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: I have added a small paragraph at the start and also added an example schema for a record.

KapoorHitesh:
@amberarrow have made modifications please review.

KapoorHitesh:
@amberarrow I have checked for the appearance of this file in chrome, could you please review?

amberarrow:
Yes, will do in the next few days.

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,240 @@
...
+## Operator Information
+1. Operator location:**_malhar-contrib_**
+2. Available since:**_3.8.0_**
+3. Operator state:**_Evolving_**
+4. Java Package:[com.datatorrent.contrib.parser.FixedWidthParser](https://github.com/apache/apex-malhar/blob/master/contrib/src/main/java/com/datatorrent/contrib/parser/FixedWidthParser.java)
+
+
+## <a name="props"></a>Properties of FixedWidthParser
+
+Data in a fixed-width text file is arranged in rows and columns, with one entry per row. Fixed width record is one row in a fixed-width file.  Each column has a fixed width, specified in characters, which determines the maximum amount of data it can contain.  No delimiters are used to separate the fields in the file.  Instead, smaller quantities of data are padded with predifined fixed padding character to fill the allotted space, such that the start of a given column can always be specified as an offset from the beginning of a line.
```
amberarrow: Instead .... => Instead, the data is left- or right- justified in its column as specified by the `alignment` value of the schema and the remaining space is filled by the padding character also specified in the schema.

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,240 @@
...
+          "type": "Character",
+          "length": "1"
+        }
+      ]
+    }
+    </value>
+</property>
+```
+Where {OperatorName} is the name of the Operator and {ApplicationName} is the name of the application.
+As explained earlier padding is a character used to specify padding used in the incoming records to fill the alloted maximum space if required. User has the flexibility to specify a single padding character for the entire file or the user can choose to provide separate padding character for separate fields (columns of the record). Similar to padding character user also has the flexibility to define the alignment of the incoming records, user can choose whether the record is left,centre or right aligned.
```
amberarrow: Looks like the padding value for the field overrides the global value; this should be stated explicitly. Also, can the user specify the alignment on a per-field basis ? Again, if this is the case, it should be stated.

Are "left" , "center" and "right" the only valid values for alignment ? Are those values case-sensitive ?
Finally, are both the American spelling "center" and the British spelling "centre" acceptable ?

Also suggest using "field width" instead of "alloted maximum space"

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,240 @@
...
+
+The sample json schema for a records having 3 fields as 'Occupation', 'Age' and 'Gender' having corresponding maximum space as 20,2 and 6, with corresponding padding character as '#','$' and '@' and with corresponding alignment as 'left', 'centre' and 'right' will be as follows:
```
amberarrow: a records => records
fields as => fields
having corresponding maximum space as => with field widths
with corresponding padding character as => padding characters
with corresponding alignment as => and alignments

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,240 @@
...
+```
+
+
+## Platform Attributes that influences operator behavior
+
+| **Attribute** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *TUPLE_CLASS* | TUPLE_CLASS attribute on output port which tells operator the class of POJO which need to be emitted | Class| Yes |
+
+User can set this property in properties.xml in the following way
```
amberarrow: properties.xml => `properties.xml`
way => way:

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,240 @@
...
+
+```xml
+<property>
+    <name>dt.application.{ApplicationName}.operator.{OperatorName}.port.out.attr.TUPLE_CLASS</name>
+    <value>com.datatorrent.tutorial.fixedwidthparser.Ad</value>
+</property>
+```
+Where {OperatorName} is the name of the Operator and {ApplicationName} is the name of the application and "com.datatorrent.tutorial.fixedwidthparser.Ad" is the fully qualified name of the Tuple class.
```
amberarrow: This sentence can be move above the XML example and modified slightly:

"In the examples below, {OperatorName} is the name of the Operator, {ApplicationName} is the name of the application and "com.datatorrent.tutorial.fixedwidthparser.Ad" is the fully qualified name of the Tuple class."

That way, the duplicated description below can be omitted.

docs/operators/fixedWidthParserOperator.md
```diff
@@ -0,0 +1,240 @@
...
+  * Boolean
+  * Date
+  * Float
+
+
+## Ports
+
+| **Port** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *in* | Tuples that needs to be parsed are recieved on this port | byte[] | Yes |
```
amberarrow: recieved => received
{A good way to remember this is: "i before e except after c when the sound is eee" though there are many exceptions}

amberarrow:
Rest looks good. Please rebase after addressing above comments to resolve conflicts.

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: There could only be left , right or centre alignment. The values are case insensitive hence not mentioning anything about the case. 
Have modified the text to take care of remaining comments

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Done

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: done

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Done

docs/operators/fixedWidthParserOperator.md
KapoorHitesh: Done

KapoorHitesh:
@amberarrow  have rebased and done the changes, please review.

PR #535, opened 2017-01-17
APEXMALHAR-2383 Documentation for Jdbc output operator

KapoorHitesh:
@amberarrow please review


docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,154 @@
...
+This operator internally extends the functionality of AbstractPassThruTransactionableStoreOutputOperator.This is the base implementation for a pass-through output adapter of a transactional store.This implementation provides the "exactly once" guarantee."Pass-through" means it does not wait for end window to write to the store. It will begin transaction at begin window and write to the store as the tuples come and commit the transaction at end window.
+
+**Note**: For enforcing exactly once semantics a table named dt_meta must exist in the database. The SQL to create the same is as follows
+"CREATE TABLE IF NOT EXISTS dt_meta (dt_app_id VARCHAR(100) NOT NULL, dt_operator_id INT NOT NULL, dt_window BIGINT NOT NULL, UNIQUE(dt_app_id,dt_operator_id,dt_window))"
```
tweise: Please change the table to apex_meta

KapoorHitesh: @tweise changing the table may cause a problem for the applications already using it as it is hard coded.
If you still want it to change let me know I will open a separate Jira and do the needful.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,154 @@
...
+     <name>dt.operator.{OperatorName}.attr.PARTITIONER</name>
+     <value>com.datatorrent.common.partitioner.StatelessPartitioner:2</value>
+   </property>
+  ```
+
+where {OperatorName} is the name of the JdbcPOJOInsertOutputOperator operator.
+Above lines will partition JdbcPOJOInsertOutputOperator statically 2 times. Above value can be changed accordingly to change the number of static partitions.
+
+
+#### Dynamic Paritioning
```
trusli: Typo on Dynamic Partitioning

KapoorHitesh: Thank you for the input. I have corrected it.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,154 @@
+JDBC Transactional POJO Output Operator
+==============
+
+## Operator Objective
+This operator receives stream of POJOs as input and using JDBC connection will insert the POJOs in a database table where each POJO corresponds to a row in the database table. The main features of this operator are persisting data to the database table and fault tolerance. This operator creates a transaction at the start of window, executes batches of sql updates, and closes the transaction at the end of the window. Each tuple corresponds to an SQL update statement. The operator groups the updates in a batch and submits them with one call to the database. Batch processing improves performance considerably. The size of a batch is configured by batchSize property. The tuples in a window are stored in check-pointed collection which is cleared in the endWindow(). This is needed for the recovery. The operator writes a tuple exactly once in the database, which is why only when all the updates are executed, the transaction is committed in the end window call. To execute/fire the insert queries on the database table, this operator uses PreparedStatement, a precompiled SQL statement.
```
amberarrow: receives stream => receives a stream

Both "insert" and "update" are mentioned in the text here and the semantics need to be clarified
as to what happens if a row for a tuple already exists: will it fail or use some sort of "upsert" query ?

Seems like too much implementation detail for a first paragraph; suggest splitting out the
implementation details into a separate section named "Implementation Notes" or something
similar. This introductory paragraph can just be something like:
"This operator receives an input stream of POJOs and inserts them as rows in a database table in
a fault-tolerant way."

Suggest using short lines instead of one very long line since it allows greater precision for review comments.

KapoorHitesh: @amberarrow  thank you for the inputs. I have modified the content please have a look at it and let me know your thoughts. I am not elaborating on "as to what happens if a row for a tuple already exists" because we are not handling it any special way (we are not making any use of upsert queries), if the underlying table allows duplicate rows it will be added else it will fail with an SQLException. We are claiming that operator will ignore replayed tuples in case of auto recovery.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,159 @@
+JDBC Transactional POJO Output Operator
+==============
+
+## Operator Objective
+This operator receives an input stream of POJOs and inserts them as rows in a database table in a fault-tolerant way.
+
+## Overview
+The main features of this operator are persisting data to the database table and fault tolerance. This operator creates a transaction at the start of window, executes batches of sql updates, and closes the transaction at the end of the window. Each tuple corresponds to an SQL update statement. The operator groups the updates in a batch and submits them with one call to the database. Batch processing improves performance considerably. The size of a batch is configured by batchSize property. The tuples in a window are stored in check-pointed collection which is cleared in the endWindow(). This is needed for the recovery. The operator writes a tuple exactly once in the database, which is why only when all the updates are executed, the transaction is committed in the end window call.
```
amberarrow: start of window => start of each window
configured by batchSize => configured by `batchSize`
in check-pointed => in a checkpointed
cleared in the endWindow() => cleared in the `endWindow()` callback
This is needed for the recovery. => omit
which is why  .... call. => omit {seems out of place in an overview paragraph}

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+This operator internally extends the functionality of AbstractPassThruTransactionableStoreOutputOperator.This is the base implementation for a pass-through output adapter of a transactional store.This implementation provides the "exactly once" guarantee."Pass-through" means it does not wait for end window to write to the store. It will begin transaction at begin window and write to the store as the tuples come and commit the transaction at end window.
```
amberarrow: AbstractPassThruTransactionableStoreOutputOperator.This => `AbstractPassThruTransactionableStoreOutputOperator`. This
store.This => store. This
guarantee."Pass-through" => guarantee. "Pass-through"
begin transaction at begin window => begin a transaction at `beginWindow`
transaction at end window => transaction at `endWindow`

Not clear if "This implementation" refers to the current class or the base class; suggest rewriting so that it is clear which class is meant by the various occurrences of "it".

Suggest splitting this paragraph into 2 or 3 smaller ones to aid readability.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+
+**Note**: For enforcing exactly once semantics a table named 'dt_meta' must exist in the database. The SQL to create the same is as follows
+"CREATE TABLE IF NOT EXISTS dt_meta (dt_app_id VARCHAR(100) NOT NULL, dt_operator_id INT NOT NULL, dt_window BIGINT NOT NULL, UNIQUE(dt_app_id,dt_operator_id,dt_window))". The operator will fail if this table is missing.
```
amberarrow: The operator will fail if this table is missing. => omit {since it is now a parenthetic note above}
Put the SQL in a markdown code block so it is highlighted
Do the column names have to match exactly what is in the above statement ?

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+
+**Note**: Additionally this operator assumes that the underlying database/table in which records are to be added supports transactions. If the database/table does not supports transactions then a tuple may be inserted in a table more than once in case of auto recovery from a failure (violation of exactly once semantics).
```
amberarrow: does not supports => does not support

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Operator Information
+1. Operator location: ***malhar-library***
+2. Available since: ***0.9.4***
+3. Java Packages:
+    * Operator: ***[com.datatorrent.lib.db.jdbc.AbstractJdbcTransactionableOutputOperator](https://www.datatorrent.com/docs/apidocs/com/datatorrent/lib/db/jdbc/AbstractJdbcTransactionableOutputOperator.html)***
```
amberarrow: Shouldn't this point to the derived class and not the base class ?

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+
+## How to Use?
+The tuple type in the abstract class is a generic parameter. Concrete subclasses need to choose an appropriate class (such as String or an appropriate concrete java class). Also implement a couple of abstract methods (if not using AbstractJdbcPOJOOutputOperator): `setStatementParameters(PreparedStatement statement, T tuple)` to set the parameter of the insert/update statement (which is a PreparedStatement) with values from the tuple and `getUpdateCommand()` the sql statement to update a tuple in the database. Note that these methods are also defined in AbstractJdbcPOJOOutputOperator.
```
amberarrow: in the abstract class -- which one ?
This whole paragraph is not clear because this document is about a concrete class so the discussion about abstract classes is confusing; if your are talking about a specific superclass, you need to make that clear and also clarify why you are discussing the base class.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+
+### <a name="AbstractJdbcPOJOOutputOperatorProps"></a>Properties of AbstractJdbcPOJOOutputOperator
+Several properties are available to configure the behavior of this operator and they are summarized in the table below.
+
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *tablename* | Name of the table where data is to be inserted | String | Yes | N/A |
+| *fieldInfos*| JdbcFieldInfo maps a store column to a POJO field name| List | Yes | N/A |
+Those properties can be set like this:
```
amberarrow: This sentence is being shown as part of the table above; please reformat so it renders correctly.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+    <value>
+      {
+      "sqlType": 0,
+      "columnName":"NAME",
+      "pojoFieldExpression": "name",
+      "type":"STRING"
+      }
+    </value>
+  </property>
+```
```
amberarrow: Above XML is not indented properly.

Also, is the POJO restricted to have only primitive type fields ? What if it has, say an "Address" field ? 
What is the mapping from the various Java primitive types to the values of the "type" field above ? Why is sqlType always 0 ? These things need to be explained.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+
+<property>
+  <name>dt.operator.{OperatorName}.prop.store.password</name>
+  <value>mypassword</value>
+</property>
+```
+### Abstract Methods
+`void setStatementParameters(PreparedStatement statement, T tuple)`:Sets the parameter of the insert/update statement with values from the tuple.
+`String getUpdateCommand()`:Gets the statement which insert/update the table in the database.
+
```
amberarrow: What is the point of the above 2 lines ?

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+## Operator Information
+1. Operator location: ***malhar-library***
+2. Available since: ***0.9.4***
+3. Java Packages:
+    * Operator: ***[com.datatorrent.lib.db.jdbc.AbstractJdbcTransactionableOutputOperator](https://www.datatorrent.com/docs/apidocs/com/datatorrent/lib/db/jdbc/AbstractJdbcTransactionableOutputOperator.html)***
+
+## How to Use?
+The tuple type in the abstract class is a generic parameter. Concrete subclasses need to choose an appropriate class (such as String or an appropriate concrete java class). Also implement a couple of abstract methods (if not using AbstractJdbcPOJOOutputOperator): `setStatementParameters(PreparedStatement statement, T tuple)` to set the parameter of the insert/update statement (which is a PreparedStatement) with values from the tuple and `getUpdateCommand()` the sql statement to update a tuple in the database. Note that these methods are also defined in AbstractJdbcPOJOOutputOperator.
+
```
amberarrow: Suggest reorganizing this section as follows: Add a sentence like this:
"As shown in the UML diagram below, this class has several superclasses, so its behavior can be configured via properties of those classes as described below."

Seems like the UML diagram actually belongs here.

Then, list the various tables as you have below and finally show the XML fragment with all of those properties in one place.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,159 @@
...
+    </value>
+  </property>
+```
+
+## Platform Attributes that influences operator behavior
+
+| **Attribute** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *input.TUPLE_CLASS* | TUPLE_CLASS attribute on input port which tells operator the class of POJO which is being received  | Class| Yes |
+
```
amberarrow: Include XML fragment for setting this.

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Thanks Done.

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Thanks Done.

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Done. Yes the column names need to match exactly.

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: By this sentence I hope you mean "Those properties can be set like this", I think it is formatted correctly.

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: These methods need to be defined in the base class.

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Done. Removed unwanted sentence

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Changed again. Adding Uml diagram after describing a concrete implementation

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Changed the indentation.  Changed the example sqlType.

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Over here I am describing 'AbstractJdbcTransactionableOutputOperator'

KapoorHitesh:
@amberarrow thank you for your patience in reviewing this. I have made the changes please let me know your thoughts on it.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+This operator internally extends the functionality of `AbstractPassThruTransactionableStoreOutputOperator`. AbstractPassThruTransactionableStoreOutputOperator is the base implementation for a pass-through output adapter of a transactional store. AbstractPassThruTransactionableStoreOutputOperator implementation provides the "exactly once" guarantee. "Pass-through" means it does not wait for end window to write to the store. It will begin transaction at `beginWindow` and write to the store as the tuples come and commit the transaction at `endWindow`.
```
amberarrow: begin transaction => begin the transaction

Start a new paragraph for "This operator ...."

Naming the base class multiple times, especially such a long name, seems needlessly verbose. How about replacing the first 2 sentences with:

An (indirect) base class for this operator is `AbstractPassThruTransactionableStoreOutputOperator` which implements a pass-through output adapter for a transactional store; it guarantees exactly-once semantics.

Actually, as I think about it some more, the part that is most confusing (and hence the part that most needs explanation) is what functional elements are present in each step of the hierarchy so that the user knows what to override and/or configure should the need arise. There is some hand-waving near the end of the document but that is not enough.

I think the reader will benefit substantially if we add a  summary of each class in the hierarchy in this overview section. It can be either a table or a bullet list but it should explain what each step of the inheritance hierarchy adds. For example:

BaseOperator: Contains empty implementations of methods from the Operator interface
AbstractTransactionableStoreOutputOperator: A skeleton implementation of an output operator that writes to a transactional store; the tuple type and store type are generic parameters. Defines an input port whose process method invokes the processTuple() abstract method. Exactly-once semantics are not guaranteed and must be provided by subclasses if needed.
AbstractPassThruTransactionableStoreOutputOperator: Simple extension of the above base class wich adds exactly-once semantics by starting a transaction in beginWindow() and committing it in endWindow().
AbstractJdbcTransactionableOutputOperator: (focus of this document) Adds support for JDBC by using an instance of JdbcTransactionalStore as the store. Also adds support for processing tuples in batches and provides an implementation of the proessTuple() abstract method mentioned above.
... etc.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+
+**Note**: For enforcing exactly once semantics a table named 'dt_meta' must exist in the database. The sample SQL to create the same is as follows
```
amberarrow: Use backquotes around dt_meta for consistency.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
+JDBC Transactional POJO Output Operator
+==============
+
+## Operator Objective
+This operator receives an input stream of POJOs and inserts them as rows in a database table in a fault-tolerant way.
+
+## Overview
+The main features of this operator are persisting data to the database table and fault tolerance. This operator creates a transaction at the start of each window, executes batches of sql updates, and closes the transaction at the end of the window. Each tuple corresponds to an SQL update statement. The operator groups the updates in a batch and submits them with one call to the database. Batch processing improves performance considerably. The size of a batch is configured by `batchSize` property. The tuples in a window are stored in a check-pointed collection which is cleared in the `endWindow()`. The operator writes a tuple exactly once in the database.
```
amberarrow: Mention the specific name of the class you are discussing, e.g.

The main features of this operator (`AbstractJdbcTransactionableOutputOperator`) are ....

Once you do this, you can remove the repetitive use of this long name, for example, the title for the subsection below can read: "Properties of this operator"

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Operator Information
+1. Operator location: ***malhar-library***
+2. Available since: ***0.9.4***
+3. Java Packages:
+    * Operator: ***[com.datatorrent.lib.db.jdbc.AbstractJdbcTransactionableOutputOperator](https://www.datatorrent.com/docs/apidocs/com/datatorrent/lib/db/jdbc/AbstractJdbcTransactionableOutputOperator.html)***
+
+## How to Use?
+ Concrete subclasses need to choose an appropriate class (such as String or an appropriate concrete java class). Also implement a couple of abstract methods (if not using AbstractJdbcPOJOOutputOperator): `setStatementParameters(PreparedStatement statement, T tuple)` to set the parameter of the insert/update statement (which is a PreparedStatement) with values from the tuple and `getUpdateCommand()` the sql statement to update a tuple in the database. Note that these methods are also defined in AbstractJdbcPOJOOutputOperator.
```
amberarrow: Clarify choose a class for what purpose, e.g.

... choose an appropriate class (....) for the generic parameters.

{Also clarify how/where the 2 generic parameters get resolved in the hierarchy}

sql => SQL

That last sentence needs an explanation: Why is the reader being asked to note this fact ? How is he expected to use this information ? Why are we asking the reader to implement these methods if a base class already provides them ?

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+<property>
+  <name>dt.operator.{OperatorName}.port.input.attr.TUPLE_CLASS</name>    
+    <value>com.example.mydtapp.PojoEvent</value>
+</property>
+```
+
+One of the concrete implementation is provided in Malhar as [JdbcPOJOInsertOutputOperator](https://github.com/apache/apex-malhar/blob/master/library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInsertOutputOperator.java).  The incoming tuples will be inserted in the table using PreparedStatement of the base class,  which is formed in activate() method of this operator.
+As shown in the UML diagram below, this class has several superclasses, so its behavior can be configured via properties of those classes as described below.
+![JdbcPOJOInsertOutputOperator.png](images/jdbcoutput/operatorsClassDiagram.png)
+
```
amberarrow: The diagram shows the hierarchy ending at `AbstractPassThruTransactionableStoreOutputOperator` but that class extends `AbstractTransactionableStoreOutputOperator` which again has a base class and so forth; please show the complete hierarchy.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+## How to Use?
+ Concrete subclasses need to choose an appropriate class (such as String or an appropriate concrete java class). Also implement a couple of abstract methods (if not using AbstractJdbcPOJOOutputOperator): `setStatementParameters(PreparedStatement statement, T tuple)` to set the parameter of the insert/update statement (which is a PreparedStatement) with values from the tuple and `getUpdateCommand()` the sql statement to update a tuple in the database. Note that these methods are also defined in AbstractJdbcPOJOOutputOperator.
+
+Several properties are available to configure the behavior of this operator and they are summarized in the table below.
+### <a name="AbstractJdbcTransactionableOutputOperatorProps"></a>Properties of AbstractJdbcTransactionableOutputOperator
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *batchSize* | Maximum number of tuples to insert in a single call (see explanation above). | int | No | 1000 |
+
+#### <a name="JdbcTransactionalStore"></a>Properties of JDBC Store (BackendStore)
```
amberarrow: BackendStore => The way it is written, it looks like a Java class name but I don't think there is such a class; either omit it or provide better explanatory text, e.g. "the store which is present in .... and performs the actual JDBC operations"

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+| **Property** | **Description** | **Type** | **Mandatory** | **Default Value** |
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *databaseDriver* |JDBC Driver class for connection to JDBC Store. This driver should be there in class path | String | Yes | N/A |
```
amberarrow: be there in class path => be present in the class path

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+| *databaseUrl* | Database url of the form jdbc:subprotocol:subname | String | Yes | N/A |
```
amberarrow: url => URL

What is "subname" ? The actual URL provided in the example below does not match this pattern. Please locate a defining spec on the net, use the pattern from it, and provide a link to it.

docs/operators/jdbcPOJOInsertOutputOperator.md
amberarrow: The point of my question is that the answer should be provided in the document, i.e. add text after the title saying something like:

These methods are defined as abstract in ....; implementations need to be provided by concrete base classes.

Please verify if this statement is actually correct; see my comment earlier.

docs/operators/jdbcPOJOInsertOutputOperator.md
amberarrow: I'm seeing this line as part of the table using mkdocs; what viewer are you using ?

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+    <value>
+      {
+        "sqlType": 4,
+        "columnName":"NAME",
+        "pojoFieldExpression": "name",
+        "type":"STRING"
+      }
+    </value>
+  </property>
+```
```
amberarrow: Please adjust indentation to be consistent throughout this document.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+
+
+## Platform Attributes that influences operator behavior
```
amberarrow: influences => influence

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+
+| **Attribute** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *input.TUPLE_CLASS* | TUPLE_CLASS attribute on input port which tells operator the class of POJO which is being received  | Class| Yes |
+Those properties can be set like this:
```
amberarrow: Same problem with this line: mkdocs shows it as a line of the table.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+
+```xml
+<property>
+  <name>dt.operator.{OperatorName}.port.input.attr.TUPLE_CLASS</name>    
+    <value>com.example.mydtapp.PojoEvent</value>
+</property>
+```
+
+One of the concrete implementation is provided in Malhar as [JdbcPOJOInsertOutputOperator](https://github.com/apache/apex-malhar/blob/master/library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInsertOutputOperator.java).  The incoming tuples will be inserted in the table using PreparedStatement of the base class,  which is formed in activate() method of this operator.
+As shown in the UML diagram below, this class has several superclasses, so its behavior can be configured via properties of those classes as described below.
```
amberarrow: Please provide a table of all the properties from superclasses that the user can set and descriptions for each; if you feel there are too many, at least choose the ones that the user is most likely to want to change and describe those.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+#### Static Partitioning
+Only static partitioning is supported for this operator.
+This can be achieved in 2 ways
+
+1. Specifying the partitioner and number of partitions in the populateDAG() method
+  ```java
+  JdbcPOJOInsertOutputOperator jdbcPOJOInsertOutputOperator = dag.addOperator("jdbcPOJOInsertOutputOperator", JdbcPOJOInsertOutputOperator.class);
+  StatelessPartitioner<JdbcPOJOInsertOutputOperator> partitioner1 = new StatelessPartitioner<JdbcPOJOInsertOutputOperator>(2);
+  dag.setAttribute(jdbcPOJOInsertOutputOperator, Context.OperatorContext.PARTITIONER, partitioner1);
+  ```
```
amberarrow: The above Java fragment is not formatted properly by mkdocs.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+
+2. Specifying the partitioner in properties file.
+  ```xml
+   <property>
+     <name>dt.operator.{OperatorName}.attr.PARTITIONER</name>
+     <value>com.datatorrent.common.partitioner.StatelessPartitioner:2</value>
+   </property>
+  ```
```
amberarrow: This XML fragment is not properly rendered by mkdocs

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+
+where {OperatorName} is the name of the JdbcPOJOInsertOutputOperator operator.
+Above lines will partition JdbcPOJOInsertOutputOperator statically 2 times. Above value can be changed accordingly to change the number of static partitions.
+
+
+#### Dynamic Partitioning
+Not supported.
+
+## Example
+Example for Jdbc output operator can be found at:
```
amberarrow: Rewrite:
An example application using this operator can be found at:

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,171 @@
...
+[https://github.com/DataTorrent/examples/tree/master/tutorials/fileToJdbc]. This example shows how to read files from HDFS, parse into POJOs and then insert into a table in MySQL.
```
amberarrow: Please format this link so it renders properly with mkdocs.

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: I am using atom and I don't see them as a part of table.

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: I have mentioned "if not using AbstractJdbcPOJOOutputOperator" because the user may choose to directly use the base class (for e.g. as done here https://www.datatorrent.com/blog/end-to-end-exactly-once-with-apache-apex/) 

Changed sql to SQL

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Appears correct to me :)

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: I have described those properties at each level, so I don't feel the need to describe the same properties again.

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: It appears good to me again :)

KapoorHitesh:
@amberarrow have made the requested changes

docs/operators/jdbcPOJOInsertOutputOperator.md
amberarrow: Please fix so they look right in mkdocs -- we use mkdocs to render.

docs/operators/jdbcPOJOInsertOutputOperator.md
amberarrow: They need to look right in mkdocs.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,178 @@
...
+dag.setAttribute(jdbcPOJOInsertOutputOperator, Context.OperatorContext.PARTITIONER, partitioner1);
+```
+
+2. Specifying the partitioner in properties file.
+```xml
+<property>
+  <name>dt.operator.{OperatorName}.attr.PARTITIONER</name>
+  <value>com.datatorrent.common.partitioner.StatelessPartitioner:2</value>
+</property>
+```
```
amberarrow: The above numbered list has both items numbered "1"; please fix so it looks right in mkdocs.

docs/operators/jdbcPOJOInsertOutputOperator.md
amberarrow: Needs to look right with mkdocs

docs/operators/jdbcPOJOInsertOutputOperator.md
amberarrow: Needs to look right with mkdocs

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Attaching the file I get after using mkdocs I don't see "Those properties can be set like this:" line as a part of table
[index.html.tar.gz](https://github.com/apache/apex-malhar/files/837561/index.html.tar.gz)

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Hi Ram I am unable to see the issue. Can you please elaborate?

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Looks correct to me attaching output from mkdocs 
[index.html.tar.gz](https://github.com/apache/apex-malhar/files/837570/index.html.tar.gz)

KapoorHitesh:
@amberarrow  I see the expected output from mkdocs as well. Please correct me if I am missing something. Attaching the output.
[index.html.tar.gz](https://github.com/apache/apex-malhar/files/837574/index.html.tar.gz)

amberarrow:
http://imgur.com/a/giJp9

Please take a look at the two images at the above URL. The first shows what the table looks like in Chrome visiting localhost:8000 after running the `mkdocs` server locally. It clearly shows the text as the last row of the table.

The second shows what the `index.html` you uploaded looks like when viewed in Chrome; again, it is clear that the text is part of the table since the first column has been sized to hold the entire sentence "Those properties can be set like this:".

Looking into the actual content of your `index.html`, we see near line 468:

```
<tr>
<td>Those properties can be set like this:</td>
<td></td>
<td></td>
<td></td>
<td></td>
</tr>
```

which again shows that normal text that should follow the table, is being included as the last row of the table.

KapoorHitesh:
@amberarrow  have added an additional line in between the end of table and the problematic sentence, this will probably solve this, I still can't see this issue at my end. Please review

amberarrow:
I don't understand how this is possible; Does the HTML I showed above is from _your_ copy of index.html so you can surely see the issue in this file. When you view this file in Chrome does it look different from the screenshot I posted ? If so can you post the screenshot ? Finally can you check what version of mkdocs you are running ? Can you run "mkdocs serve" locally and post a screenshot of what it looks like in Chrome ?

KapoorHitesh:
@amberarrow Yes in chrome I can see this issue.  I didn't checked that earlier in chrome (which I should have). However after adding an extra line it seems that this issue is fixed (I have verified that in chrome mozilla and also checked out the html file). Apologies for neglecting chrome (and html) earlier.

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Changed the text in it, I am unable to figure out why this is happening hence avoiding the numbered list.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,179 @@
+JDBC Transactional POJO Output Operator
+==============
+
+## Operator Objective
+This operator receives an input stream of POJOs and inserts them as rows in a database table in a fault-tolerant way.
+
+## Overview
+The main features of this operator (`AbstractJdbcTransactionableOutputOperator`)are persisting data to the database table and fault tolerance. This operator creates a transaction at the start of each window, executes batches of sql updates, and closes the transaction at the end of the window. Each tuple corresponds to an SQL update statement. The operator groups the updates in a batch and submits them with one call to the database. Batch processing improves performance considerably. The size of a batch is configured by `batchSize` property. The tuples in a window are stored in a check-pointed collection which is cleared in the `endWindow()`. The operator writes a tuple exactly once in the database.
```
amberarrow: )are => ) are
sql updates => SQL updates
in the `endWindow()` => in each `endWindow()` call.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,179 @@
...
+
+An (indirect) base class for this operator is `AbstractPassThruTransactionableStoreOutputOperator` which implements a pass-through output adapter for a transactional store; it guarantees exactly-once semantics. "Pass-through" means it does not wait for end window to write to the store. It will begin transaction at `beginWindow` and write to the store as the tuples come and commit the transaction at `endWindow`.
+
+The overall heirarchy is described in the the following diagram:
+
+![JdbcPOJOInsertOutputOperator.png](images/jdbcoutput/operatorsClassDiagrams.png)
+
+AbstractTransactionableStoreOutputOperator: A skeleton implementation of an output operator that writes to a transactional store; the tuple type and store type are generic parameters. Defines an input port whose process method invokes the processTuple() abstract method. Exactly-once semantics are not guaranteed and must be provided by subclasses if needed.
```
amberarrow: Please highlight the operator name in this list using backquotes or bold text in each item in this list.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,179 @@
...
+
+AbstractPassThruTransactionableStoreOutputOperator: Simple extension of the above base class wich adds exactly-once semantics by starting a transaction in beginWindow() and committing it in endWindow().
```
amberarrow: wich => which
beginWindow() => `beginWindow()`
endWindow() => `endWindow()`

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,179 @@
...
+
+AbstractJdbcTransactionableOutputOperator: (focus of this document) Adds support for JDBC by using an instance of JdbcTransactionalStore as the store. Also adds support for processing tuples in batches and provides an implementation of the proessTuple() abstract method mentioned above.
```
amberarrow: proessTuple() => `processTuple()`

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Operator Information
+1. Operator location: ***malhar-library***
+2. Available since: ***0.9.4***
+3. Java Packages:
+    * Operator: ***[com.datatorrent.lib.db.jdbc.AbstractJdbcTransactionableOutputOperator](https://www.datatorrent.com/docs/apidocs/com/datatorrent/lib/db/jdbc/AbstractJdbcTransactionableOutputOperator.html)***
+
+## How to Use?
+ Concrete subclasses need to choose an appropriate class (such as String or an appropriate concrete java class). Also implement a couple of abstract methods (if not using AbstractJdbcPOJOOutputOperator): `setStatementParameters(PreparedStatement statement, T tuple)` to set the parameter of the insert/update statement (which is a PreparedStatement) with values from the tuple and `getUpdateCommand()` the SQL statement to update a tuple in the database. Note that these methods are also defined in AbstractJdbcPOJOOutputOperator.
```
amberarrow: As I mentioned in an earlier comment, the point of the first sentence is not clear: Choose a class for what purpose ? Please explain.

the SQL statement => to return the SQL statement

Note that .... . => Note that subclasses of AbstractJdbcPOJOOutputOperator need not define these methods since they are already defined in that class.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,179 @@
...
+  <value>mypassword</value>
+</property>
+```
+### Abstract Methods
+These methods are defined as abstract in AbstractJdbcTransactionableOutputOperator `void setStatementParameters(PreparedStatement statement, T tuple)`:Sets the parameter of the insert/update statement with values from the tuple.
+`String getUpdateCommand()`:Gets the statement which insert/update the table in the database.
+
+
+## AbstractJdbcPOJOOutputOperator
+This is the abstract implementation extending the functionality of AbstractJdbcTransactionableOutputOperator that serves as base class for inserting rows in a table using a JDBC store. It has the definition for the abstract methods in AbstractJdbcTransactionableOutputOperator. It can be further extended to modify functionality or add new capabilities. This class has an input port, to recieve the records in the form of tuples, so concrete sub classes won't need to provide the same, and processes/inserts each input tuple as a database table record one by one . You need to set the input port attribute TUPLE_CLASS to define your [POJO](https://en.wikipedia.org/wiki/Plain_Old_Java_Object) class name to define Object type.
```
amberarrow: port to, recieve => port to receive
sub classes => subclasses
record one by one . => record one by one. {suggest omitting "one by one" since does not add anything}

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,179 @@
...
+| -------- | ----------- | ---- | ------------------ | ------------- |
+| *tablename* | Name of the table where data is to be inserted | String | Yes | N/A |
+| *fieldInfos*| JdbcFieldInfo maps a store column to a POJO field name| List | Yes | N/A |
+
+Those properties can be set like this:
+
+```xml
+<property>
+  <name>dt.operator.{OperatorName}.prop.tablename</name>
+  <value>tablename</value>
```
amberarrow: Please use something that resembles an actual table name such as "ResultTable"; otherwise, users may be confused by the fact that this value is identical to the last segment  of the `<name>` element above.

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+
+## Platform Attributes that influence operator behavior
+
+| **Attribute** | **Description** | **Type** | **Mandatory** |
+| -------- | ----------- | ---- | ------------------ |
+| *input.TUPLE_CLASS* | TUPLE_CLASS attribute on input port which tells operator the class of POJO which is being received  | Class| Yes |
+
+Those properties can be set like this:
```
amberarrow: properties => attributes

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,179 @@
...
+
+```xml
+<property>
+  <name>dt.operator.{OperatorName}.port.input.attr.TUPLE_CLASS</name>    
+  <value>com.example.mydtapp.PojoEvent</value>
+</property>
+```
+
+One of the concrete implementation is provided in Malhar as [JdbcPOJOInsertOutputOperator](https://github.com/apache/apex-malhar/blob/master/library/src/main/java/com/datatorrent/lib/db/jdbc/JdbcPOJOInsertOutputOperator.java).  The incoming tuples will be inserted in the table using PreparedStatement of the base class,  which is formed in activate() method of this operator.
```
amberarrow: One of the concrete => A concrete
activate() => `activate()`

docs/operators/jdbcPOJOInsertOutputOperator.md
amberarrow: Why "input.TUPLE_CLASS" rather than just "TUPLE_CLASS" ?

docs/operators/jdbcPOJOInsertOutputOperator.md
```diff
@@ -0,0 +1,179 @@
...
+
+where {OperatorName} is the name of the JdbcPOJOInsertOutputOperator operator.
+Above lines will partition JdbcPOJOInsertOutputOperator statically 2 times. Above value can be changed accordingly to change the number of static partitions.
+
+
+#### Dynamic Partitioning
+Not supported.
+
+## Example
+An example application using this operator can be found at:
```
amberarrow: Please check the appearance of this link in Chrome using mkdocs; this item is still pending from an earlier review comment.

amberarrow:
If the main focus of this document is the class `AbstractJdbcTransactionableOutputOperator` why is the file itself named `jdbcPOJOInsertOutputOperator.md` ?

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: done

docs/operators/jdbcPOJOInsertOutputOperator.md
KapoorHitesh: Modified it.

KapoorHitesh:
@amberarrow have made the changes please review

docs/operators/jdbcPOJOInsertOutputOperator.md
amberarrow: The spelling error for "receive" noted above was still present but I went ahead and merged it.

PR #538, opened 2017-01-18
APEXMALHAR-2390 and APEXMALHAR-2391 List of Operator documentation

trusli:
Sort operator list in alphabetical order, correct a few typos and add missing JDBC Poll Input doc


amberarrow:
Typo is first commit message: ddd => add
If you squash the two commits, you can use a new commit message.

While you're at it, might as well replace "Transform Operator" with "Transformer" to be consistent
with other entries like "Enricher" and "Deduper" (the "Operator" is redudent since the whole list is
under the "Operators" title). "Windowed Operator" is the sole exception since it seems a bit special
and I can't think of a better name.

PR #539, opened 2017-01-19
APEXMALHAR-2221 Fixed check style violations in Benchmark

oliverwnk:
@tweise @sandeshh please review and merge as appropriate

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -143,14 +143,6 @@
            <skip>true</skip>
          </configuration>
        </plugin>
-       <plugin>
-         <groupId>org.apache.maven.plugins</groupId>
-         <artifactId>maven-checkstyle-plugin</artifactId>
-         <configuration>
-           <maxAllowedViolations>281</maxAllowedViolations>
-           <logViolationsToConsole>${checkstyle.console}</logViolationsToConsole>
-         </configuration>
-       </plugin>
     </plugins>
   </build>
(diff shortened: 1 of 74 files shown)


benchmark/pom.xml
```diff
@@ -147,7 +147,7 @@
          <groupId>org.apache.maven.plugins</groupId>
```
tweise: There shouldn't be a need to configure the plugin here after these changes. It is inherited from the parent pom. Please check library module.

tweise:
Since these are all formatting changes, please amend the commit like so:
``` 
git commit --amend --author "Apex Dev <dev@apex.apache.org>"
```
See http://apex.apache.org/contributing.html#opening-pull-requests

benchmark/src/test/java/com/datatorrent/benchmark/util/serde/GenericSerdePerformanceTest.java
```diff
@@ -7,7 +7,7 @@
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
  *
- *   http://www.apache.org/licenses/LICENSE-2.0
```
tweise, line 10: This change does not look right, see license header in other files (and header.txt).

tweise:
Thanks! Next time please include the JIRA number in the commit message.

PR #544, opened 2017-01-31
APEXMALHAR-2397 #resolve Removing DAG.GATEWAY_CONNECT_ADDRESS which i…

sashadt:
…s causing evaluation failures during apex get-app-package-info call

### DIFF
--- a/examples/pi/src/main/java/org/apache/apex/examples/pi/ApplicationAppData.java
+++ b/examples/pi/src/main/java/org/apache/apex/examples/pi/ApplicationAppData.java
@@ -18,9 +18,6 @@
  */
 package org.apache.apex.examples.pi;
 
-import java.net.URI;
-
-import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 
 import com.datatorrent.api.DAG;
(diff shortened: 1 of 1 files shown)


tweise:
PR would need to be rebased. 

How is the application supposed to work without the URI?

sanjaypujare:
@sashadt  @tweise is right. You should only remove the check for dag.getValue(DAG.GATEWAY_CONNECT_ADDRESS) being null and throwing the RuntimeException since that causes problem in constructing the DAG in get-app-package-info command. The rest of the code is required for the app to work properly since the operators need the URI to be set. If the URI is not correct i.e. connectable then those operators will throw an exception in setup/activate and that ought to be enough for error-checking.

sashadt:
@tweise There is no need to set the `uri` variable on `wsQuery` or `wsResult` because their internal implementation uses `uriHelper`, which sets this to a default value, which is correct.  Setting the `uri` is optional, and seems to be an override, which is not needed.

@sanjaypujare I'm not following your comment about "only remove the check for `dag.getValue(DAG.GATEWAY_CONNECT_ADDRESS)` ", because that is the key to getting the `uri` in the frist place.  If we don't make that call, do you have an alternative strategy for finding the correct `uri`?

In summary, I don't see a problem with this PR, and I don't see why `uri` needs to be set, if I don't want to provide an override, and am happy with the defaults.

sanjaypujare:
@sashadt  my comment was based on @tweise 's comment. I did look at uriHelper based on your clarification and it does look like your changes should work so I have no issues.

sanjaypujare:
@sashadt  you'll still need to rebase though...

tweise:
@sashadt makes sense. Please rebase the PR.

sashadt:
@tweise, @sanjaypujare Rebased.  Please merge.

PR #545, opened 2017-01-31
APEXMALHAR-2376 Add Common Log support in LogParser operator

akshay-harale:
https://issues.apache.org/jira/browse/APEXMALHAR-2376

### DIFF
--- a/contrib/src/main/java/org/apache/apex/malhar/contrib/parser/CommonLogParser.java
+++ b/contrib/src/main/java/org/apache/apex/malhar/contrib/parser/CommonLogParser.java
@@ -1,0 +1,84 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.contrib.parser;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.apex.malhar.contrib.parser.log.CommonLog;
+import org.apache.hadoop.classification.InterfaceStability;
+
+import com.datatorrent.api.Context;
+
+/**
+ * Operator that parses a log string tuple against the
+ * a specified json schema and emits POJO on a parsed port and tuples that could not be
+ * parsed on error port.<br>
+ * <b>Properties</b><br>
+ * <b>jsonSchema</b>:schema as a string<br>
+ * <b>pojoClass</b>:Pojo class in case of user specified schema<br>
+ * <b>Ports</b> <br>
+ * <b>in</b>:input tuple as a String. Each tuple represents a log<br>
+ * <b>parsedOutput</b>:tuples that are validated against the specified schema are emitted
+ * as POJO on this port<br>
+ * <b>err</b>:tuples that do not confine to log format are emitted on this port as
+ * KeyValPair<String,String><br>
+ * Key being the tuple and Val being the reason.
+ */
+@InterfaceStability.Evolving
+public class CommonLogParser extends LogParser
+{
+  private static final Logger logger = LoggerFactory.getLogger(CommonLogParser.class);
+
+  private String schema="{\n" +
+    "  \"fields\": [{\n" +
+    "    \"field\": \"host\",\n" +
+    "    \"regex\": \"^([0-9.]+)\"\n" +
+    "  }, {\n" +
+    "    \"field\": \"rfc931\",\n" +
+    "    \"regex\": \"(\\\\S+)\"\n" +
+    "  }, {\n" +
+    "    \"field\": \"username\",\n" +
+    "    \"regex\": \"(\\\\S+)\"\n" +
+    "  }, {\n" +
+    "    \"field\": \"datetime\",\n" +
+    "    \"regex\": \"\\\\[(.*?)\\\\]\"\n" +
+    "  },{\n" +
+    "    \"field\": \"request\",\n" +
+    "    \"regex\": \"\\\"((?:[^\\\"]|\\\")+)\\\"\"\n" +
+    "  },{\n" +
+    "    \"field\": \"statusCode\",\n" +
+    "    \"regex\": \"(\\\\d{3})\"\n" +
+    "  },{\n" +
+    "    \"field\": \"bytes\",\n" +
+    "    \"regex\": \"(\\\\d+|-)\"\n" +
+    "  }]\n" +
+    "}";
+
+
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    this.setLogFileFormat(schema);
+    super.setup(context);
+    super.setClazz(CommonLog.class);
+  }
+
+
+}
--- a/contrib/src/main/java/org/apache/apex/malhar/contrib/parser/log/CommonLog.java
+++ b/contrib/src/main/java/org/apache/apex/malhar/contrib/parser/log/CommonLog.java
@@ -1,0 +1,120 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.contrib.parser.log;
+
+import org.apache.apex.malhar.contrib.parser.LogParser;
+import org.apache.hadoop.classification.InterfaceStability;
+
+/**
+ * This is default log format parser for <a href="https://en.wikipedia.org/wiki/Common_Log_Format">Common log </a>
+ * To use this format with {@link LogParser} operator just mention the property "logFileFormat" as "COMMON"
+ */
+@InterfaceStability.Evolving
+public class CommonLog
+{
+  private String host;
+  private String rfc931;
+  private String username;
+  private String datetime;
+  private String request;
+  private String statusCode;
+  private String bytes;
+
+  @Override
+  public String toString()
+  {
+    return "CommonLog [ Host : " + this.getHost() +
+      ", rfc931 : " + this.getRfc931() +
+      ", userName : " + this.getUsername() +
+      ", dateTime : " + this.getDatetime() +
+      ", request : " + this.getRequest() +
+      ", statusCode : " + this.getStatusCode() +
+      ", bytes : " + this.getBytes() + " ]";
+  }
+
+  public String getHost()
+  {
+    return host;
+  }
+
+  public void setHost(String host)
+  {
+    this.host = host;
+  }
+
+  public String getRfc931()
+  {
+    return rfc931;
+  }
+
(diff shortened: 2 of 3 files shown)


akshay-harale:
Closing and reopening this due to jenkins failure.

akshay-harale:
reopening this due to jenkins failure.

chinmaykolhatkar:
@akshay-harale A general design comment.

There is already a logFileFormat parameter, can you not use that to do COMMON log format parsing?

akshay-harale:
@chinmaykolhatkar we are using the same parameter to get the type. If user give COMMON log file format then he dont need to provide regex for the same. So we are having default regex for the COMMON log file format.

chinmaykolhatkar:
@akshay-harale Ok.. I got your point. But I don't think this is the right way.

Create a new operator, lets say "CommonLogParser" or something, as extending from LogParser and force-override the logFileFormat variable to common log format in its empty constructor and all constructors. If you want to harden it even further, override the setter method of logFileFormat and make it private. This is much cleaner and object-oriented approach.

This way, one who want to deal with generic format can work with LogParser and one who needs to work with common log format can use CommonLogParser.

akshay-harale:
@chinmaykolhatkar thanks for reply. Your point makes more sense. Instead of confusing user lets create separate operator for Common log format.

chinmaykolhatkar:
Hi @akshay-harale , Any update on this?

akshay-harale:
@chinmaykolhatkar yes. I have updated this PR with one more operator as CommonLogParser.

chinmaykolhatkar:
Hi @akshay-harale , I think this is a good start for the change. But I think we can simplify the code even further.

Lets see from top down. As you said before, when you would set logFileFormat property of LogParser, rest all works fine. In that case here is what I would expect in CommonLogParser:
1. Extend CommonLogParser from LogParser. You've already done that.
2. Override the setup method and in that set the logFileFormat using setLogFileFormat method to a static value which reflects common log format. You can have a public static final String variable in this just so that the values are not hardcoded and are picked from variable.
3. After setLogFileFormat is called, call super.setup().

That's one I think you would need to do, rest all will be taken care by super class.

Please me know what you think about this.

akshay-harale:
Hello @chinmaykolhatkar, actually in LogParser operator "logFileFormat" is to get the schema details in JSON format. Inside the LogParser there is on more method setupLog() which parse this schema from "logFileFormat" and it is called from LogParser.setup(). 
Now in my case I need to modify setupLog() api and hard code common log format field. Instead of doing all this I just creating object of CommonLog format in CommonLogParser.setup() and emitting object from processTuple().

chinmaykolhatkar:
@akshay-harale , I get your point. But this is still quite complex way. Moreover you've not really extended LogParser operator.

From what I see you can do that you need to do by just extending the LogParser.. Let LogParser to do its work and just set logFileFormat which represents the common log format.

If above is correct, all you need to create a single class (no other class is required), which extends from LogParser and set the value of logFileFormat in setup of this new operator and then call setup of super to do that rest of the work.

akshay-harale:
Hello @chinmaykolhatkar 
I have updated the PR with changes.

chinmaykolhatkar:
Hi @akshay-harale, Thanks for the update. I personally don't think there is any need to modify LogParser.java.
Do you possibly want to have an offline discussion about this? Ping me when you have time, we can have a quick call and run through the changes. As the outcome of ur discussion we'll update this PR.

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/CommonLogParser.java
```diff
@@ -0,0 +1,85 @@
...
+ * <b>pojoClass</b>:Pojo class in case of user specified schema<br>
+ * <b>Ports</b> <br>
+ * <b>in</b>:input tuple as a String. Each tuple represents a log<br>
+ * <b>parsedOutput</b>:tuples that are validated against the specified schema are emitted
+ * as POJO on this port<br>
+ * <b>err</b>:tuples that do not confine to log format are emitted on this port as
+ * KeyValPair<String,String><br>
+ * Key being the tuple and Val being the reason.
+ */
+public class CommonLogParser extends LogParser
```
chinmaykolhatkar: Please add @Evolving annotation to this class as this is a new one.
The annotation is : @InterfaceStability.Evolving

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/CommonLogParser.java
```diff
@@ -0,0 +1,85 @@
...
+{
+  private static final Logger logger = LoggerFactory.getLogger(CommonLogParser.class);
+
+  private Class pojoClass =CommonLog.class;
```
chinmaykolhatkar: You don't need this variable. See comment at line 81.

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/CommonLogParser.java
```diff
@@ -0,0 +1,85 @@
...
+    "  }]\n" +
+    "}";
+
+
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    this.setLogFileFormat(schema);
+    super.setup(context);
+    super.setClazz(pojoClass);
```
chinmaykolhatkar: You could simply do super.setClazz(CommonLog.class)... You don't need to have a vaeiable pojoClass.

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/CommonLogParser.java
```diff
@@ -0,0 +1,85 @@
...
+ * KeyValPair<String,String><br>
+ * Key being the tuple and Val being the reason.
+ */
+public class CommonLogParser extends LogParser
+{
+  private static final Logger logger = LoggerFactory.getLogger(CommonLogParser.class);
+
+  private Class pojoClass =CommonLog.class;
+
+  private String schema="{\n" +
```
chinmaykolhatkar: Do you want to make this variable "public static final String"

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/LogParser.java
```diff
@@ -19,14 +19,19 @@
 package org.apache.apex.malhar.contrib.parser;
```
chinmaykolhatkar: There is no functional change in this file. Its just that your IDE has adjusted the whitespaces... Please revert this file to master.

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/LogSchemaDetails.java
```diff
@@ -23,6 +23,7 @@
 import org.codehaus.jettison.json.JSONObject;
```
chinmaykolhatkar: There is no functional change in this file. It seems that your IDE has adjusted the whitespaces... Please revert this file to master.

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/log/CommonLog.java
```diff
@@ -0,0 +1,125 @@
...
+
+import org.joda.time.DateTime;
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
+
+/**
+ * This is default log format parser for <a href="https://en.wikipedia.org/wiki/Common_Log_Format">Common log </a>
+ * To use this format with {@link LogParser} operator just mention the property "logFileFormat" as "COMMON"
+ */
+public class CommonLog
```
chinmaykolhatkar: Can you make this @Evolving same as CommonLogParser?

contrib/src/test/java/org/apache/apex/malhar/contrib/parser/LogParserTest.java
```diff
@@ -18,16 +18,16 @@
  */
```
chinmaykolhatkar: There is no functional change in this file. It seems that your IDE has adjusted the whitespaces... Please revert this file to master.

contrib/src/test/resources/commonLogSchema.json
```diff
@@ -0,0 +1,24 @@
+{
```
chinmaykolhatkar: Where is this file used? If not used, please remove it.

library/pom.xml
```diff
@@ -308,6 +308,16 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+          <groupId>com.fasterxml.jackson.core</groupId>
```
chinmaykolhatkar: What is the purpose of this change? If this is unintensional, please revert this file. As such we should be very careful in modifying pom.xml as this might break things at other places.

chinmaykolhatkar (changes requested):
@akshay-harale The changes functionally looks good now.
There are some minor cosmatic changes required though. 
Can you do following so that its ready to merge:
1. Out of the 8 files you've modifed, I see only following files are valid for this PR:
CommonLogParser.java
CommonLogParserTest.java
CommonLog.java

If this is correct, can you please revert them?

2. Please check the minor comments that I have put and modify the changes. Please add "Done" as a reply to comment for the comments which are taken care of.

3. Squash and rebase the changes with github/master.

Once this is done, we'll be ready to merge the code.

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/CommonLogParser.java
akshay-harale: Done

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/CommonLogParser.java
akshay-harale: Done

library/pom.xml
akshay-harale: Done. Removed these changes. It was required while running test cases in IDE.

contrib/src/test/resources/commonLogSchema.json
akshay-harale: Done

contrib/src/test/java/org/apache/apex/malhar/contrib/parser/LogParserTest.java
akshay-harale: Done

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/log/CommonLog.java
akshay-harale: Done

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/LogSchemaDetails.java
akshay-harale: Done

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/LogParser.java
akshay-harale: Done

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/CommonLogParser.java
akshay-harale: Done

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/CommonLogParser.java
akshay-harale: Done

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/LogParser.java
```diff
@@ -209,7 +209,8 @@ public LogSchemaDetails getLogSchemaDetails() {
    * Set log schema details like (fields and regex)
    * @param logSchemaDetails
    */
-  public void setLogSchemaDetails(LogSchemaDetails logSchemaDetails) {
+  public void setLogSchemaDetails(LogSchemaDetails logSchemaDetails)
```
chinmaykolhatkar: There is no functional change here.. Please revert this file.

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/LogSchemaDetails.java
```diff
@@ -180,7 +180,6 @@ public void setPattern(String pattern)
      */
     private String regex;
 
-
```
chinmaykolhatkar: THere is no functional change here.. Please revert this file.

chinmaykolhatkar (changes requested):
Hi @akshay-harale , There are still 2 files namely LogParser.java & LogSchemaDetails.java which do not contain any changes functionally. Please revert and squash the commits.. I'll merge it then.

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/LogParser.java
akshay-harale: Done

contrib/src/main/java/org/apache/apex/malhar/contrib/parser/LogSchemaDetails.java
akshay-harale: Done

PR #549, opened 2017-02-13
APEXMALHAR-2380 Add MutablePair for Kinensis Operator for Recovery State

deepak-narkhede:
Add MutablePair for Kinensis Operator for Recovery State which will avoid default serialization caused by KinesisPair.

Tested with Kinesis to S3 app.

### DIFF
--- a/contrib/src/main/java/com/datatorrent/contrib/kinesis/AbstractKinesisInputOperator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/kinesis/AbstractKinesisInputOperator.java
@@ -38,12 +38,11 @@ import org.slf4j.LoggerFactory;
 
 import org.apache.apex.malhar.lib.wal.WindowDataManager;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.MutablePair;
 
 import com.amazonaws.services.kinesis.model.Record;
 import com.amazonaws.services.kinesis.model.Shard;
 import com.amazonaws.services.kinesis.model.ShardIteratorType;
-import com.esotericsoftware.kryo.DefaultSerializer;
-import com.esotericsoftware.kryo.serializers.JavaSerializer;
 import com.google.common.collect.Sets;
 
 import com.datatorrent.api.Context.OperatorContext;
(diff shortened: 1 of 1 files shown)


contrib/src/main/java/com/datatorrent/contrib/kinesis/AbstractKinesisInputOperator.java
```diff
@@ -58,16 +32,27 @@
...
+import java.io.IOException;
+import java.lang.reflect.Array;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
```
chaithu14: Why are you changing the import order?

deepak-narkhede: @chaithu14 . It was not intentional, IDE did some import optimizations. I've reverted them.

contrib/src/main/java/com/datatorrent/contrib/kinesis/AbstractKinesisInputOperator.java
```diff
@@ -565,11 +554,11 @@ public void emitTuples()
       outputPort.emit(tuple);
       if(!currentWindowRecoveryState.containsKey(shardId))
       {
-        currentWindowRecoveryState.put(shardId, new KinesisPair<String, Integer>(recordId, 1));
+        currentWindowRecoveryState.put(shardId, new MutablePair<String, Integer>(recordId, 1));
```
chaithu14: I would suggest optimize this if block as below:
```java
MutablePair<String, Integer> shardOffsetAndCount = currentWindowRecoveryState.get(shardId);
if (shardOffsetAndCount == null) {
  currentWindowRecoveryState.put(shardId, new MutablePair<String, Integer>(recordId, 1));
} else {
  shardOffsetAndCount.setRight(shardOffsetAndCount.right+1);
}
```

deepak-narkhede: @chaithu14 . Thanks, Done.

chaithu14:
Looks okay to me.

PR #553, opened 2017-02-15
upgrade httpclient version

mattqzhang:


tweise:
@PramodSSImmaneni please open a new PR that closes this when the changes in core are merged.

tweise:
@vrozov can you please have a look at it?

contrib/pom.xml
```diff
@@ -604,10 +604,10 @@
       <optional>true</optional>
     </dependency>
     <dependency>
-      <!-- Kinesis depends on HttpPatch which requires httpclient 4.3.5 -->
+      <!-- Kinesis depends on HttpPatch which requires httpclient 4.3.6 -->
```
vrozov, line 607: What Kinesis actually depends on? Why does it need to be documented here instead of a JIRA or a commit?

contrib/pom.xml
```diff
@@ -604,10 +604,10 @@
...
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
-      <version>4.3.5</version>
+      <version>4.3.6</version>
```
vrozov, line 610: Will Kinesis work with 4.3.6? May it require a more recent version?

pramodin: Unsure we will need to check this

contrib/pom.xml
pramodin: Probably at both places.

tweise:
I think the update needs to be done anyways due to APEXCORE-656

@vrozov I don't think @mattqzhang is working on this any longer, can you take it over?

contrib/pom.xml
vrozov: Both places are OK, but not in pom.xml

vrozov:
@tweise No, I will not be able to pick this up.

contrib/pom.xml
vrozov: The comment is misleading. Kinesis version that contrib optionally depends on depends on 4.3.0 and does not require 4.3.5 or 4.3.6. IMO, it will be better to upgrade Kinesis dependency to at least 1.10.77 that depends on 4.3.6 httpclient.

contrib/pom.xml
chaithu14: I verified the KinesisInputOperator with httpclient = 4.3.6 and Kinesis dependency = 1.9.10. It's working fine.

PR #560, opened 2017-02-23
APEXMALHAR-2422 During recovery don't relay on the size returned by the API

sandeshh:
…t to work.

@otterc , @tushargosavi please review.

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
@@ -20,6 +20,7 @@ package org.apache.apex.malhar.lib.wal;
 
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
+import java.io.EOFException;
 import java.io.IOException;
 import java.util.EnumSet;
 import java.util.HashSet;
@@ -426,6 +427,36 @@ public class FileSystemWAL implements WAL<FileSystemWAL.FileSystemWALReader, Fil
       readOrSkip(true);
     }
 
+    private static int readLen(final DataInputStream inputStream) throws IOException
+    {
+      if (inputStream == null) {
+        return -1;
+      }
+
+      int len = inputStream.read();
+
+      if (len < 0) {
+        return len;
+      }
+
+      len = len << 24;
+
+      for (int i = 2;i >= 0;--i) {
+        int ch = inputStream.read();
+        if (ch < 0) {
+          throw new EOFException();
+        }
+
+        len += (ch << 8 * i);
+      }
+
+      if (len < 0) {
+        throw new IOException("Negative length");
+      }
+
+      return len;
+    }
+
     private Slice readOrSkip(boolean skip) throws IOException
     {
       if (currentPointer == null) {
@@ -442,11 +473,9 @@ public class FileSystemWAL implements WAL<FileSystemWAL.FileSystemWALReader, Fil
           inputStream = getInputStream(currentPointer);
         }
 
-        if (inputStream != null && currentPointer.offset <
-            fileSystemWAL.fileContext.getFileStatus(currentOpenPath).getLen()) {
-          int len = inputStream.readInt();
-          Preconditions.checkState(len >= 0, "negative length");
+        int len = readLen(inputStream);
 
+        if (len != -1) {
           if (!skip) {
             byte[] data = new byte[len];
             inputStream.readFully(data);
--- a/library/src/test/java/org/apache/apex/malhar/lib/wal/FileSystemWALTest.java
+++ b/library/src/test/java/org/apache/apex/malhar/lib/wal/FileSystemWALTest.java
@@ -18,8 +18,9 @@
  */
 package org.apache.apex.malhar.lib.wal;
 
-import java.io.File;
 import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Random;
 
 import org.junit.Assert;
@@ -31,11 +32,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.apex.malhar.lib.utils.FileContextUtils;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileContext;
+import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 
 import com.datatorrent.lib.util.KryoCloneUtils;
-import com.datatorrent.lib.util.TestUtils;
 import com.datatorrent.netlet.util.Slice;
 
 public class FileSystemWALTest
@@ -53,12 +55,21 @@ public class FileSystemWALTest
   {
     private String targetDir;
     FileSystemWAL fsWAL = new FileSystemWAL();
+    Configuration conf = new Configuration();
+    FileSystem fs;
 
     @Override
     protected void starting(Description description)
     {
-      TestUtils.deleteTargetTestClassFolder(description);
       targetDir = "target/" + description.getClassName() + "/" + description.getMethodName();
+
+      try {
+        fs = FileSystem.get(new URI(targetDir + "/WAL"), conf);
+        fs.delete(new Path(targetDir), true);
+      } catch (IOException | URISyntaxException e) {
+        throw new RuntimeException(e);
+      }
+
       fsWAL = new FileSystemWAL();
       fsWAL.setFilePath(targetDir + "/WAL");
     }
@@ -66,7 +77,11 @@ public class FileSystemWALTest
     @Override
     protected void finished(Description description)
     {
-      TestUtils.deleteTargetTestClassFolder(description);
+      try {
+        fs.delete(new Path(targetDir), true);
+      } catch (IOException e) {
+        throw  new RuntimeException(e);
+      }
     }
   }
 
@@ -100,8 +115,8 @@ public class FileSystemWALTest
     fsWALWriter.rotate(true);
     testMeta.fsWAL.beforeCheckpoint(0);
     testMeta.fsWAL.committed(0);
-    File walFile = new File(testMeta.fsWAL.getPartFilePath(0));
-    Assert.assertEquals("WAL file created ", true, walFile.exists());
+    Path walFile = new Path(testMeta.fsWAL.getPartFilePath(0));
+    Assert.assertEquals("WAL file created ", true, testMeta.fs.isFile(walFile));
 
     FileSystemWAL.FileSystemWALReader fsWALReader = testMeta.fsWAL.getReader();
     assertNumTuplesRead(fsWALReader, numTuples);
@@ -133,9 +148,6 @@ public class FileSystemWALTest
     fsWALWriter.rotate(true);
     testMeta.fsWAL.beforeCheckpoint(0);
     testMeta.fsWAL.committed(0);
-
-    File walFile = new File(testMeta.fsWAL.getPartFilePath(0));
-    Assert.assertEquals("WAL file size ", totalBytes, walFile.length());
 
     FileSystemWAL.FileSystemWALReader fsWALReader = testMeta.fsWAL.getReader();
 


library/pom.xml
```diff
@@ -378,6 +378,11 @@
         </exclusion>
       </exclusions> 
     </dependency>
+      <dependency>
+          <groupId>org.apache.hadoop</groupId>
+          <artifactId>hadoop-hdfs</artifactId>
+         <version>2.6.1</version>
+      </dependency>
```
tushargosavi: will this copy hadoo-hdfs jar in app package? see if it can be made provided scope.

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
```diff
@@ -680,9 +681,7 @@ protected void flush() throws IOException
           //until the stream is closed on the local FS, readers don't see any data.
           close();
         } else {
-          Syncable syncableOutputStream = (Syncable)outputStream;
-          syncableOutputStream.hflush();
-          syncableOutputStream.hsync();
+          ((HdfsDataOutputStream)outputStream).hsync(EnumSet.of(SyncFlag.UPDATE_LENGTH));
```
tushargosavi: If I have < 2.6.1 hadoop, then will this work? Is it possible to detect if outputSteams supports hsync with parameter and then only call this else we use default hsync.

sandeshh: HdfsDataOutputStream is available in older version as well, but the minimum supported version for Apex-core is 2.6, is it different for Malhar?

Using default sync could be dangerous as it is not working as expected.

library/pom.xml
sandeshh: Apex-archetype skips the hadoop-hdfs jar in app package, I have tested that. I will explore provided scope.

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
tweise: The previous code used to work on previous versions of Hadoop, which are still supported with the apex core version that malhar currently depends on. Have you researched if there is a way to fix this without the added dependency? Any pointers on what changed in HDFS to cause this?

sandeshh: Here is the explanation for the HSync
https://issues.cloudera.org/browse/DISTRO-696

Behaviour here is, first write is successfully reflected but the subsequent writes are not. 

In the previous version, before WAL, each window data was written to separate file, so it worked.

tweise: HDFS based WAL is used in other places, this is not limited to WindowDataManager.

tweise: Also the example is using append, there is no need to rely on append for the WAL.

sandeshh: I have updated the HDFS library version.

Please refer to following link
http://stackoverflow.com/questions/32231105/why-is-hsync-not-flushing-my-hdfs-file

Just doing hsync won't update the file length with NameNode. We have different options of implementing the solution,

1. Close & re-open the wal file, this will update the length with NameNode and some more changes as rest of the code assumes that Stream is open.

2. Don't depend on the length provided by the fileContext, as mentioned in the SO link

3. Pass the flag while calling HSync ( no other changes ). 

Earlier versions of the HDFS WAL could have used one of the first 2 options.

tweise: HDFS dependency is the least desirable solution. Why do we need to rely on the length from the file? Can we not track the offset in the operator?

sandeshh: If we don't specify flag and don't use HDFS dependency, During recovery, code should not rely on the file length returned by the APIs as namenode has wrong file length, this behavior is confusing.  Ideally, HSYNC should have updated the NameNode, so I consider passing the flag as a cleaner approach, as it updates the Namenode and the data is persisted.

@PramodSSImmaneni please let me know your take as well.

tweise: It isn't important whether the namenode has the metadata or not as long as you can read the data you need from the WAL itself.

library/pom.xml
```diff
@@ -378,6 +378,12 @@
         </exclusion>
       </exclusions> 
     </dependency>
+    <dependency>
+        <groupId>org.apache.hadoop</groupId>
+        <artifactId>hadoop-hdfs</artifactId>
+       <version>2.2.0</version>
```
pramodin: You can use the hadoop.version variable

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
pramodin: Looks like you need this to address an issue you've uncovered that relates to the original problem explained in the JIRA of not getting the correct largest recovery window. Could you explain what that issue is exactly? Also if including the functionality provide by the dependency is the way to do this, please add a comment in the code describing the problem, in case in the future there is a substitute in the common api.

tweise: @sandeshh if you are referring to the use of length in readOrSkip, then this looks incorrect. The WAL needs to be self-sufficient and not rely on data that is stored elsewhere. This should be done by sequentially reading the file till there is no more data. Further, namenode operations should be avoided, especially when they occur on every streaming window. The flush/sync as it stands is sufficient to save the data and keep the file open, rest should be addressed when reading the WAL, which is the infrequent operation.

sandeshh: @tweise will make the changes you suggested.

pramodin: Looking at the code closer, the files whose size isn't reflected correctly with name node are still tmp files that would be finalized later at which point their size will reflect correctly as they would be closed. It may be ok to not have the size be accurately reflected for these tmp files. How about further investigation on the complexity of using the contents of the file itself, specifically what would happen if there was a failure when writing data and there is partial data for a window.

sandeshh: @PramodSSImmaneni I will answer your questions after doing the further investigation.

sandeshh:
@PramodSSImmaneni @tweise please review. Changed the unit test to be compatible with the Hadoop filesystem API as well. Allows to run the unit test on HDFS in future.

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
```diff
@@ -442,8 +442,7 @@ private Slice readOrSkip(boolean skip) throws IOException
           inputStream = getInputStream(currentPointer);
         }
 
-        if (inputStream != null && currentPointer.offset <
-            fileSystemWAL.fileContext.getFileStatus(currentOpenPath).getLen()) {
+        if (inputStream != null && inputStream.available() != 0) {
```
tweise: Please check the documentation for available. It can return 0 even when there is more data to read (blocking). I'm sure you have tested this behavior with HDFS, but maybe it will be necessary to actually try to read from the stream till you get an EOF to be sure that there is indeed no more data to read.

library/src/test/java/org/apache/apex/malhar/lib/wal/FileSystemWALTest.java
```diff
@@ -53,20 +55,33 @@ private static Slice getRandomSlice(int len)
...
+    Configuration conf = new Configuration();
+    FileSystem fs;
 
     @Override
     protected void starting(Description description)
     {
-      TestUtils.deleteTargetTestClassFolder(description);
       targetDir = "target/" + description.getClassName() + "/" + description.getMethodName();
+
+      try {
```
tweise, line 66: I suspect these changes are motivated by your local attempts to run the test against HDFS. If so, it may help others to document what else is needed for it. We cannot by default depend on HDFS but is is useful to know what other setup would be needed to manually test with it.

sandeshh: Yes, that is the motivation for these changes.  The current implementation of the WAL has different code path for a local file system so this issue was not caught for a long time. We need a way to run these tests on HDFS or Mock FS. As a first step, unit tests should only be using Hadoop's FileSystem api. I am planning to propose this in the mailing list.

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
sandeshh: Done.

library/src/test/java/org/apache/apex/malhar/lib/wal/FileSystemWALTest.java
tweise: If you refer to tests that specifically test behavior that relies on HDFS then yes (it is an integration test), but expecting every other unit test to use Hadoop API is certainly not appropriate.

tweise:
Can you summarize on the JIRA what testing was done?

sandeshh:
I have listed the tests done to verify this fix on the jira.

library/src/main/java/org/apache/apex/malhar/lib/wal/FileSystemWAL.java
```diff
@@ -442,9 +443,16 @@ private Slice readOrSkip(boolean skip) throws IOException
           inputStream = getInputStream(currentPointer);
         }
 
-        if (inputStream != null && currentPointer.offset <
-            fileSystemWAL.fileContext.getFileStatus(currentOpenPath).getLen()) {
-          int len = inputStream.readInt();
+        int ch1;
+
+        if (inputStream != null && (ch1 = inputStream.read()) != -1) {
```
tweise: How is this different from just doing readInt() ?

sandeshh: with readInt(), reading of every file ends with exception, that is not the case with read().

tweise: readInt will throw EOFException, isn't that what you need?

sandeshh: There is a logic to read the next segment, which will be skipped if the exception is thrown at the EOF. So EOFException is required only when the write to WAL was incomplete, which also indicates no next Segment.

Another aspect I was looking was, EOF should not be indicated by an Exception (Effective Java- Item 57)

tweise: The reference here shouldn't be the book but the existing API that we depend on and need to work with optimally. So if relying on the EOFException raised would be more suitable, then we should do so. @vrozov any thoughts?

Overall this fix is critical and we agree on the approach now, so we don't need to delay over splitting hairs.

vrozov: @tweise I don't see a big difference between using readInt() and read(). Both are suitable in that place.
@sandeshh Effective Java, IMO, refers to designing API, so it suggests to prefer returning null or -1 over raising an exception. In the case of readInt() there is no way to return a marker that will indicate an invalid value. Note that in this particular case, you may want to see if there is a library that allows reading length, not int. You may also implement helper method readLenght instead of inlining the logic.

sandeshh: I have implemented the helper function.

library/src/test/java/org/apache/apex/malhar/lib/wal/FileSystemWALTest.java
vrozov: Can the test rely on MiniDFSCluster?

tweise: It could, this might be a good follow-up JIRA.

tweise:
@sandeshh can you run a quick rebase, I will merge it then.

sandeshh:
@tweise rebase done.

PR #561, opened 2017-02-24
APEXMALHAR-2415 Taking join on multiple columns in PojoInnerJoin

KapoorHitesh:
@tushargosavi please review and merge. Have rebased the code.

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/PojoInnerJoin.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/PojoInnerJoin.java
@@ -23,8 +23,10 @@ import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.apex.malhar.lib.window.MergeAccumulation;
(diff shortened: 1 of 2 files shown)


library/src/test/java/org/apache/apex/malhar/lib/window/accumulation/PojoInnerJoinTest.java
```diff
@@ -142,6 +192,43 @@ public void setDep(String dep)
     }
   }
 
+  public static class TestOutMultipleKeysClass
```
tushargosavi: Instead of adding new Pojo can you use existing Pojo with additional field.

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/PojoInnerJoin.java
```diff
@@ -212,26 +213,31 @@ private void createSetters()
...
-    } else {
-      for (String field : map2.keySet()) {
-        if (!field.equals(key2)) {
-          map1.put(field, map2.get(field));
-        }
+    Set<String> keySetStream2 = new HashSet<>();
+    for (int i = 0; i < keys.length; i = i+2) {
+      String key1=keys[i];
+      String key2=keys[i+1];
+      keySetStream2.add(key2);
```
tushargosavi: keySetStream2 can computed only once during setup.

PR #571, opened 2017-03-06
APEXMALHAR-2539 Solace input operators

pramodin:
Solace input operators to read data from solace messaging system. 

This was built as a team effort between Solace and DataTorrent, hence the multiple authors in the commit. The authors have been listed in the order of the size of the contribution.

### DIFF
--- a/solace/pom.xml
+++ b/solace/pom.xml
@@ -1,0 +1,227 @@
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache.apex</groupId>
+    <artifactId>malhar</artifactId>
+    <version>3.8.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>malhar-solace</artifactId>
+  <name>Apache Apex Malhar Solace Support</name>
+  <packaging>jar</packaging>
+
+  <properties>
+    <!--checkstyle.console>false</checkstyle.console-->
+    <solace.version>10.0.1</solace.version>
+    <maven.test.skip>true</maven.test.skip>
+  </properties>
+
+
+  <build>
+    <plugins>
+      <!-- Publish tests jar -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <version>2.4</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>test-jar</goal>
+            </goals>
+            <phase>package</phase>
+          </execution>
+        </executions>
+      </plugin>
+     <!-- create resource directory for xml javadoc-->
+     <plugin>
+         <groupId>org.apache.maven.plugins</groupId>
+         <artifactId>maven-antrun-plugin</artifactId>
+         <executions>
+           <execution>
+               <id>createJavadocDirectory</id>
+               <phase>generate-resources</phase>
+               <configuration>
+                   <tasks>
+                     <delete dir="${project.build.directory}/generated-resources/xml-javadoc"/>
+                     <mkdir dir="${project.build.directory}/generated-resources/xml-javadoc"/>
+                   </tasks>
+               </configuration>
+               <goals>
+                   <goal>run</goal>
+               </goals>
+           </execution>
+         </executions>
+      </plugin>
+     <!-- generate javdoc -->
+     <plugin>
+       <groupId>org.apache.maven.plugins</groupId>
+       <artifactId>maven-javadoc-plugin</artifactId>
+       <executions>
+         <!-- generate xml javadoc -->
+         <execution>
+           <id>xml-doclet</id>
+           <phase>generate-resources</phase>
+           <goals>
+             <goal>javadoc</goal>
+           </goals>
+           <configuration>
+             <doclet>com.github.markusbernhardt.xmldoclet.XmlDoclet</doclet>
+             <additionalparam>-d ${project.build.directory}/generated-resources/xml-javadoc -filename ${project.artifactId}-${project.version}-javadoc.xml</additionalparam>
+             <useStandardDocletOptions>false</useStandardDocletOptions>
+             <docletArtifact>
+               <groupId>com.github.markusbernhardt</groupId>
+               <artifactId>xml-doclet</artifactId>
+               <version>1.0.4</version>
+             </docletArtifact>
+           </configuration>
+         </execution>
+         <!-- generate default javadoc jar with custom tags -->
+         <execution>
+           <id>attach-sources</id>
+           <goals>
+             <goal>jar</goal>
+           </goals>
+           <configuration>
+             <skip>true</skip>
+             <tags>
+               <tag>
+                 <name>customTag1</name>
+                 <placement>a</placement>
+                 <head>Custom Tag One:</head>
+               </tag>
+               <tag>
+                 <name>customTag2</name>
+                 <placement>a</placement>
+                 <head>Custom Tag two:</head>
+               </tag>
+               <tag>
+                 <name>customTag3</name>
+                 <placement>a</placement>
+                 <head>Custom Tag three:</head>
+               </tag>
+             </tags>
+           </configuration>
+         </execution>
+       </executions>
+     </plugin>
+     <!-- Transform xml javadoc to stripped down version containing only class/interface comments and tags-->
+     <plugin>
+       <groupId>org.codehaus.mojo</groupId>
+       <artifactId>xml-maven-plugin</artifactId>
+       <version>1.0</version>
+       <executions>
+         <execution>
+           <id>transform-xmljavadoc</id>
+           <phase>generate-resources</phase>
+           <goals>
+             <goal>transform</goal>
+           </goals>
+         </execution>
+       </executions>
+       <configuration>
+         <transformationSets>
+           <transformationSet>
+             <dir>${project.build.directory}/generated-resources/xml-javadoc</dir>
+             <includes>
+               <include>${project.artifactId}-${project.version}-javadoc.xml</include>
+             </includes>
+             <stylesheet>XmlJavadocCommentsExtractor.xsl</stylesheet>
+             <outputDir>${project.build.directory}/generated-resources/xml-javadoc</outputDir>
+           </transformationSet>
+         </transformationSets>
+       </configuration>
+     </plugin>
+     <!-- copy xml javadoc to class jar -->
+     <plugin>
+       <artifactId>maven-resources-plugin</artifactId>
+       <version>2.6</version>
+       <executions>
+         <execution>
+           <id>copy-resources</id>
+           <phase>process-resources</phase>
+           <goals>
+             <goal>copy-resources</goal>
+           </goals>
+           <configuration>
+             <outputDirectory>${basedir}/target/classes</outputDirectory>
+             <resources>
+               <resource>
+                 <directory>${project.build.directory}/generated-resources/xml-javadoc</directory>
+                 <includes>
+                   <include>${project.artifactId}-${project.version}-javadoc.xml</include>
+                 </includes>
+                 <filtering>true</filtering>
+               </resource>
+             </resources>
+           </configuration>
+         </execution>
+       </executions>
+     </plugin>
+     <plugin>
+       <artifactId>maven-surefire-plugin</artifactId>
+       <configuration>
+         <argLine>-Xmx2048m</argLine>
+       </configuration>
+     </plugin>
+    </plugins>
+  </build>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.solacesystems</groupId>
+      <artifactId>sol-jms</artifactId>
+      <version>10.0.1</version>
+    </dependency>
+    <dependency>
+      <groupId>com.solacesystems</groupId>
+      <artifactId>sol-jcsmp</artifactId>
+      <version>10.0.1</version>
+    </dependency>
+    <dependency>
+      <groupId>com.solacesystems</groupId>
+      <artifactId>sol-common</artifactId>
+      <version>10.0.1</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>malhar-library</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>malhar-library</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+      <classifier>tests</classifier>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>apex-common</artifactId>
+      <version>${apex.core.version}</version>
+      <type>jar</type>
+    </dependency>
+  </dependencies>
+</project>
--- a/solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
+++ b/solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
@@ -1,0 +1,305 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.solace;
+
+import java.io.IOException;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.solacesystems.jcsmp.BytesXMLMessage;
+import com.solacesystems.jcsmp.Consumer;
+import com.solacesystems.jcsmp.JCSMPChannelProperties;
+import com.solacesystems.jcsmp.JCSMPException;
+import com.solacesystems.jcsmp.JCSMPFactory;
+import com.solacesystems.jcsmp.JCSMPProperties;
+import com.solacesystems.jcsmp.JCSMPReconnectEventHandler;
+import com.solacesystems.jcsmp.JCSMPSession;
+import com.solacesystems.jcsmp.SessionEvent;
+import com.solacesystems.jcsmp.SessionEventArgs;
+import com.solacesystems.jcsmp.SessionEventHandler;
+import com.solacesystems.jcsmp.XMLMessageListener;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.InputOperator;
+import com.datatorrent.api.Operator;
+import com.datatorrent.common.util.BaseOperator;
+import com.datatorrent.lib.io.IdempotentStorageManager;
+import com.datatorrent.netlet.util.DTThrowable;
+
+@SuppressWarnings("unused")
+public abstract class AbstractSolaceBaseInputOperator<T> extends BaseOperator implements
+    InputOperator, Operator.ActivationListener<Context.OperatorContext>, Operator.CheckpointNotificationListener
+{
+
+  private static final Logger logger = LoggerFactory.getLogger(AbstractSolaceBaseInputOperator.class);
+
+  @NotNull
+  protected JCSMPProperties properties = new JCSMPProperties();
+  protected int connectRetries = -1;
+  protected int reconnectRetries = -1;
+  @Min(1)
+  protected int unackedMessageLimit;
+
+  protected FSOpsIdempotentStorageManager idempotentStorageManager = new FSOpsIdempotentStorageManager();
+
+  protected transient JCSMPFactory factory;
+  protected transient JCSMPSession session;
+
+  //Consumer not being currently used and implemented by the specific operators
+  //protected transient Consumer consumer;
+  protected transient Consumer reliableConsumer;
+
+  protected transient int operatorId;
+  protected transient long currentWindowId;
+  protected transient long lastCompletedWId;
+
+  protected transient int emitCount;
+
+  protected transient volatile boolean drFailover = false;
+  protected transient volatile boolean tcpDisconnected = false;
+
+  //protected transient BlockingQueue<BytesXMLMessage> unackedMessages; // hosts the Solace messages that need to be acked when the streaming window is OK to remove
+  //protected LinkedList<Long> inFlightMessageId = new LinkedList<Long>(); //keeps track of all in flight IDs since they are not necessarily sequential
+  // Messages are received asynchronously and collected in a queue, these are processed by the main operator thread and at that time fault tolerance
+  // and idempotency processing is done so this queue can remain transient
+  protected transient ArrayBlockingQueue<BytesXMLMessage> arrivedTopicMessagesToProcess;
+
+  //protected transient com.solace.dt.operator.DTSolaceOperatorInputOutput.ArrayBlockingQueue<BytesXMLMessage> arrivedMessagesToProcess;
+
+  private transient ReconnectCallbackHandler rcHandler;
+
+  private transient MessageHandler cbHandler;
+
+  protected transient int reconnectRetryMillis = 0;
+
+  protected transient volatile Throwable throwable;
+
+  protected static final int DEFAULT_BUFFER_SIZE = 500;
+
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    operatorId = context.getId();
+    logger.debug("OperatorID: {}", operatorId);
+
+    rcHandler = new ReconnectCallbackHandler();
+    cbHandler = new MessageHandler(arrivedTopicMessagesToProcess);
+
+    factory = JCSMPFactory.onlyInstance();
+
+    //Required for HA and DR to try forever if set to "-1"
+    JCSMPChannelProperties channelProperties = (JCSMPChannelProperties)this.properties.getProperty(JCSMPProperties.CLIENT_CHANNEL_PROPERTIES);
+    channelProperties.setConnectRetries(this.connectRetries);
+    channelProperties.setReconnectRetries(this.reconnectRetries);
+
+    reconnectRetryMillis = channelProperties.getReconnectRetryWaitInMillis();
+
+    try {
+      session = factory.createSession(this.properties, null, new SessionHandler());
+    } catch (JCSMPException e) {
+      DTThrowable.rethrow(e);
+    }
+
+    //logger.debug("Properties Raw: \n{}", properties.toProperties());
+    logger.debug("Properties:\n" + properties.toString());
+    //logger.debug("\n===============================================\n");
+
+    idempotentStorageManager.setup(context);
+    lastCompletedWId = idempotentStorageManager.getLargestRecoveryWindow();
+    //logger.debug("Largest Completed: " + lastCompletedWId);
+  }
+
+  @Override
+  public void beforeCheckpoint(long l)
+  {
+  }
+
+  @Override
+  public void checkpointed(long arg0)
+  {
+  }
+
+  @Override
+  public void committed(long window)
+  {
+    try {
+      idempotentStorageManager.deleteUpTo(operatorId, window);
+    } catch (IOException e) {
+      DTThrowable.rethrow(e);
+    }
+
+  }
+
+  protected T processMessage(BytesXMLMessage message)
+  {
+    T tuple = convert(message);
+    if (tuple != null) {
+      emitTuple(tuple);
+    }
+    return tuple;
+  }
+
+  @Override
+  public void activate(Context.OperatorContext context)
+  {
+    try {
+      session.connect();
+      reliableConsumer = session.getMessageConsumer(rcHandler, cbHandler);
+      //consumer = getConsumer();
+      reliableConsumer.start();
+    } catch (JCSMPException e) {
+      DTThrowable.rethrow(e);
+    }
+  }
+
+  @Override
+  public void deactivate()
+  {
+    reliableConsumer.close();
+  }
+
+  @Override
+  public void teardown()
+  {
+    idempotentStorageManager.teardown();
+    session.closeSession();
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
+    super.beginWindow(windowId);
+    this.currentWindowId = windowId;
+  }
+
+  protected abstract T convert(BytesXMLMessage message);
+
+  protected abstract void emitTuple(T tuple);
+
+  public void setProperties(JCSMPProperties properties)
+  {
+    this.properties = properties;
+  }
+
+  public JCSMPProperties getProperties()
+  {
+    return properties;
+  }
+
+  public IdempotentStorageManager getIdempotentStorageManager()
+  {
+    return idempotentStorageManager;
+  }
+
+  public void setUnackedMessageLimit(int unackedMessageLimit)
+  {
+    this.unackedMessageLimit = unackedMessageLimit;
+  }
+
+  public int getUnackedMessageLimit()
+  {
+    return unackedMessageLimit;
+  }
+
+  public void setConnectRetries(int connectRetries)
+  {
+    this.connectRetries = connectRetries;
+    logger.debug("connectRetries: {}", this.connectRetries);
+  }
+
+  public void setReconnectRetries(int reconnectRetries)
+  {
+    this.reconnectRetries = reconnectRetries;
+    logger.debug("reconnectRetries: {}", this.reconnectRetries);
+  }
+
+  public void setReapplySubscriptions(boolean state)
+  {
+    this.properties.setBooleanProperty(JCSMPProperties.REAPPLY_SUBSCRIPTIONS, state);
+  }
+
+  private class ReconnectCallbackHandler implements JCSMPReconnectEventHandler
+  {
+    @Override
+    public void postReconnect() throws JCSMPException
+    {
+
+      logger.info("Solace client now Reconnected --  possibe Solace HA or DR fail-over");
+      tcpDisconnected = false;
+
+    }
+
+    @Override
+    public boolean preReconnect() throws JCSMPException
+    {
+      drFailover = false;
+      logger.info("Solace client now in Pre Reconnect state -- possibe Solace HA or DR fail-over");
+      tcpDisconnected = true;
+      return true;
+    }
+  }
+
+  private class SessionHandler implements SessionEventHandler
+  {
+    public void handleEvent(SessionEventArgs event)
+    {
+      logger.info("Received Session Event %s with info %s\n", event.getEvent(), event.getInfo());
+
+      // Received event possibly due to DR fail-ver complete
+      if (event.getEvent() == SessionEvent.VIRTUAL_ROUTER_NAME_CHANGED) {
+        drFailover = true; // may or may not need recovery
+        tcpDisconnected = false;
+      }
+    }
+  }
+
+  protected class MessageHandler implements XMLMessageListener
+  {
+
+    BlockingQueue<BytesXMLMessage> messageQueue;
+
+    MessageHandler(BlockingQueue<BytesXMLMessage> messageQueue)
+    {
+      this.messageQueue = messageQueue;
+    }
+
+    @Override
+    public void onException(JCSMPException e)
+    {
+      throwable = e;
+      DTThrowable.rethrow(e);
+    }
+
+    @Override
+    public void onReceive(BytesXMLMessage msg)
+    {
+      try {
+        messageQueue.put(msg);
+      } catch (InterruptedException e) {
+        DTThrowable.rethrow(e);
+      }
+    }
+
+  }
+
+}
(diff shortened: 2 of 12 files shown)


pramodin:
@sanjaypujare please see

solace/pom.xml
```diff
@@ -0,0 +1,224 @@
...
+       <configuration>
+         <argLine>-Xmx2048m</argLine>
+       </configuration>
+     </plugin>
+    </plugins>
+  </build>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.solacesystems</groupId>
```
tweise, line 194: What's the license situation?

pramodin: Good point. The license info is linked here
https://mvnrepository.com/artifact/com.solacesystems/sol-jcsmp/10.0.1
and available here
http://dev.solace.com/solace-license.txt

I am going through it now.

sanjaypujare: I looked at the license which is "solely for the Licensee’s internal business purposes". But we are just referencing the software as a dependency in this pom file which does not subject us to any redistribution or bundling terms. The source code here itself has Apache license so I think we are okay.

tweise: Please provide the references for "I think we are okay".

One potential issue is that users that are not aware of licensing restrictions unknowingly create restrictions for their own product, as it would be the case with xGPL.

sanjaypujare: Reference http://dev.solace.com/solace-license.txt and section 2.2:

"2.2 License of APIs. Provided Licensee complies with this Agreement and any terms that SOLACE provides, SOLACE grants to Licensee a non-exclusive, royalty free license, during the term of this Agreement, to download, install and use, the applicable application programming interfaces that may be made available by SOLACE with the Software (“APIs”) solely to create interfaces between the Software and the Licensee’s software or third party software on Licensee’s systems."

We are creating the interface between Solace and our software/3rd party software.

Also the Solace binary jars are made available on the maven repository and that combined with the "internal business use" license described in 1.1 the users of these operators are allowed to use the jars. Only for the redistribution scenario the developers may need to be aware of this license IMO.

solace/pom.xml
```diff
@@ -0,0 +1,224 @@
...
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache.apex</groupId>
+    <artifactId>malhar</artifactId>
+    <version>3.7.0-SNAPSHOT</version>
```
sanjaypujare: => 3.8.0-SNAPSHOT

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+import com.solacesystems.jcsmp.JCSMPReconnectEventHandler;
+import com.solacesystems.jcsmp.JCSMPSession;
+import com.solacesystems.jcsmp.SessionEvent;
+import com.solacesystems.jcsmp.SessionEventArgs;
+import com.solacesystems.jcsmp.SessionEventHandler;
+import com.solacesystems.jcsmp.XMLMessageListener;
+import com.datatorrent.api.Context;
+import com.datatorrent.api.InputOperator;
+import com.datatorrent.api.Operator;
+import com.datatorrent.api.Operator.CheckpointListener;
```
sanjaypujare: This is deprecated, so can we use CheckpointNotificationListener instead? Why use deprecated class in new code?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+import com.datatorrent.common.util.BaseOperator;
+import com.datatorrent.lib.io.IdempotentStorageManager;
```
sanjaypujare: This too is deprecated so why not use WindowDataManager instead?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+
+@SuppressWarnings("unused")
+public abstract class AbstractSolaceBaseInputOperator<T> extends BaseOperator implements
+    InputOperator, Operator.ActivationListener<Context.OperatorContext>, CheckpointListener
+{
+
+  private static final Logger logger = LoggerFactory.getLogger(AbstractSolaceBaseInputOperator.class);
+
+  @NotNull
+  protected JCSMPProperties properties = new JCSMPProperties();
```
sanjaypujare, line 57: We should take a look at this properties variable and its NotNull constraint and where all this variable is set. My concerns are:

- here it is initialized with an empty JCSMProperties object only to address the NotNull constraint
- the setProperties() method just overwrites the existing variable with a new value
- there is a place where a property is set in the existing JCSMProperties object JCSMPProperties.REAPPLY_SUBSCRIPTIONS.
- if setProperties() is called after setReapplySubscriptions() won't it clobber the earlier setting?

To solve this can we change setProperties() to copy properties from the argument to the member variable as my comment there says?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+
+  protected abstract void emitTuple(T tuple);
+
+  protected abstract Consumer getConsumer() throws JCSMPException;
+
+  protected abstract void clearConsumer() throws JCSMPException;
+
+  public void setProperties(JCSMPProperties properties)
+  {
+    this.properties = properties;
```
sanjaypujare: Can we change this to:

this.properties.fromProperties(properties.toProperties());

Or a better way to copy properties from the argument to the properties data member?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+  int spinMillis;
+
+  protected transient int reconnectRetryMillis = 0;
+
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+
+    operatorId = context.getId();
+    logger.info("OperatorID from Base class: {}", operatorId);
```
sanjaypujare: Why logger.info() and not logger.debug()? Also the "from Base class" comment is not clear. Do you mean AbstractSolaceBaseInputOperator class? Is this logger line useful?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+    InputOperator, Operator.ActivationListener<Context.OperatorContext>, CheckpointListener
+{
+
+  private static final Logger logger = LoggerFactory.getLogger(AbstractSolaceBaseInputOperator.class);
+
+  @NotNull
+  protected JCSMPProperties properties = new JCSMPProperties();
+  protected String connectRetries;
+  protected String reconnectRetries;
+  protected String unackedMessageLimit;
```
sanjaypujare: All 3 variables (connectRetries, reconnectRetries and unackedMessageLimit) sound like a number and they are indeed used as integers ultimately to set integral values on other objects. I would like these to be treated as int's here as well instead of String's. I can't think of a reason why they are String's now.

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    arrivedTopicMessagesToProcess = new ArrayBlockingQueue<BytesXMLMessage>(Integer.parseInt(this.unackedMessageLimit));
+
+    this.context = context;
+
+    //setup info for HA and DR at the transport level
+    super.setConnectRetries(this.connectRetries);
+    super.setReconnectRetries(this.reconnectRetries);
```
sanjaypujare: The above 2 set* calls don't make any sense other than the fact that the call generates a logger.info line. Note that connectRetries and reconnectRetries are defined only in AbstractSolaceBaseInputOperator so 

super.setConnectRetries(this.connectRetries)

assigns a variable value to itself via a function call. Need a reason for this code.

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+
+
+  //protected IdempotentStorageManager idempotentStorageManager = new IdempotentStorageManager.FSIdempotentStorageManager();
```
sanjaypujare: Any reason to have this commented line around since it declares the same idempotentStorageManager variable?  Best to remove this line

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+    //Required for HA and DR to try forever if set to "-1"
+    JCSMPChannelProperties channelProperties = (JCSMPChannelProperties)this.properties.getProperty(JCSMPProperties.CLIENT_CHANNEL_PROPERTIES);
+    channelProperties.setConnectRetries(Integer.parseInt(this.connectRetries));
+    channelProperties.setReconnectRetries(Integer.parseInt(this.reconnectRetries));
+
+    reconnectRetryMillis = channelProperties.getReconnectRetryWaitInMillis();
+
+
+    try {
+      session = factory.createSession(this.properties, null, new PrintingSessionEventHandler());
```
sanjaypujare: Is it more appropriate to move this to activate() where you createSession just before connect()?

Also the 2nd argument is null, but is it better to pass factory.getDefaultContext() ?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+
+
+  //protected IdempotentStorageManager idempotentStorageManager = new IdempotentStorageManager.FSIdempotentStorageManager();
+  protected FSOpsIdempotentStorageManager idempotentStorageManager = new FSOpsIdempotentStorageManager();
+
+  protected transient JCSMPFactory factory;
+  protected transient JCSMPSession session;
+
+  protected transient Consumer consumer;
+  protected transient Consumer reliableConsumer;
```
sanjaypujare: Should have code comments to describe the purpose of having 2 consumers with one being called reliableConsumer.

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+    BytesXMLMessage message;
+    try {
+      while (emitCount < DEFAULT_BUFFER_SIZE && ((message = (BytesXMLMessage)super.arrivedTopicMessagesToProcess.poll(10, TimeUnit.MILLISECONDS)) != null)) {
+        processMessage(message);
+        emitCount++;
+      }
+    } catch (InterruptedException e) {
+      DTThrowable.rethrow(e);
+    }
+    emitCount++;
```
sanjaypujare: Why this extra increment of emitCount? Is this correct? If yes, can we have a comment about why this is required?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceGuaranteedIdempotentInputOperator.java
```diff
@@ -0,0 +1,649 @@
...
+
+    inFlightRecoveryMessages.clear();
+    inFlightRecoveryMessagesPartition.clear();
+    inFlightRecoveryMessagesDR.clear();
+
+
+    //super.setUnackedMessageLimit(this.unackedMessageLimit);
+    //setup info for HA and DR at the transport level
+    super.setConnectRetries(this.connectRetries);
+    super.setReconnectRetries(this.reconnectRetries);
```
sanjaypujare: Similar to above, why have these redundant calls?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceGuaranteedIdempotentInputOperator.java
```diff
@@ -0,0 +1,649 @@
...
+      currentWindowRecoveryState.clear();
+      ackCompleted = ackMessages();
+      LOG.debug("acking messages completed successfully: " + ackCompleted);
+    }
+
+  }
+
+
+  @Override
+  public void emitTuples()
```
sanjaypujare, line 278: Can we break this 160 line function into smaller functions? My suggestions follow. Also please get rid of blank lines that serve no purpose

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceGuaranteedIdempotentInputOperator.java
```diff
@@ -0,0 +1,649 @@
...
+
+            }
+          } else {
+            doneDupsPartitioned = true;
+            donePartitionCheck = false; //Get ready in case another partition restarts and results in replayed messages
+          }
+
+
+          // Checking for duplicates after recovery from DR looking for redelivered messages
+          if (super.DRFailover == true && !(message.getRedelivered()) && doneDupsDR == true && donePartitionCheck == false) {
```
sanjaypujare: Remove "== true" everywhere .

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+    lastCompletedWId = idempotentStorageManager.getLargestRecoveryWindow();
+    //logger.debug("++++++++++++++++++++Largest Completed: " + lastCompletedWId);
+
+
+  }
+
+  @Override
+  public void checkpointed(long arg0)
+  {
+    // TODO Auto-generated method stub
```
sanjaypujare: Pls remove the autogenerated comment,

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+  {
+    try {
+      idempotentStorageManager.deleteUpTo(operatorId, window);
+    } catch (IOException e) {
+      DTThrowable.rethrow(e);
+    }
+
+  }
+
+  protected T processMessage(BytesXMLMessage message)
```
sanjaypujare, line 153: This processMessage is overridden in subclasses AbstractSolaceDirectInputOperator and AbstractSolaceGuaranteedIdempotentInputOperator and those methods also call this one as super.processMessage(..). Unless this processMessage(..) is going to be used in future by other subclasses without overriding it can we call this method something else (say processMessageHelper) so it is clearer

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+
+  @Override
+  public void activate(Context.OperatorContext context)
+  {
+    try {
+      session.connect();
+      reliableConsumer = session.getMessageConsumer(rcHandler, cbHandler);
+      //consumer = getConsumer();
```
sanjaypujare, line 168: remove the commented code

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceGuaranteedIdempotentInputOperator.java
```diff
@@ -0,0 +1,649 @@
...
+      try {
+        Thread.sleep(super.reconnectRetryMillis);
+      } catch (InterruptedException e) {
+        DTThrowable.rethrow(e);
+      }
+      if (sleepCounter % 10 == 0) {
+        LOG.info("Sleeping for another 30 seconds waiting for TCP reconnect for Solace with milliseconds sleep per cycle = {}", super.reconnectRetryMillis);
+      }
+      LOG.info("Queued messages to process: {}", arrivedMessagesToProcess.size());
+    }
```
sanjaypujare: The above block of 14 lines can be a separate method?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceGuaranteedIdempotentInputOperator.java
```diff
@@ -0,0 +1,649 @@
...
+            inFlightRecoveryMessagesDR.clear();
+            super.DRFailover = false;
+            DR_COUNTER = 0;
+            DR_COUNTER_SIZE = 2 * DEFAULT_BUFFER_SIZE;
+            LOG.info("Cleared in flight recovery messages, no more possible duplicate messages detected after DR fail over");
+          }
+
+
+          if (goodToGo) {
+            //if the redelivery flag is no no longer on the messages we can dispose of the inFLightRecoveryMessages
```
sanjaypujare, line 417: this whole block up to line 419 can be refactored into a separate function.

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceGuaranteedIdempotentInputOperator.java
```diff
@@ -0,0 +1,649 @@
...
+  protected transient LinkedList<T> inFlightRecoveryMessages = new LinkedList<T>(); // used by partition that restarts for duplicate detection
+  protected transient LinkedList<T> inFlightRecoveryMessagesPartition = new LinkedList<T>(); // used by partitions that didn't restart for duplicate detection
+  protected transient LinkedList<T> inFlightRecoveryMessagesDR = new LinkedList<T>(); // used by partition that detect DR fail over for duplicate detection
+
+  private transient Context.OperatorContext context;
+  private transient int partitionCount = 0;
+
+  protected static final transient int DEFAULT_BUFFER_SIZE = 500;
+  protected static transient int DR_COUNTER_SIZE = 2 * DEFAULT_BUFFER_SIZE;
+  protected transient int DR_COUNTER = 0;
```
sanjaypujare: Why are the above 2 variables defined in upper case when they are not constants? Can we change to camel-case? Also DR_COUNTER_SIZE is class static but I am not sure if that is intended.

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceGuaranteedIdempotentInputOperator.java
sanjaypujare: Also is the reference to super. necessary?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+  protected transient int operatorId;
+  protected long windowId;
+  protected transient long lastCompletedWId;
+
+  protected transient int emitCount;
+
+  protected transient volatile boolean DRFailover = false;
+  protected transient volatile boolean TCPDisconnected = false;
+
+  protected transient BlockingQueue<BytesXMLMessage> unackedMessages; // hosts the Solace messages that need to be acked when the streaming window is OK to remove
```
sanjaypujare: unackedMessages is not used anywhere in the project. What is the purpose? Can we remove it? If it needs to be kept can we have some comments about why it is there and why it is transient?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+  protected LinkedList<Long> inFlightMessageId = new LinkedList<Long>(); //keeps track of all in flight IDs since they are not necessarily sequential
```
sanjaypujare: This too is not used anywhere. Can we remove it? Also note this is not transient...

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+  protected transient ArrayBlockingQueue<BytesXMLMessage> arrivedTopicMessagesToProcess;
```
sanjaypujare: This is declared transient. It will be good to describe (in a code comment) how you don't lose messages even if this is transient (e.g. similar to Kafka's EARLIEST or APPLICATION setting for the offset).

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+
+  //protected transient com.solace.dt.operator.DTSolaceOperatorInputOutput.ArrayBlockingQueue<BytesXMLMessage> arrivedMessagesToProcess;
+
+  private transient ReconnectCallbackHandler rcHandler = new ReconnectCallbackHandler();
+
+  private transient CallbackMessageHandler cbHandler = new CallbackMessageHandler();
+
+  int spinMillis;
```
sanjaypujare: Why is this not transient? It is initialized in setup()

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+    public void onException(JCSMPException e)
+    {
+      DTThrowable.rethrow(e);
+
+    }
+
+    @Override
+    public void onReceive(BytesXMLMessage msg)
+    {
+      arrivedTopicMessagesToProcess.add(msg);
```
sanjaypujare: Couple of things:

- would put(msg) be better than add() since put is blocking and that is what you want when the queue is full (depends on the contract of XMLMessageListener)

- even with add() if it throws IllegalStateException (when the queue is full) who is going to handle it? Won't there be a loss of messages otherwise?

- don't we need to synchronize on the queue?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceGuaranteedIdempotentInputOperator.java
```diff
@@ -0,0 +1,649 @@
...
+
+  @Override
+  public void emitTuples()
+  {
+    if (currentWindowId <= idempotentStorageManager.getLargestRecoveryWindow()) {
+      return;
+    }
+    //If in HA or DR fail-over, block until Solace TCP connection is reestablished so recovery windows are not lost as empty windows
+    int sleepCounter = 0;
+    while (super.TCPDisconnected == true) {
```
sanjaypujare: => while (TCPDisconnected)

unless there is some advantage in using this idiom that I am not aware of

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceGuaranteedIdempotentInputOperator.java
```diff
@@ -0,0 +1,649 @@
...
+import com.datatorrent.api.InputOperator;
+import com.datatorrent.api.Operator;
+import com.datatorrent.api.Partitioner;
+import com.datatorrent.api.annotation.OperatorAnnotation;
+import com.datatorrent.common.partitioner.StatelessPartitioner;
+import com.datatorrent.lib.io.IdempotentStorageManager;
+import com.datatorrent.netlet.util.DTThrowable;
+
+@OperatorAnnotation(checkpointableWithinAppWindow = false)
+public abstract class AbstractSolaceGuaranteedIdempotentInputOperator<T> extends AbstractSolaceBaseInputOperator<T> implements InputOperator, Operator.ActivationListener<Context.OperatorContext>
```
sanjaypujare, line 55: Javadocs for this class to explain what this class does and how it is different from other input operators?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+
+  }
+
+  /*
```
sanjaypujare: Can we remove all this commented code? If there is some other value in this commented code can we instead use code documentation to achieve that result?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+
+
+  @Override
+  protected Consumer getConsumer() throws JCSMPException
+  {
+    // TODO Auto-generated method stub
+    return null;
+  }
+
+  /*
```
sanjaypujare: remove commented code?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+  protected void removeSubscription(Topic topic)
+  {
+    try {
+      session.removeSubscription(topic);
+    } catch (JCSMPException e) {
+      DTThrowable.rethrow(e);
+    }
+  }
+
+  /*
```
sanjaypujare, line 143: Remove commented code?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+
+  public String getUnackedMessageLimit()
+  {
+    return unackedMessageLimit;
+  }
+
+  public void setConnectRetries(String connectRetries)
+  {
+    this.connectRetries = connectRetries;
+    logger.info("+++++++++++++++++++reconnectRetries: {}", this.connectRetries);
```
sanjaypujare: Do we want to keep this logger.info in production code? Why is showing this more important than other properties?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  public void setReconnectRetries(String reconnectRetries)
+  {
+    this.reconnectRetries = reconnectRetries;
+    logger.info("+++++++++++++++++++reconnectRetries: {}", this.reconnectRetries);
```
sanjaypujare: Ditto for this logger.info.

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+
+  public class ReconnectCallbackHandler implements JCSMPReconnectEventHandler
+  {
+
+
+    @Override
+    public void postReconnect() throws JCSMPException
+    {
+
+      logger.info("++++++++++++++Solace client now Reconnected --  possibe Solace HA or DR fail-over +++++++++++++");
```
sanjaypujare: The leading and trailing "++++" in the message is not the standard format in other log messages in malhar. We should standardize on log messages

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+      DTThrowable.rethrow(e);
+    }
+  }
+
+  @Override
+  public void deactivate()
+  {
+    try {
+      consumer.stop();
+      clearConsumer();
```
sanjaypujare: clearConsumer() is a no-op everywhere. Can we remove the definition and call unless there is a reason for it in the future?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
```diff
@@ -0,0 +1,319 @@
...
+      consumer.close();
+      reliableConsumer.close();
```
sanjaypujare: Why is there no

reliableConsumer.stop();

?

Also the deactivate() and activate() are not symmetric in terms of how consumer is initialized and closed so it is difficult to prove the correctness of the logic. Is there any way we can make it symmetric?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceGuaranteedIdempotentInputOperator.java
```diff
@@ -0,0 +1,649 @@
...
+    if (payload != null) {
+      currentWindowRecoveryState.put(message.getMessageIdLong(), payload);
+    }
+    recentMessage = message;
+    return payload;
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
```
sanjaypujare, line 190: Call to super.beginWindow() is missing which is highly suspicious. Note that assignment to super.windowId doesn't then happen but super.windowId is used elsewhere in the class. Also see my comment about possible duplication between super.windowId and this.currentWindowId.

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceGuaranteedIdempotentInputOperator.java
```diff
@@ -0,0 +1,649 @@
...
+  protected transient EndpointType endpointType = EndpointType.QUEUE;
+  private transient Endpoint endpoint;
+  @NotNull
+  public transient FlowCallbackMessageHandler FlowHandler = new FlowCallbackMessageHandler();
+  @NotNull
+  protected transient EndpointProperties endpointProperties = new EndpointProperties();
+  private transient BytesXMLMessage recentMessage = null;
+
+  private transient long[] operatorRecoveredWindows;
+  protected transient long currentWindowId;
```
sanjaypujare: There is a "windowId" variable in AbstractSolaceBaseInputOperator (super class). So this currentWindowId serves the same purpose it looks like so this seems to be duplication which is also error prone.

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+import com.datatorrent.netlet.util.DTThrowable;
+
+public abstract class AbstractSolaceDirectInputOperator<T> extends AbstractSolaceBaseInputOperator<T> implements InputOperator, Operator.ActivationListener<com.datatorrent.api.Context.OperatorContext>
+{
+
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractSolaceDirectInputOperator.class);
+
+  private final transient AtomicReference<Throwable> throwable;
+  private transient Context.OperatorContext context;
+  protected transient long currentWindowId;
```
sanjaypujare: Does it serve the same purpose as windowId in AbstractSolaceBaseInputOperator (super class)? If there is duplication can we eliminate it?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+import com.datatorrent.api.Operator;
+import com.datatorrent.netlet.util.DTThrowable;
+
+public abstract class AbstractSolaceDirectInputOperator<T> extends AbstractSolaceBaseInputOperator<T> implements InputOperator, Operator.ActivationListener<com.datatorrent.api.Context.OperatorContext>
+{
+
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractSolaceDirectInputOperator.class);
+
+  private final transient AtomicReference<Throwable> throwable;
+  private transient Context.OperatorContext context;
```
sanjaypujare: This is set but not used anywhere. Can we remove it?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+import com.datatorrent.api.InputOperator;
+import com.datatorrent.api.Operator;
+import com.datatorrent.netlet.util.DTThrowable;
+
+public abstract class AbstractSolaceDirectInputOperator<T> extends AbstractSolaceBaseInputOperator<T> implements InputOperator, Operator.ActivationListener<com.datatorrent.api.Context.OperatorContext>
+{
+
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractSolaceDirectInputOperator.class);
+
+  private final transient AtomicReference<Throwable> throwable;
```
sanjaypujare: Not used anywhere. Can we remove?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+  private transient Context.OperatorContext context;
+  protected transient long currentWindowId;
+
+  @NotNull
+  protected String topicName;
+  private transient Topic topic;
+  private transient BytesXMLMessage recentMessage = null;
```
sanjaypujare: recentMessage is assigned to but never used. Can we remove?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+
+  private transient List<T> messages = new ArrayList<T>();
```
sanjaypujare: Never used anywhere. Can we remove?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+  protected final transient Map<Long, T> currentWindowRecoveryState;
```
sanjaypujare: This is marked transient but created in the constructor (not setup). Looks highly suspicious.

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+  protected static final transient int DEFAULT_BUFFER_SIZE = 500;
```
sanjaypujare: transient not necessary for a static final integer constant. @PramodSSImmaneni ?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+    arrivedTopicMessagesToProcess = new ArrayBlockingQueue<BytesXMLMessage>(Integer.parseInt(this.unackedMessageLimit));
+
+    this.context = context;
+
+    //setup info for HA and DR at the transport level
+    super.setConnectRetries(this.connectRetries);
+    super.setReconnectRetries(this.reconnectRetries);
+    super.setReapplySubscriptions(true);
+
+    super.setup(context);
```
sanjaypujare: ideally super.setup(..) should be at the top unless there is a logical reason for it to be where it is now

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+    topic = factory.createTopic(topicName);
+    addSubscription(topic);
```
sanjaypujare: We can move this to activate() (esp if session creation happens in super.activate())

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
```diff
@@ -0,0 +1,238 @@
...
+  {
+    throwable = new AtomicReference<Throwable>();
+    currentWindowRecoveryState = Maps.newLinkedHashMap();
+  }
+
+  @Override
+  public void beginWindow(long windowId)
+  {
+    currentWindowId = windowId;
+    super.beginWindow(windowId);
```
sanjaypujare: Can we move this to the top unless logic requires it?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceGuaranteedIdempotentInputOperator.java
```diff
@@ -0,0 +1,649 @@
...
+public abstract class AbstractSolaceGuaranteedIdempotentInputOperator<T> extends AbstractSolaceBaseInputOperator<T> implements InputOperator, Operator.ActivationListener<Context.OperatorContext>
+{
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractSolaceGuaranteedIdempotentInputOperator.class);
+
+  @NotNull
+  protected String endpointName;
+
+
+  @NotNull
+  protected transient EndpointType endpointType = EndpointType.QUEUE;
```
sanjaypujare: endpointType not used anywhere. Pls remove

tweise:
Commit should have a single author so that the contributor stats are not messed up. Work can be split into multiple commits if that is necessary.

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
pramodin: Having the properties variable initialized helps in configuring the operator from properties from a property file, for example <name>operator.prop.properties.properties['name']</name><value>...</value>

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
pramodin: It is like down converting and up converting back again. Any specific advantages to this?

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
pramodin: This is an offline call so I think it is ok here. It should also be ok to leave the second argument as null because internally it makes that assignment when it is null.

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceBaseInputOperator.java
pramodin: It is a synchronized queue

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
pramodin: It holds the idempotent state for the current window before it is saved.

solace/src/main/java/org/apache/apex/malhar/solace/AbstractSolaceDirectInputOperator.java
pramodin: Additional property reapply subscription is being set

pramodin:
@tweise to your comment, in this case unfortunately individual contributions weren't clearly kept track of and hence the single commit. I don't think contributor stats should be the reason to not allow a collaborative commit, that stats information and graph generated by other tools, that is mainly tracking information and not central to the working or contents of the project, being a little inaccurate should be of concern. These will be very few and far between and will be a small minority and would probably appear at the bottom of most of those stats.

pramodin:
Went through one more round of refactoring and changes.

PR #572, opened 2017-03-06
APEXMALHAR-2429 Ambiguity in passing key parameter to Join accumulation

KapoorHitesh:
@chinmaykolhatkar please review.

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
@@ -20,7 +20,6 @@ package org.apache.apex.malhar.lib.window.accumulation;
 
 import java.lang.reflect.Field;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -43,29 +42,33 @@ import com.datatorrent.lib.util.PojoUtils;
 public abstract class AbstractPojoJoin<InputT1, InputT2>
     implements MergeAccumulation<InputT1, InputT2, List<List<Map<String, Object>>>, List<?>>
 {
-  protected final String[] keys;
-  protected final Class<?> outClass;
+  protected String[] keys;
+  protected Class<?> outClass;
   private transient Map<String,PojoUtils.Getter> gettersStream1;
   private transient Map<String,PojoUtils.Getter> gettersStream2;
   private transient Map<String,PojoUtils.Setter> setters;
   protected transient Set<String> keySetStream2;
   protected transient Set<String> keySetStream1;
+  protected transient String[] leftKeys;
+  protected transient String[] rightKeys;
 
   public AbstractPojoJoin()
   {
-    keys = new String[]{};
+    leftKeys = new String[]{};
+    rightKeys = new String[]{};
     outClass = null;
   }
 
-  public AbstractPojoJoin(Class<?> outClass, String... keys)
+  public AbstractPojoJoin(Class<?> outClass, String[] leftKeys, String[] rightKeys)
   {
-    if (keys.length % 2 != 0) {
-      throw new IllegalArgumentException("Wrong number of keys.");
+    if (leftKeys.length != rightKeys.length) {
+      throw new IllegalArgumentException("Number of keys in left stream should match in right stream");
     }
-
-    this.keys = Arrays.copyOf(keys, keys.length);
+    this.leftKeys = leftKeys;
+    this.rightKeys = rightKeys;
     this.outClass = outClass;
   }
+
 
   private void createSetters()
   {
@@ -174,9 +177,10 @@ public abstract class AbstractPojoJoin<InputT1, InputT2>
       createSetters();
       keySetStream2 = new HashSet<>();
       keySetStream1 = new HashSet<>();
-      for (int i = 0; i < keys.length; i = i + 2) {
-        keySetStream1.add(keys[i]);
-        keySetStream2.add(keys[i + 1]);
+      int lIndex = getLeftStreamIndex();
+      for (int i = 0; i < leftKeys.length; i++) {
+        keySetStream1.add(lIndex == 0 ? leftKeys[i] : rightKeys[i]);
+        keySetStream2.add(lIndex == 1 ? leftKeys[i] : rightKeys[i]);
       }
     }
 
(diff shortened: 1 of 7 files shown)


library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
```diff
@@ -50,10 +50,14 @@
...
+  protected transient String[] leftKeys;
+  protected transient String[] rightKeys;
 
   public AbstractPojoJoin()
   {
     keys = new String[]{};
+    leftKeys = new String[]{};
+    rightKeys = new String[]{};
     outClass = null;
   }
```
chinmaykolhatkar, line 61: Why is this method required? I think we can remove that.

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
```diff
@@ -50,10 +50,14 @@
   private transient Map<String,PojoUtils.Setter> setters;
   protected transient Set<String> keySetStream2;
   protected transient Set<String> keySetStream1;
+  protected transient String[] leftKeys;
```
chinmaykolhatkar, line 52: As leftKeys and rightKeys are added, keys variable can be removed.

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/PojoFullOuterJoin.java
```diff
@@ -44,6 +44,12 @@ public PojoFullOuterJoin(int num, Class<?> outClass, String... keys)
     super(outClass,keys);
   }
 
+  public PojoFullOuterJoin(Class<?> outClass, String[] leftKeys, String[] rightKeys)
```
chinmaykolhatkar: This is a correct ctor... Please remove the previous one i.e. public PojoFullOuterJoin(int num, Class<?> outClass, String... keys)... I'm not somehow not able to add comment on line no 42.

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/PojoInnerJoin.java
```diff
@@ -41,6 +41,11 @@ public PojoInnerJoin(int num, Class<?> outClass, String... keys)
```
chinmaykolhatkar: Ctor at line 38 (i.e. the old ctor) should call the new ctor of super.

chinmaykolhatkar: OR it can call new ctor is PojoInnerJoin itself.

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/PojoLeftOuterJoin.java
```diff
@@ -43,6 +43,12 @@ public PojoLeftOuterJoin(int num, Class<?> outClass, String... keys)
 
   }
 
+  public PojoLeftOuterJoin(Class<?> outClass, String[] leftKeys, String[] rightKeys)
```
chinmaykolhatkar: Remove previous ctor at line 40... This is the right one to be used.

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/PojoRightOuterJoin.java
```diff
@@ -43,6 +43,12 @@ public PojoRightOuterJoin(int num, Class<?> outClass, String... keys)
 
   }
 
+  public PojoRightOuterJoin(Class<?> outClass, String[] leftKeys, String[] rightKeys)
```
chinmaykolhatkar: Remove previous ctor at line 40... This is the correct one that should nly be  exposed to user,

library/src/test/java/org/apache/apex/malhar/lib/window/accumulation/PojoInnerJoinTest.java
```diff
@@ -259,4 +259,31 @@ public void PojoInnerJoinTestMultipleKeys()
     Assert.assertEquals("Josh", testOutClass.getUName());
     Assert.assertEquals(12, testOutClass.getAge());
   }
+
+  @Test
+  public void PojoInnerJoinTestSeparateLeftAndRightKeys()
```
chinmaykolhatkar, line 264: No need to add a new test... Just modify the existing test to user the new constructor.

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/PojoInnerJoin.java
```diff
@@ -41,6 +41,11 @@ public PojoInnerJoin(int num, Class<?> outClass, String... keys)
 
   }
 
+  public PojoInnerJoin(Class<?> outClass, String[] leftKeys, String[] rightKeys)
```
chinmaykolhatkar: As this new one is added now, but previous can;t be removed because of backward compatibility, you could mark the old ctor as Deprecated.

library/src/test/java/org/apache/apex/malhar/lib/window/accumulation/PojoOuterJoinTest.java
```diff
@@ -277,4 +277,85 @@ public void PojoFullOuterJoinTest()
     Assert.assertEquals("Bob", testOutClass.getUName());
     Assert.assertEquals(0, testOutClass.getAge());
   }
+
+  @Test
+  public void PojoRightOuterJoinTestSeparateLeftAndRightKeys()
```
chinmaykolhatkar: No need to add a new test... Just modify the existing test to user the new constructor.

chinmaykolhatkar (changes requested):
Some minor changes.. Please take care of it. Squash and rebase too...

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
KapoorHitesh: Correct but then this will affect backward compatibility for PojoInnerJoin

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
KapoorHitesh: Not sure but may be for kyro serialization. Please suggest.

chinmaykolhatkar:
Looks good to me. Please rebase with master and I can merge it.

KapoorHitesh:
@chinmaykolhatkar have rebased it please merge.

PR #577, opened 2017-03-11
APEXMALHAR-2306 Introduce TestOperatorContext factory method to enable mocking.

tweise:


tweise:
@davidyan74  or @vrozov please merge. I will do the switch to Mockito change in a separate PR.

davidyan74:
LGTM, @vrozov any comments?

library/src/test/java/com/datatorrent/lib/helper/OperatorContextTestHelper.java
```diff
@@ -44,22 +42,26 @@ protected DateFormat initialValue()
 
   };
 
-  public static class TestIdOperatorContext extends TestContext implements OperatorContext
+  public abstract static class TestIdOperatorContext extends TestContext implements OperatorContext
```
vrozov: Is it necessary to have this class as public? Can the method `of()` be moved to the parent class as `mockContextOf(int id)` or `testContextOf(int id)`?  This should allow changing the class to private (implementation details)? Should it have a private constructor?

tweise: I think the class should remain public to be able to extend it. I see the need to possibly add builder style setters in the future to make it easier to customize the context for test cases and hence I prefer to keep the class exposed. I don't have a strong opinion on where the factory methods reside (one level up would make it easier to use though), it wasn't the focus of the PR. If you think it is important enough, I can make that change, I would also rename the class to `TestOperatorContext` then.

tweise: Moving factory methods up would actually result in longer lines (OperatorContextTestHelper.contextOf() vs. TestIdOperatorContext.of()) and potentially the need for extra import statement, so does not look worth it.

vrozov: I disagree. Test classes should use OperatorContext interface only and exposing TestIdOperatorContext promote incorrect usage. There is no need for extra import statement. The method can be imported statically.

tweise: I changed the class name to MockOperatorContext.

I disagree that test classes should use only the OperatorContext interface. They don't do it now and I just gave you a reason why it may be useful to keep it that way. I don't see how that promotes incorrect usage. The OperatorContext interface is for the operator, not for the test code of an operator. In any case, this falls outside of the scope of this PR and can be further discussed separately.

vrozov: The name of the class is not that important - IMO, it should not be visible at all.

Please explain what is the reason to have it exposed. I agree that in the future it may be necessary to provide builder style setters, but they can and should be provided on the builder class (can be the parent or a helper class). The builder class by itself should not implement OperatorContext interface. It should be mocked. I don't see why test cases should use MockOperatorContext instead of OperatorContext.

benchmark/src/test/java/com/datatorrent/benchmark/memsql/MemsqlInputBenchmarkTest.java
```diff
@@ -77,8 +77,8 @@ public void testMethod() throws SQLException, IOException
     attributeMap.put(OperatorContext.PROCESSING_MODE, ProcessingMode.AT_LEAST_ONCE);
     attributeMap.put(OperatorContext.ACTIVATION_WINDOW_ID, -1L);
     attributeMap.put(DAG.APPLICATION_ID, APP_ID);
-    OperatorContextTestHelper.TestIdOperatorContext context =
-        new OperatorContextTestHelper.TestIdOperatorContext(OPERATOR_ID, attributeMap);
+    OperatorContextTestHelper.MockOperatorContext context =
```
vrozov, line 80: What is the reason to use MockOperatorContext instead of OperatorContext?

tweise: You would need to ask that question to the original author. You see it's not introduced by this PR. As I said, it should be a separate discussion, it is not the scope of this PR (regardless whether we agree or not).

vrozov: Why not to use this PR that already modifies a large number of files to fix the usage? As this line needs to be changed, it should be changed to use OperatorContext interface instead of MockOperatorContext.

tweise: It is extra work that I'm not willing to do because I don't have the time and it isn't related to what I want to get done (allowing Malhar to compile with the newer version of core). I don't consider it important either and frankly I think that this discussion is taking time from both of us that would be much better spent on other PRs.

vrozov: If the only goal is to make Malhar to compile with the newer version of Apex, it should be sufficient to add getName() method to the original class and not to modify a large number of files. If the goal is to allow proper mocking of the OperatorContext, the OperatorContext interface should be used. A partial solution does not sound right to me.

tweise: The changes in this PR will allow mocking of the interface, which I have already tested with subsequent change to turn on Mockito. There won't be any need to change code when changing the core version number and that's the issue that I seek to address. I understand that you would like to see other changes piggybacked, but I'm afraid that I cannot work on those.

vrozov: @tweise I can make the necessary changes, will you be OK with the proposed approach?

tweise: You mean changing individual tests to use the OperatorContext interface instead of MockOperatorContext. I'm OK with it but I don't want this PR to be blocked. If you can complete it today then fine, otherwise please work on it separately.

vrozov: Will complete today

tweise:
#580

PR #578, opened 2017-03-13
*Review only* Apexmalhar 2434.jms output operator fix

sanjaypujare:


library/src/main/java/com/datatorrent/lib/io/jms/JMSTransactionableStore.java
```diff
@@ -44,6 +44,7 @@
 
   private transient MessageProducer producer;
   private transient MessageConsumer consumer;
+  private static String OPERATOR_ID = "operatorId";
```
sanjaypujare, line 47: @oliverwnk  add final ...

PR #579, opened 2017-03-13
*Review only* Apexmalhar 2434.jms output operator fix

oliverwnk:


library/src/main/java/com/datatorrent/lib/io/jms/JMSTransactionableStore.java
```diff
@@ -44,6 +44,7 @@
 
   private transient MessageProducer producer;
   private transient MessageConsumer consumer;
+  private static String OPERATOR_ID = "operatorId";
```
sanjaypujare: Make it "private static final ..."

library/src/main/java/com/datatorrent/lib/io/jms/JMSTransactionableStore.java
```diff
@@ -175,7 +178,7 @@ public void connect() throws IOException
       }
 
       Queue queue = getBase().getSession().createQueue(queueName);
-      QueueBrowser browser = getBase().getSession().createBrowser(queue);
+      QueueBrowser browser = getBase().getSession().createBrowser(queue,"operatorId=" + String.valueOf(getOperatorId()));
```
sanjaypujare: String.valueOf is unnecessary. So the following should suffice:

"operatorId=" + getOperatorId()

library/src/main/java/com/datatorrent/lib/io/jms/JMSTransactionableStore.java
```diff
@@ -186,14 +189,15 @@ public void connect() throws IOException
       }
 
       producer = getBase().getSession().createProducer(queue);
-      consumer = getBase().getSession().createConsumer(queue);
+      consumer = getBase().getSession().createConsumer(queue, "operatorId=" + String.valueOf(getOperatorId()));
```
sanjaypujare: Same as above: you can remove String.valueOf()

library/src/main/java/com/datatorrent/lib/io/jms/JMSTransactionableStore.java
```diff
@@ -229,6 +233,7 @@ public boolean isConnected()
 
   private String getQueueName(String appId, int operatorId)
   {
-    return appId + "-" + operatorId;
+    return getSubject() + "." + "metadata";
```
sanjaypujare: For now it is okay but it is preferable to just store the queue name instead of composing it in getQueueName.

library/src/main/java/com/datatorrent/lib/io/jms/JMSTransactionableStore.java
sanjaypujare: Why use hardcoded operatorId here instead of OPERATOR_ID variable?

library/src/main/java/com/datatorrent/lib/io/jms/JMSTransactionableStore.java
sanjaypujare: Why use hardcoded operatorId here instead of OPERATOR_ID variable?

PR #581, opened 2017-03-14
APEXMALHAR-2400 removing dependency of output fieldnames to input fie…

KapoorHitesh:
…ld names @chinmaykolhatkar please review.

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
@@ -33,7 +33,10 @@ import com.google.common.base.Throwables;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Multimap;
 
+import com.datatorrent.lib.util.KeyValPair;
 import com.datatorrent.lib.util.PojoUtils;
+
+import static org.apache.apex.malhar.lib.window.accumulation.AbstractPojoJoin.STREAM.LEFT;
 
 /**
  * Join Accumulation for Pojo Streams.
@@ -47,9 +50,14 @@ public abstract class AbstractPojoJoin<InputT1, InputT2>
   protected Class<?> outClass;
   private transient Map<String,PojoUtils.Getter> gettersStream1;
   private transient Map<String,PojoUtils.Getter> gettersStream2;
-  private transient Map<String,PojoUtils.Setter> setters;
+  protected transient Map<String,PojoUtils.Setter> setters;
+  protected transient Map<String, KeyValPair<STREAM, String>> outputToInputMap;
   protected transient String[] leftKeys;
   protected transient String[] rightKeys;
+  public enum STREAM
+  {
+    LEFT, RIGHT
+  }
 
   public AbstractPojoJoin()
   {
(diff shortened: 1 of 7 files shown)


library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
```diff
@@ -68,6 +76,11 @@ public AbstractPojoJoin(Class<?> outClass, String[] leftKeys, String[] rightKeys
     this.outClass = outClass;
   }
 
+  public AbstractPojoJoin(Class<?> outClass, String[] leftKeys, String[] rightKeys, Map<String, KeyValPair<STREAM, String>> outputToInputMap)
```
chinmaykolhatkar: Please add javadoc for this method. Also can you please add javadoc for the other constructors too?

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
```diff
@@ -47,9 +50,14 @@
   protected Class<?> outClass;
   private transient Map<String,PojoUtils.Getter> gettersStream1;
   private transient Map<String,PojoUtils.Getter> gettersStream2;
-  private transient Map<String,PojoUtils.Setter> setters;
+  protected transient Map<String,PojoUtils.Setter> setters;
+  protected transient Map<String, KeyValPair<STREAM, String>> outputToInputMap;
   protected transient String[] leftKeys;
   protected transient String[] rightKeys;
+  protected enum STREAM
```
chinmaykolhatkar: I think this enum should be public. Otherwise how would anyone access this from outside of this package?

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
```diff
@@ -220,8 +233,24 @@ protected void setObjectForResult(Map<String,PojoUtils.Getter> stream, Object in
         }
         for (Object lObj:left) {
           for (Object rObj:right) {
-            setObjectForResult(leftGettersStream, lObj,o);
-            setObjectForResult(rightGettersStream, rObj,o);
+            if (outputToInputMap != null) {
```
chinmaykolhatkar, line 245: Instead of adding a condition here.. If this is null, then the constructor should populate this approapriately and always use it.

chinmaykolhatkar (changes requested):
Can you take care of review comments and rebase and squash?

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
KapoorHitesh: Not always as this may be a scenario when user wants to add all the fields of output POJO and the mapping is based on field names.

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
KapoorHitesh: Done

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
KapoorHitesh: Done

KapoorHitesh:
@chinmaykolhatkar have made the requested changes have rebased it, please review and merge.

PR #582, opened 2017-03-15
APEXMALHAR-2445 During recovery no need to write to WindowDataManger i…

sandeshh:
…f window id <= LargestCompleted Window ID

@amberarrow please review.

### DIFF
--- a/kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
+++ b/kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
@@ -182,7 +182,11 @@ public class KafkaSinglePortExactlyOnceOutputOperator<T> extends AbstractKafkaOu
   @Override
   public void endWindow()
   {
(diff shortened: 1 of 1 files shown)


kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
```diff
@@ -182,6 +182,10 @@ public void teardown()
   @Override
   public void endWindow()
   {
+    if (windowId <= windowDataManager.getLargestCompletedWindow()) {
+      return;
+    }
+
```
amberarrow: The check for "windowId > windowDataManager.getLargestCompletedWindow()" in the next line is now redundant.

amberarrow:
Seems like the commit message should say "During recovery" rather than "After recovery"

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaSinglePortExactlyOnceOutputOperator.java
sandeshh: Done.

PR #567, opened 2017-02-28
*DO NOT MERGE - REVIEW_ONLY* - Changes in File Input / Output operators for Batch

bhupeshchawda:
3 Changes:
1. Pointing to apex-core 3.6.0-SNAPSHOT for Control Tuple Support
2. Modifying Windowed Operator to use Custom Control Tuples instead of Control port
3. Changes in AbstractFileInputOperator and AbstractFileOutputOperator for supporting file based batches

PR #583, opened 2017-03-16
APEXMALHAR-2447 Notification from AbstractFileInputOperator when scanning directory

prasannapramod:
APEXMALHAR-2447 Added notification in AbstractFileInputOperator, when scanning the directory for the first time. Implementations can use this information in different ways such as notifying downstream operators with a control tuple that a new directory has been scanned.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
@@ -25,6 +25,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Serializable;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
@@ -710,6 +711,23 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
   }
 
   /**
+   * Notifies that the directory is being scanned.<br>
+   * Override this method to custom handling. Will be called once
+   */
+  protected void visitDirectory(Path filePath)
+  {
+  }
+
+  private void checkVisitedDirectory(Path path)
+  {
+    String pathString = path.toString();
+    if (!processedFiles.contains(pathString)) {
+      visitDirectory(path);
+      processedFiles.add(pathString);
+    }
+  }
+
+  /**
    * Scans the directory for new files.
    */
   protected void scanDirectory()
@@ -718,10 +736,19 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
       Set<Path> newPaths = scanner.scan(fs, filePath, processedFiles);
 
       for (Path newPath : newPaths) {
-        String newPathString = newPath.toString();
-        pendingFiles.add(newPathString);
-        processedFiles.add(newPathString);
-        localProcessedFileCount.increment();
+        try {
+          FileStatus fileStatus = fs.getFileStatus(newPath);
+          if (fileStatus.isDirectory())  {
+            checkVisitedDirectory(newPath);
+          } else {
+            String newPathString = newPath.toString();
+            pendingFiles.add(newPathString);
+            processedFiles.add(newPathString);
+            localProcessedFileCount.increment();
+          }
+        } catch (IOException e) {
+          throw new RuntimeException(e);
+        }
       }
 
       lastScanMillis = System.currentTimeMillis();
@@ -1059,6 +1086,11 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
       LinkedHashSet<Path> pathSet = Sets.newLinkedHashSet();
       try {
         LOG.debug("Scanning {} with pattern {}", filePath, this.filePatternRegexp);
+        if (!consumedFiles.contains(filePath.toString())) {
+          if (fs.isDirectory(filePath)) {
+            pathSet.add(filePath);
+          }
+        }
         FileStatus[] files = fs.listStatus(filePath);
         for (FileStatus status : files) {
           Path path = status.getPath();
--- a/library/src/test/java/com/datatorrent/lib/io/fs/AbstractFileInputOperatorTest.java
+++ b/library/src/test/java/com/datatorrent/lib/io/fs/AbstractFileInputOperatorTest.java
@@ -20,7 +20,6 @@ package com.datatorrent.lib.io.fs;
 
 import java.io.ByteArrayOutputStream;
 import java.io.File;
-
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
@@ -59,7 +58,6 @@ import com.datatorrent.api.Context;
 import com.datatorrent.api.DefaultPartition;
 import com.datatorrent.api.Partitioner.Partition;
 import com.datatorrent.api.StatsListener;
-
 import com.datatorrent.lib.io.fs.AbstractFileInputOperator.DirectoryScanner;
 import com.datatorrent.lib.partitioner.StatelessPartitionerTest.PartitioningContextImpl;
 import com.datatorrent.lib.testbench.CollectorTestSink;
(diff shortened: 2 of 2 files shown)


prasannapramod:
can you assign someone to review this pull request? thanks

prasannapramod:
@sanjaypujare please review

bhupeshchawda:
Is the purpose of this JIRA to identify new directories? The code seems to be calling the hook `visitDirectory()` the first time the path is scanned.  This is irrespective of whether new directories are identified..

prasannapramod:
To give a notification for the top level directory that can be used to do
activities pertaining to the directory itself such as the one specified in
the JIRA APEXMALHAR-2447. Are you asking about notification when new
sub-directories are created under the top level directory?

bhupeshchawda:
@prasannapramod What I mean is if `scanDirectory` only runs once, it would not notice any new directory which is created after the path is scanned. This is basically kind of a batch use case where you scan the path once and process anything that is there. Once done, you do not scan for any new files / directories in that path. 
Is there any specific use case you are trying to address? I did some changes to see if we can achieve the batch use case. Please see if the following helps: https://github.com/apache/apex-malhar/pull/567/files#diff-8b18d6df947f93d70436ad9b32645463 
This would be taken up as part of https://issues.apache.org/jira/browse/APEXMALHAR-2451

prasannapramod:
@bhupeshchawda As per my understanding the JIRA is about, Input operator does not read any tuples if the input directory is empty. So output directory is not created.  I provided a fix in form of notification in AbstractFileInputOperator and the implementations can use it for whatever purpose they seem fit.

prasannapramod:
@DT-Priyanka please review.

DT-Priyanka:
Can you please put some clear comments on how it should be used by concrete sub-classes. And update our concrete implementations to do it.
Also, we have handled similar case in FileSplitterInput class, check TimeBasedDirectoryScanner.scan() method. Just refer it once to see if something similar could help.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -710,11 +711,23 @@ public void emitTuples()
...
+  {
+  }
+
+  /**
    * Scans the directory for new files.
    */
   protected void scanDirectory()
   {
     if (System.currentTimeMillis() - scanIntervalMillis >= lastScanMillis) {
+      if (!visitedDirectory) {
```
DT-Priyanka: This will not take care of empty child directory. e.g. you had a directory with following structure
a
|
1.txt, 2.txt, c

and dir "c" is empty  we need to send a control tuple for "c" as well.

DT-Priyanka (changes requested):
We need to add test also for these changes.

prasannapramod:
@DT-Priyanka I looked into this and to reliably know about empty folders independent of the scanner being used I would need to modify the scanner contract to return folder paths along with filepaths. Otherwise, I would need to make changes specifically in the directory scanner which would not available if some other scanner is used. What do you suggest?

DT-Priyanka:
I didn't quiet understand that. Can you please elaborate the approach first and then limitation. Sorry I might have forgot some discussions we had around this before.

prasannapramod:
@DT-Priyanka scanDirectory() method in AbstractFileInputOperator, only returns files, not folders, which means when the directory is empty whether parent directory or subdirectory is empty I cannot get the path as it doesn't return folder path. For ex:  Let's consider directory 'a' ,'b' parent directories. Directory 'c' is inside directory 'b'. Now, we have 2 empty directories. Directory 'a' & directory 'b/c'. But scanDirectory() method doesn't return this folder path. So, we cannot return notify about the empty directories. But when a file placed inside directory 'a', then the scanner returns file path. It's not returning folder paths because it's not in scanner contract to return it.

Two things are possible, first is to make a change in the scanner to do the callback but then it only works for that scanner and user set their own scanner then there will be no callback. 
Second is to change the contract of the scan method to also return folder paths along with file paths and also return folder paths even if the folder is empty.

DT-Priyanka:
scanDirectory method does scan for empty folder as well, we don't have any check to filter out directories in scan method. You can write a quick test and ses "scanDirectory" method does list empty folders as well.

prasannapramod:
@DT-Priyanka In line number 718, scanDirectory is calling scan method and the if block in the scan method on line 1075 is not adding the folder to the set containing paths being return. In line number 1080, it's adding normal files only. So you can see scanner is not returning folders. So should I change the contract?

DT-Priyanka:
Oops, sorry I was looking into someother branch of malhar (3.5) Just took a look back, the behaviour was changed while fixing https://issues.apache.org/jira/browse/APEXMALHAR-2374
We need to understand the downside of sending back folder paths. And see which is the best way to fix it.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
ilganeli: Could address this by storing a set of visited directory URIs and then relying on the provided DirectoryScanner to handle a recursive traversal if necessary

DT-Priyanka:
Here is my suggestion,
In scan method we do add all discovered directory paths to "pathSet", so they will be added to "pendingFiles". While iterating over "pedningFiles" in emitTuples method you can check for "isDirectory", provided it's true, you can always send a control tuple or you can do it selectively when directory is empty. I will say better send it always.
Let me know if this does sounds good to you.

prasannapramod:
@DT-Priyanka I already had this on the mind and did necessary changes in the branch. So will go ahead rise a pull request with this change. Thanks

prasannapramod:
@DT-Priyanka please see.

DT-Priyanka:
Please change branch name to APEXMALHAR jira name

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -710,6 +711,23 @@ public void emitTuples()
...
+   * Override this method to custom handling. Will be called once
+   */
+  protected void visitDirectory(Path filePath)
+  {
+  }
+
+  private void checkVisitedDirectory(Path path)
+  {
+    String pathString = path.toString();
+    if (!processedFiles.contains(pathString)) {
```
DT-Priyanka, line 724: haven't we already done this check on line 1089 ?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -710,6 +711,23 @@ public void emitTuples()
   }
 
   /**
+   * Notifies that the directory is being scanned.<br>
+   * Override this method to custom handling. Will be called once
+   */
+  protected void visitDirectory(Path filePath)
```
DT-Priyanka, line 717: can you please implement this in one of the subclasses / in a subclass written in test and write a unit test around it? There is not way it is tested as of now.

prasannapramod:
@DT-Priyanka regarding chnaging branch name, doing so will require me to close pull request and reopen a new one. Instead can i use the same branch but ensure the commit message and pull request header have the apexmalhar jira number.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
prasannapramod: Since scanner is pluggable and user might implement there own scanner and may not necessary check for the path properly in there scanner. So this check provides additional safety.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
prasannapramod: added unit test.

DT-Priyanka:
@prasannapramod  it is okay to close this PR and open new one. But please rename the branch, as per recent discussions it's important that we follow these guidelines.

Code looks okay to me. Once you rename the branch I should merge it. In new PR please provide reference to this one.

prasannapramod:
Pull request is closed as new pull request is raised against apex malhar.

prasannapramod:
https://github.com/apache/apex-malhar/pull/619

PR #584, opened 2017-03-20
APEXMALHAR-2450 Change unique counter to not emit empty maps.

tweise:


tweise:
@amberarrow this is a very minor change but makes a big difference when in wordcount the result files are not filled with {}s !

amberarrow:
Seems like this could be a breaking change if an application is depending on seeing those empty maps as an indicator of something special in the input stream; could they not be filtered out by a downstream operator ?

tweise:
Which application? The only practical use case for unique counter are demo applications. And even those don't work well with the current behavior, which is why I propose the change.

And tests are passing... 

On a more serious note: This operator has no perspective beyond simple examples, it is superseded by windowed accumulations. And when there is no input, then any sensible operator performing windowed aggregation should not emit output.

amberarrow:
That's one way of looking at it.

Another is that once a library is released, we have no idea what sort of applications are out there. The fact that is _possible_ to write an application that relies on the current behavior and and might break with the new behavior should be enough to exercise caution with such changes.

Could you create a ticket so the discussion can continue there ? If there is consensus, we can act on it.

sanjaypujare:
Why change the behavior if the operator anyway has no use beyond simple examples? To get the modified behavior in the Wordcount DAG can one insert another operator that just discards empty maps (i.e. empty {} )? Or even better if the operator behavior is made configurable using a property so the default behavior is legacy and the new behavior is available with a property (that Wordcount sets).

tweise:
@sanjaypujare For examples it is important that they are concise and intuitive, not quirky and convoluted. I would not want to see extra output, operators or configuration properties in the most basic thing that one could run and that we advertise!

sanjaypujare:
I only suggested this in order to not break backward compatibility. 

We also need to honor the "contract" of the operator as documented in the source code at the top:

```
 * This operator counts the number of times a tuple exists in a window.&nbsp;A map from tuples to counts is emitted at the end of each window.
 * Counts the number of times a key exists or is added in that given window; Count is emitted for the modified or added keys at end of window in a single HashMap.
``` 

So the "contract" does NOT say that a map is emitted only if it is non-empty. So my comments/concerns are as follows:

- does everyone agree that we don't need to worry about backward compatibility?
- for this PR, shouldn't we make the "contract" and behavior consistent?
- not sure if the part that says "Count is emitted for the modified or added keys at the end of window...". If not right that may need to be fixed as well.

tweise:
Good catch WRT documentation. I can update it if we are otherwise in
agreement.

amberarrow:
Absent any further objections, I'm OK with merging it once the issues @sanjaypujare identified with javadoc comments are fixed.

tweise:
@amberarrow updated the PR

PR #585, opened 2017-03-20
APEXMALHAR-2233 Moving DataTorrent examples to Malhar examples

prasannapramod:
This contains the DataTorrent examples minus the ingestion related examples. Those will be consolidated in some form and brought separately in another PR.

I have done the following.

a. Submitted each example as a commit under the original author.
b. Changed examples to conform to the malhar example application and java package structures and not duplicating the entire application pom.xml
c. Updated dependencies in pom.xml for each example to individually specify all the required runtime transitive dependencies as opposed to referring to contrib or library wholesale.
d. Tested all examples to ensure tests are successful.

Review only because of a TODO to add copyright and fix checkstyle. Will do that real soon and commit it as Apex Dev.

@ashwinchandrap @tweise  @amberarrow please see


examples/csvformatter/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+
+    <!-- change these to the appropriate values -->
+    <name>CSVFormatter Application</name>
+    <description>Applications to showcase different formatters</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.apex</groupId>
+            <artifactId>malhar-contrib</artifactId>
+            <version>3.7.0-SNAPSHOT</version>
```
tweise: should not be repeated

examples/parser/README.md
```diff
@@ -0,0 +1,37 @@
+This project contains applications showcasing different **Parsers** and **Formatters** present in the Malhar Library. For all the apps, parameters are configurable in META_INF/properties.xml.
+
+* **Json Parser App**  
+This app showcases **Json Parser**. Data generator sends Json data to the Json Parser which emits each record as POJO on the *output* port. The parser also has *parsedOutput* port that outputs each record as JSONObject and *error* port that emits a error records as key value pair.
+
+* **Csv Parser App**
+
+This application showcases how to use [CsvParser](https://datatorrent.com/docs/apidocs/com/datatorrent/contrib/parser/CsvParser.html) from [Apex Malhar](https://github.com/apache/apex-malhar) library. The CsvParser converts your delimited data to a key-value pair map or concrete java class also know as [POJO](https://en.wikipedia.org/wiki/Plain_Old_Java_Object). The parser emits key-value pair map on *parsedOutput* port. It emits POJO on *out* and error records on *err* port.
```
tweise: links should be fixed

tweise (changes requested):

tweise:
It would be more practical to review if the examples are added one by one. I would also prefer that checkstyle and other cleanup happens outside the Apex commit history. Commit messages should be updated to remove SPOI references. Why not (at least for remaining examples) have original authors bring their code in shape and submit as individual PRs?

prasannapramod:
Made checkstyle fixes and added license headers. Ready for review and merge.

prasannapramod:
@tweise for the remaining examples I am not sure if the original authors can work on bringing the examples. I will submit them in separate PRs as you suggested and original authors could possibly participate in the discussions.

tweise:
I would prefer to see one commit for each example that contains the final contribution, just as it would be the case for any other PR. This should be straightforward, as each example is a separate directory. Noise like this should not be imported into the repository but squashed:  https://github.com/apache/apex-malhar/pull/585/commits/7a09d7ab39b48c46f897362ce6b534985b12e044

prasannapramod:
@tweise i will do that for remaining examples. I already done quite a bit of work for migrating these examples into new structure and don't want to redo the effort. I will address the other comments you have made.

tweise:
@prasannapramod let me know when you are done with the changes. I will have a look at adjusting the commits then.

prasannapramod:
@tweise I am done doing changes for non-ingestion examples. May I know what commits you are trying to adjust? Thanks.

tweise:
@prasannapramod I sent you an updated PR. Please check and update the branch with it. Any further examples need to be submitted as individual PRs following the contributor guidelines.

Also can you check that all examples are part of the build? Parser seems to be missing.

Test pass locally, travis failed because of too much logging. Another reason why these things needs to be reviewed one by one, proper tests need to be present, logging needs to be cleaned up etc.

pramodin:
@tweise I don't see why the original history needs to be re-written and I am not in favor of doing that. To not have a large number of commits, the entire commit history of the example were not brought over. Instead there were single commits for each example for the most part with the primary author, and a couple of commits that adapted it to the project conventions with the checkstyle as apex dev.

tweise:
@PramodSSImmaneni the original history is irrelevant for this repository. With the changes I have made the contribution is attributed to the original author. The alternative is to have the original author do the same work and submit individual PRs.

prasannapramod:
@tweise  With apex-core 3.4 tests are failing for examples such as exactly-once, maprapp, unifiers and also fixedwidthparser, xmlparser which are grouped under parser example. I am working on them.

pramodin:
@tweise why do we need to change what the author contributed and in effect re-write their contribution when there is no other problem like references to copyrighted code? Why can't the code contribution be preserved along with the attribution the way it was written by the author? The subsequent changes done on top ensure that the final code is compliant with the current project conventions.

tweise:
@PramodSSImmaneni if you are interested in history, you can look at the original repository. Adjusting license headers and fixing checkstyle issues (which to me are mechanical changes) all happen prior for contributions making their way in here and are to be squashed, and that is done by the original author. See contributor guidelines.

tweise:
@prasannapramod if that is the case perhaps it better to deal with the parser tests in a separate JIRA? Malhar is going to move to 3.5 core soon probably.

prasannapramod:
@tweise  Yes. That is the plan for failing tests.

pramodin:
@tweise the contribution guidelines point 8. clearly states that when there are multiple authors that attribution be preserved for the work done by them and could require multiple commits. I see there are three types of commits, first the authors who contributed the individual examples, second @prasannapramod that did the project migration which probably includes package structure changes, pom.xml changes such as adapting to the examples maven project structure, dependency changes that I see in the PR, and third changes that can be considered trivial like checkstyle and copyright addition. So I think the structure you proposed doesn't capture this correctly and the current PR structure does.

tweise:
@PramodSSImmaneni regarding the guidelines: that's one way of reading them. Another is that if a feature is developed by multiple authors then those should be reflected through the commits in the PR.

The feature here are the example applications. Package renames and code formatting to me is not material to the feature. I looked at the commit that I think you refer to as "third changes":  https://github.com/apache/apex-malhar/pull/585/commits/e2f5fa90b0bb4abb2958d8304bc49fbf428bcfbf

Most of the diff is package rename and repeated removal of boilerplate code. I had to do similar cleanup work in the past that became part of someone else's PR without my commit appearing in it and I don't consider that a problem. I think a better migration plan on your end could have been to first cleanup all the formatting and perform other mechanical changes and then apply smaller changes on top of it.

pramodin:
@tweise What you are suggesting, as a general policy, will not be welcoming to folks wanting to make bulk contributions to apex (such as an existing project) and in the long run not good for our project. While some might be ok to contribute in this form, as you suggested yourself in the example, others will feel disincentivized to have to do the work to adapt the contribution to the project without any attribution. Furthermore, they may be disinclined to do the work in the first place or have the patience to do it, if they are required to modify existing commits to get the contribution in. I would like this to be left as a choice to the contributor which approach they want to take and allow both options.

tweise:
@PramodSSImmaneni The contributor guidelines are for *individual* contributions. There is no general policy for "bulk contributions". Those will always need to be discussed on a case by case basis. And they involve IP clearance. Furthermore, per examples that I have seen, the import usually removes the history. I think that is necessary for multiple reasons, one of them being that it is nearly impossible to review each and every commit to determine whether it is appropriate to have in the ASF repository. The safest and easiest approach is to flatten everything and bring it in as single commit. In this specific case, it is possible to represent it as a sequence of commits from contributors that happen to be existing committers. 

Who did the work is tracked in JIRA, which is also a good place to look at when discussing potential new committers.

The second point you made about the extra work to modify the commits is mute in this case, because I already did that work.

pramodin:
@tweise I have trouble with "insisting" that the original commits be re-written, when IP issues are not a concern and the commits have already been squashed into independent items with different authors, as it is in this case. Rewriting commits (not squashing) will put undue burden on the contributor. The choice of whether to do this or not should be available to the contributor and the same with whether to drop attributions and flatten it to just one commit. If the submission has an unreasonable number of commits, like the flume PR was, it is reasonable to insist that commits be redone to reduce them. Even though I was on the receiving end of that PR to make those changes, I am fine with it.

If you insist that the PR strictly follow the format you have suggested and otherwise you would be -1 on it, it is disappointing, but putting the contributing party hat on, we would be disinclined to make this contribution.

tweise:
@PramodSSImmaneni it is not clear to me what exactly you have trouble with. When I originally suggested to do the changes we are discussing here, the response was that much effort was put in and don't want to redo it:

> I already done quite a bit of work for migrating these examples into new structure and don't want to redo the effort.

So I picked up the PR at the current stage and did the changes myself, which was rather simple in this case. So the work is done, yet you are still telling me that it is too much work.

Beyond this PR: It is not up the contributor to decide if and how bulk changes are imported into the ASF repository. It is actually strange that you would suggest that, but the discussion also doesn't belong here.

tweise:
WRT the flume PR, if you want I can do a similar rewrite along the lines that I suggested there some time back as well.

pramodin:
@tweise I don't see why the PR in it's current form, which practically follows the guidelines and best practices, is not sufficient and you keep insisting on making changes to recast it, whilst there is no real benefit. If you have substantive changes that affect the working of these examples and I am missing those, let me know and we will be glad to apply them on the PR. I also think, being one of the first bulk changes, the procedures in the PR and the related discussion happening here helps formalize the guidelines for submitting bulk changes that would be in keeping with ASF and project guidelines while also being helpful to contributors. Note that my intent was not to have a discussion here and I am responding to your questions.  Regd flume, it is not an issue, we have aleady made the changes and are close to submitting it. I don't think I have anything more to add to this, so further exchange would not be necessary, let us know if you are -1 on this or you would be ok with going ahead.

tweise:
@PramodSSImmaneni I see a benefit in making the changes that I suggested, otherwise I would have not taken it up to implement them myself. I think they bring the feature in as clean commits with references to a JIRA that allows to track the origin easily. I would have preferred the original contributors do that on their own so it is fully reflected in JIRA as well as the commit history, but then the bulk operation was already on its way and what I did was an improvement on top of it. It is up to the contributor to decide what goes into the proposed PR. If I was the contributor I would be happy to incorporate an improvement and I have done so in similar situations.

Now to the review part. I don't consider the improvement as absolutely required to merge this PR, or in other words, your refusal to incorporate it as a reason to block the PR. I have myself advocated the creation of these examples and I am going to merge either the PR in its present form or with the update that I provided. If you "insist" on the former, then you still need to rebase.

tweise:
@prasannapramod can you please reference the follow-up JIRAs. I see the following issues that need to be taken care of:
- parser examples excluded from build
- build failures in Travis, probably due to excessive logging

tweise:
@PramodSSImmaneni @prasannapramod any update on the follow-up? Have the JIRAs been created? The Travis build hasn't been successful since the examples were merged, so it should be looked into (IMO prior to adding further examples).

prasannapramod:
@tweise JIRA 's created. 
https://issues.apache.org/jira/browse/APEXMALHAR-2465
https://issues.apache.org/jira/browse/APEXMALHAR-2464

PR #587, opened 2017-03-21
APEXMALHAR-2453 Sort Accumulation for Windowed operators

ajaygit158-zz:


ajaygit158-zz:
@chinmaykolhatkar Kindly review

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/Sort.java
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Sort accumulation
+ */
+@InterfaceStability.Evolving
+public class Sort<T> implements Accumulation<T, List<T>, List<T>>
+{
+  boolean sortDesc;
+  Comparator<T> comparator;
+  Comparator<T> descComparator;
```
chinmaykolhatkar: Please consider renaming this to reverseComparator or something similar...

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/Sort.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @Override
+  public List<T> accumulate(List<T> accumulatedValue, T input)
+  {
+    if (comparator == null) {
+      throw new RuntimeException("Comparator not provided, Tuple cannot be compared");
+    }
+    Comparator<T> accComparator = sortDesc ? descComparator : comparator;
+    //binarySearch returns location if input exists else returns (-(insertion point) - 1)
+    int index = Collections.binarySearch(accumulatedValue, input, accComparator);
```
chinmaykolhatkar: Next 3 lines is common piece of code in accumulate and merge.. Can you create a method, lets say insertElement(T input, List<T> accumulation) and call this method from these 2 places?

chinmaykolhatkar: I think doing Collections.binarySearch can be a bit costly for every incoming tuple.
Can this be done in more optimal way?

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/Sort.java
```diff
@@ -0,0 +1,101 @@
...
+{
+  boolean sortDesc;
+  Comparator<T> comparator;
+  Comparator<T> descComparator;
+
+  /**
+   * @param sortDesc
+   *          true if descending sort order is required
+   */
+  public void setSortDesc(boolean sortDesc)
```
chinmaykolhatkar: Instead of having setSortDesc and setCompartor, I suggest to have a constructor for SortAccumulation which has these 2 as parameters. Also you would need to add an empty constructor for kryo purpose.

chinmaykolhatkar (changes requested):
Some minor changes and few questions.
Can you please look into it? Thanks!!

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/Sort.java
ajaygit158-zz: This will be suboptimal since the elements in the arraylist will have to be moved by 1 when inserting the element. I couldnot think of any predefined collections which could help us improve optimality. We can probably improve optimality by using SkipLists.

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/Sort.java
ajaygit158-zz: Renamed to reverseComparator

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/Sort.java
ajaygit158-zz: Replaced the setters with constructor.

library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/Sort.java
ajaygit158-zz: factored out the common code

ajaygit158-zz:
@chinmaykolhatkar Incorporated your review comments.

ajaygit158-zz:
@chinmaykolhatkar Reminder!

chinmaykolhatkar (approved):
LGTM... Will merge in a day if no further comments.

chinmaykolhatkar:
Build failed because of unrelated error. The travis command passed locally. Hence merging.

PR #591, opened 2017-03-24
APEXMALHAR-2455 Create example for Kafka 0.9 API exactly-once output …

oliverwnk:
@sandeshh please review and merge as appropriate


tweise:
This should be integrated into kafka examples after #595 is merged.

examples/kafka-exactly-once/pom.xml
```diff
@@ -0,0 +1,345 @@
...
+        <executions>
+          <execution>
+            <id>enforce-provided-dependencies</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <bannedDependencies>
+                  <excludes>
```
tweise, line 94: The pom.xml would need to be refactored. But I will go away anyways when you make this incremental to the Kafka examples PR.

PR #595, opened 2017-03-29
Apexmalhar 2458 Move kafka related examples from datatorrent examples to apex-malhar examples

prasannapramod:
@tweise @amberarrow @ashwinchandrap please review


tweise:
@prasannapramod please rebase this PR, update version number. Are the tests adequate and do they pass? If everything is ready please make it ready for merge so that the related PR 591 can proceed.

examples/kafka/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+    kafka2hdfs is a example Use of New Kafka Input Operator.
+    hdfs2kafka is a simple application to transfer data from HDFS to Kafka
+  </description>
+  
+  <dependencies>
+    <!-- add your dependencies here -->
+    <dependency>
+      <groupId>org.apache.apex</groupId>
+      <artifactId>malhar-kafka</artifactId>
+      <version>3.6.0</version> <!--check with pramod once-->
```
tweise: this should use ${project.version}

examples/kafka/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+      <version>${apex.core.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka_2.11</artifactId>
+      <version>0.9.0.1</version>
+    </dependency>
+    <!-- Ask pramod about this having 2 versions of kafka -->
+    <dependency>
```
tweise: I would prefer to have the examples based on Kafka 0.9 operators.

tweise:
@prasannapramod would it be possible to take this PR forward before opening more example PRs?

prasannapramod:
@tweise @amberarrow @ashwinchandrap please review fixed issues.

examples/kafka/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+  <parent>
+    <groupId>org.apache.apex</groupId>
+    <artifactId>malhar-examples</artifactId>
+    <version>3.8.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>malhar-examples-kafka</artifactId>
+  <packaging>jar</packaging>
+
+  <!-- change these to the appropriate values -->
```
tweise: remove this comment?

examples/kafka/pom.xml
tweise: still to be addressed

examples/kafka/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+  
+  <dependencies>
+    <!-- add your dependencies here -->
+    <dependency>
+      <groupId>org.apache.apex</groupId>
+      <artifactId>malhar-kafka</artifactId>
+      <version>3.6.0</version> <!--check with pramod once-->
+      <!--version>${project.parent.version}</version-->
+      <exclusions>
+        <exclusion>
```
tweise: why is this necessary?

examples/kafka/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.apache.apex</groupId>
+      <artifactId>apex-engine</artifactId>
+      <version>${apex.core.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <!-- -->
+    <dependency> 
+      <groupId>org.apache.kafka</groupId>
```
tweise: Why is this redeclared and not coming from malhar-kafka? Please update malhar-kafka if necessary.

tweise (changes requested):

examples/kafka/pom.xml
prasannapramod: @tweise I was doing some reading to understand why this was done. Looks like kafka-clients is the new java implementation for the kafka client API and kafka_<scala_version> is the scala client which we have been using in the older operators for some time.

I can only venture a guess, maybe the author of new kafka operator wanted to make sure that they are API compatible with java API and possibly switch completely to java API and hence they included kafka-clients as a dependency in the kafka operator project pom.xml. However, the author may have found that the java clients don't work correctly for all the functionality that we are using and hence, for example, they are using the scala client for the operator tests, as evidenced by the test dependency pointing to the scala client. I believe the original author of this example also is following the same pattern. They excluded the java client and included the scala one in the dependencies.

prasannapramod: An additional data point. I am looking at 0.9 kafka operator dependencies in the kafka operator pom.xml. I see that there is an "optional" dependency to the Scala api along with the regular dependency to the java api. There is a test dependency to scala api as well like I mentioned above for runtime. 

It didn't seem right that both apis are included as dependencies, albeit one is marked optional. I commented out the scala dependency and noticed that while the operators built fine the tests didn't compile. Surmising that the optional dependency to the scala api for compile was added to successfully build the tests.

examples/kafka/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+    kafka2hdfs is a example Use of New Kafka Input Operator.
+    hdfs2kafka is a simple application to transfer data from HDFS to Kafka
+  </description>
+  
+  <dependencies>
+    <!-- add your dependencies here -->
+    <dependency>
+      <groupId>org.apache.apex</groupId>
+      <artifactId>malhar-kafka</artifactId>
+      <version>3.6.0</version>
```
tweise: change to project.version

examples/kafka/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+    <artifactId>malhar-examples</artifactId>
+    <version>3.8.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>malhar-examples-kafka</artifactId>
+  <packaging>jar</packaging>
+
+  <name>Apache Apex Malhar Kafka examples</name>
+  <description>
+    kafka2hdfs is a example Use of New Kafka Input Operator.
```
tweise: This description could be improved.

examples/kafka/README.md
```diff
@@ -0,0 +1,16 @@
+## Kafka to HDFS example :
+
+This sample application show how to read lines from a Kafka topic using the new (0.9)
+Kafka input operator and write them out to HDFS using rolling files with a bounded size.
+
+The output files start out with a `.tmp` extension and get renamed when they reach the
+size bound.  Additional operators to perform parsing, aggregation or filtering can be
+inserted into this pipeline as needed.
+
+## HDFS to kafka example :
```
tweise: kafka => "Kafka"

examples/kafka/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+  </description>
+  
+  <dependencies>
+    <!-- add your dependencies here -->
+    <dependency>
+      <groupId>org.apache.apex</groupId>
+      <artifactId>malhar-kafka</artifactId>
+      <version>3.6.0</version> 
+      <exclusions>
+        <exclusion>
```
tweise: I don't see the comments that arrived via email on this PR now. But since you have already researched this, can we somehow capture what should be done in malhar-kafka as follow-up in a separate JIRA. It seem confusing that the user would have to make this exclusion when using the dependency.

tweise: It may be helpful to add appropriate comments in the malhar-kafka pom.xml

prasannapramod:
@tweise addressed the comments. please review.

tweise:
LGTM, please rebase the PR.

PR #600, opened 2017-04-05
APEXMALHAR-2459 1)Refactor the existing Kafka Input Operator. 2)Added the support of KafkaInputOperator using 0.10 consumer API

chaithu14:


chaithu14:
@siyuanh @tushargosavi @tweise Please review and merge.

tushargosavi:
@chaithu14 
Can you fix following error while compilation.

[ERROR] Failed to execute goal on project malhar-examples-kafka: Could not resolve dependencies for project org.apache.apex:malhar-examples-kafka:jar:3.8.0-SNAPSHOT: Failure to find org.apache.apex:malhar-kafka:jar:3.8.0-SNAPSHOT in http://repository.apache.org/snapshots was cached in the local repository, resolution will not be reattempted until the update interval of apache.snapshots has elapsed or updates are forced -> [Help 1]

tweise:
Probably that is an issue with backward compatibility that should be not be affected unless there was a reason and discussion to the contrary.

chaithu14:
@tushargosavi : Will hit this issue, when re-base the PR. I will fix it and will update the PR.

@tweise: I renamed the artifactId from malhar-kafka to malhar-kafka09. But, the kafka operator is still under "org.apache.apex.malhar.kafka" package, which i didn't change. Will this change comes under backward compatibility? If yes, then I will make the changes in the following way:
1) Copy the existing malhar-kafka to malhar-kafka09
2) Mark the malhar-kafka as deprecated.

tweise:
I don't see any need to rename the existing module and certainly not to copy/duplicate it.

chaithu14:
@tweise To make consistent between the two versions(Kafka-0.9, Kafka-0.10), i renamed the existing module.

tweise:
I don't think that is appropriate as it breaks existing usage of the module.

chaithu14:
@tweise  Thanks for you comments. I realized after reading the document https://apex.apache.org/docs/apex/compatibility/, the above changes comes under backward compatible.
I will retain the existing module and will update the PR.

chaithu14:
I looked into other projects for fixing similar issue without duplicating the code. 
Below is the summary:
1) Renamed the existing operators and module. For example, AbstractKafkaInputOperator renamed to AbstractKafkaInput09Operator, .... malhar-kafka to malhar-kafka09, etc.
2) Create a malhar-kafka module and mark them as deprecated. In this module, added the existing operators which  extends from *09Operator. For example, AbstractKafkaInputOperator extends from AbstractKafkaInput09Operator and mark them as deprecated. This way, all the operators in this module marked them as deprecated and are compatible.

@tweise @tushargosavi Please share your thoughts on this.

kafka/kafka-common/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
```diff
@@ -230,17 +228,17 @@ public void run()
...
   protected void handleNoOffsetForPartitionException(NoOffsetForPartitionException e,
-      KafkaConsumer<byte[], byte[]> consumer)
+      AbstractKafkaConsumer consumer)
   {
     // if initialOffset is set to EARLIST or LATEST
     // and the application is run as first time
     // then there is no existing committed offset and this error will be caught
     // we need to seek to either beginning or end of the partition
     // based on the initial offset setting
-    AbstractKafkaInputOperator.InitialOffset io =
```
tweise: Retain the original class name. I see no need to change it.

kafka/kafka010/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaInputOperator.java
```diff
@@ -0,0 +1,35 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.kafka;
+
+import java.util.Properties;
+
+public abstract class AbstractKafkaInputOperator extends AbstractBaseKafkaInputOperator
```
tweise: Include 010 here instead.

kafka/kafka09/pom.xml
```diff
@@ -0,0 +1,240 @@
...
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka_2.11</artifactId>
+      <version>0.9.0.0</version>
```
tweise: use latest patch

tweise:
@chaithu14 can you look at:

https://issues.apache.org/jira/browse/APEXMALHAR-2162
https://issues.apache.org/jira/browse/APEXMALHAR-2446
http://apache-apex-developers-list.78491.x6.nabble.com/EmbeddedKafka-Server-for-testing-kafka-related-code-td10672.html

tweise:
This will also require update to the documentation (including embedded links)

kafka/kafka-common/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
chaithu14: AbstractKafkaInputOperator spitted into 2 level hierarchy. AbstractBaseKafkaInputOperator consists of common design which is under kafka-common module and contains the createConsumer(...) abstract method. AbstractKafkaInputOperator extends from AbstractBaseKafkaInputOperator which is under malhar-kafka* modules.

kafka/kafka010/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaInputOperator.java
chaithu14: Done.

kafka/kafka09/pom.xml
chaithu14: Done.

chaithu14:
@tweise Sure. I look into it and will open a separate PR for the above JIRA's.

chaithu14:
@tweise @tushargosavi Updated the PR. Please review.

kafka-connectors/kafka-common/src/main/java/org/apache/apex/malhar/kafka/AbstractBaseKafkaInputOperator.java
```diff
@@ -77,12 +77,12 @@
  * @since 3.3.0
  */
 @InterfaceStability.Evolving
-public abstract class AbstractKafkaInputOperator implements InputOperator,
+public abstract class AbstractBaseKafkaInputOperator implements InputOperator,
```
tweise: Please undo this rename.

chaithu14: This change is needed.
Changed the class hierarchy from AbstractKafkaInputOperator <-- KafkaSinglePortInputOperator to
AbstractBaseKafkaInputOperator <-- AbstractKafkaInputOperator(malhar-kafka) <-- KafkaSinglePortInputOperator(malhar-kafka).
Please note, the functionality of AbstractKafkaInputOperator is same in both the hierarchies.
AbstractBaseKafkaInputOperator which is under malhar-common package contains createConsumer(...) abstract method.
AbstractKafkaInputOperator which is in malhar-kafka(supports 0.9 version) package creates the type of consumer based on the kafka version i.e, 0.9 consumer wrapper.

kafka-connectors/kafka010/src/test/java/org/apache/apex/malhar/kafka/KafkaConsumerPropertiesTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestWatcher;
+import org.junit.runner.Description;
+
+import org.apache.kafka.common.KafkaException;
+
+import com.datatorrent.api.Context;
+
+public class KafkaConsumerPropertiesTest
```
tushargosavi: This test seems to be repeat. Can it be moved to common package?

kafka-connectors/kafka010/src/main/java/org/apache/apex/malhar/kafka/AbstractKafka010InputOperator.java
```diff
@@ -0,0 +1,35 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.kafka;
+
+import java.util.Properties;
+
+public abstract class AbstractKafka010InputOperator extends AbstractBaseKafkaInputOperator
```
tushargosavi: I feel, it should be easy for user to move from 0.9 to 0.10 version by just changing the application pom.xml , If the class name is different then source code change is also needed , @tweise what do you think about following class structures.

- AbstractKafkaInputOperator  (In malhar-kafka010 artifact for 0.10 kafka)
- AbstractKafkaInputOperator (In malhar-kafka artifact for 0.9 kafka)
- AbstractBaseKafkaInputOperator (in malhar-kafka-common artifact)

tushargosavi (commented):
@chaithu14 How was testing done? Have your verified read/write from kafka 0.9 and 0.10 versions? Also how backward compatibility was tested?

tushargosavi:
@chaithu14 Documentation update is needed with this change.

Also there are no output operators in malhar-kafka10 , If I want to use output operators with kafka 10, what is the procedure? will adding dependency on malhar-kafka and excluding kaka-client (0.9) version will work?

kafka-connectors/kafka010/src/test/java/org/apache/apex/malhar/kafka/KafkaConsumerPropertiesTest.java
chaithu14: Yes. You are right. Common test code can be moved to malhar-common package.

kafka-connectors/kafka010/src/main/java/org/apache/apex/malhar/kafka/AbstractKafka010InputOperator.java
chaithu14: +1 for @tushargosavi point. 
@tweise Please share your opinion on this. If you are okay, i will rename to AbstractKafkaInputOperator.

chaithu14:
@tushargosavi Below are my comments:

Testing the operator
===============
I tested kafka2hdfs example application which is under examples package. This application is referring to the 0.9 version of kafka client. 
I tested the same example by changing the artifactId from malhar-kafka to malhar-kafka010. It worked fine.

Documentation Update
=================
I think once the hierarchy and class names are fine, then it's good to start the documentation update. So, I will work on documentation update once this PR is merged. If you are okay with this, will create the JIRA for the same.

Output operators in Kafka-10
=====================
This JIRA is about the KafkaInputOperator. In this PR, there is no output operator for Kafka 0.10.  I will create JIRA for this task.

kafka-connectors/kafka010/src/main/java/org/apache/apex/malhar/kafka/AbstractKafka010InputOperator.java
chaithu14: @tweise Please share your opinion on the above class name.

chaithu14:
@tushargosavi @tweise Created JIRA's for the above two tasks(APEXMALHAR-2496, APEXMALHAR-2497) and updated the PR. Please review and merge.

kafka-connectors/kafka-common/pom.xml
```diff
@@ -0,0 +1,223 @@
+<!--
```
tweise: I see no reason to change the top level directory name.

kafka-connectors/kafka-common/pom.xml
```diff
@@ -0,0 +1,223 @@
...
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka_2.11</artifactId>
+      <version>0.9.0.0</version>
```
tweise: Should probably use latest patch.

kafka-connectors/kafka09/pom.xml
```diff
@@ -0,0 +1,240 @@
...
+
+  <parent>
+    <groupId>org.apache.apex</groupId>
+    <artifactId>malhar-kafka-connectors</artifactId>
+    <version>3.8.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>malhar-kafka</artifactId>
+  <packaging>jar</packaging>
+  <name>Apache Apex Malhar Kafka Support using 0.9.0.0 Consumer API</name>
```
tweise: 0.9.0.0 -> 0.9

kafka-connectors/kafka-common/src/main/java/org/apache/apex/malhar/kafka/AbstractBaseKafkaInputOperator.java
tweise: I don't see any need for the rename. I also don't see the need for derived abstract classes when the only difference is the creation of the consumer (you can add that to KafkaSinglePortInputOperator). Also, existing abstract class was evolving, the derived classes you created are not. Any new classes should be evolving.

chaithu14:
@tweise Incorporated review comments. Please review the changes.

chaithu14:
@tweise @tushargosavi Please review. If you are okay with the changes, I will squash the commits and then update the PR.

kafka/kafka010/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumer010.java
```diff
@@ -0,0 +1,200 @@
...
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.clients.consumer.OffsetCommitCallback;
+import org.apache.kafka.common.Metric;
+import org.apache.kafka.common.MetricName;
+import org.apache.kafka.common.TopicPartition;
+
+/**
+ * Wrapper for 0.10.0.0 version of Kafka consumer
```
tweise: 0.10.x - check other occurrences

kafka/kafka010/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,50 @@
...
+# under the License.
+#
+
+log4j.rootLogger=DEBUG,CONSOLE
+
+log4j.appender.CONSOLE=org.apache.log4j.ConsoleAppender
+log4j.appender.CONSOLE.layout=org.apache.log4j.PatternLayout
+log4j.appender.CONSOLE.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %c{2} %M - %m%n
+log4j.appender.CONSOLE.threshold=INFO
+#log4j.appender.CONSOLE.threshold=${test.log.console.threshold}
```
tweise: why is this commented out

kafka/kafka09/pom.xml
```diff
@@ -0,0 +1,240 @@
...
+        <executions>
+          <execution>
+            <goals>
+              <goal>test-jar</goal>
+            </goals>
+            <phase>package</phase>
+          </execution>
+        </executions>
+      </plugin>
+      <!-- create resource directory for xml javadoc-->
```
tweise: do we really need a repetition of this and other boilerplate code in every module vs. having it defined in the parent

tweise (changes requested):

kafka/kafka010/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumer010.java
chaithu14: I verified and updated the version.

kafka/kafka09/pom.xml
chaithu14: Yes. It's not needed in every module. I moved it to parent pom.

kafka/kafka010/src/test/resources/log4j.properties
chaithu14: This is commented out even in the master branch.

tweise: Unless you are aware why let's make it consistent with other modules.

chaithu14: Done.

sandeshh:
Any updates on this?

chaithu14:
@tweise I addressed your comments. If you are fine with the changes, please merge it.

sandeshh:
@chaithu14 @tushargosavi Can we merge this PR?, I have a customer waiting for this fix.

tushargosavi:
@chaithu14 I have tested the changes with following configuration
- old application using kafka
- Just updated the dependency in pom.xml with new malhar and was able to run app without any code changes.
- Just updated dependency in pom to xml to use kafka 10 module and was able to run the application with  kafka 0.10 server.

The PR looks good to me. 

Can you fix the test failures? I will merge the pull request after that, If there are no more objections.

chaithu14:
@tushargosavi Thanks for reviewing and testing the changes.
  I'll investigate the test failures and update it. Looks like the Zookeeper client is not started.

tweise:
The output operator should be in the common package, unless there was a change in the producer API.

chaithu14:
@tweise Created separate JIRA(APEXMALHAR-2496) for output. I will open a separate PR for this. I would suggest, first let's merge this PR because there are lot of file changes happened in this PR. 

@tushargosavi I ran the unit tests in cluster, it's working fine. But, these are failing in travis/jenkins. So, I reverted the re-factoring the unit test. But, still the tests are failing. Looks like there is a Kafka server setup issue in travis and also lot's of noisy in logs. Please help me out, how to debug and fix this issue.

tweise:
@chaithu14 this PR is moving around the output operator as well. Unless it does not work with 0.10, please leave it in the common module.

chaithu14:
@tweise Ok. Will verify and add it in common module.

chaithu14:
@tushargosavi @tweise I moved the abstract Kafka output operator to common-package and fixed the test failures.
Please review and merge.

tweise:
Why is KafkaSinglePortOutputOperator not in common? Also can you please fix the parameterized type warning?

tweise:
Why is it necessary to duplicate the input operator test? They seem nearly identical:
```
diff kafka/kafka09/src/test/java/org/apache/apex/malhar/kafka/KafkaInputOperatorTest.java kafka/kafka010/src/test/java/org/apache/apex/malhar/kafka/KafkaInputOperatorTest.java
394,396c394
<       (hasMultiPartition ? "," + l + TEST_KAFKA_BROKER_PORT[0] : "") +
<       (hasMultiCluster ? ";" + l + TEST_KAFKA_BROKER_PORT[1] : "") +
<       (hasMultiCluster && hasMultiPartition ? "," + l + TEST_KAFKA_BROKER_PORT[1] : "");
---
>       (hasMultiCluster ? ";" + l + TEST_KAFKA_BROKER_PORT[1] : "");
```

tweise:
test this please

chaithu14:
@tweise Thanks a lot for reviewing. Yes. KafkaSinglePortOutputOperator can be in kafka-common module, but, not the KafkaSinglePortExactlyOnceOutputOperator. I'll move it.

Fix the parameterized type warning: I'm unable to figure it out. Could you please point it out the code.

Regarding duplicate the input operator test: If i move it to kafka-common module, then it will test only 0.9.* version. I want to test both the versions, that's why, i added it in both modules. If the tests be re-factored, then the tests are failing. I investigated on this, and trying to integrate different embedded servers, but, it's taking time to fix. So, I'll work as a separate task for refactoring the tests.

tweise:
So how about you just keep the test in common and then just extend from it in Kafka010? 

Any idea why the build in Jenkins fails?

chaithu14:
@tweise I tried it, but mostly it's failing due to the creation of Kafka server in travis/jenkins. That's why trying to integrate with below different embedded server's. 
1) EmbeddedKafka(org.apache.apex.malhar.kafka) which is in Kafka module. Limitation is single cluster. Using this, we can't able to test the multi-cluster test cases which are in KafkaInputOperatorTest.
2) KafkaUnit(https://github.com/chbatey/kafka-unit): Limitation - max number of connections allowed from a particular host is 16. Using this, we can test the multi-cluster test cases. But, due to the limitation of connections, can't able to ran all the tests.

It's taking time to fix this, so, will take this as a separate task.

chaithu14:
@tweise I don't have any idea about the Jenkins failed. But, the Kafka tests are passed.

tweise:
It is OK to improve the tests later, but I don't understand why the same code in two different locations would work vs the same unit test code executed from two different test cases not?

chaithu14:
@tweise I don't have an idea about that. But, the failure is due to the server setup.
Please let me know, if there are no more comments, I'll squash the commits and update the PR.

tweise:
Do squash the commits anyways.

What is the issue with the server setup? Why does it work with duplicate code in two places vs. same code called from two places?

chaithu14:
@tweise In the cluster, it's working fine. But, failing it in travis. I don't remembered the exact build, but might be failing due to the below error:
ERROR server.KafkaApis error - [KafkaApi-0] error when handling request Name: TopicMetadataRequest; Version: 0; CorrelationId: 10; ClientId: producer-1; Topics: testtopic0

chaithu14:
@tweise I squashed the commits and updated the PR. If there are no more comments, please merge it.

tweise:
Please address test code in follow-up PR. Output test should probably also be in common, input test code should be de-duped. We need solid test coverage and reliable CI.

chaithu14:
Sure Thomas.

tweise:
@chaithu14 can you point me to the JIRA that will address the test related issues? Did you capture the related info from this PR?  

It looks like the Kafka tests are flaky and that impacts the CI, see:

https://travis-ci.org/apache/apex-malhar/builds/249401536?utm_source=github_status&utm_medium=notification

```
java.lang.AssertionError: received END_TUPLES more than expected.
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.apache.apex.malhar.kafka.KafkaInputOperatorTest$CollectorModule.endWindow(KafkaInputOperatorTest.java:232)
	at com.datatorrent.stram.engine.GenericNode.processEndWindow(GenericNode.java:153)
	at com.datatorrent.stram.engine.GenericNode.run(GenericNode.java:397)
	at com.datatorrent.stram.engine.StreamingContainer$2.run(StreamingContainer.java:1428)
```

chaithu14:
@tweise I haven't created a new JIRA for test related issues. I'll use APEXMALHAR-2162 for tracking these issues. I'm working on re-factor the unit tests and integration with embedded server.

PR #605, opened 2017-04-17
APEXMALHAR-2473 Support for global cache meta information in db Cache…

oliverwnk:
…Manager

  1. Uses Component interface and newly implemented CacheContext to pass properties to the Stores by calling setup(CacheContext).
  2. APEXMALHAR-2474: FSLoader implements Component to get numInitLinesToCache from CacheManager and use it in initial load.
	Add implementation of get() function so data will also be loaded after initial load.
  3. APEXMALHAR-2475: CacheStore implements Component for passing readOnly and numInitLinesToCache.
	Added NO_EVICTION expire strategy. This strategy will be set in setup(CacheContext) if readOnly is true.

@PramodSSImmaneni: please review

### DIFF
--- a/contrib/src/main/java/com/datatorrent/contrib/enrich/AbstractEnricher.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/enrich/AbstractEnricher.java
@@ -73,9 +73,10 @@ public abstract class AbstractEnricher<INPUT, OUTPUT> extends BaseOperator imple
   /**
    * Helper variables.
    */
-  private transient CacheManager cacheManager;
   protected transient List<FieldInfo> lookupFieldInfo = new ArrayList<>();
   protected transient List<FieldInfo> includeFieldInfo = new ArrayList<>();
+
+  private CacheManager cacheManager = new NullValuesCacheManager();
 
   /**
    * This method needs to be called by implementing class for processing a tuple for enrichment.
@@ -154,7 +155,6 @@ public abstract class AbstractEnricher<INPUT, OUTPUT> extends BaseOperator imple
   {
     super.setup(context);
 
-    cacheManager = new NullValuesCacheManager();
     CacheStore primaryCache = new CacheStore();
 
     // set expiration to one day.
@@ -185,13 +185,18 @@ public abstract class AbstractEnricher<INPUT, OUTPUT> extends BaseOperator imple
     try {
       cacheManager.initialize();
     } catch (IOException e) {
-      throw new RuntimeException("Unable to initialize primary cache", e);
+      throw new RuntimeException("Unable to initialize cache manager", e);
     }
   }
 
   @Override
   public void deactivate()
   {
+    try {
+      cacheManager.close();
+    } catch (IOException e) {
+      throw new RuntimeException("Unable to close cache manager", e);
+    }
   }
 
   /**
@@ -319,5 +324,15 @@ public abstract class AbstractEnricher<INPUT, OUTPUT> extends BaseOperator imple
   public void setCacheSize(int cacheSize)
   {
     this.cacheSize = cacheSize;
+  }
+
+  public CacheManager getCacheManager()
+  {
+    return cacheManager;
+  }
+
+  public void setCacheManager(CacheManager cacheManager)
+  {
+    this.cacheManager = cacheManager;
   }
 }
--- a/contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
@@ -28,7 +28,6 @@ import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import org.apache.commons.io.IOUtils;
 import org.apache.hadoop.classification.InterfaceStability;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
@@ -37,6 +36,8 @@ import org.apache.hadoop.fs.Path;
 
 import com.esotericsoftware.kryo.NotNull;
 import com.google.common.collect.Maps;
+
+import com.datatorrent.api.Component;
 import com.datatorrent.lib.db.cache.CacheManager;
 import com.datatorrent.lib.util.FieldInfo;
 
@@ -51,7 +52,7 @@ import com.datatorrent.lib.util.FieldInfo;
  * @since 3.4.0
  */
 @InterfaceStability.Evolving
-public abstract class FSLoader extends ReadOnlyBackup
+public abstract class FSLoader extends ReadOnlyBackup implements Component<CacheManager.CacheContext>
 {
   @NotNull
   private String fileName;
@@ -59,6 +60,7 @@ public abstract class FSLoader extends ReadOnlyBackup
   private transient Path filePath;
   private transient FileSystem fs;
   private transient boolean connected;
+  private transient int numInitCachedLines = -1;
 
   private static final Logger logger = LoggerFactory.getLogger(FSLoader.class);
 
@@ -75,40 +77,29 @@ public abstract class FSLoader extends ReadOnlyBackup
   @Override
   public Map<Object, Object> loadInitialData()
   {
-    Map<Object, Object> result = null;
-    FSDataInputStream in = null;
-    BufferedReader bin = null;
-    try {
+    Map<Object, Object> result;
+    try (
+      FSDataInputStream in = fs.open(filePath);
+      BufferedReader bin = new BufferedReader(new InputStreamReader(in));
+    ) {
       result = Maps.newHashMap();
-      in = fs.open(filePath);
-      bin = new BufferedReader(new InputStreamReader(in));
       String line;
-      while ((line = bin.readLine()) != null) {
+      int linesCount = 0;
+      while ((line = bin.readLine()) != null && (numInitCachedLines < 0 || linesCount < numInitCachedLines)) {
         try {
           Map<String, Object> tuple = extractFields(line);
           if (tuple != null && !tuple.isEmpty()) {
             result.put(getKey(tuple), getValue(tuple));
           }
         } catch (Exception parseExp) {
-          logger.info("Unable to parse line {}", line);
+          throw new RuntimeException(parseExp);
         }
+        ++linesCount;
       }
     } catch (IOException e) {
       throw new RuntimeException(e);
-    } finally {
-      if (bin != null) {
-        IOUtils.closeQuietly(bin);
-      }
-      if (in != null) {
-        IOUtils.closeQuietly(in);
-      }
-      try {
-        fs.close();
-      } catch (IOException e) {
-        throw new RuntimeException(e);
-      }
     }
-    logger.debug("loading initial data {}", result.size());
+    logger.debug("loading initial data, size: {}", result.size());
     return result;
   }
 
@@ -145,7 +136,42 @@ public abstract class FSLoader extends ReadOnlyBackup
   @Override
   public Object get(Object key)
   {
+    try (
+      FSDataInputStream in = openWithRetryFsInputStream(filePath);
+      BufferedReader bin = new BufferedReader(new InputStreamReader(in));
+      ) {
+      String line;
+      while ((line = bin.readLine()) != null) {
+        try {
+          Map<String, Object> tuple = extractFields(line);
+          if (tuple != null && !tuple.isEmpty() && getKey(tuple).equals(key)) {
+            logger.debug("Found line in FS {}", getValue(tuple));
+            return getValue(tuple);
+          }
+        } catch (Exception parseExp) {
+          throw new RuntimeException(parseExp);
+        }
+      }
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
     return null;
+  }
+
+  private FSDataInputStream openWithRetryFsInputStream(Path filePath)
+  {
+    FSDataInputStream in;
+    try {
+      in = fs.open(filePath);
+    } catch (IOException e) {
+      try {
+        this.connect();
+        in = fs.open(filePath);
+      } catch (IOException ex) {
+        throw new RuntimeException(ex);
+      }
+    }
+    return in;
   }
 
   @Override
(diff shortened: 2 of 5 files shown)


contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
```diff
@@ -83,7 +87,11 @@ public void setFileName(String fileName)
       in = fs.open(filePath);
       bin = new BufferedReader(new InputStreamReader(in));
       String line;
+      int linesCount = 0;
       while ((line = bin.readLine()) != null) {
+        if ( numInitCachedLines > 0 && linesCount >= numInitCachedLines) {
```
pramodin: Can this be added to the while condition

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
```diff
@@ -145,6 +149,34 @@ private Object getKey(Map<String, Object> tuple)
   @Override
   public Object get(Object key)
   {
+    FSDataInputStream in = null;
+    BufferedReader bin = null;
+    try {
+      this.connect();
```
pramodin: Why open file system for each call?

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
```diff
@@ -145,6 +149,34 @@ private Object getKey(Map<String, Object> tuple)
...
+      in = fs.open(filePath);
+      bin = new BufferedReader(new InputStreamReader(in));
```
pramodin: You can also do... try with close

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
```diff
@@ -145,6 +149,34 @@ private Object getKey(Map<String, Object> tuple)
...
+      String line;
+      while ((line = bin.readLine()) != null) {
+        try {
+          Map<String, Object> tuple = extractFields(line);
+          if (tuple != null && !tuple.isEmpty() && getKey(tuple).equals(key)) {
+            logger.debug("Found line in FS {}", getValue(tuple));
+            return getValue(tuple);
+          }
+        } catch (Exception parseExp) {
+          logger.info("Unable to parse line {}", line);
```
pramodin: warn level

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
pramodin: Also shouldn't fs be already initialized?

contrib/src/main/java/com/datatorrent/contrib/enrich/AbstractEnricher.java
```diff
@@ -73,10 +73,11 @@
   /**
    * Helper variables.
    */
-  private transient CacheManager cacheManager;
   protected transient List<FieldInfo> lookupFieldInfo = new ArrayList<>();
   protected transient List<FieldInfo> includeFieldInfo = new ArrayList<>();
 
+  private CacheManager cacheManager = new NullValuesCacheManager();
```
pramodin, line 79: Enricher is initializing cacheManager during activate but doesn't seem to be closing it in deactivate.

library/src/main/java/com/datatorrent/lib/db/cache/CacheManager.java
```diff
@@ -210,6 +272,54 @@ public String getRefreshTime()
...
   }
 
+
+  /**
+   * Used by {@link CacheManager} to pass properties to its stores which implement {@link Component}.
+   */
+  public static class CacheContext implements Context
+  {
+    public static final transient Attribute<Boolean> READ_ONLY_ATTR = new Attribute<>((false));
+    public static final transient Attribute<Integer> NUM_INIT_CACHED_LINES_ATTR = new Attribute<>((-1));
```
pramodin, line 285: The checks in other places might be straightforward if it is initialized to Integer.MAX_VALUE.

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
```diff
@@ -51,14 +53,16 @@
...
+public abstract class FSLoader extends ReadOnlyBackup implements Component<CacheManager.CacheContext>
 {
   @NotNull
   private String fileName;
 
   private transient Path filePath;
   private transient FileSystem fs;
   private transient boolean connected;
+  private transient int numInitCachedLines = -1;
+  private transient CacheManager.CacheContext cacheContext;
```
pramodin: Is there a need to save the cache context?

library/src/main/java/com/datatorrent/lib/db/cache/CacheManager.java
```diff
@@ -182,6 +217,32 @@ public String getRefreshTime()
...
+   * the time. If the time is not set, cache is not refreshed.
+   *
+   * @param numInitCachedLines number of lines to initially from
+   */
+  public void setNumInitCachedLines(int numInitCachedLines)
+  {
+    this.numInitCachedLines = numInitCachedLines;
+  }
+
+  public boolean getReadOnlyAttribute()
```
pramodin: How about just readOnlyData instead of Attribute in name

library/src/main/java/com/datatorrent/lib/db/cache/CacheManager.java
```diff
@@ -56,27 +61,55 @@
...
   {
+    Class[] parameters = new Class[1];
+    parameters[0] = CacheContext.class;
+
+    if (primary instanceof Component) {
+      try {
+        primary.getClass().getMethod("setup", parameters);
+        attributeMap.put(CacheContext.READ_ONLY_ATTR, readOnlyAttribute);
+        attributeMap.put(CacheContext.NUM_INIT_CACHED_LINES_ATTR, numInitCachedLines);
+        ((Component)primary).setup(new CacheContext(attributeMap));
```
pramodin: How about reusing the CacheContext instance between the primary and backup initialization. Looks like same attributeMap is being re-written.

library/src/main/java/com/datatorrent/lib/db/cache/CacheManager.java
oliverwnk: I see, it would slightly change the behavior, e.g. in FSLoader it has no upper bound if num_init_cached_lines_attr was not set. With the change Integer.MAX_VALUE would always be the default limit there

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
oliverwnk: Okay, I added a function where it tries to open a stream with the initialized file system. If it happens to fail it will try to initialize file system once again.

library/src/main/java/com/datatorrent/lib/db/cache/CacheManager.java
oliverwnk: Might be necessary to leave as is to keep functionality of CacheStore (primary store) to not limit maxCacheSize explicitly when numInitCacheLines is not set, but rise maxCacheSize if a greater numInitCacheLines is chosen.

contrib/src/main/java/com/datatorrent/contrib/enrich/AbstractEnricher.java
```diff
@@ -185,13 +185,18 @@ public void activate(Context context)
     try {
       cacheManager.initialize();
     } catch (IOException e) {
-      throw new RuntimeException("Unable to initialize primary cache", e);
+      throw new RuntimeException("Unable to initialize cache store", e);
```
pramodin: Why would be just cache store initialization failure, the backup could fail as well, isn't it. Shouldn't this be unable to initialize cache manager.

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
```diff
@@ -75,40 +77,29 @@ public void setFileName(String fileName)
   @Override
   public Map<Object, Object> loadInitialData()
   {
-    Map<Object, Object> result = null;
-    FSDataInputStream in = null;
-    BufferedReader bin = null;
-    try {
+    Map<Object, Object> result;
+    try (
+      FSDataInputStream in = fs.open(filePath);
```
pramodin, line 82: Where is the input stream closed?

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
```diff
@@ -75,40 +77,29 @@ public void setFileName(String fileName)
...
+      int linesCount = 0;
+      while ((line = bin.readLine()) != null && (numInitCachedLines < 0 || linesCount < numInitCachedLines)) {
         try {
           Map<String, Object> tuple = extractFields(line);
           if (tuple != null && !tuple.isEmpty()) {
             result.put(getKey(tuple), getValue(tuple));
           }
         } catch (Exception parseExp) {
-          logger.info("Unable to parse line {}", line);
+          logger.warn("Unable to parse line {}", line);
```
pramodin: Pass exception as well

pramodin: Since this is initialization, you might as well fail with a runtime exception.

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
```diff
@@ -145,9 +136,44 @@ private Object getKey(Map<String, Object> tuple)
   @Override
   public Object get(Object key)
   {
+    try (
+      FSDataInputStream in = openFsInputStream(filePath);
```
pramodin: Close input stream when done

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
```diff
@@ -145,9 +136,44 @@ private Object getKey(Map<String, Object> tuple)
...
+      String line;
+      while ((line = bin.readLine()) != null) {
+        try {
+          Map<String, Object> tuple = extractFields(line);
+          if (tuple != null && !tuple.isEmpty() && getKey(tuple).equals(key)) {
+            logger.debug("Found line in FS {}", getValue(tuple));
+            return getValue(tuple);
+          }
+        } catch (Exception parseExp) {
+          logger.warn("Unable to parse line {}", line);
```
pramodin: Fail with runtime exception if unable to parse line.

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
```diff
@@ -145,9 +136,44 @@ private Object getKey(Map<String, Object> tuple)
...
+        }
+      }
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
     return null;
   }
 
+  private FSDataInputStream openFsInputStream(Path filePath)
```
pramodin: Maybe call it check open or something that indicates it is not a mere open but will try to handle failures and retry

library/src/main/java/com/datatorrent/lib/db/cache/CacheStore.java
```diff
@@ -111,14 +119,26 @@ public void connect() throws IOException
...
+
     if (entryExpiryStrategy == ExpiryType.EXPIRE_AFTER_ACCESS) {
       cacheBuilder.expireAfterAccess(entryExpiryDurationInMillis, TimeUnit.MILLISECONDS);
     } else if (entryExpiryStrategy == ExpiryType.EXPIRE_AFTER_WRITE) {
       cacheBuilder.expireAfterWrite(entryExpiryDurationInMillis, TimeUnit.MILLISECONDS);
     }
     cache = cacheBuilder.build();
+
+    if (entryExpiryStrategy == ExpiryType.NO_EVICTION) {
+      return;
```
pramodin, line 139: In disconnect, cleanupSchedule is referenced as well, in disconnect check for not null

library/src/test/java/com/datatorrent/lib/db/cache/CacheManagerTest.java
```diff
@@ -118,6 +122,40 @@ public boolean isConnected()
...
   }
 
+  private static class DummyComponentBackupStore extends DummyBackupStore implements Component<CacheManager.CacheContext>
+  {
+    CacheManager.CacheContext cacheContext;
+
+    @Override
+    public Map<Object, Object> loadInitialData()
+    {
+      Map<Object, Object> result = new HashedMap();
```
pramodin: Why not use the standard HashMap?

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
oliverwnk: try-with-resources should close it automatically

contrib/src/main/java/com/datatorrent/contrib/enrich/AbstractEnricher.java
oliverwnk: I meant generally a store of the cache, but I see the confusion with 'CacheStore'. I will change it accordingly.

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
oliverwnk: How about openWithRetryFsInputStream(..)?

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
```diff
@@ -137,7 +137,7 @@ private Object getKey(Map<String, Object> tuple)
   public Object get(Object key)
   {
     try (
-      FSDataInputStream in = openFsInputStream(filePath);
+      FSDataInputStream in = openWithRetryFsInputStream(filePath);
```
pramodin: Looks like stream is still not being closed.

pramodin:
squash commits and rebase

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
oliverwnk: Try-with-resources should close it automatically.

contrib/src/main/java/com/datatorrent/contrib/enrich/FSLoader.java
```diff
@@ -145,9 +136,44 @@ private Object getKey(Map<String, Object> tuple)
...
 
+  private FSDataInputStream openWithRetryFsInputStream(Path filePath)
+  {
+    FSDataInputStream in = null;
+    try {
+      in = fs.open(filePath);
+    } catch (IOException e) {
+      try {
+        this.connect();
+        fs.open(filePath);
```
pramodin: in = fs.open(filePath)

PR #617, opened 2017-05-07
APEXMALHAR-2481 support lambda expressions with high level API

tweise:
@davidyan74 @siyuanh please review

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/function/Function.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/function/Function.java
@@ -35,7 +35,7 @@ import com.datatorrent.lib.util.KeyValPair;
  * @since 3.4.0
  */
 @InterfaceStability.Evolving
-public interface Function
+public interface Function extends java.io.Serializable
 {
   /**
    * If the {@link Function} implements this interface.
--- a/library/src/main/java/org/apache/apex/malhar/lib/function/FunctionOperator.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/function/FunctionOperator.java
@@ -30,8 +30,11 @@ import org.apache.apex.malhar.lib.utils.ByteArrayClassLoader;
 import org.apache.apex.malhar.lib.utils.TupleUtil;
 import org.apache.apex.malhar.lib.window.Tuple;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.mutable.MutableObject;
 import org.apache.hadoop.classification.InterfaceStability;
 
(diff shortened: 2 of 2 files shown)


library/src/main/java/org/apache/apex/malhar/lib/function/Function.java
```diff
@@ -35,7 +35,7 @@
  * @since 3.4.0
  */
 @InterfaceStability.Evolving
-public interface Function
+public interface Function extends java.io.Serializable
```
davidyan74, line 38: Would kryo's @DefaultSerializer work for interfaces? If so, maybe we should add it here?

davidyan74 (approved):

library/src/main/java/org/apache/apex/malhar/lib/function/Function.java
tweise: Kryo cannot load a lambda class. Therefore the problem described below applies, even if the annotation is recognized, which I'm not sure about either.

tweise:
```
Caused by: com.esotericsoftware.kryo.KryoException: Unable to find class: com.example.myapexapp.ApplicationWithHighLevelAPI$$Lambda$4/1847509784
Serialization trace:
value (org.apache.commons.lang3.mutable.MutableObject)
statefulF (org.apache.apex.malhar.lib.function.FunctionOperator$FlatMapFunctionOperator)
	at com.esotericsoftware.kryo.util.DefaultClassResolver.readName(DefaultClassResolver.java:138)
	at com.esotericsoftware.kryo.util.DefaultClassResolver.readClass(DefaultClassResolver.java:115)
	at com.esotericsoftware.kryo.Kryo.readClass(Kryo.java:641)
	at com.esotericsoftware.kryo.serializers.ObjectField.read(ObjectField.java:99)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.read(FieldSerializer.java:528)
	at com.esotericsoftware.kryo.Kryo.readObject(Kryo.java:679)
	at com.esotericsoftware.kryo.serializers.ObjectField.read(ObjectField.java:106)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.read(FieldSerializer.java:528)
	at com.esotericsoftware.kryo.Kryo.readClassAndObject(Kryo.java:761)
	at com.datatorrent.common.util.FSStorageAgent.retrieve(FSStorageAgent.java:192)
	at com.datatorrent.stram.plan.logical.LogicalPlan$OperatorMeta.readObject(LogicalPlan.java:898)
```

PR #618, opened 2017-05-14
Upgrading APEXCORE dependency to version 3.6.0

ajaygit158-zz:


ajaygit158-zz:
Travis build is successful. Jenkins build failed due to Jenkins' internal issue. Closing and reopening again.

ajaygit158-zz:
Travis build stuck without producing any output. Reopening PR.

pramodin:
Can you see there seem to be some other changes in this PR

tweise:
Changes other than the version number would need explanation.

contrib/src/test/java/com/datatorrent/contrib/hbase/HBaseTransactionalPutOperatorTest.java
```diff
@@ -92,6 +92,12 @@ public int getWindowsFromCheckpoint()
         {
           return 0;
         }
+
```
vrozov: use com.datatorrent.stram.engine.OperatorContext:
```
      final DefaultAttributeMap attributes = new DefaultAttributeMap();
      attributes.put(OperatorContext.PROCESSING_MODE, ProcessingMode.AT_LEAST_ONCE);
      thop.setup(new OperatorContext(0, "hbase", attributes, null));
```

library/src/main/java/org/apache/apex/malhar/lib/dedup/TimeBasedDedupOperator.java
```diff
@@ -186,7 +186,7 @@ public void deactivate()
   @Override
   protected void putManagedState(Object tuple)
   {
-    ((ManagedTimeUnifiedStateImpl)managedState).put(getTime(tuple), getKey(tuple), new Slice(new byte[0]));
+    ((ManagedTimeUnifiedStateImpl)managedState).put(getTime(tuple), getKey(tuple), new Slice(null));
```
vrozov: not required for the upgrade. Please use separate JIRA and PR.

library/src/main/java/com/datatorrent/lib/math/RunningAverage.java
```diff
@@ -64,9 +64,9 @@
     @Override
     public void process(Number tuple)
     {
-      double sum = (count * average) + tuple.doubleValue();
-      count++;
-      average = sum / count;
+      double sum = (RunningAverage.this.count * average) + tuple.doubleValue();
```
vrozov, line 67: @bhupeshchawda This is a problem in the DefaultInputPort. We should keep count as private and provide protected methods to update the count.

library/src/test/java/com/datatorrent/lib/util/TestUtils.java
```diff
@@ -69,7 +69,7 @@ public String getDir()
     {
       String methodName = desc.getMethodName();
       String className = desc.getClassName();
-      return "target/" + className + "/" + methodName;
+      return "target/tests/" + className + "/" + methodName;
```
vrozov: Is it required for the upgrade?

stream/src/test/java/org/apache/apex/malhar/stream/api/impl/ApexStreamImplTest.java
```diff
@@ -80,7 +82,7 @@ public void testAddOperator()
       // Assert the stream is from first operator to second operator
       Assert.assertEquals("first", stream.getSource().getOperatorMeta().getName());
       Assert.assertTrue(1 == stream.getSinks().size());
-      Assert.assertEquals("second", stream.getSinks().get(0).getOperatorWrapper().getName());
+      Assert.assertEquals("second", ((List<InputPortMeta>)(stream.getSinks())).get(0).getOperatorMeta().getName());
```
vrozov: Do not cast, use iterator.

library/pom.xml
```diff
@@ -366,6 +366,37 @@
...
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-common</artifactId>
+      <version>${hadoop.version}</version>
+      <type>test-jar</type>
+      <scope>test</scope>
+      <exclusions>
+        <exclusion>
```
vrozov, line 376: exclude all dependencies

contrib/src/test/java/com/datatorrent/contrib/hbase/HBaseTransactionalPutOperatorTest.java
tweise: Should be replaced with MockContext
https://github.com/apache/apex-malhar/pull/577/files#diff-ebae1f3ab77d8bc7c7c38d21abf4db32L22

tweise (changes requested):

library/src/main/java/org/apache/apex/malhar/lib/dedup/TimeBasedDedupOperator.java
ajaygit158-zz: We cannot create a separate PR as Apex 3.4.0 was using netlet 1.2.1 which will throw NullPointerException if null is passed.

library/src/main/java/com/datatorrent/lib/math/RunningAverage.java
bhupeshchawda: @vrozov I created: https://issues.apache.org/jira/browse/APEXCORE-722
While making changes, I see that the change would break semantic versioning, as DefaultInputPort is not Evolving.

library/pom.xml
ajaygit158-zz: I could not get what you mean by exclude all dependencies. I tried with 
`
<exclusion>
           <groupId>*</groupId>
           <artifactId>*</artifactId>
         </exclusion>
`
This gives compilation issues.

library/src/main/java/org/apache/apex/malhar/lib/dedup/TimeBasedDedupOperator.java
vrozov: Exclude this change from the version upgrade PR, it is not necessary. After upgrade, change `new Slice(new byte[0])` to `new Slice(null, 0, 0)`.

library/pom.xml
vrozov: Please provide more details, for example, what classes are required for compilation to succeed.

library/src/main/java/com/datatorrent/lib/math/RunningAverage.java
vrozov: Let me bring the issue to the community attention. We will need to agree how to proceed.

contrib/src/test/java/com/datatorrent/contrib/hbase/HBaseTransactionalPutOperatorTest.java
```diff
@@ -55,44 +59,11 @@ public void testAtleastOnce() throws Exception {
...
-
-        @Override
-        public int getWindowsFromCheckpoint()
-        {
-          return 0;
-        }
-      });
+      AttributeMap.DefaultAttributeMap attributeMap = new AttributeMap.DefaultAttributeMap();
+      attributeMap.put(OperatorContext.PROCESSING_MODE, ProcessingMode.AT_LEAST_ONCE);
+      attributeMap.put(OperatorContext.ACTIVATION_WINDOW_ID, -1L);
```
vrozov: Why is it necessary?

pramodin:
I am sorry but why don't we have a separate PR for updating just the version dependency?

library/src/main/java/com/datatorrent/lib/math/RunningAverage.java
vrozov: @ajaygit158 There are other occurrences where DefaultInputPort.count is updated instead of Operator's count.

library/src/main/java/org/apache/apex/malhar/lib/dedup/TimeBasedDedupOperator.java
ajaygit158-zz: @vrozov The test "DeduperTimeBasedPOJOImplTest" is failing if we dont do this change as a part of upgrade. Apex-core 3.6.0 is using netlet 1.3.0 which throws IllegalArgumentException from Slice().

library/src/main/java/com/datatorrent/lib/math/RunningAverage.java
ajaygit158-zz: Corrected other occurrences. Next PR update should reflect the changes.

contrib/src/test/java/com/datatorrent/contrib/hbase/HBaseTransactionalPutOperatorTest.java
ajaygit158-zz: This was as @tweise had suggested to use mockOperatorContext. Should we move it to different PR?
We will still have to make some changes in this file since there are compilation issues.

library/pom.xml
ajaygit158-zz: I have updated the pom to exclude most of the dependencies with successful build.

ajaygit158-zz:
@PramodSSImmaneni There are a few compilation and test failures due to upgrade. We will limit the changes to bare minimum for what is required.

library/src/test/java/org/apache/apex/malhar/lib/fs/GenericFileOutputOperatorTest.java
```diff
@@ -166,6 +166,7 @@ public void runTestApplication() throws Exception
     for (int i = 0; i < MAX && (!outputFile.exists()); ++i) {
       Thread.sleep(1000);
     }
+    lc.shutdown();
```
ajaygit158-zz: I have raised PR for this change separately (APEXMALHAR-2491). I required this for successful travis builds

stream/src/test/java/org/apache/apex/malhar/stream/api/impl/ApexStreamImplTest.java
```diff
@@ -80,7 +81,11 @@ public void testAddOperator()
       // Assert the stream is from first operator to second operator
       Assert.assertEquals("first", stream.getSource().getOperatorMeta().getName());
       Assert.assertTrue(1 == stream.getSinks().size());
-      Assert.assertEquals("second", stream.getSinks().get(0).getOperatorWrapper().getName());
+      Collection<InputPortMeta> portMetaList = stream.getSinks();
+      Assert.assertEquals(1, portMetaList.size());
+      for (InputPortMeta inputPortMeta : portMetaList) {
+        Assert.assertEquals("second", inputPortMeta.getOperatorMeta().getName());
+      }
```
ajaygit158-zz: stream.getSinks() was earlier returning List<InputPortMeta>. With 3.6.0, its returning Collection<InputPortMeta>

library/pom.xml
```diff
@@ -366,6 +366,189 @@
...
+          <artifactId>log4j</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>commons-collections</groupId>
+          <artifactId>commons-collections</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>javax.servlet</groupId>
```
ajaygit158-zz, line 383: Certain tests were failing/stuck due to ClassNotFound error.
java.lang.NoClassDefFoundError: javax/servlet/AsyncContext
This class is available in above dependency version > 3

contrib/src/test/java/com/datatorrent/contrib/hbase/HBaseTransactionalPutOperatorTest.java
vrozov: I do not question the usage of `OperatorContextTestHelper.mockOperatorContext`. My question is regarding `ACTIVATION_WINDOW_ID` and `APPLICATION_ID` attributes that were not part of the OperatorContext before.

library/src/main/java/org/apache/apex/malhar/lib/dedup/TimeBasedDedupOperator.java
```diff
@@ -186,7 +186,7 @@ public void deactivate()
   @Override
   protected void putManagedState(Object tuple)
   {
-    ((ManagedTimeUnifiedStateImpl)managedState).put(getTime(tuple), getKey(tuple), new Slice(new byte[0]));
+    ((ManagedTimeUnifiedStateImpl)managedState).put(getTime(tuple), getKey(tuple), new Slice(null, 0, 0));
```
vrozov: The change is not required for the upgrade. Open another JIRA and PR to review Slice usage (including how empty Slices are constructed).

stream/src/test/java/org/apache/apex/malhar/stream/api/impl/ApexStreamImplTest.java
```diff
@@ -80,7 +81,11 @@ public void testAddOperator()
       // Assert the stream is from first operator to second operator
       Assert.assertEquals("first", stream.getSource().getOperatorMeta().getName());
       Assert.assertTrue(1 == stream.getSinks().size());
-      Assert.assertEquals("second", stream.getSinks().get(0).getOperatorWrapper().getName());
+      Collection<InputPortMeta> portMetaList = stream.getSinks();
+      Assert.assertEquals(1, portMetaList.size());
```
vrozov: Remove assert on line 83, it checks for the same condition. Avoid 'List' in the name, it is a Collection.

vrozov:
Please see http://apex.apache.org/contributing.html. I would also prefer to include JIRA not only into PR, but also into commit message.

library/src/main/java/org/apache/apex/malhar/lib/dedup/BoundedDedupOperator.java
```diff
@@ -156,8 +156,7 @@ protected Slice getKey(Object tuple)
   protected void putManagedState(Object tuple)
   {
     Slice key = getKey(tuple);
-    ((ManagedTimeStateImpl)managedState).put(getBucketId(key), DEFAULT_CONSTANT_TIME,
-        key, new Slice(new byte[0]));
+    ((ManagedTimeStateImpl)managedState).put(getBucketId(key), DEFAULT_CONSTANT_TIME, key, new Slice(null, 0, 0));
```
ajaygit158-zz: The changes related to Slice are part of different commit (APEXMALHAR-2492) which need to be merged before this PR.

vrozov:
Unit tests fail due to unhandled exceptions in the DefaultEventLoop when Buffer Server terminates that lead to DefaultEventLoop thread termination. When StramLocalCluster starts the next time, the DefaultEventLoop thread should be started, but it is not due to an incorrect reference count. Fix for the incorrect reference count is available https://github.com/DataTorrent/Netlet/pull/62.

vrozov:
Add netlet-1.3.1 (just released) as the explicit dependency with test scope.

library/pom.xml
vrozov: The exclusion list is too large, exclude all dependencies and add missing one explicitly.

library/pom.xml
```diff
@@ -366,6 +366,189 @@
...
+      <artifactId>javax.servlet-api</artifactId>
+      <version>3.0.1</version>
```
vrozov: Why 3.0.1 and not 3.0.0? What is the latest?

library/src/main/java/org/apache/apex/malhar/lib/dedup/BoundedDedupOperator.java
vrozov: Why APEXMALHAR-2492 needs to be merged before?

ajaygit158-zz: Apart from the correct usage of Slice, APEXMALHAR-2492 is also the set of changes required to ensure support for newer netlet library (>= 1.3.0) which is used by apex-core 3.6.0.
The above change is the minimum required change since build with apex-core 3.6.0/netlet 1.3.0 fails without this change.

vrozov: I don't see why Slice(new byte[0]) would not work with the 1.3.x version of netlet. The constructor that takes byte array as a single argument still should work.

library/pom.xml
```diff
@@ -366,6 +366,36 @@
...
+      <groupId>org.apache.avro</groupId>
+      <artifactId>avro</artifactId>
+      <version>1.7.4</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.datatorrent</groupId>
+      <artifactId>netlet</artifactId>
+      <version>1.3.1</version>
+    </dependency>
```
vrozov: test scope

library/src/main/java/org/apache/apex/malhar/lib/dedup/BoundedDedupOperator.java
ajaygit158-zz: The test doesnot fail at this point. It fails for the test DeduperTimeBasedPOJOImplTest. 
This is because we try to create a BufferSlice (extends Slice) object in this code flow
`public BufferSlice(byte[] array, int offset, int length)`
`{`
`     super(array, offset, length);`
`}`
The above constructor call fails with IllegalArgumentException

library/pom.xml
ajaygit158-zz: Compilation fails with only test scope. We have files like ManagedStateImpl and BufferSlice which refer netlet.

library/src/main/java/org/apache/apex/malhar/lib/dedup/BoundedDedupOperator.java
vrozov: If it does not fail, why the change is required?

ajaygit158-zz:
Jenkins build is stuck in some loop for test POJOPartitionJoinOperatorTest. The following log appears continuously.
```
2017-05-24 21:17:21,686 [container-3] DEBUG stram.StreamingContainerManager processOperatorDeployStatus - heartbeat PTOperator[id=8,name=Join.outputPort#unifier,state=PENDING_UNDEPLOY] PENDING_UNDEPLOY/ACTIVE container-3
2017-05-24 21:17:21,686 [container-3] DEBUG stram.StreamingContainerManager getHeartbeatResponse - PTContainer[id=7(container-3),state=ACTIVE] deferred requests due to concurrent plan change.
```

library/src/main/java/org/apache/apex/malhar/lib/dedup/BoundedDedupOperator.java
ajaygit158-zz: I guess I am not able to put forward my point correctly. 
Below is the trace for the failing test which indicates that the Slice created above is causing the test to fail.
```
java.lang.IllegalArgumentException: Invalid slice: offset=0, length=0 array.length=0
	at com.datatorrent.netlet.util.Slice.<init>(Slice.java:43)
	at org.apache.apex.malhar.lib.utils.serde.BufferSlice.<init>(BufferSlice.java:48)
	at org.apache.apex.malhar.lib.utils.serde.BufferSlice.<init>(BufferSlice.java:58)
	at org.apache.apex.malhar.lib.utils.serde.SliceUtils.toBufferSlice(SliceUtils.java:111)
	at org.apache.apex.malhar.lib.state.managed.Bucket$DefaultBucket.put(Bucket.java:421)
	at org.apache.apex.malhar.lib.state.managed.AbstractManagedStateImpl.putInBucket(AbstractManagedStateImpl.java:286)
	at org.apache.apex.malhar.lib.state.managed.ManagedTimeUnifiedStateImpl.put(ManagedTimeUnifiedStateImpl.java:72)
	at org.apache.apex.malhar.lib.dedup.TimeBasedDedupOperator.putManagedState(TimeBasedDedupOperator.java:189)
	at org.apache.apex.malhar.lib.dedup.AbstractDeduper.processAuxiliary(AbstractDeduper.java:316)
	at org.apache.apex.malhar.lib.dedup.AbstractDeduper.handleIdleTime(AbstractDeduper.java:293)
	at org.apache.apex.malhar.lib.dedup.DeduperTimeBasedPOJOImplTest.testDedup(DeduperTimeBasedPOJOImplTest.java:98)
```

library/pom.xml
vrozov: provided

library/src/main/java/org/apache/apex/malhar/lib/dedup/BoundedDedupOperator.java
vrozov: I see.

library/pom.xml
ajaygit158-zz: Updated scope to provided

ajaygit158-zz:
@vrozov The travis build failed and logs show a netlet exception. The test failing is POJOPartitionJoinOperatorTest. In Jenkins also, this test was stuck.
https://travis-ci.org/apache/apex-malhar/builds/235913509#L4454
Can you have a look?

vrozov:
@ajaygit158 It is not a netlet exception. It is an IOException that is raised when downstream operators close connections to the upstream buffer server. By itself should not cause a problem or a test failure.

ajaygit158-zz:
Seems like an intermittent issue. Build was successful on personal repository
https://travis-ci.org/ajaygit158/apex-malhar/builds/235913496
Closing and reopening PR to initiate rebuild.

PR #619, opened 2017-05-15
APEXMALHAR-2447 Notification from AbstractFileInputOperator when scanning directory

prasannapramod:
APEXMALHAR-2447 Added notification in AbstractFileInputOperator, when scanning the directory for the first time. Implementations can use this information in different ways such as notifying downstream operators with a control tuple that a new directory has been scanned.

@DT-Priyanka please see.


library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -25,6 +25,7 @@
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Serializable;
+
```
DT-Priyanka, line 28: you should avoid adding or removing blank lines, or in general avoid any formatting changes around the code you are touching.

PR #621, opened 2017-05-17
APEXMALHAR-2492 Correct usage of empty Slice in Malhar Library

ajaygit158-zz:
@vlad There is another piece of code where there is possibility of creating a wrong empty Slice. 
https://github.com/apache/apex-malhar/blob/master/library/src/test/java/org/apache/apex/malhar/lib/wal/FileSystemWALTest.java#L113
Should we change the getRandomSlice() implementation if len = 0?

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/dedup/BoundedDedupOperator.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/dedup/BoundedDedupOperator.java
@@ -156,8 +156,7 @@ public class BoundedDedupOperator extends AbstractDeduper<Object>
   protected void putManagedState(Object tuple)
   {
     Slice key = getKey(tuple);
-    ((ManagedTimeStateImpl)managedState).put(getBucketId(key), DEFAULT_CONSTANT_TIME,
-        key, new Slice(new byte[0]));
+    ((ManagedTimeStateImpl)managedState).put(getBucketId(key), DEFAULT_CONSTANT_TIME, key, new Slice(null, 0, 0));
   }
 
   protected int getBucketId(Slice key)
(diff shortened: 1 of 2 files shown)


ajaygit158-zz:
Travis build was successful in personal git repository. 
https://travis-ci.org/ajaygit158/apex-malhar/builds/233114232
Jenkins build failed with "Error creating local cluster"
Closing and reopening PR.

library/src/main/java/org/apache/apex/malhar/lib/utils/serde/BufferSlice.java
```diff
@@ -34,7 +32,7 @@
 public class BufferSlice extends Slice
 {
   private static final long serialVersionUID = -471209532589983329L;
-  public static final BufferSlice EMPTY_SLICE = new BufferSlice(ArrayUtils.EMPTY_BYTE_ARRAY);
+  public static final BufferSlice EMPTY_SLICE = new BufferSlice(null, 0, 0);
```
ajaygit158-zz: This specific change causes SpillableMapImplTest to fail with null pointer exception (null buffer). This null buffer is passed to an append method in Tfile.java (belongs to hadoop-commons)

vrozov:
Is BufferSlice still necessary?

ajaygit158-zz:
BufferSlice code mentions the reason for creating BufferSlice 
`com.datatorrent.netlet.util.Slice has problem with the hashCode(), so override here`
I see that Slice in netlet 1.3.0 the same hash function. I will mark this as deprecated and replace all usages with Slice.
Also, point to note is that if we replace all usages of BufferSlice with Slice, this PR should go only after apex-core upgrade to 3.6.0

PR #627, opened 2017-05-28
APEXMALHAR-2504 Allow for customization of TwitterSampleInput ConfigurationBuilder

tweise:
@amberarrow please have a look.

### DIFF
--- a/contrib/src/main/java/com/datatorrent/contrib/twitter/TwitterSampleInput.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/twitter/TwitterSampleInput.java
@@ -56,19 +56,19 @@ public class TwitterSampleInput implements InputOperator, ActivationListener<Ope
   /**
    * This is the output port on which the twitter status information is emitted.
    */
-  public final transient DefaultOutputPort<Status> status = new DefaultOutputPort<Status>();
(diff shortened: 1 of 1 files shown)


contrib/src/main/java/com/datatorrent/contrib/twitter/TwitterSampleInput.java
```diff
@@ -204,6 +205,13 @@ public void onException(Exception ex)
     }
   }
 
+  /**
+   * Allow derived classes to customize the configuration
+   */
+  protected void setupConfigurationBuilder(ConfigurationBuilder cb)
+  {
```
amberarrow: Would be better to move the configuration code here from setup(); derived classes would then have the option to invoke or ignore the superclass method.

amberarrow:
Ready to merge once the comment is addressed.

contrib/src/main/java/com/datatorrent/contrib/twitter/TwitterSampleInput.java
tweise: Agreed, also looks much cleaner.

tweise:
@amberarrow PR updated

amberarrow:
Good to go.

PR #628, opened 2017-05-28
APEXMALHAR-2502 Fix Kudu output operator for some minor fixes

ghost:


ghost:
@tweise / @amberarrow  : Could anyone of you please review this minor change.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/BaseKuduOutputOperator.java
```diff
@@ -65,11 +67,27 @@
 
   public BaseKuduOutputOperator() throws IOException, ClassNotFoundException
   {
+    windowDataManager = new FSWindowDataManager();
+    initConnectionBuilderProperties(DEFAULT_CONNECTION_PROPS_FILE_NAME);
+  }
+
+  public BaseKuduOutputOperator(WindowDataManager windowManager, String configFileInClasspath)
```
tweise: Instead of adding specialized constructors, how about adding setters for the individual properties?

amberarrow:
Need a matching JIRA and the commit message should have the JIRA number as a prefix.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/BaseKuduOutputOperator.java
ghost: There was already a getter/setter in the abstract class for managing the window manager.  The default implementation extends from this abstract class.  I am not sure if this change is worth it or an overkill.

ghost:
@amberarrow   - Updated the branch with the correct JIRA number in the commit message.

contrib/src/main/java/org/apache/apex/malhar/contrib/kudu/BaseKuduOutputOperator.java
tweise: Additional constructor looks like overkill when same can be accomplished with the setter. I will wait for you to rebase and then merge this.

ghost:
@tweise  - Rebased to the latest. I also removed the extra constructors and added a safety check for timestamp propagation. Could you please review and merge ?

PR #629, opened 2017-05-29
APEXMALHAR-2505 retain original field order in SnapshotSchema

tweise:
@amberarrow please review


amberarrow:
This change itself looks OK and can be merged; why is the Travis build failing though ?

PR #630, opened 2017-05-30
APEXMALHAR-2451 Batch support for File I/O operators

ajaygit158-zz:
@bhupeshchawda Kindly Review

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
@@ -124,6 +124,34 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
   protected transient long skipCount = 0;
   private transient OperatorContext context;
 
+  /**
+   * The time between sending batch tuples. Default value is 1 hour. A batch
+   * interval could consist of multiple scan intervals
+   */
+  private long batchIntervalMillis = 60 * 60 * 1000; //1 hour
+
+  /**
+   * By setting this property to true, a single scan will be performed every
+   * batchIntervalMillis. All files made available in this scan period will form
+   * a part of same batch. For each batch, start batch and end batch control
+   * tuples will be emitted before and after the batch respectively. Note that
+   * batch tuple processing currently works only in single operator instance
+   * scenarios i.e without partitions. However, batch processing with parallel
+   * partitions will work
+   */
+  private boolean emitBatchTuples = false;
+
+  private transient boolean fileClosed = false;
+  private transient boolean startBatchEmittedInCurrentWindow = false;
+  private transient boolean endBatchEmittedInCurrentWindow = false;
+
+  /**
+   * End Batch can only be emitted if start batch was emitted earlier. This is
+   * set to true only after a start batch is emitted
+   */
+  private boolean canEmitEndBatch = false;
+
+
   private final BasicCounters<MutableLong> fileCounters = new BasicCounters<MutableLong>(MutableLong.class);
   protected MutableLong globalNumberOfFailures = new MutableLong();
   protected MutableLong localNumberOfFailures = new MutableLong();
@@ -137,6 +165,7 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
   private WindowDataManager windowDataManager = new WindowDataManager.NoopWindowDataManager();
   protected transient long currentWindowId;
   protected final transient LinkedList<RecoveryEntry> currentWindowRecoveryState = Lists.newLinkedList();
+  protected final transient WindowRecoveryEntry currentWindowState = new WindowRecoveryEntry();
   protected int operatorId; //needed in partitioning
 
   /**
@@ -535,6 +564,9 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
     if (windowId <= windowDataManager.getLargestCompletedWindow()) {
       replay(windowId);
     }
+    fileClosed = false;
+    startBatchEmittedInCurrentWindow = false;
+    endBatchEmittedInCurrentWindow = false;
   }
 
   @Override
@@ -542,12 +574,12 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
   {
     if (currentWindowId > windowDataManager.getLargestCompletedWindow()) {
       try {
-        windowDataManager.save(currentWindowRecoveryState, currentWindowId);
+        windowDataManager.save(currentWindowState, currentWindowId);
       } catch (IOException e) {
         throw new RuntimeException("saving recovery", e);
       }
     }
-    currentWindowRecoveryState.clear();
+    currentWindowState.clear();
     if (context != null) {
       pendingFileCount.setValue(pendingFiles.size() + failedFiles.size() + unfinishedFiles.size());
 
@@ -569,9 +601,18 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
       Map<Integer, Object> recoveryDataPerOperator = windowDataManager.retrieveAllPartitions(windowId);
 
       for (Object recovery : recoveryDataPerOperator.values()) {
-        @SuppressWarnings("unchecked")
-        LinkedList<RecoveryEntry> recoveryData = (LinkedList<RecoveryEntry>)recovery;
 
+        WindowRecoveryEntry windowRecoveryEntry = (WindowRecoveryEntry)recovery;
+        List<RecoveryEntry> recoveryData = windowRecoveryEntry.fileRecoveryInfo;
+
+        if (windowRecoveryEntry.startBatchEmitted) {
+          emitStartBatchControlTuple();
+          return;
+        }
+        if (windowRecoveryEntry.endBatchEmitted) {
+          emitEndBatchControlTuple();
+          return;
+        }
         for (RecoveryEntry recoveryEntry : recoveryData) {
           if (scanner.acceptFile(recoveryEntry.file)) {
             //The operator may have continued processing the same file in multiple windows.
@@ -583,11 +624,13 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
               }
               processedFiles.add(recoveryEntry.file);
               //removing the file from failed and unfinished queues and pending set
+              boolean isFailedFile = false;
               Iterator<FailedFile> failedFileIterator = failedFiles.iterator();
               while (failedFileIterator.hasNext()) {
                 FailedFile ff = failedFileIterator.next();
                 if (ff.path.equals(recoveryEntry.file) && ff.offset == recoveryEntry.startOffset) {
                   failedFileIterator.remove();
+                  isFailedFile = true;
                   break;
                 }
               }
@@ -604,7 +647,12 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
                 pendingFiles.remove(recoveryEntry.file);
               }
               inputStream = retryFailedFile(new FailedFile(recoveryEntry.file, recoveryEntry.startOffset));
-
+              if (skipCount == 0) {
+                emitStartFileControlTuple(false);
+              }
+              if (isFailedFile) {
+                emitStartFileControlTuple(isFailedFile);
+              }
               while (--skipCount >= 0) {
                 readEntity();
               }
@@ -614,6 +662,7 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
                 emit(line);
               }
               if (recoveryEntry.fileClosed) {
+                emitEndFileControlTuple(false);
                 closeFile(inputStream);
               }
             } else {
@@ -623,6 +672,7 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
                 emit(line);
               }
               if (recoveryEntry.fileClosed) {
+                emitEndFileControlTuple(false);
                 closeFile(inputStream);
               }
             }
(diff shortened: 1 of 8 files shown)


library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
```diff
@@ -871,6 +878,15 @@ protected void processTuple(INPUT tuple)
   }
 
   /**
+   * @param controlTuple
+   * @return
+   */
+  public boolean processControlTuple(ControlTuple controlTuple)
```
bhupeshchawda, line 884: Can we keep this abstract? That way all implementations need to implement it.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -124,6 +124,34 @@
   protected transient long skipCount = 0;
   private transient OperatorContext context;
 
+  /**
+   * The time between sending batch tuples. Default value is 1 hour. A batch
+   * interval could consist of multiple scan intervals
+   */
+  private long batchIntervalMillis = 60 * 60 * 1000; //1 hour
```
bhupeshchawda, line 131: `static final BATCH_INTERVAL_MILLIS`?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -569,9 +601,18 @@ protected void replay(long windowId)
...
 
       for (Object recovery : recoveryDataPerOperator.values()) {
-        @SuppressWarnings("unchecked")
-        LinkedList<RecoveryEntry> recoveryData = (LinkedList<RecoveryEntry>)recovery;
 
+        WindowRecoveryEntry windowRecoveryEntry = (WindowRecoveryEntry)recovery;
+        List<RecoveryEntry> recoveryData = windowRecoveryEntry.fileRecoveryInfo;
+
+        if (windowRecoveryEntry.startBatchEmitted) {
+          emitStartBatchControlTuple();
```
bhupeshchawda, line 609: It seems something is wrong here. Do we want to emit multiple start and end batch tuples in the replay code?
This might mess up the downstream operators. Remember that all this is done in a single window.
The right thing to do here would be to do a replay but also at the same speed. We do have the files and offsets that are completed in each window. Perhaps we can keep repeating this until all data is replayed. Note we do not increment the windowDataManager state during this time. Once all replayed data is emitted, then in the next window we can start with reading new data.
@PramodSSImmaneni @vrozov thoughts?

bhupeshchawda (commented):
@ajaygit158 I have added a few adhoc comments. I went through the changes but still missing some things. Can you document the behavior of the operator as to which use cases are handled and which properties a user can set to achieve a particular use case? It will help other reviewers as well.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
ajaygit158-zz: Shouldn't this be configurable?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileOutputOperator.java
ajaygit158-zz: will it be fine to add a blank operation to all subclasses?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
bhupeshchawda: yes, I mean add a static final constant and use it for your variable..

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
ajaygit158-zz: @bhupeshchawda for every window that is replayed, replay() performs operations which were done only in that window earlier. Hence, replay ensures idempotency here. You can have a look at the tests added for batch.
Let me know if my understanding regarding your concern is wrong.

ajaygit158-zz:
Below is the behavior for supporting batch in FileInputOperator
Batches in fileInput operator are currently optional. To enable batch support, we set emitBatchTuples to true. By default, one batch is a set of files obtained after scanning the input directory once. The time interval between two consecutive batches (and hence two consecutive scans) is controlled via "batchIntervalMillis"

For explaining the code changes, I will add review comments.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -137,6 +165,7 @@
   private WindowDataManager windowDataManager = new WindowDataManager.NoopWindowDataManager();
   protected transient long currentWindowId;
   protected final transient LinkedList<RecoveryEntry> currentWindowRecoveryState = Lists.newLinkedList();
```
ajaygit158-zz, line 167: currentWindowRecoveryState is no more useful. It is replaced with WindowRecoveryEntry instance declared below. Removing this was causing semanticVersioning issues. Apart from the state of files which was stored earlier, we also store whether batch tuples were emitted in that window.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -604,7 +647,12 @@ protected void replay(long windowId)
                 pendingFiles.remove(recoveryEntry.file);
               }
               inputStream = retryFailedFile(new FailedFile(recoveryEntry.file, recoveryEntry.startOffset));
-
+              if (skipCount == 0) {
```
ajaygit158-zz, line 650: If the file processing failed just after opening the file without reading any records, we emit startFileControlTuple. Since we are reading from beginning, it indicates file processing had not failed earlier. This information is conveyed via the parameter (false) passed.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -604,7 +647,12 @@ protected void replay(long windowId)
...
+                emitStartFileControlTuple(false);
+              }
+              if (isFailedFile) {
+                emitStartFileControlTuple(isFailedFile);
+              }
```
ajaygit158-zz, line 655: If we are processing a file whose processing had failed earlier, we will emit a start file CT with a status (true) indicating that file processing had failed earlier.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -569,9 +601,18 @@ protected void replay(long windowId)
...
-        LinkedList<RecoveryEntry> recoveryData = (LinkedList<RecoveryEntry>)recovery;
 
+        WindowRecoveryEntry windowRecoveryEntry = (WindowRecoveryEntry)recovery;
+        List<RecoveryEntry> recoveryData = windowRecoveryEntry.fileRecoveryInfo;
+
+        if (windowRecoveryEntry.startBatchEmitted) {
+          emitStartBatchControlTuple();
+          return;
+        }
+        if (windowRecoveryEntry.endBatchEmitted) {
```
ajaygit158-zz, line 612: Note that one window will emit only one among StartBatch and EndBatch control tuples.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -614,6 +662,7 @@ protected void replay(long windowId)
                 emit(line);
               }
               if (recoveryEntry.fileClosed) {
+                emitEndFileControlTuple(false);
```
ajaygit158-zz, line 665: If file was closed in the window, we emit endBatch CT. false indicates that entire file processing completed. (A file which encounters failure midway while processing will emit control tuple with status true)

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -658,14 +713,17 @@ public void emitTuples()
           }
         } else if (!unfinishedFiles.isEmpty()) {
           retryFailedFile(unfinishedFiles.poll());
+          //For an unfinished file, we wont emit start file control tuple since it was already emitted before partitioning
```
ajaygit158-zz, line 716: Batch operations in this operator cannot support partitioning.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -701,11 +759,13 @@ public void emitTuples()
           }
         }
       } catch (IOException e) {
+        //An exception indicating file I/O encountered exception
+        emitEndFileControlTuple(true);
```
ajaygit158-zz, line 763: If file processing failed, indicate to downstream operators that entire file is not yet processed.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -732,7 +792,34 @@ private void checkVisitedDirectory(Path path)
...
+    if (emitBatchTuples) {
+      if (canEmitEndBatch) {
+        /*
+         * Emit EndBatch and mark canEmitEndBatch to false to avoid multiple emits of End Batch
+         */
+        emitEndBatchControlTuple();
+        endBatchEmittedInCurrentWindow = true;
+        currentWindowState.endBatchEmitted = true;
+        canEmitEndBatch = false;
+      }
```
ajaygit158-zz, line 805: A scan is tried again only after all files are processed. Also, canEmitEndBatch is true only if start batch is emitted earlier. 
Need to add a "return" statement here to ensure startBatch also does not get emitted in same window.

ajaygit158-zz (commented):
Added comments for understanding the changes made

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
bhupeshchawda: I see. Let me look at the tests..

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
bhupeshchawda: So do both start batch and end batch tuples have a end window delivery mechanism?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
bhupeshchawda: So do both start batch and end batch tuples have a end window delivery mechanism?

bhupeshchawda:
Where do I specify when to stop and end my application? We should have something for shutting down the app after one or more batches.

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
ajaygit158-zz: No start batch tuples don't have end window delivery mechanism. This condition was required to ensure replay of windows works fine.

bhupeshchawda: so, you mean a batch will at least span two windows? start in first and end in second?

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
tweise, line 131: Do you have a use case for this? I would think that batch boundary is determined by input, not by processing time. Where are the design aspects documented? The JIRA doesn't have any information.

bhupeshchawda: In this case the use case is processing time batches, for example, an hourly scan through the directory is part of a single batch
@ajaygit158 please add the use cases to the jira

tweise: Typically you have hourly input files or directories. Also, even if there was a real use case for this, it doesn't seem implemented correctly. There is no abstraction for triggering the batch tuple and it relies on system clock.

bhupeshchawda: Hourly batch is just an example, you can have even smaller intervals as batches.
Regarding the abstraction for a batch tuple, what do you suggest? We have predefined types of batch control tuples. For a start, we wanted to have file control tuples - emitted when a file starts and ends - which is a useful event for a downstream operator.
With regards to batch specification here, there is just a single use case for scanning every n time units and reading whatever is present as a batch.

tweise: The point is that typically the batch boundary is determined based on the input, not on system clock. What is lacking here is the necessary abstraction for triggering a batch tuple.

bhupeshchawda: Yes Thomas, we discussed this. Ajay would be closing this PR and adding the design / discussion points on the JIRA soon.

PR #631, opened 2017-06-01
APEXMALHAR-2366 #resolve #comment Apply BloomFilter to Bucket, use internal BloomFilter

pramodin:
@bhupeshchawda please see, this is to finish up the work started in https://github.com/apache/apex-malhar/pull/521

### DIFF
--- a/library/src/test/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilterTest.java
+++ b/library/src/test/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilterTest.java
@@ -1,0 +1,191 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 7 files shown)


bhupeshchawda (approved):

bhupeshchawda:
Will merge this today.

bhupeshchawda:
Can you please add the JIRA id to the second commit message?

library/src/test/java/org/apache/apex/malhar/lib/state/managed/SliceBloomFilterTest.java
```diff
@@ -0,0 +1,191 @@
...
+      public void populateDAG(DAG dag, Configuration conf)
+      {
+      }
+    };
+
+    lma.prepareDAG(app, conf);
+
+    // Create local cluster
+    final LocalMode.Controller lc = lma.getController();
+    lc.run(3000);
```
tweise, line 187: @bhupeshchawda can you please look at this test and remove the timing dependency?

bhupeshchawda: This test was supposed to just test serialization, not a timing dependency as such. 
But perhaps the test is not sufficient. Will look into it.

PR #632, opened 2017-06-02
APEXMALHAR-2488 APEXMALHAR-2503 Deprecated inner join operator and co…

bhupeshchawda:
…mmented out the tests

@tweise @chaithu14 Please see comments on https://issues.apache.org/jira/browse/APEXMALHAR-2488 and https://issues.apache.org/jira/browse/APEXMALHAR-2503


tweise:
Tests, if they are not removed, can be annotated it @ Ignore

However, given the operator is actually broken, why would we keep it around?

bhupeshchawda:
The operator is broken, but could be potentially used with minor hacks to the managed state code. If at all someone is using it, then removing this would cause problems for them.
I'll mark tests as @Ignore.

tweise:
I think this operator being used in a real application is a long shot, but I'm OK with keeping it deprecated for now. What are you planning to do about the innerjoin example?

bhupeshchawda:
I think we can remove the inner join example; I'll include it in the same PR.
Also, we need an example for doing inner join using the windowed merge operator. Will create a JIRA for that as well.

PR #633, opened 2017-06-09
APEXMALHAR-2506 Added the monitoring to Kafka Consumer threads, if on…

sandeshh:
…e of them dies then the Kafka operator is killed so that it can recover from the proper state.

@PramodSSImmaneni please review.

### DIFF
--- a/kafka/kafka-common/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaInputOperator.java
+++ b/kafka/kafka-common/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaInputOperator.java
@@ -340,6 +340,10 @@ public abstract class AbstractKafkaInputOperator implements InputOperator,
 
   protected void processConsumerError()
   {
+    if (consumerWrapper.hasAnyKafkaReaderThreadDied()) {
+      throw new RuntimeException("Kafka Consumer threads exited.");
+    }
+
     if (consumerError != null) {
       logger.error("Error in consumer, terminating");
       throw Throwables.propagate(consumerError);
--- a/kafka/kafka-common/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
+++ b/kafka/kafka-common/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
@@ -21,6 +21,7 @@ package org.apache.apex.malhar.kafka;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
@@ -32,6 +33,7 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.slf4j.Logger;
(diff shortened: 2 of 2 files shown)


kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
```diff
@@ -90,6 +92,19 @@
 
   private boolean waitForReplay = false;
 
+  final List<Future<?>> kafkaConsumreThreads = new ArrayList<>();
```
vrozov: fix typo in kafkaConsumerThreads

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
```diff
@@ -90,6 +92,19 @@
...
+
+  public boolean areKafkaThreadsAreRunning()
+  {
+    for (Future<?> future : kafkaConsumreThreads) {
```
vrozov: Not directly related to this PR, but it does not seem right to use execution service for never ending tasks. It may lead to resource starvation.

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
sandeshh: Done.

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
sandeshh: We can have a separate discussion about. How do you want to it?

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
```diff
@@ -90,6 +92,19 @@
 
   private boolean waitForReplay = false;
 
+  private final List<Future<?>> kafkaConsumerThreads = new ArrayList<>();
+
+  public boolean doesAnyKafkaReaderThreadDied()
```
tweise: does -> has

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaInputOperator.java
```diff
@@ -255,6 +255,10 @@ public void emitTuples()
       }
     }
     emitCount += count;
+
+    if (!consumerWrapper.doesAnyKafkaReaderThreadDied()) {
```
tweise: This may not be the correct handling for all situations. What happens, for example, on partition or lead broker change? Or a transient connection issue? Downstream resets may be very costly and should only be triggered when really necessary.

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
sandeshh: Done.

kafka/src/main/java/org/apache/apex/malhar/kafka/AbstractKafkaInputOperator.java
sandeshh: If any of these thread dies because of the exception, how to handle that cleanly?

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
```diff
@@ -90,6 +92,19 @@
 
   private boolean waitForReplay = false;
 
+  private final List<Future<?>> kafkaConsumerThreads = new ArrayList<>();
+
+  public boolean hasAnyKafkaReaderThreadDied()
+  {
+    for (Future<?> future : kafkaConsumerThreads) {
+      if (future.isDone() || future.isCancelled()) {
+        return false;
```
tweise: You have flipped true/false and also need to fix the check when calling this method.

tweise:
Check if there are conditions under which the consumer thread exits that is not an error but expected (such as partition change).

kafka/src/main/java/org/apache/apex/malhar/kafka/KafkaConsumerWrapper.java
sandeshh: fixed it.

sandeshh:
There is no exit strategy for consumer thread, the assumption is that they are running forever.

sandeshh:
@tweise I have addressed all the review comments. Can you please review and merge the PR?

tweise:
What testing was done? (Please document in the JIRA.)

sandeshh:
@tweise I have documented my observation. This fix was done to address an intermittent issue.

PR #636, opened 2017-06-26
APEXMALHAR-2510 adding all matching keys to result for join accumulation

KapoorHitesh:
@chinmaykolhatkar @ShunxinLu please review.

### DIFF
--- a/library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
+++ b/library/src/main/java/org/apache/apex/malhar/lib/window/accumulation/AbstractPojoJoin.java
@@ -53,9 +53,9 @@ public abstract class AbstractPojoJoin<InputT1, InputT2>
   private transient Map<String,PojoUtils.Getter> gettersStream1;
   private transient Map<String,PojoUtils.Getter> gettersStream2;
   protected transient Map<String,PojoUtils.Setter> setters;
-  protected transient Map<String, KeyValPair<STREAM, String>> outputToInputMap;
-  protected transient String[] leftKeys;
-  protected transient String[] rightKeys;
+  protected Map<String, KeyValPair<STREAM, String>> outputToInputMap;
+  protected String[] leftKeys;
+  protected String[] rightKeys;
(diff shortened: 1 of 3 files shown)


ShunxinLu:
I think this is good. Thanks for the quick response.

chinmaykolhatkar:
@Hitesh-Scorpio Can you check why the builds are failing?

KapoorHitesh:
@chinmaykolhatkar  In Jenkins I am getting the build success message but still it is failing with "Warning: this build has no associated authentication, so build permissions may be lacking, and downstream projects which cannot even be seen by an anonymous user will be silently skipped
Finished: UNSTABLE".

Travis is passing here as well as on my local machine. Should I close and reopen the PR?

--Hitesh

ShunxinLu:
@Hitesh-Scorpio I tested the fix in an application. Even though the previous problem we discussed is fixed, the app still seems to be still unable to behave correctly. I dug deeper into the new problem and it looks like that Kryo creates couple more instances of join accumulation by calling its default constructor. And this wipes out all the keys (left, right) stored in the first place, so that when doing accumulation, the left and right keys are always empty, and therefore the inner join is not being performed correctly. 
Making the default constructor do nothing and changing the member fields of AbstractPojoJoin to static kind of fix the issue. But I wonder if that's the correct way to do this. Could you please take a look into this problem as well? Thanks!

KapoorHitesh:
@ShunxinLu  I think you can open a separate JIRA for that issue and we can discuss it separately.

ShunxinLu:
@Hitesh-Scorpio Agree. Will open a JIRA shortly. @chinmaykolhatkar I believe this PR is ready to merge.

chinmaykolhatkar:
@Hitesh-Scorpio @ShunxinLu  Thanks for resolution on this. I'll merge this PR in couple of days.

chinmaykolhatkar (approved):

vrozov:
@Hitesh-Scorpio @ShunxinLu @chinmaykolhatkar closed due to inactivity

PR #638, opened 2017-06-29
APEXMALHAR-2515 Operator maturity - HBase output operator Multi Table feature.

prasannapramod:
Implemented HBase output operator multi-table insertion feature.

@venkateshkottapalli @tushargosavi @sanjaypujare @PramodSSImmaneni please see.

### DIFF
--- a/contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseAppendOutputOperator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseAppendOutputOperator.java
@@ -20,15 +20,17 @@ package com.datatorrent.contrib.hbase;
 
 import java.io.IOException;
 
-import org.apache.hadoop.hbase.client.Append;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.hadoop.hbase.client.Append;
+import org.apache.hadoop.hbase.client.HTable;
+
 import com.datatorrent.netlet.util.DTThrowable;
-import com.datatorrent.lib.db.AbstractStoreOutputOperator;
 
 /**
- * A base implementation of a StoreOutputOperator operator that stores tuples in HBase columns and offers non-transactional append.&nbsp; Subclasses should provide implementation for appending operations. <br>
+ * A base implementation of a StoreOutputOperator operator that stores tuples in HBase columns and offers 
+ * non-transactional append.&nbsp; Subclasses should provide implementation for appending operations. <br>
  * <p>
  * <br>
  * This class provides a HBase output operator that can be used to store tuples
@@ -47,8 +49,7 @@ import com.datatorrent.lib.db.AbstractStoreOutputOperator;
  *            The tuple type
  * @since 1.0.2
  */
-public abstract class AbstractHBaseAppendOutputOperator<T>
-extends AbstractStoreOutputOperator<T, HBaseStore> {
+public abstract class AbstractHBaseAppendOutputOperator<T> extends AbstractHBaseOutputOperator<T> {
   private static final transient Logger logger = LoggerFactory
       .getLogger(AbstractHBaseAppendOutputOperator.class);
 
@@ -57,10 +58,10 @@ extends AbstractStoreOutputOperator<T, HBaseStore> {
   }
 
   @Override
-  public void processTuple(T tuple) {
+  public void processTuple(T tuple, HTable table) {
     Append append = operationAppend(tuple);
     try {
-      store.getTable().append(append);
+      table.append(append);
     } catch (IOException e) {
       logger.error("Could not append tuple", e);
       DTThrowable.rethrow(e);
--- a/contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseOutputOperator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseOutputOperator.java
@@ -1,0 +1,86 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.contrib.hbase;
+
+import org.apache.hadoop.hbase.client.HTable;
+
+import com.datatorrent.api.Operator;
+import com.datatorrent.lib.db.AbstractStoreOutputOperator;
+
+/**
+ * A base implementation of a HBase output operator that stores tuples in HBase and offers 
+ * non-transactional Insert.&nbsp; Subclasses should provide implementation for specific Hbase operations.
+ */
+public abstract class AbstractHBaseOutputOperator<T> extends AbstractStoreOutputOperator<T, HBaseStore> implements OutputAdapter.Driver<T>, Operator.CheckpointNotificationListener
+{
+  private transient OutputAdapter<T> outputAdapter;
+  
+  public AbstractHBaseOutputOperator()
+  {
+    outputAdapter = new OutputAdapter<T>(store, this);
+  }
+
+  @Override
+  public void processTuple(T tuple)
+  {
+    outputAdapter.processTuple(tuple);
+  }
+  
+  @Override
+  public abstract void processTuple(T tuple, HTable table);
+
+  /**
+   * Get the table name for tuple.
+   * 
+   * Implementations can override this method to return the table name where the tuple should be written to.
+   * Return null to write to default table
+   * @param tuple The tuple
+   * @return The table name
+   */
+  @Override
+  public String getTableName(T tuple)
+  {
+    return null;
+  }
+
+  @Override
+  public void errorTuple(T tuple)
+  {
+
+  }
+
+  @Override
+  public void beforeCheckpoint(long l)
+  {
+    outputAdapter.flushTuples();
+  }
+
+  @Override
+  public void checkpointed(long l)
+  {
+
+  }
+
+  @Override
+  public void committed(long l)
+  {
+
+  }
+  
+}
--- a/contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
@@ -19,21 +19,15 @@
 package com.datatorrent.contrib.hbase;
 
 import java.io.IOException;
-import java.io.InterruptedIOException;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
 
-import org.apache.hadoop.hbase.client.Append;
-import org.apache.hadoop.hbase.client.HTable;
-import org.apache.hadoop.hbase.client.RetriesExhaustedWithDetailsException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.hadoop.hbase.client.Append;
+import org.apache.hadoop.hbase.client.HTable;
+
 import com.datatorrent.api.Context.OperatorContext;
-import com.datatorrent.api.Operator.ProcessingMode;
 import com.datatorrent.netlet.util.DTThrowable;
-import com.datatorrent.lib.db.AbstractAggregateTransactionableStoreOutputOperator;
 
 /**
  * A base implementation of an AggregateTransactionableStoreOutputOperator operator that stores tuples in HBase columns and provides batch append.&nbsp; Subclasses should provide implementation for appending operations. <br>
@@ -46,8 +40,6 @@ import com.datatorrent.lib.db.AbstractAggregateTransactionableStoreOutputOperato
  * for the tuple in the table.<br>
  *
  * <br>
- * This class provides batch append where tuples are collected till the end
- * window and they are appended on end window
  *
  * Note that since HBase doesn't support transactions this store cannot
  * guarantee each tuple is written only once to HBase in case the operator is
@@ -63,15 +55,20 @@ import com.datatorrent.lib.db.AbstractAggregateTransactionableStoreOutputOperato
  *            The tuple type
  * @since 1.0.2
  */
-public abstract class AbstractHBaseWindowAppendOutputOperator<T> extends AbstractAggregateTransactionableStoreOutputOperator<T, HBaseWindowStore> {
+public abstract class AbstractHBaseWindowAppendOutputOperator<T> extends AbstractHBaseWindowOutputOperator<T> {
   private static final transient Logger logger = LoggerFactory.getLogger(AbstractHBaseWindowAppendOutputOperator.class);
-  private List<T> tuples;
   private transient ProcessingMode mode;
+
+  /**
+   * Processing mode being a common platform feature should be set as an attribute.
+   * Even if the property is set, it wouldn't effect how the platform would process that feature for the operator. */
+  @Deprecated
   public ProcessingMode getMode()
   {
     return mode;
   }
 
+  @Deprecated
   public void setMode(ProcessingMode mode)
   {
     this.mode = mode;
@@ -79,34 +76,17 @@ public abstract class AbstractHBaseWindowAppendOutputOperator<T> extends Abstrac
 
   public AbstractHBaseWindowAppendOutputOperator() {
     store = new HBaseWindowStore();
-    tuples = new ArrayList<T>();
   }
 
   @Override
-  public void storeAggregate() {
-    HTable table = store.getTable();
-    Iterator<T> it = tuples.iterator();
-    while (it.hasNext()) {
-      T t = it.next();
-      try {
-        Append append = operationAppend(t);
-        table.append(append);
-      } catch (IOException e) {
-        logger.error("Could not output tuple", e);
-        DTThrowable.rethrow(e);
-      }
-
-    }
+  public void processTuple(T tuple, HTable table) {
     try {
-      table.flushCommits();
-    } catch (RetriesExhaustedWithDetailsException e) {
-      logger.error("Could not output tuple", e);
-      DTThrowable.rethrow(e);
-    } catch (InterruptedIOException e) {
+      Append append = operationAppend(tuple);
+      table.append(append);
+    } catch (IOException e) {
       logger.error("Could not output tuple", e);
       DTThrowable.rethrow(e);
     }
-    tuples.clear();
   }
 
   /**
@@ -121,19 +101,11 @@ public abstract class AbstractHBaseWindowAppendOutputOperator<T> extends Abstrac
   public abstract Append operationAppend(T t);
 
   @Override
-  public void processTuple(T tuple) {
-    tuples.add(tuple);
-  }
-
-  @Override
   public void setup(OperatorContext context)
   {
-    mode=context.getValue(context.PROCESSING_MODE);
+    mode=context.getValue(OperatorContext.PROCESSING_MODE);
     if(mode==ProcessingMode.EXACTLY_ONCE){
       throw new RuntimeException("This operator only supports atmost once and atleast once processing modes");
-    }
-    if(mode==ProcessingMode.AT_MOST_ONCE){
-      tuples.clear();
     }
     super.setup(context);
   }
--- a/contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowOutputOperator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowOutputOperator.java
@@ -1,0 +1,96 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.datatorrent.contrib.hbase;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.hadoop.hbase.client.HTable;
+
+import com.datatorrent.api.Operator;
+import com.datatorrent.lib.db.AbstractPassThruTransactionableStoreOutputOperator;
+
+/**
+ * A base implementation of an AggregateTransactionableStoreOutputOperator that stores tuples in HBase
+ * and provides batch insert.&nbsp; Subclasses should provide implementation for specific Hbase operations. <br>
+ */
+public abstract class AbstractHBaseWindowOutputOperator<T> extends AbstractPassThruTransactionableStoreOutputOperator<T, HBaseWindowStore> implements OutputAdapter.Driver<T>,  Operator.CheckpointNotificationListener
+{
+  /**
+   * Both at checkpoint window and end window, flush the tuples as application window may not align with
+   * end window especially when it is more than one streaming window
+   */
+  
+  private static final Logger logger = LoggerFactory.getLogger(AbstractHBaseWindowOutputOperator.class);
+
+  private transient OutputAdapter<T> outputAdapter;
+
+  public AbstractHBaseWindowOutputOperator()
+  {
+    outputAdapter = new OutputAdapter<T>(store, this);
+  }
+
+  @Override
+  public void processTuple(T tuple)
+  {
+    outputAdapter.processTuple(tuple);
+  }
+
+  @Override
+  public abstract void processTuple(T tuple, HTable table);
+
+  /**
+   * Get the table name for tuple.
+   *
+   * Implementations can override this method to return the table name where the tuple should be written to.
+   * Return null to write to default table
+   * @param tuple The tuple
+   * @return The table name
+   */
+  @Override
+  public String getTableName(T tuple)
+  {
+    return null;
+  }
+
+  @Override
+  public void errorTuple(T tuple)
+  {
+
+  }
+
+  @Override
+  public void beforeCheckpoint(long l)
+  {
+    outputAdapter.flushTuples();
+  }
+
+  @Override
+  public void checkpointed(long l)
+  {
+
+  }
+
+  @Override
+  public void committed(long l)
+  {
+
+  }
+
+}
--- a/contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowPutOutputOperator.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowPutOutputOperator.java
@@ -19,21 +19,15 @@
 package com.datatorrent.contrib.hbase;
 
 import java.io.IOException;
-import java.io.InterruptedIOException;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
 
-
-import org.apache.hadoop.hbase.client.HTable;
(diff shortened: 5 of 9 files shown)


contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseAppendOutputOperator.java
```diff
@@ -20,15 +20,17 @@
 
 import java.io.IOException;
 
-import org.apache.hadoop.hbase.client.Append;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.hadoop.hbase.client.Append;
```
sanjaypujare, line 26: Why did the import order change? Is the new order correct as per the import-order guideline or was the old one better?

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseAppendOutputOperator.java
```diff
@@ -57,10 +58,10 @@ public AbstractHBaseAppendOutputOperator() {
   }
```
sanjaypujare: store is not used any more in the new version. Is it necessary to create it here? If a subclass needs it, it can create it, right?

contrib/src/main/java/com/datatorrent/contrib/hbase/OutputAdapter.java
```diff
@@ -0,0 +1,77 @@
...
+ * Delegating handling to an adapter so that the same adapter can be reused 
+ * from different inheritance hierarchies.
+ * Created by lakshmi on 6/27/17.
+ */
+public class OutputAdapter<T>
+{
+  private static final Logger logger = LoggerFactory.getLogger(OutputAdapter.class);
+  
+  HBaseStore store;
+  OutputAdapter.Driver driver;
```
sanjaypujare: You don't need the qualifier OutputDriver. since Driver is local to this class.

sanjaypujare: Also can we get rid of warnings like the below:

OutputAdapter.Driver is a raw type. References to generic type OutputAdapter<T>.Driver<T> should be parameterized

contrib/src/main/java/com/datatorrent/contrib/hbase/OutputAdapter.java
```diff
@@ -0,0 +1,77 @@
...
+  public void flushTuples()
+  {
+    try {
+      store.flushTables();
+    } catch (InterruptedIOException | RetriesExhaustedWithDetailsException e) {
+      throw Throwables.propagate(e);
+    }
+  }
+
+  interface Driver<T>
```
sanjaypujare: The Driver always contains the store since Driver is implemented by most HBase operators you have modified. Why can't we make this more modular/OO by encapsulating the store also in the Driver? In this case Driver becomes an abstract class as follows:

```
abstract class Driver<T>
  {
      HBaseStore store;
      Driver(HBaseStore store) {
         this.store = store;
     }    
     void processTuple(T tuple) {
        String tableName = getTableName(tuple);
        HTable table = store.getTable(tableName);
        if (table == null) {
          logger.debug("No table found for tuple {}", tuple);
          errorTuple(tuple);
          return;
       }
       processTuple(tuple, table);
     }
    void processTuple(T tuple, HTable table); 
    String getTableName(T tuple);
    void errorTuple(T tuple);
  }
```

Then OutputAdapter<T>.processTuple(T tuple) simply calls driver.processTuple(T tuple) as the logic has moved into the latter (and belongs there)? This also makes it easy to implement the batch/window modes that currently have been removed from those classes.

contrib/src/main/java/com/datatorrent/contrib/hbase/OutputAdapter.java
```diff
@@ -0,0 +1,77 @@
...
+  HBaseStore store;
+  OutputAdapter.Driver driver;
+  
+  public OutputAdapter(HBaseStore store, OutputAdapter.Driver driver)
+  {
+    this.store = store;
+    this.driver = driver;
+  }
+
+  public void processTuple(T tuple)
```
sanjaypujare, line 48: See comment below for `Driver<T>`

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseOutputOperator.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+package com.datatorrent.contrib.hbase;
+
+import org.apache.hadoop.hbase.client.HTable;
+
+import com.datatorrent.api.Operator;
+import com.datatorrent.lib.db.AbstractStoreOutputOperator;
+
+/**
+ * Created by lakshmi on 6/27/17.
```
sanjaypujare: More Javadoc to describe this class?

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
```diff
@@ -121,20 +98,12 @@ public void storeAggregate() {
...
 
   @Override
-  public void processTuple(T tuple) {
-    tuples.add(tuple);
-  }
-
-  @Override
   public void setup(OperatorContext context)
   {
     mode=context.getValue(context.PROCESSING_MODE);
```
sanjaypujare: context.PROCESSING_MODE => OperatorContext.PROCESSING_MODE

(to get rid of "The static field Context.OperatorContext.PROCESSING_MODE should be accessed in a static way")

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
```diff
@@ -63,50 +55,35 @@
  *            The tuple type
  * @since 1.0.2
  */
-public abstract class AbstractHBaseWindowAppendOutputOperator<T> extends AbstractAggregateTransactionableStoreOutputOperator<T, HBaseWindowStore> {
+public abstract class AbstractHBaseWindowAppendOutputOperator<T> extends AbstractHBaseWindowOutputOperator<T> {
   private static final transient Logger logger = LoggerFactory.getLogger(AbstractHBaseWindowAppendOutputOperator.class);
-  private List<T> tuples;
   private transient ProcessingMode mode;
+  
+  @Deprecated
```
sanjaypujare: More info about why deprecated? What is the alternative?

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
```diff
@@ -46,8 +40,6 @@
  * for the tuple in the table.<br>
  *
  * <br>
- * This class provides batch append where tuples are collected till the end
- * window and they are appended on end window
```
sanjaypujare, line 50: With this change, the operator is no different from `AbstractHBaseAppendOutputOperator`. So:

- why can't we support batch append even with multi-table support? Pls see comments below.
- Isn't the class name `...WindowAppend...` misleading if there is no batch or window append support? Can we make it deprecated and prevent creation of this operator?

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
```diff
@@ -63,50 +55,35 @@
...
     this.mode = mode;
   }
 
   public AbstractHBaseWindowAppendOutputOperator() {
     store = new HBaseWindowStore();
-    tuples = new ArrayList<T>();
   }
 
   @Override
-  public void storeAggregate() {
```
sanjaypujare, line 86: By getting rid of this method/logic we are no longer supporting batch/window mode. Why can't we keep this method and do the following?
```
   Iterator<T> it = tuples.iterator();		
   while (it.hasNext()) {		
       T t = it.next();
       HTable table = store.getTable(getTableName(t));
       try {		
         Append append = operationAppend(t);		
         table.append(append);		
       } catch (IOException e) {		
         logger.error("Could not output tuple", e);		
         DTThrowable.rethrow(e);		
       }				
     }
```
Coupled with my comments on Adapter/Driver this may become even easier.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
```diff
@@ -121,20 +98,12 @@ public void storeAggregate() {
...
-  @Override
   public void setup(OperatorContext context)
   {
     mode=context.getValue(context.PROCESSING_MODE);
     if(mode==ProcessingMode.EXACTLY_ONCE){
       throw new RuntimeException("This operator only supports atmost once and atleast once processing modes");
     }
-    if(mode==ProcessingMode.AT_MOST_ONCE){
-      tuples.clear();
-    }
```
sanjaypujare, line 137: Should re-instate this once batch/window mode is re-instated.

contrib/src/main/java/com/datatorrent/contrib/hbase/OutputAdapter.java
sanjaypujare: Javadocs for this interface?

sanjaypujare: It will be better to separate the functionality of processTuple/errorTuple and getTableName into 2 separate classes. I don't like the idea of individual operators being made responsible to map a tuple to a table-name. How about another interface like
```
interface TupleToTableMapper<T>
  {
    String getTableName(T tuple);
  }
```
And users of our HBase output operators just provide an instance of `TupleToTableMapper`? So declare it in `AbstractHBaseOutputOperator`?

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowOutputOperator.java
```diff
@@ -0,0 +1,95 @@
...
+
+import com.datatorrent.api.Operator;
+import com.datatorrent.lib.db.AbstractPassThruTransactionableStoreOutputOperator;
+
+/**
+ * Created by lakshmi on 6/27/17.
+ */
+public abstract class AbstractHBaseWindowOutputOperator<T> extends AbstractPassThruTransactionableStoreOutputOperator<T, HBaseWindowStore> implements OutputAdapter.Driver<T>,  Operator.CheckpointNotificationListener
+{
+  /**
```
sanjaypujare: This comment looks a bit out of place here.  May be we can move it to class Javadocs and/or to the respective methods such as endWindow or beforeCheckpoint?

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowOutputOperator.java
```diff
@@ -0,0 +1,95 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.hadoop.hbase.client.HTable;
+
+import com.datatorrent.api.Operator;
+import com.datatorrent.lib.db.AbstractPassThruTransactionableStoreOutputOperator;
+
+/**
+ * Created by lakshmi on 6/27/17.
```
sanjaypujare: More javadocs for the class?

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowOutputOperator.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @Override
+  public void processTuple(T tuple)
+  {
+    outputAdapter.processTuple(tuple);
+  }
+
+  @Override
+  public abstract void processTuple(T tuple, HTable table);
```
sanjaypujare, line 56: What is the advantage in declaring this `abstract` method here with `@Override` annotation? Subclasses are anyway required to implement this because of `OutputAdapter.Driver` interface unless this idiom is common or has some advantages.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowOutputOperator.java
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Get the table name for tuple.
+   *
+   * Implementations can override this method to return the table name where the tuple should be written to.
+   * Return null to write to default table
```
sanjaypujare, line 62: Is this correct? Following the code of OutputAdapter.processTuple(T) and HBaseMultiTableStore.getTable(String) I don't see how a "default table" is returned for null value.

sanjaypujare: Instead why can't we leave this as `abstract` ?

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowOutputOperator.java
sanjaypujare: Also I don't see endWindow() being defined and calling flush

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowPutOutputOperator.java
```diff
@@ -46,8 +40,6 @@
  * the tuple in the table.<br>
  *
  * <br>
- * This class provides a batch put where tuples are collected till the end
- * window and they are put on end window
  *
```
sanjaypujare, line 43: All same comments for `AbstractHBaseWindowAppendOutputOperator` apply here as well.

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseMultiTableStore.java
```diff
@@ -0,0 +1,128 @@
...
+import com.google.common.cache.LoadingCache;
+import com.google.common.cache.RemovalListener;
+import com.google.common.cache.RemovalNotification;
+
+/**
+ * Created by lakshmi on 6/25/17.
+ */
+public class HBaseMultiTableStore extends HBaseWindowStore
+{
+  /**
```
sanjaypujare: Move this comment up as the Java doc for the class with some more explanation about what this class does?

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseMultiTableStore.java
```diff
@@ -0,0 +1,128 @@
...
+      {
+        unloadTable(notification.getValue());
+      }
+    };
+    tableCache = CacheBuilder.<String, HTable>newBuilder().maximumSize(maxOpenTables).removalListener(removalListener).build(cacheLoader);
+  }
+  
+  protected HTable loadTable(String tableName) throws IOException
+  {
+    if ((allowedTableNames != null) && !ArrayUtils.contains(allowedTableNames, tableName)) {
```
sanjaypujare: If allowedTableNames is initialized to `new String[0]` we don't need to check for null

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseMultiTableStore.java
```diff
@@ -0,0 +1,128 @@
...
+public class HBaseMultiTableStore extends HBaseWindowStore
+{
+  /**
+   * The tableName and table from HBaseWindowStore will be used as the meta table to store
+   * recovery information
+   */
+  
+  private static final Logger logger = LoggerFactory.getLogger(HBaseMultiTableStore.class);
+
+  protected String[] allowedTableNames;
```
sanjaypujare: HBaseStore also has `allowedTableNames` and this is a subclass of HBaseStore. Why do we need duplication?

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseMultiTableStore.java
```diff
@@ -0,0 +1,128 @@
...
+  
+  protected transient LoadingCache<String, HTable> tableCache;
+
+  @Min(1)
+  protected int maxOpenTables = Integer.MAX_VALUE;
```
sanjaypujare: Isn't MAX_VALUE too high to be default value for maxOpenTables?

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseWindowStore.java
```diff
@@ -132,8 +140,11 @@ public void beginTransaction() {
 
   @Override
   public void commitTransaction() {
-    // HBase does support transactions so this method left empty
-
+    try {
+      flushTables();
```
sanjaypujare, line 144: Is it correct to just flushTables for commitTransaction when we don't have any other transaction semantics implemented? e.g. beginTransactions, rollback etc are empty and do nothing. So implementing commitTransaction() like this is misleading?

sanjaypujare: Also these other functions have a comment "HBase does support transactions so this method left empty" . Seems to be a typo with "not" missing. Can we fix these comemnts?

sanjaypujare (changes requested):
Pls address comments

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseAppendOutputOperator.java
prasannapramod: In apex checkstyle, org is listed before org.apache. IDE ordered it accordingly. Maybe the older imports were before the checkstyle configuration was created.

contrib/src/main/java/com/datatorrent/contrib/hbase/OutputAdapter.java
prasannapramod: The problem is AbstractHBaseOutputOperator series and AbstractHBaseWindowOutputOperator series operators belong to two different hierarchies that originate from generic store operators which are a common framework across different connectors such as jdbc, couch, memcache etc. The encapsulation was a way to reuse commonly needed functionality without sacrificing the hierarchy or backward compatibility.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
prasannapramod: Processing mode being a common platform feature should be set as an attribute so not sure why it was made available as a property. Even if the property is set, it wouldn't effect how the platform would process that feature for the operator.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
prasannapramod: I was also thinking earlier they look similar and couldn't figure out a better way to avoid this. Even though the concrete implementations look similar they will function differently because of the operators they extend. Also, with the changes, both will support multiple tables.

The difference between the non-window and the window operator is this. In the case of restart from failure, the non-window operator writes the tuples from all the windows replayed from the checkpoint. The windowed operator doesn't write tuples belonging to those windows after the checkpoint that were completely processed in the prior run, before the operator failed. It only repeats the partial writes for the window during which the operator failed. This reduces the number of duplicates in cases where the writes will result in new entries in HBase. Exactly once is not being done because transaction support is not available in HBase. This functionality requires the user to provide additional resources from hbase such as a meta column family etc, which is why I think there were two operators created and not a single one.

In either case when tuples are written, they are not flushed immediately to hbase server for performance reasons. They are given to hbase api which would internally flush it when some internal conditions are satisfied like the buffer filling up. However, we do want to make sure all pending tuples are flushed in certain conditions to ensure no data loss in case of operator failure. The non-window operator explicitly flushes before checkpoint and the window operator flushes during end of window because fully processed windows will be skipped on replay after failure.

In the earlier implementation of windowed operator, I noticed that the operator was batching all the tuples within the operator into one big batch and then sending all the tuples to the HBase api at the end of the window. I think this is not necessary as the HBase api internally would batch and flush appropriately. Batching at the operator level on a per window basis will introduce latency and will also significantly increase memory usage. That is why I changed it and one of the reasons now the concrete operators look similar.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
prasannapramod: Please see my description above. The window name is not for batching of tuples but rather for fault recovery behavior. The batching of tuples is not optimal as I described in the last paragraph above and hence I changed it. This also means the class hierarchy is slightly different but I don't think it will be a problem as I don't see anybody extending the class and overriding this method. Also since the operator is in contrib it is evolving in nature.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
prasannapramod: Please see above, batching introduces unnecessary issues and is not needed.

contrib/src/main/java/com/datatorrent/contrib/hbase/OutputAdapter.java
prasannapramod: It is possible to do it but do we want to introduce another interface just for getting the table name. It would have to be an additional adapter that is passed into the OutputAdapter. Also it would have to a property in the operator with a default implementation instantiated within the operator that enables the existing behavior of returning the same table for all tuples for compatibility. Is this extra complexity needed in every case? If some operator wants the pluggability they could implement the getTableName method and delegate it to another adapter. Other operators like file output operator just have the operator return the filename as well.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowOutputOperator.java
prasannapramod: It is a developer comment for anyone working on this operator in future and not for end users, hence didn't put it in javadoc. The endWindow is being implemented by the parent class AbstractPassThruTransactionableStoreOutputOperator which will result in flush getting called.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowOutputOperator.java
prasannapramod: Will add

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseOutputOperator.java
prasannapramod: Will add

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowOutputOperator.java
prasannapramod: Mainly for easy indication to developers extending this class that they need to implement the abstract method. I can remove it.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowOutputOperator.java
prasannapramod: The HBaseStore will return default table for the null table name. The reason we can't leave it abstract and upto the implementations is that it will break existing concrete operators extending from the operator as they are not implementing this method. By default, it should be compatible with existing operators from an api and functional perspective.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowPutOutputOperator.java
prasannapramod: Please see above

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseMultiTableStore.java
prasannapramod: I am sorry, I need to remove this class. It was part of my initial trials and no longer needed.

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseMultiTableStore.java
prasannapramod: Sorry about the confusion on HBaseMultiTableStore. I will remove it.

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseWindowStore.java
prasannapramod: The reason for doing this as part of commitTransaction is that we are piggybacking on the mechanism to minimize rewrites on a replay from failure. Please see the long description above on the workings of these operators. There are existing common store operators that implement the fault tolerance mechanism by using the transaction pattern. If transactions are supported it can result in exactly once otherwise it will help in reducing writing duplicates significantly.

prasannapramod:
@sanjaypujare Addressed comments and made changes where necessary. Please see.

sanjaypujare:
@prasannapramod both build checks are failing. Are these pre-existing failures?

prasannapramod:
Jenkins, re-test this, please.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseAppendOutputOperator.java
sanjaypujare: Hmmm, does it mean both the old and new orders are checkstyle compliant?

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
sanjaypujare: Can we add Javadoc with this explanation so readers know why it is deprecated? That was the intent of my original comment

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseWindowStore.java
```diff
@@ -126,14 +134,17 @@ public void connect() throws IOException {
 
   @Override
   public void beginTransaction() {
-    // HBase does support transactions so this method left empty
+    // HBase does not support transactions so this method left empty
```
sanjaypujare, line 137: Similar change needed in rollbackTransaction() and isInTransaction() . And possibly others. Pls check everywhere for this comment and fix it.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
sanjaypujare: Thanks for the explanation - makes sense.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowAppendOutputOperator.java
sanjaypujare: ok

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowOutputOperator.java
sanjaypujare: Actually keep it. I can see the value of having both abstract and Override annotation here.

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseWindowOutputOperator.java
sanjaypujare: ok

contrib/src/main/java/com/datatorrent/contrib/hbase/HBaseWindowStore.java
sanjaypujare: ok

contrib/src/main/java/com/datatorrent/contrib/hbase/OutputAdapter.java
sanjaypujare: ok

sanjaypujare (commented):
I have a couple of comments requesting small changes - mainly docs/Javadocs. Once they are done I'll approve

prasannapramod:
@sanjaypujare @PramodSSImmaneni Added JavaDocs as suggested. Please see.

pramodin:
LGTM

contrib/src/main/java/com/datatorrent/contrib/hbase/AbstractHBaseOutputOperator.java
sanjaypujare: I don't see any Javadoc for this class yet

sanjaypujare (commented):
@prasannapramod I still see missing Javadocs in a few places and where you commented you will add them. Waiting for them

prasannapramod:
@sanjaypujare I added JavaDocs for all newly created classes now. Hoping there is sufficient documentation for methods that I changed. Please see

sanjaypujare (approved):
LGTM. @PramodSSImmaneni you can merge

pramodin:
@prasannapramod can you fix the commit message to include jira number

PR #639, opened 2017-07-03
APEXMALHAR-2513 fixes for JdbcPollOperator

tweise:
Fix result set extraction to use column index.
Exit static partitions after work is done.
Propagate all exceptions from poller task.
Add comment regarding SQL dialect support.
Fix poller example properties.
Documentation fixes.

@devtagare please review


devtagare:
LGTM

PR #643, opened 2017-07-10
APEXMALHAR-2519 Memory fix in the benchmark app properties

SaumyaMohan:


benchmark/src/main/resources/META-INF/properties.xml
```diff
@@ -28,7 +28,7 @@
   -->
   <property>
     <name>dt.attr.MASTER_MEMORY_MB</name>
-    <value>1024</value>
+    <value>2048</value>
```
vrozov, line 31: It is an application master memory. Do you need to change operators memory requirement as well?

PR #646, opened 2017-07-18
APEXMALHAR-2525 Fix yahoofinance example

florianschmidt1994:
This fixes the yahoofinance example in the malhar-examples artifact.
The StockTickInput Operator uses stock ticker symbols to fetch stock
information from the yahoo finance api. Because the default tickers
included "YHOO", which does not exist anymore, the StockTickInput
operator would crash because it fails to parse a "N/A" response from the
api into a number.

This is fixed with this commit by changing all "YHOO" tickers to "AABA",
which is the new ticker of Altaba Inc., that Yahoo is now traded under


sanjaypujare:
Sorry, didn't see this PR earlier. YHOO=>AABA is a simple fix to only fix the data input. How about improving the logic to avoid the NumberFormatException when the string is N/A?

PR #647, opened 2017-07-18
APEXMALHAR-2526 FunctionOperator fails to resolve class from context class loader.

tweise:


tweise:
test this please

tweise:
test this please

tweise:
@davidyan74 please have a look

library/src/main/java/org/apache/apex/malhar/lib/utils/KryoJavaSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.KryoException;
+import com.esotericsoftware.kryo.io.Input;
+import com.esotericsoftware.kryo.serializers.JavaSerializer;
+import com.esotericsoftware.kryo.util.ObjectMap;
+
+/**
+ * Extension of {@JavaSerializer} for consistent class loading behavior.
+ *
+ * <p>This class can be removed after upgrade to a Kryo release > 4.0
```
davidyan74, line 35: Should we file a JIRA so that we don't lose track of this?

davidyan74 (approved):

library/src/main/java/org/apache/apex/malhar/lib/utils/KryoJavaSerializer.java
tweise: Done: https://issues.apache.org/jira/browse/APEXCORE-768

PR #649, opened 2017-07-20
APEXMALHAR-2513 JdbcPollInputOperator does not respect poll interval

tweise:

PR #652, opened 2017-07-27
APEXMALHAR-2514 JDBC poll offset rebase support.

tweise:
based on #649

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
@@ -24,14 +24,15 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Executors;
-import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
@@ -85,7 +86,7 @@ import static org.jooq.impl.DSL.field;
  * The operator uses jOOQ to build the SQL queries based on the discovered {@link org.jooq.SQLDialect}.
  * Note that some of the dialects (including Oracle) are only available in commercial
  * jOOQ distributions. If the dialect is not available, a generic translation is applied,
- * you can post-process the generated SQL by overriding {@link #buildRangeQuery(int, int)}.
+ * you can post-process the generated SQL by overriding {@link #buildRangeQuery(Object, int, int)}.
  *
  * @displayName Jdbc Polling Input Operator
  * @category Input
@@ -98,13 +99,14 @@ import static org.jooq.impl.DSL.field;
 public abstract class AbstractJdbcPollInputOperator<T> extends AbstractStoreInputOperator<T, JdbcStore> implements
     ActivationListener<OperatorContext>, Partitioner<AbstractJdbcPollInputOperator<T>>
 {
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractJdbcPollInputOperator.class);
   private static int DEFAULT_QUEUE_CAPACITY = 4 * 1024;
   private static int DEFAULT_POLL_INTERVAL = 10 * 1000;
   private static int DEFAULT_FETCH_SIZE = 20000;
   private static int DEFAULT_BATCH_SIZE = 2000;
   private static int DEFAULT_SLEEP_TIME = 100;
   private static int DEFAULT_RESULT_LIMIT = 100000;
-  private int pollInterval = DEFAULT_POLL_INTERVAL; //in miliseconds
+  private int pollInterval = DEFAULT_POLL_INTERVAL; //in milliseconds
   private int queueCapacity = DEFAULT_QUEUE_CAPACITY;
   private int fetchSize = DEFAULT_FETCH_SIZE;
   /**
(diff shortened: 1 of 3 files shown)


tweise:
@sandeshh would you like to review this PR?

tweise:
@bhupeshchawda can you review?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -136,13 +139,24 @@
...
   protected boolean isPollerPartition;
 
-  protected transient MutablePair<Integer, Integer> currentWindowRecoveryState;
-
   private transient int lastOffset;
+  private transient Object prevKey;
+  private transient Object lastKey;
+
+  /**
+   * The candidate key/offset pair identified by the fetch thread
```
amberarrow: Suggest using "poller thread"  instead of "fetch thread" since the thread class is called "DBPoller", reader will be able correlate the two more easily.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -750,6 +813,24 @@ public void setResultLimit(int resultLimit)
...
   }
 
-  private static final Logger LOG = LoggerFactory.getLogger(AbstractJdbcPollInputOperator.class);
+  public boolean isRebaseOffset()
+  {
+    return rebaseOffset;
+  }
+
+  /**
+   * Whether the query should automatically be be augmented with a WHERE
```
amberarrow: typo: be be => be

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -246,6 +275,25 @@ public void emitTuples()
...
   protected abstract void emitTuple(T tuple);
 
+  /**
+   * Visible to subclasses to allow for custom offset saving and initialization.
+   */
+  protected static class WindowData
+  {
+    public Object key;
+    public int lowerBound;
+    public int upperBound;
```
amberarrow: Can these fields be made protected ?

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -423,11 +466,15 @@ public void emitReplayedTuples(PreparedStatement ps)
...
+    final List<KeyValPair<Integer, Integer>> partitionRanges;
     try {
       store.connect();
       dslContext = createDSLContext();
-      partitionToRangeMap = getPartitionedQueryRangeMap(getPartitionCount());
+      if (getPartitionCount() > 0) {
+        partitionRanges = getPartitionedQueryRanges(getPartitionCount());
+      } else {
+        partitionRanges = new ArrayList<>(0);
+      }
```
amberarrow: Looks better if this check is moved into getPartitionedQueryRanges; also makes that function more robust by avoiding divide-by-zero errors in future maintenance changes.

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -464,52 +512,66 @@ public void partitioned(
...
     int rowCount = 0;
     try {
-      rowCount = getRecordsCount();
+      rowCount = getRecordsCount(null);
     } catch (SQLException e) {
       LOG.error("Exception in getting the record range", e);
     }
 
-    HashMap<Integer, KeyValPair<Integer, Integer>> partitionToQueryMap = new HashMap<>();
+    List<KeyValPair<Integer, Integer>> partitionToQueryMap = new ArrayList<>();
```
amberarrow: partitionToQueryMap => partitionToQueryList

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -282,23 +330,44 @@ public void deactivate()
...
+  protected int insertDbDataInQueue(PreparedStatement preparedStatement) throws SQLException, InterruptedException
   {
+    int resultCount = 0;
     preparedStatement.setFetchSize(getFetchSize());
     ResultSet result = preparedStatement.executeQuery();
     if (result.next()) {
       do {
-        while (execute && !emitQueue.offer(getTuple(result))) {
+        T obj = getTuple(result);
+        while (execute && !emitQueue.offer(obj)) {
```
amberarrow: Not part of this PR but seems like emitQueue does not need to be a LinkedBlockingDeque since we are not using the deque feature of adding at both ends or removing from both.

amberarrow:
Made a few comments; once addressed, merge can proceed

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
```diff
@@ -282,23 +330,44 @@ public void deactivate()
...
-  protected void insertDbDataInQueue(PreparedStatement preparedStatement) throws SQLException, InterruptedException
+  protected int insertDbDataInQueue(PreparedStatement preparedStatement) throws SQLException, InterruptedException
   {
+    int resultCount = 0;
     preparedStatement.setFetchSize(getFetchSize());
     ResultSet result = preparedStatement.executeQuery();
     if (result.next()) {
       do {
-        while (execute && !emitQueue.offer(getTuple(result))) {
+        T obj = getTuple(result);
```
amberarrow: Can we check the result for null and skip it ? Then we won't need to null check inside the loop in emitTuples().

library/src/main/java/com/datatorrent/lib/db/jdbc/AbstractJdbcPollInputOperator.java
tweise: I modified the loop as there is no reason to check for result.next() first and then have a do .. while.

tweise:
@amberarrow comments are addressed. Please mark the PR approved or merge it.

amberarrow:
Approved.

PR #653, opened 2017-07-28
APEXMALHAR-2489 Change algorithm for running average

florianschmidt1994:
The current algorithm for calculating the running average was subject to
a potential overflow, because part of the formula required the
average value (average) to be multiplied with the number of processed
tuples (count). `average * count` would for example overflow when e.g.
`average > Double.MAX_VALUE` and `count >=2`

This PR changes the formula used to the one described on
http://www.heikohoffmann.de/htmlthesis/node134.html, where such a
multiplication is not necessary anymore. It also adds a unit test which
checks that there is not overflow occuring anymore

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/math/RunningAverage.java
+++ b/library/src/main/java/com/datatorrent/lib/math/RunningAverage.java
@@ -64,9 +64,10 @@ public class RunningAverage extends BaseOperator
     @Override
     public void process(Number tuple)
     {
-      double sum = (RunningAverage.this.count * average) + tuple.doubleValue();
-      RunningAverage.this.count++;
-      average = sum / RunningAverage.this.count;
+      // Floating mean as explained on http://www.heikohoffmann.de/htmlthesis/node134.html
+      double firstPart = (1.0 / (++RunningAverage.this.count));
+      double secondPart = (tuple.doubleValue() - average);
+      average += (firstPart * secondPart);
     }
   };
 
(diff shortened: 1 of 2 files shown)


florianschmidt1994:
This is already an improvement to what was there before (see pr message), but according to @vrozov there are still some corner cases where issues with overflows could occur. Unfortunately I have not been able to find / reproduce those (yet), maybe give me another hint?

library/src/main/java/com/datatorrent/lib/math/RunningAverage.java
```diff
@@ -64,9 +64,10 @@
     @Override
     public void process(Number tuple)
     {
-      double sum = (RunningAverage.this.count * average) + tuple.doubleValue();
-      RunningAverage.this.count++;
-      average = sum / RunningAverage.this.count;
+      // Floating mean as explained on http://www.heikohoffmann.de/htmlthesis/node134.html
+      double firstPart = (1.0 / (++RunningAverage.this.count));
+      double secondPart = (tuple.doubleValue() - average);
+      average = average + (firstPart * secondPart);
```
vrozov: average +=? Why two parts is necessary?

florianschmidt1994: No reason, just a matter of style. I will change it.

tweise:
@vrozov can the PR be merged?

PR #654, opened 2017-07-28
APEXMALHAR-2500 Fix javadoc for FSInputModule

florianschmidt1994:
The javadoc stated that FSInputModule is an abstract class, which it is
not. This PR fixes the JavaDoc


florianschmidt1994:
Test me again, Jenkins!

florianschmidt1994:
After checking for all occurrences of the pattern "abstract class", this seems to be the only case where there is such a mismatch. @sanjaypujare could you review?

sanjaypujare (approved):
LGTM

PR #655, opened 2017-07-30
APEXMALHAR-2529 Allow subclasses of AbstractFileInputOperator to have…

davidyan74:
… their own logic to advance the streaming window

@tweise Please take a look.

### DIFF
--- a/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
+++ b/library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
@@ -680,7 +680,7 @@ public abstract class AbstractFileInputOperator<T> implements InputOperator, Par
 
       try {
         int counterForTuple = 0;
-        while (counterForTuple++ < emitBatchSize) {
+        while (!suspendEmit() && counterForTuple++ < emitBatchSize) {
           T line = readEntity();
           if (line == null) {
             LOG.info("done reading file ({} entries).", offset);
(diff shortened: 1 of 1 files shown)


tweise:
LGTM, will merge on Monday if there is no other feedback.

tweise:
test this please

tweise (approved):

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
```diff
@@ -711,6 +711,14 @@ public void emitTuples()
   }
 
   /**
+   * Whether or not to advance the window now regardless of emitBatchSize.
+   */
+  protected boolean advanceWindow()
```
tweise: The method does not "advance the window" and "window" by itself is ambiguous. Would "suspendEmit()" or something similar better convey the purpose? Maybe also expand on that in the javadoc?

davidyan74: Should it be suspendEmitTillNextWindow or something like that? I think suspendEmit can be confusing...

davidyan74:
test this please

library/src/main/java/com/datatorrent/lib/io/fs/AbstractFileInputOperator.java
tweise: Doesn't till when emit is suspended depend on the logic you implement?

davidyan74: You're right. Changes applied.

PR #656, opened 2017-07-31
APEXMALHAR-2233 Migrate exactly-once examples.

tweise:
@amberarrow please review


examples/exactly-once/src/main/java/com/example/myapexapp/Application.java
```diff
@@ -0,0 +1,103 @@
+/**
```
amberarrow: Add license (multiple files)

examples/exactly-once/src/main/java/com/example/myapexapp/Application.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Override
+    protected void setStatementParameters(PreparedStatement statement, KeyValPair<String, Integer> tuple) throws SQLException
+    {
+      statement.setString(1, tuple.getKey());
+      statement.setInt(2, tuple.getValue());
+    }
+  }
+
+  public static class UniqueCounter<K> extends BaseUniqueKeyCounter<K>
```
amberarrow: A comment here describing why the built-in UniqueCounter class is not suitable would be helpful.

amberarrow:
(Missed the README, so ignore earlier comment).

Other than the minor items noted, this looks good.

examples/exactly-once/src/main/java/com/example/myapexapp/Application.java
tweise: @amberarrow there are multiple commits, please look at the net diff. (CI would fail if license header was missing.)

tweise:
@amberarrow made the changes

tweise:
@amberarrow comments are addressed

amberarrow:
Looks good to me.

PR #659, opened 2017-08-02
APEXMALHAR-2534 Provided fix for TopNWordsWithQueries example.

prasannapramod:
In TopNWordsWithQueries, the file meta information is not being sent with every tuple, however, the application is attempting to store output per input file. The input file information is being sent via separate port the only way it will work is if the file contents from different files don't get mixed up in the same window. The fix ensures that a new file always starts in a new window.

@amberarrow @PramodSSImmaneni please see

### DIFF
--- a/examples/wordcount/src/main/java/org/apache/apex/examples/wordcount/LineReader.java
+++ b/examples/wordcount/src/main/java/org/apache/apex/examples/wordcount/LineReader.java
@@ -41,6 +41,7 @@ import com.datatorrent.lib.io.fs.AbstractFileInputOperator;
 public class LineReader extends AbstractFileInputOperator<String>
 {
   private static final Logger LOG = LoggerFactory.getLogger(LineReader.class);
+  private boolean flag = true;
 
   /**
    * Output port on which lines from current file name are emitted
@@ -104,6 +105,7 @@ public class LineReader extends AbstractFileInputOperator<String>
     if (control.isConnected()) {
       LOG.info("readEntity: EOF for {}", path);
       final String name = path.getName();    // final component of path
+      flag = false;
       control.emit(name);
     }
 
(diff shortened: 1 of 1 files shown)


amberarrow:
Looks good, I'll merge in a day or two.

examples/wordcount/src/main/java/org/apache/apex/examples/wordcount/LineReader.java
```diff
@@ -104,15 +116,37 @@ protected String readEntity() throws IOException
...
   }
 
   @Override
+  public void emitTuples()
+  {
+    if (flag) {
+      super.emitTuples();
+    } else {
+      try {
+        Thread.sleep(spinMillis);
```
tweise: why this? engine will handle it when nothing is emitted.

prasannapramod:
@amberarrow @tweise please see.

examples/wordcount/src/main/java/org/apache/apex/examples/wordcount/LineReader.java
```diff
@@ -57,6 +60,13 @@
 
   private Path path;
 
+  @Override
+  public void setup(Context.OperatorContext context)
```
tweise: remove this override?

tweise (changes requested):

prasannapramod:
@amberarrow @tweise please see

tweise (approved):

examples/wordcount/src/main/java/org/apache/apex/examples/wordcount/LineReader.java
```diff
@@ -41,6 +41,8 @@
 public class LineReader extends AbstractFileInputOperator<String>
 {
   private static final Logger LOG = LoggerFactory.getLogger(LineReader.class);
+  private boolean flag = true;
+  transient int spinMillis;
```
sandeshh: Why spinMillis ?

prasannapramod: fixed.

PR #660, opened 2017-08-04
APEXMALHAR-2532 Suppress messages in Transform Application Test

ShunxinLu:
@tweise Please review.

### DIFF
--- a/examples/transform/src/main/java/org/apache/apex/examples/transform/Application.java
+++ b/examples/transform/src/main/java/org/apache/apex/examples/transform/Application.java
@@ -19,6 +19,7 @@
 
 package org.apache.apex.examples.transform;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


examples/transform/src/main/java/org/apache/apex/examples/transform/Application.java
```diff
@@ -19,25 +19,72 @@
...
 import com.datatorrent.lib.transform.TransformOperator;
 
-@ApplicationAnnotation(name = "TransformExample")
 /**
  * @since 3.7.0
  */
+@ApplicationAnnotation(name = "TransformExample")
 public class Application implements StreamingApplication
 {
+  public static class Collector extends BaseOperator
```
tweise, line 48: Instead of adding all of this, I have an alternative proprosal:

- Use the FunctionOperator instead of Console...
- Add member outputFunction to the application class (default console)
- Override the outputFunction in the test to collect in test private static collection.

That will keep the code more concise yet allow proper testing.

tweise (changes requested):

tweise:
Fixed with #674

PR #661, opened 2017-08-08
APEXMALHAR-2535: change type of timeout variables from int to long

tushargosavi:
Fix NPE during teardown, when refreshtime is not set
Allow setting expiration type

### DIFF
--- a/contrib/src/main/java/com/datatorrent/contrib/enrich/AbstractEnricher.java
+++ b/contrib/src/main/java/com/datatorrent/contrib/enrich/AbstractEnricher.java
@@ -25,11 +25,13 @@ import java.util.List;
 import org.apache.hadoop.classification.InterfaceStability;
 
 import com.esotericsoftware.kryo.NotNull;
+
 import com.datatorrent.api.Context;
 import com.datatorrent.api.Operator;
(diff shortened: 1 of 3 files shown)


contrib/src/main/java/com/datatorrent/contrib/enrich/AbstractEnricher.java
```diff
@@ -66,8 +68,9 @@
   /**
    * Optional parameters for enricher.
    */
-  private int cacheExpirationInterval = 1 * 60 * 60 * 1000;  // 1 hour
-  private int cacheCleanupInterval = 1 * 60 * 60 * 1000; // 1 hour
+  private long cacheExpirationInterval = 1 * 60 * 60 * 1000;  // 1 hour
+  private long cacheCleanupInterval = 1 * 60 * 60 * 1000; // 1 hour
+  private ExpiryType expiryType = ExpiryType.EXPIRE_AFTER_WRITE;
```
DT-Priyanka: can we rename variable as expiryAfterWrite ?

DT-Priyanka:
Looks good to me, will merge in 2 days if there is no objection from community.

tushargosavi:
retest this please

PR #662, opened 2017-08-14
APEXMALHAR-2517 relocate com.datatorrent packages to org.apache.apex

tweise:


davidyan74:
This is long overdue. LGTM.

pramodin:
Please give some time to look into this as there are a number of changes

tweise:
@PramodSSImmaneni this PR is difficult to rebase but the changes are trivial (imports and checkstyle fixes). I would like to wrap it up in a day or two.

amolhkekre:
I am -1 on this as users this will break binary compatibility with current users, and will require code changes. For current users it is not backward incompatibility. The approach suggested in #664 is not adequate as it will require existing apps to stick to older version of Malhar. This change should be taken up in a major upgrade.

tweise:
@amolhkekre I will call out the following WRT your veto on this PR:
1. I did not see a contribution from you to the preceding 6 months old discussion thread
2. You seem to suggest that suggested change (2 PRs) break binary compatibility ("as users this will break binary compatibility with current users"). However, the proposed solution doesn't. Granted current guarantees in malhar are confusing because in reality it is evolving and stricter guarantees would mostly bring development to a halt. However, this has been discussed on the list and when it comes to a veto on the PR then IMO the onus is on you to have the context.

vrozov:
Please resolve conflicts and update pom.

amolhkekre:
Thomas,
The thread has been taken up on Apex user forums. I believe a decision to
update the major version was taken, under which backward compatibility need
not hold. So do proceed.

Thks,
Amol



E:amol@datatorrent.com | M: 510-449-2606 | Twitter: @*amolhkekre*

www.datatorrent.com

pom.xml
```diff
@@ -139,7 +139,10 @@
               <excludes>
                 <exclude>@org.apache.hadoop.classification.InterfaceStability$Evolving</exclude>
                 <exclude>@org.apache.hadoop.classification.InterfaceStability$Unstable</exclude>
-                <exclude>com.datatorrent.contrib.parquet</exclude>
+                <exclude>com.datatorrent.lib</exclude>
```
vrozov: Are those excludes required if the major version is changed?

tweise: No, they won't be needed. Should I make the version change in the same commit?

vrozov: My take - yes. Is there a reason not to?

tweise:
@vrozov added version number change and also removed obsolete samples directory. Will squash these commits once approved.

apps/filecopy/src/main/java/com/datatorrent/apps/copy/HDFSFileCopyApp.java
```diff
@@ -19,14 +19,14 @@
 
 package com.datatorrent.apps.copy;
```
vrozov, line 20: Should this be relocated to org.apache.malhar.apps.copy (possibly as part of a follow-up PR). If yes, can you file JIRA?

vrozov (approved):
LGTM +1. Please squash commits.

tweise:
test this please

tweise:
test this please

PR #665, opened 2017-08-15
APEXMALHAR-2034 Adding new Avro Module to encapsulate Container File to GenericRecord to POJO transformation

SaumyaMohan:

PR #666, opened 2017-08-16
APEXMALHAR-2034 Adding new Avro Module to encapsulate Container File to Avro GenericRecord to POJO transformations

SaumyaMohan:
Adding link to closed PR: https://github.com/apache/apex-malhar/pull/665

New PR is created to ensure forked branch starts with JIRA name


tweise:
Why is a new Maven project added when goal is to just add the Avro operator/module that combines existing operators? Wouldn't it be more appropriate to add that class to the existing package? 

@sandeshh Adding the new project will require more work, including history preserving renames, proper dependency management, documentation, meaningful test coverage for the CI and other things that justify such invasive change. 

Most importantly, I think that it would warrant an upfront discussion. What we don't need is a lot of code moved around when the net benefit is small.

Also, will this module work with partitioning? Let's say Kafka -> Avro -> POJO all in a parallel partition?

SaumyaMohan:
@tweise Creating a new maven module for Avro is in accordance with the JIRA https://issues.apache.org/jira/browse/APEXMALHAR-1843. Since a new 'apex module' is being created for Avro, we decided to add it strategically in accordance with the above JIRA.

We have ensured that the files that are moved to the new location retain the GIT history.
Unit test for the new 'apex module' has been added
Current users referring to the old location of Avro operators will continue to be able to use them as the operators in old location are not deleted but marked deprecated, and they extend the operators in the new location. This has been tested by the running a sample app using old location operators.

sandeshh:
Following combination, Kafka -> Avro -> POJO, is not supported. Avro operator is an input operator which extends AbstractFileInputOperator, so parallel partitioning behavior is same as AbstractFileInputOperator's behavior.

SaumyaMohan:
I've updated the JIRA https://issues.apache.org/jira/browse/APEXMALHAR-2034 with all the tests and checks done to ensure proper dependencies, expected behavior of Avro module, and backward compatibility.

avro/src/test/java/org/apache/apex/malhar/avro/AvroFileToPojoModuleTest.java
```diff
@@ -0,0 +1,269 @@
...
+    }
+  }
+
+  private void writeAvroFile(File outputFile) throws IOException
+  {
+
+    DatumWriter<GenericRecord> datumWriter = new GenericDatumWriter<GenericRecord>(
+        new Schema.Parser().parse(AVRO_SCHEMA));
+
+    DataFileWriter<GenericRecord> dataFileWriter = new DataFileWriter<GenericRecord>(datumWriter);
```
sandeshh: Use Try-with-resources structure

avro/src/test/java/org/apache/apex/malhar/avro/AvroFileToPojoModuleTest.java
```diff
@@ -0,0 +1,269 @@
...
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+    }
+  }
+
+  @Rule
+  public TestMeta testMeta = new TestMeta();
+
+  private void createAvroInput(int cnt)
```
sandeshh: Code duplication? please create util functions

avro/src/test/java/org/apache/apex/malhar/avro/AvroFileToPojoModuleTest.java
```diff
@@ -0,0 +1,269 @@
...
+      rec.put("orderId", cnt * 1L);
+      rec.put("customerId", cnt * 2);
+      rec.put("total", cnt * 1.5);
+      rec.put("customerName", "*" + cnt + "*");
+      cnt--;
+      recordList.add(rec);
+    }
+  }
+
+  private void writeAvroFile(File outputFile) throws IOException
```
sandeshh: Where is the cleanup of the files?

tweise:
@SaumyaMohan addition of the composite Avro operator does not require the creation of a new Maven project and code refactor. You can separate the two if you like, but the project changes need to be discussed on the mailing list first, @sandeshh should be aware of that.

avro/src/test/java/org/apache/apex/malhar/avro/AvroFileToPojoModuleTest.java
SaumyaMohan: the file cleanup is done by the overloaded finished() method

avro/src/test/java/org/apache/apex/malhar/avro/AvroFileToPojoModuleTest.java
SaumyaMohan: fixed

avro/src/test/java/org/apache/apex/malhar/avro/AvroFileToPojoModuleTest.java
sandeshh: Let us not worry about refactoring in this jira. You can create a separate jira.

sandeshh:
Will merge it tomorrow.

tweise:
-1 on the PR as it is. I see no justification to create a new Maven project for this change. You can add the composite operator to the existing package.

SaumyaMohan:
@tweise thanks for your feedback. I just want to understand why a strategic move to a new package as per existing JIRA https://issues.apache.org/jira/browse/APEXMALHAR-1843 is being discouraged. I understand the functional change is small but through this PR we have tried to not only improve the functionality but also take this opportunity to improve the project structure which was already encouraged in the APEXMALHAR-1843 (unless that JIRA should not be referred to by developers).

We have taken proper steps to ensure file history and backward compatibility is maintained.

We have also started an email chain sending our justification for this move. 
Please could you respond with your concerns on that email so all Apache Apex developers are on the same page on what approach to take for development tasks?

tweise:
Yes, we can continue discussing this on the email thread. It is the recommended process for changes of this extend, to avoid contributors wasting efforts.

SaumyaMohan:
@tweise if I've already taken the initiative to make the strategic structural change (as per existing JIRA as mentioned above), please could you elaborate what's the technical reason why this should not be done?

I just want to understand if the move to new package is something not considered strategic or it should not be done right now.

tweise:
@SaumyaMohan you should really use the mailing list to discuss this ahead of taking up the work, that's part of the collaboration process.

It is correct that the project structure needs to become more modular and in principle this is the right move. But it is not just about moving around existing code. Contrib generally implies lack of maturity/stability and when I tried to use the Avro operator in the past, it was frankly not ready for prime time. Promotion to a separate module needs to come with a base line in readiness, otherwise the entire move is just cosmetical. So before taking up this kind of initiative, you should make the case why it makes sense at this time and propose what needs to be done to provide adequate value to a user. I also agree that the JIRA isn't very helpful, it was created a long time ago and could benefit from some more context based on where the project is today.

As already stated, you could break up your PR into addressing the problem that you originally set out to solve (add new operator) as one change and new project creation separately.

avro/src/main/java/org/apache/apex/malhar/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,172 @@
...
+import com.datatorrent.api.DefaultOutputPort;
+import com.datatorrent.lib.io.fs.AbstractFileInputOperator;
+
+/**
+ * <p>
+ * Avro File Input Operator
+ * </p>
+ * A specific implementation of the AbstractFileInputOperator to read Avro
+ * container files.<br>
+ * No need to provide schema,its inferred from the file<br>
```
tweise: It's not "inferred", the schema is stored in the file and that is standard Avro.

avro/src/main/java/org/apache/apex/malhar/avro/AvroFileInputOperator.java
```diff
@@ -0,0 +1,172 @@
...
+        recordCount++;
+        return record;
+      }
+    } catch (AvroRuntimeException are) {
+      LOG.error("Exception in parsing record for file - " + super.currentFile + " at offset - " + offset, are);
+      if (errorRecordsPort.isConnected()) {
+        errorRecordsPort.emit("FileName:" + super.currentFile + ", Offset:" + offset);
+      }
+      errorCount++;
+      throw new AvroRuntimeException(are);
```
tweise: why is the original exception not thrown

avro/pom.xml
```diff
@@ -0,0 +1,260 @@
...
+    </build>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.codehaus.janino</groupId>
+            <artifactId>janino</artifactId>
+            <version>2.7.8</version>
+        </dependency>
+        <dependency>
+            <groupId>org.python</groupId>
```
tweise: why is this needed?

avro/pom.xml
```diff
@@ -0,0 +1,260 @@
...
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <!-- generate javdoc -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-javadoc-plugin</artifactId>
```
tweise: why all of this clutter? what is it needed for?

avro/pom.xml
```diff
@@ -0,0 +1,260 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.apex</groupId>
+            <artifactId>apex-common</artifactId>
+            <version>${apex.core.version}</version>
+            <type>jar</type>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
```
tweise: why are these dependencies declared here?

avro/pom.xml
```diff
@@ -0,0 +1,260 @@
...
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.7.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro</artifactId>
+            <version>1.7.4</version>
+            <exclusions>
```
tweise: why the exclusions?

avro/src/main/java/org/apache/apex/malhar/avro/AvroFileToPojoModule.java
```diff
@@ -0,0 +1,92 @@
...
+ * to ensure exactly once semantics with a HDFS backed WAL.
+ *
+ * @displayName AvroFileToPojoModule
+ * @category Input
+ * @tags fs, file,avro, input operator, generic record, pojo
+ *
+ * @since
+ */
+
+public class AvroFileToPojoModule implements Module
```
tweise: What value does "Module" in the name add?

avro/src/main/java/org/apache/apex/malhar/avro/AvroFileToPojoModule.java
```diff
@@ -0,0 +1,92 @@
...
+  public final transient ProxyOutputPort<GenericRecord> errorPort = new ProxyOutputPort<>();
+  //output ports from AvroFileInputOperator
+  public final transient ProxyOutputPort<String> completedAvroFilesPort = new ProxyOutputPort<>();
+  public final transient ProxyOutputPort<String> avroErrorRecordsPort = new ProxyOutputPort<>();
+
+  private AvroFileInputOperator avroFileInputOperator = new AvroFileInputOperator();
+  Class<?> pojoClass = null;
+
+  @Override
+  public void populateDAG(DAG dag, Configuration configuration)
```
tweise: I asked before how parallel partitioning works with this?

avro/src/main/java/org/apache/apex/malhar/avro/AvroRecordHelper.java
```diff
@@ -0,0 +1,126 @@
...
+      convertedValue = convertAndResolveUnionToPrimitive(schema, key, value);
+    } else {
+      convertedValue = convertValueToAvroPrimitive(type, key, value);
+    }
+
+    return convertedValue;
+
+  }
+
+  private static Object convertValueToAvroPrimitive(Type type, String key, String value) throws ParseException
```
tweise: Why is there an assumption that every value is String, when conversion is from/to POJO. This also does not support all types that are supported in Avro? Latest Avro version is 1.8.x

avro/src/test/java/org/apache/apex/malhar/avro/AvroFileToPojoModuleTest.java
```diff
@@ -0,0 +1,266 @@
...
+ * <p>
+ * In this class the emitTuples method is called twice to process the first
+ * input, since on begin window 0 the operator is setup & stream is initialized.
+ * The platform calls the emitTuples method in the successive windows
+ * </p>
+ */
+public class AvroFileToPojoModuleTest
+{
+
+  private static final String AVRO_SCHEMA = "{\"namespace\":\"abc\"," + ""
```
tweise: why is this when schemas can be stored as JSON file resources

avro/src/test/java/org/apache/apex/malhar/avro/AvroFileToPojoModuleTest.java
tweise: we actually should worry about this and other issues when promoting an operator to a new module

tweise (changes requested):
I just looked through parts of this PR and there are quite a few issues with it. How can this be declared as ready for merge? As discussed, it is better to focus on one thing and do that right then dealing with a mountain of inherited code issues while attempting to address unrelated issues at once.

contrib/pom.xml
```diff
@@ -678,5 +678,10 @@
       <version>1.3.0</version>
       <optional>true</optional>
     </dependency>
+    <dependency>
```
tweise: This doesn't belong here. Not everyone using contrib wants to pull in the Avro module.

avro/src/main/java/org/apache/apex/malhar/avro/AvroFileInputOperator.java
sandeshh: Few of these comments are on the existing code, which we were not planning to change as a part of this PR. We wanted to improve this operator step by step. Anyways, now it is up to @SaumyaMohan to decide whether she wants to continue with this PR.

SaumyaMohan:
@tweise most of the issues you pointed out are in the existing code which I am not changing. It seems it is easier to make the code change in the old contrib location so I will go ahead and do that and not move it to new location for this PR.

Regarding your question on how parallel partitioning will work for Avro, @sandeshh already replied "Avro operator is an input operator which extends AbstractFileInputOperator, so parallel partitioning behavior is same as AbstractFileInputOperator's behavior."

SaumyaMohan:
Creating a new Pull Request with only the code changes and no file move

PR #667, opened 2017-08-18
APEXMALHAR-2472 Kudu Input Operator implementation

ghost:
@tweise  / @bhupeshchawda / @amberarrow : Could you please review.

I have implemented the Kudu input operator as per the design described in the JIRA ticket given here : https://issues.apache.org/jira/browse/APEXMALHAR-2472


tweise:
@ananthc thanks for the contribution! Can you please move this to a separate Maven module? With the antlr build step it does no longer fit into contrib, and a new module will also allow for proper dependency management. Would it be possible to write some unit tests that can run as part of the CI (by mocking the Kudu dependencies, without requiring an actual Kudu cluster).

ghost:
@tweise  . Sure. I will move to a separate maven module. I will add conditional test rules for junit to ignore all of the current tests which assume a kudu cluster is present and add more tests with some mocking. 

A quick confirmation. By a new module, I understood it as that the new kudu module would be in the same directory as that of kafka ( and move one level up contrib ) . Please let me know if my understanding is wrong.

ghost:
@tweise  - I have made the following changes 
 - Kudu pacakge as a separate module 
-  A conditional trigger of unit tests based on a Junit Rule
- The Kudu cluster based tests are automatically enabled when there is a command line switch "-Dkudumasterhosts=list-of-comm-separated-host-names"

Could you please review or assign this to some one who can complete the review ?

kudu/.gitignore
```diff
@@ -0,0 +1,15 @@
+.DS_Store
```
tweise: Why this file? If needed, then only list what is in addition to top level .gitignore,

kudu/XmlJavadocCommentsExtractor.xsl
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tweise: No need for this. If someone needs it in the future, then a solution that does not duplicate in every module should be found.

kudu/pom.xml
```diff
@@ -0,0 +1,299 @@
...
+              <outputDirectory>${project.basedir}/src/main/java/org/apache/apex/malhar/kudu/sqlparser</outputDirectory>
+            </configuration>
+            <goals>
+              <goal>antlr4</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <!-- The reason we are duplicating the plugin from parent is to take care of checkstyle violations in Antlr4 autogenerated code -->
+      <plugin>
```
tweise: remove this

kudu/pom.xml
```diff
@@ -0,0 +1,299 @@
...
+            </configuration>
+            <goals>
+              <goal>run</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <!-- generate javdoc -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
```
tweise: get rid of this and related configuration, see previous comment

kudu/pom.xml
```diff
@@ -0,0 +1,299 @@
...
+      <groupId>org.apache.kudu</groupId>
+      <artifactId>kudu-client</artifactId>
+      <version>1.4.0</version>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.antlr</groupId>
+      <artifactId>antlr4-runtime</artifactId>
+      <version>4.7</version>
+      <optional>true</optional>
```
tweise: why optional?

kudu/src/main/java/org/apache/apex/malhar/kudu/AbstractKuduOutputOperator.java
```diff
@@ -497,14 +500,14 @@ public void endWindow()
       kuduSession.flush();
     } catch (KuduException e) {
       LOG.error("Could not flush kudu session on an end window boundary " + e.getMessage(), e);
-      throw new RuntimeException(e);
+      DTThrowable.rethrow(e);
```
tweise: use Throwables from guava

kudu/src/main/java/org/apache/apex/malhar/kudu/IncrementalStepScanInputOperator.java
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * <p>A Kudu input operator that provides the functionality of scanning a Kudu table by incrementing
+ * a column in steps as specified in the configuration. Each iteration scans the table as given by the
+ *  SQL expression in the properties file. Other aspects like connection details are also read from the properties file.
+ * </p>
+ *
+ * <p>
+ *   Here is an example property file format.
+ *   Place a file named kuduincrementalstepscaninputoperator.properties anywhere in the classpath ( say by
```
tweise, line 43: What's the advantage using the configuration file vs. configuring individual properties on the operator?

kudu/src/main/java/org/apache/apex/malhar/kudu/partitioner/AbstractKuduInputPartitioner.java
```diff
@@ -0,0 +1,233 @@
...
+  @Override
+  public Collection<Partition<AbstractKuduInputOperator>> definePartitions(
+      Collection<Partition<AbstractKuduInputOperator>> partitions, PartitioningContext context)
+  {
+    List<Partition<AbstractKuduInputOperator>> partitionsForInputOperator = new ArrayList<>();
+    List<KuduPartitionScanAssignmentMeta> assignmentMetaList = new ArrayList<>();
+    try {
+      assignmentMetaList.addAll(getListOfPartitionAssignments(partitions,context));
+    } catch (Exception e) {
+      LOG.error(" Aborting partition planning as Kudu meta data could not be obtained");
```
tweise: Why not wrap the exception with this message instead of logging it?

kudu/src/main/java/org/apache/apex/malhar/kudu/partitioner/KuduOneToManyPartitioner.java
```diff
@@ -0,0 +1,85 @@
...
+   * @param context The context that is provided by the framework when repartitioning is to be executed
+   * @return  Map with key as operator id and value as the a list of assignments that can be assigned to that operator.
+   */
+  @Override
+  public Map<Integer, List<KuduPartitionScanAssignmentMeta>> assign(List<KuduPartitionScanAssignmentMeta> totalList,
+      PartitioningContext context)
+  {
+    Map<Integer,List<KuduPartitionScanAssignmentMeta>> partitionAssignments = new HashMap<>();
+    int partitionCount = getNumberOfPartitions(context);
+    if ( partitionCount <= 0) {
```
tweise, line 61: What causes the partition count to be < 1? If it is a user parameter, it should be a validation error at startup.

kudu/src/main/java/org/apache/apex/malhar/kudu/partitioner/KuduPartitionScanStrategy.java
```diff
@@ -0,0 +1,29 @@
...
+ * under the License.
+ */
+package org.apache.apex.malhar.kudu.partitioner;
+
+/**
+ * Used to set the partition scan.
+ */
+public enum KuduPartitionScanStrategy
+{
+
```
tweise: remove extra blank lines

kudu/src/main/java/org/apache/apex/malhar/kudu/scanner/AbstractKuduPartitionScanner.java
```diff
@@ -0,0 +1,240 @@
...
+  public ApexKuduConnection verifyConnectionStaleness(int indexPos)
+  {
+    ApexKuduConnection apexKuduConnection = connectionPoolForThreads.get(indexPos);
+    checkNotNull(apexKuduConnection, "Null connection not expected while checking staleness of" +
+        " existing connection");
+    if (apexKuduConnection.getKuduSession().isClosed()) {
+      try {
+        apexKuduConnection.close(); // closes the wrapper
+      } catch (Exception e) {
+        LOG.error(" Could not close a possibly stale kudu connection handle ", e);
```
tweise, line 121: if it is an error, then why not throw exception?

kudu/src/main/java/org/apache/apex/malhar/kudu/scanner/KuduPartitionScannerCallable.java
```diff
@@ -0,0 +1,204 @@
...
+
+  public void setValuesInPOJO(RowResult aRow, T payload)
+  {
+    Set<String> columnsUsed = parsedQuery.getKuduSQLParseTreeListener().getListOfColumnsUsed();
+    for (String aColumnName : columnsUsed) {
+      ColumnSchema schemaForThisColumn = tableSchema.get(aColumnName);
+      if (aRow.isNull(aColumnName)) {
+        continue;
+      }
+      switch ( schemaForThisColumn.getType().getDataType().getNumber()) {
```
tweise, line 101: Why not do this with a map of setters that is initialized once?

kudu/src/main/java/org/apache/apex/malhar/kudu/sqlparser/KuduSQLExpressionLexer.java
```diff
@@ -0,0 +1,212 @@
+/**
```
tweise: Why check in the generated files?

tweise (changes requested):

kudu/.gitignore
ghost: It is not needed. Removing this.

kudu/XmlJavadocCommentsExtractor.xsl
ghost: Removed.

kudu/pom.xml
ananthc: Removed

kudu/pom.xml
ghost: Removed

kudu/pom.xml
ghost: Because of check style errors emanating from auto generated code, I had to specify the same plugin with additional configuration highly specific to the package structure of Antlr4 Kudu packages. Hence cannot inherit from parent. If we inherit form parent, kudu specific check style config has to be pushed to the parent thereby polluting it.

kudu/pom.xml
ghost: Removed the optional. This component originally resided in contrib and hence was optional. Since Kudu is its own module, it should no longer be optional as you pointed out.

kudu/src/main/java/org/apache/apex/malhar/kudu/AbstractKuduOutputOperator.java
ghost: Done.

kudu/src/main/java/org/apache/apex/malhar/kudu/IncrementalStepScanInputOperator.java
ghost: Configuration files would mean lesser code in the application class. If the template query were to be changed and redeployed , it does not mean a code change and instead replace the properties file in the class path. However I see what you mean, and hence made changes to support both models. There are additional getters/setters and the default constructor does not throw an exception if the properties file is not in the classpath.

kudu/src/main/java/org/apache/apex/malhar/kudu/partitioner/AbstractKuduInputPartitioner.java
ghost: Done.

kudu/src/main/java/org/apache/apex/malhar/kudu/partitioner/KuduPartitionScanStrategy.java
ghost: done

kudu/src/main/java/org/apache/apex/malhar/kudu/scanner/KuduPartitionScannerCallable.java
ananthc: We do not want to throw an error as this "error" state can be reached if there is no valid query that lands on this Apex partition and the client handle socket times out/becomes stale. Hence only logging the stale state and reinitialising the connection using the initially provided connect info object.

kudu/src/main/java/org/apache/apex/malhar/kudu/sqlparser/KuduSQLExpressionLexer.java
ghost: Thanks for raising this. I initially had to check them in as the IDE was not recognising the automatically generated sources. This caused error markers in the IDE until there was a maven build done properly. Removed all of auto generated files from the source tree now.

kudu/src/main/java/org/apache/apex/malhar/kudu/scanner/AbstractKuduPartitionScanner.java
ghost: We do not want to throw an error as this "error" state can be reached if there is no valid query that lands on this Apex partition for a significant amount of time and the client handle socket times out/becomes stale. Hence only logging the stale state and reinitialising the connection using the initially provided connect info object.

kudu/src/main/java/org/apache/apex/malhar/kudu/scanner/KuduPartitionScannerCallable.java
ghost: The reasoning for this is as follows: 

In the SQL expression that is supported by this module, the user can specify a mapping between a Kudu column and the Pojo field name using the "AS" clause.

For example - "Select kuducol1 AS pojofieldone from kudutable where ......"

Since we are letting a query to be supplied for each iteration and the underlying implementation can choose to map the same column in a previous query to a new Pojo Field in the next query, I had to make this parsing logic based on the query change state and not once for all queries.

kudu/src/main/java/org/apache/apex/malhar/kudu/partitioner/KuduOneToManyPartitioner.java
ghost: I have used the approach of letting the user specify the number of apex partitions in two ways;

1. As part of the application config xml ( the usual approach )
2. As part of the code ( setter method )

For the setter method approach, I am having an initial of -1 as the default value to denote that the user did not set anything ( as it can be optional ) . Once the user sets it explicitly , the value is to be used when the partitioning logic is being executed ( thus overriding the app config xml value ).  We could throw an exception in the setter method but having a forced exception at setter level looked a bit ugly. Hence the check when doing the actual planning and forcing it to a minimum value.

kudu/src/main/java/org/apache/apex/malhar/kudu/sqlparser/KuduSQLExpressionLexer.java
ghost: @tweise / @vrozov  - The build is failing on Travis because the auto-generated files do not have the license headers. The license check is failing the build. Please advise if there is an approach that is followed in such situations.

kudu/src/main/java/org/apache/apex/malhar/kudu/scanner/KuduPartitionScannerCallable.java
tweise: Upon further inspection it appears that this is done for every row, but the result columns remain the same for a given query. I don't think this needs to be changed now. I thought of building a list if extract/set functions that work based on column index instead of names.

kudu/src/main/java/org/apache/apex/malhar/kudu/scanner/AbstractKuduPartitionScanner.java
tweise: OK, when this isn't an error then why log it as error? If it is a condition the user cannot do anything about and is of no consequence, why not debug or info?

kudu/src/main/java/org/apache/apex/malhar/kudu/sqlparser/KuduSQLExpressionLexer.java
tweise: Is the solution to generate the code under the target folder and not under src? That may also eliminate the need to do special things for gitignore or checkstyle.

kudu/src/main/antlr4/KuduSQLExpression.g4
```diff
@@ -0,0 +1,154 @@
+grammar KuduSQLExpression;
```
vrozov: Apache license header?

kudu/src/main/java/org/apache/apex/malhar/kudu/scanner/AbstractKuduPartitionScanner.java
ghost: There are two aspects to these few lines of code. The context is we are trying to close a session that is stale and then building a new connection. While invoking the close() method on the connection, it is forcing a checked exception ( and my understanding is that it should not result in an exception even if the session has become stale). Hence was logging it as an error for the aspect of closing the connection ( and only if we get an exception while closing ). The next line is a log statement that that is at level Info stating that the connection is being rebuilt.

kudu/src/main/java/org/apache/apex/malhar/kudu/scanner/KuduPartitionScannerCallable.java
ghost: Perhaps I misunderstood your first comment on this line and hence giving more context. Please let me know if you think there is something more that needs to be done here. 

- The input operator scanner only returns those columns that were asked as part of the SELECT statement ( and does not return all the columns of a table ) and hence is highly dependent on the query being processed. This is for performant reasons as you can guess. 
-  Hence the setters are getting decided only at the beginning of the query processing which you are pointing out.
- This set of setters are actually in another class AbstractKuduInputOperator#extractSettersForResultObject()
- This collection of setter definitions are sent to all threads that are establishing the connection to the Kudu Tablet and retrieving the rows that this Apex operator is responsible for.
- This is done once per query and then passed to the scanner thread as part of its constructor
- In the Scanner thread , after a row is obtained from the Kudu client, only the columns that are returned are used to invoke the setters for those columns.
- We are not defining the setters but only calling the setter to set the value from the Kudu Row into the Pojo Field.
- A POJO field can be a superset of the Kudu columns in terms of mapping. I was thinking that will result in better usability of POJOs as we do not need a separate POJO across operators if required.

kudu/src/main/antlr4/KuduSQLExpression.g4
ghost: Thanks for catching this. Will add the license. It looks like Antlr4 has issues with comments starting with a double asterisk. /**. Will add the license header. Hope the license check does not fail if it looks for a double asterisk at the beginning.

kudu/src/main/java/org/apache/apex/malhar/kudu/sqlparser/KuduSQLExpressionLexer.java
ghost: Thanks for the suggestion. This is a good idea. I will explore to see how the Antlr4 plugin can be coerced to do this. I will reply as a comment as soon as I am done with this experiment.

ghost:
@tweise  - Thanks for the trick. I changed the approach to generate the code under the generated-sources folder. The IDE was giving an error ( as I believe some IDEs need specialised plugins and/or additional setup by Devs ). This was solved by using the build-helper plugin in the pom which also took care of automatically by all of the IDEs.

@vrozov  - The grammar file now has the Apache license as well.

tweise:
@ananthc there are several classes that are not marked `@Evolving`. I would suggest to mark everything evolving since this is brand new code that will be difficult to change once released unless it is marked evolving. The easiest way to accomplish this is to put a package-info file for each package and annotate that. That would also be an opportunity to provide a package level javadoc.

kudu/src/main/java/org/apache/apex/malhar/kudu/AbstractKuduOutputOperator.java
```diff
@@ -339,7 +343,7 @@ private void performCommonProcessing(Operation currentOperation, KuduExecutionCo
       kuduSession.apply(currentOperation);
     } catch (KuduException e) {
       LOG.error("Could not execute operation because " + e.getMessage(), e);
-      throw new RuntimeException(e.getMessage());
+      Throwables.propagateIfPossible(e);
```
tweise: This isn't propagating all exceptions. See javadoc, you still need to throw non propagated exceptions. Also, why use logging in this case? If you have a custom message, then throw new RuntimeException(msg, e), will be logged by the engine anyways.

kudu/src/main/java/org/apache/apex/malhar/kudu/BaseKuduOutputOperator.java
```diff
@@ -53,7 +52,7 @@
  */
 public class BaseKuduOutputOperator extends AbstractKuduOutputOperator
 {
-  public static final String DEFAULT_CONNECTION_PROPS_FILE_NAME = "kuduoutputoperator.properties";
+  public static String DEFAULT_CONNECTION_PROPS_FILE_NAME = "kuduoutputoperator.properties";
```
tweise, line 55: why not final?

kudu/src/test/java/org/apache/apex/malhar/kudu/test/KuduClusterAvailabilityTestRule.java
```diff
@@ -0,0 +1,161 @@
...
+   * @return
+   */
+  private String rewritePropsFile(String masterHosts, String originalFileName)
+  {
+    Properties kuduConnectionProperties = new Properties();
+    ClassLoader loader = Thread.currentThread().getContextClassLoader();
+    InputStream kuduPropsFileAsStream = loader.getResourceAsStream(originalFileName);
+    File tempPropertiesFile = null;
+    try {
+      tempPropertiesFile = File.createTempFile(System.getProperty("java.io.tmpdir"),".properties");
```
tweise: The first parameter is a temp file prefix, not need to get the system property.

kudu/src/test/java/org/apache/apex/malhar/kudu/test/KuduClusterAvailabilityTestRule.java
```diff
@@ -0,0 +1,161 @@
...
+        runThisTest = true;
+      }
+    }
+    if (runThisTest) {
+      // call the before class equivalent for real kudu cluster
+      if ((testContext != null ) && (testContext.kuduClusterBasedTest())) {
+        try {
+          initRealKuduClusterSetup(masterHostsValue);
+        } catch (Exception e) {
+          LOG.error(e.getMessage(), e);
```
tweise: why LOG.error? If it is a real problem, propagate exception.

kudu/src/test/java/org/apache/apex/malhar/kudu/sqltranslator/SQLToKuduPredicatesTranslatorTest.java
```diff
@@ -0,0 +1,156 @@
+/**
```
tweise, line 1: nice test coverage 👍

kudu/pom.xml
```diff
@@ -0,0 +1,164 @@
...
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>malhar-library</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+      <classifier>tests</classifier>
+    </dependency>
+    <dependency>
```
pramodin, line 113: Shouldn't the scope be provided

kudu/pom.xml
```diff
@@ -0,0 +1,164 @@
...
+      <groupId>org.codehaus.janino</groupId>
+      <artifactId>janino</artifactId>
+      <version>2.7.8</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.kudu</groupId>
+      <artifactId>kudu-client</artifactId>
+      <version>1.4.0</version>
```
pramodin: Can you make this a property, kudu.version

kudu/pom.xml
```diff
@@ -0,0 +1,164 @@
...
+      <artifactId>antlr4-runtime</artifactId>
+      <version>4.7</version>
+    </dependency>
+    <dependency>
+      <groupId>com.conversantmedia</groupId>
+      <artifactId>disruptor</artifactId>
+      <version>1.2.10</version>
+      <classifier>jdk7</classifier><!-- Set classifier to jdk8 when malhar switches to JDK 8 -->
+    </dependency>
+    <dependency>
```
pramodin: Not necessary to be addressed in this PR but we should try to put the mocking test dependencies in a parent pom.

kudu/src/main/java/org/apache/apex/malhar/kudu/scanner/KuduPartitionScannerCallable.java
tweise: Yes, the columns that need to be processed depend on the query and not what columns a given table has. This is similar to the JDBC case where you use the result set metadata and not the table metadata. My idea was that you could build a list of functions that handle both, get from result and set on object that are subsequently just referenced by index for processing the rows. Not in the scope of this PR though, we can discuss it separately.

kudu/pom.xml
tweise: No, it is a test dependency for testing only. Besides, malhar dependencies are never provided scope.

pramodin: Not malhar, referring to apex-common

tweise: Yep, and that one is another potential candidate for parent pom, in case it is used in most modules (to be investigated).

pramodin: On a tangential note, I don't see much benefit of separating api and common anymore as practically all apps use common as a dependency instead of api, and I don't know of an instance where api is used standalone. Common can be merged into api and an empty artifact that depends on the api, can be left in place for backward compatibility

kudu/src/main/java/org/apache/apex/malhar/kudu/ApexKuduConnection.java
```diff
@@ -128,8 +131,20 @@ public void setKuduClient(KuduClient kuduClient)
...
+  {
+    return builderForThisConnection;
+  }
+
+  public void setBuilderForThisConnection(ApexKuduConnectionBuilder builderForThisConnection)
   {
+    this.builderForThisConnection = builderForThisConnection;
+  }
+
+  public static class ApexKuduConnectionBuilder implements Serializable
```
pramodin, line 141: Any specific reason this is marked Serializable? By default kryo is used to serialize the state and it doesn't need classes to implement Serializable.

kudu/src/main/java/org/apache/apex/malhar/kudu/AbstractKuduOutputOperator.java
ghost: Ended up reverting to throw RuntimeException(e) because of the reasoning provided here :  https://github.com/google/guava/wiki/Why-we-deprecated-Throwables.propagate

It looks like we need up upgrade guava to a higher version if we need to use the new preferred API ( along with the RuntimeException ) . Since it might disrupt a whole lot of other things, ended up just using the RuntimeException.

kudu/src/test/java/org/apache/apex/malhar/kudu/test/KuduClusterAvailabilityTestRule.java
ghost: Fixed.

kudu/src/test/java/org/apache/apex/malhar/kudu/test/KuduClusterAvailabilityTestRule.java
ghost: Fixed.

kudu/pom.xml
ghost: Done.

kudu/src/main/java/org/apache/apex/malhar/kudu/BaseKuduOutputOperator.java
ghost: This was done primarily to account for flexible unit testing. 

Kudu cluster based tests are being enabled by just passing in the master hosts definition from the command line. The rest of the properties for Kudu cluster are being enabled via properties file definition. So a temp file is being created whenever the kudu cluster master hosts is passed from the command line. Since we are using the temp file creation API ( your review comment below ) , the file name is being controlled by the API as opposed to us locking it down. Hence removed the "final" qualifier.

kudu/src/main/java/org/apache/apex/malhar/kudu/ApexKuduConnection.java
ghost: Was more thinking of Java serialisability support. I can remove it but it would not harm either if java serialisation needs to be considered for all base operator derived classes.

kudu/src/main/java/org/apache/apex/malhar/kudu/BaseKuduOutputOperator.java
tweise: I think it would be better if environment specific parameters are configured outside of the build (you may have different environments that you want to deploy a build to). Unit tests should also be able to set operator properties to override defaults. Can be done as follow-up PR.

kudu/pom.xml
```diff
@@ -0,0 +1,167 @@
...
+      <artifactId>malhar-library</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+      <classifier>tests</classifier>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>apex-common</artifactId>
+      <version>${apex.core.version}</version>
+      <type>jar</type>
```
tweise, line 117: @ananthc this is the dependency that needs to be scope provided

tweise (changes requested):

kudu/pom.xml
ghost: Done.

kudu/src/main/java/org/apache/apex/malhar/kudu/BaseKuduOutputOperator.java
ghost: I will work on this as an iteration. There are a few more things like security that need to be implemented for this module and will take this up as part of that iteration.

In the current state, the environment is flipped by passing in the kudu cluster environment as a System property using the "-Dkudumasterhosts=xxx.yyy.zzz.aaa" approach thereby allowing configurability for just the Kudu cluster. However the table to be used, the Pojo that represents the table etc are being read from a properties file. The unit test config creates the tables and subsequently uses them in the tests. 

Just to understand your comment a bit more, we would like these configs to be picked up from environment variables instead of the System property ?

tweise: Not a system property, but operator properties. Something like this here: https://github.com/apache/apex-malhar/blob/master/examples/exactly-once/src/test/java/org/apache/apex/examples/exactlyonce/ExactlyOnceFileOutputAppTest.java#L79

tweise (approved):

tweise:
@ananthc I created tickets for test issues that have nothing to do with this PR. Two subsequent failures due to https://issues.apache.org/jira/browse/APEXMALHAR-2544

Kudu tests pass.

pramodin (approved):

PR #670, opened 2017-09-12
APEXMALHAR-2034 Adding new Avro Module to encapsulate Container File to Avro GenericRecord to POJO transformations

SaumyaMohan:


SaumyaMohan:
Link to previously closed PR: https://github.com/apache/apex-malhar/pull/666
(Avro files were moved from contrib to new maven module in old PR. Since history of files were moved too, creating this new PR to just pick the code changes)

Updated the JIRA to refer to only code changes. Removing the scope of moving Avro files to new maven module

SaumyaMohan:
@tweise please could you review and let me know if the changes look good?

contrib/src/test/java/com/datatorrent/contrib/avro/AvroFileToPojoModuleTest.java
```diff
@@ -0,0 +1,260 @@
...
+    public void setAvroFileToPojoModule(AvroFileToPojoModule avroFileToPojoModule)
+    {
+      this.avroFileToPojoModule = avroFileToPojoModule;
+    }
+
+    @Override
+    public void populateDAG(DAG dag, Configuration conf)
+    {
+      AvroFileToPojoModule avroFileToPojoModule = dag.addModule("avroFileToPojoModule", getAvroFileToPojoModule());
+      ConsoleOutputOperator consoleOutput = dag.addOperator("console", new ConsoleOutputOperator());
```
tweise, line 189: Since the application is written for the test only, there is no need to use a console operator. You can specify a collector that allows you to verify the results instead. Or, adopt the approach discussed on #660

contrib/src/test/java/com/datatorrent/contrib/avro/AvroFileToPojoModuleTest.java
```diff
@@ -0,0 +1,260 @@
...
+
+      EmbeddedAppLauncherImpl lma = new EmbeddedAppLauncherImpl();
+      Configuration conf = new Configuration(false);
+
+      AvroToPojoApplication avroToPojoApplication = new AvroToPojoApplication();
+      avroToPojoApplication.setAvroFileToPojoModule(avroFileToPojoModule);
+
+      lma.prepareDAG(avroToPojoApplication, conf);
+      EmbeddedAppLauncherImpl.Controller lc = lma.getController();
+      lc.run(10000);// runs for 10 seconds and quits
```
tweise, line 165: Since you want to verify that the operator is working, you need to check the results and the test only needs to run until these results are available. For use of non-deprecated  API and wait until results example, see https://github.com/apache/apex-malhar/blob/master/examples/exactly-once/src/test/java/org/apache/apex/examples/exactlyonce/ExactlyOnceJdbcOutputTest.java#L118

contrib/src/main/java/com/datatorrent/contrib/avro/AvroFileToPojoModule.java
```diff
@@ -0,0 +1,91 @@
...
+  public final transient ProxyOutputPort<String> avroErrorRecordsPort = new ProxyOutputPort<>();
+
+  private AvroFileInputOperator avroFileInputOperator = new AvroFileInputOperator();
+  Class<?> pojoClass = null;
+
+  @Override
+  public void populateDAG(DAG dag, Configuration configuration)
+  {
+    AvroFileInputOperator avroFileInputOperator = dag.addOperator("AvroFileInputOperator", this.avroFileInputOperator);
+    AvroToPojo avroToPojo = dag.addOperator("AvroGenericObjectToPojo", new AvroToPojo());
```
tweise, line 68: I think this question is still open: how would this operator be parallel partitioned?

Example:
```
file -> pojo \
                    -> output
file -> pojo /
```

tweise (changes requested):

pramodin:
@SaumyaMohan, @sandeshh, @tweise I am joining late to this discussion but it seems the PR originally created a new module but there was an objection that it was not discussed. Why don't you start a discussion on dev about creating a new module. Recent trend has been to put operators in their own modules, see kudu PR for example, which helps users because they can include exactly the operators they want and don't have to keep track of transitive dependencies.

tweise:
@PramodSSImmaneni there already has been related discussion on previous PRs and mailing list. In this case, we have seen that the majority of the work of creating a new Maven module is unrelated to the objective of adding the composite operator and the contributor is not ready to take up that work. Therefore I would rather see this PR completed and then, if contributor or anyone else is interested, take up new module separately, which would start on the mailing list.

pramodin:
@tweise yes that should be fine

tweise:
Merging and marking JIRA incomplete.

PR #671, opened 2017-09-27
APEXMALHAR-2542-Fix test failures by altering the location property

ghost:
@tweise  - Could you please review. The fix was to change the properties relative to the JVM Unit test launch path. Did not end up using the code snippet you provided as an example as this did not need us to make any code changes.


tweise:
In general you want to separate test setup from standalone defaults so that the user when running such example from the CLI has reasonable defaults (that's the purpose of the test snipped I pointed you to). In this case, pre-existing source property would require a change regardless, so it is OK. 

Travis build failure may be intermittent or another flaky test. Please rebase to trigger Travis CI again.

BTW did you verify manually that the enricher test "passes" since the test does not do any verification :)

examples/enricher/src/main/resources/META-INF/properties.xml
```diff
@@ -37,13 +37,13 @@
   </property>
   <property>
     <name>dt.application.EnricherAppWithJSONFile.operator.Enrich.prop.store.fileName</name>
-    <value>/tmp/circleMapping.txt</value>
+    <value>src/main/resources/circleMapping.txt</value>
```
tweise: maybe for clarity prefix with './' both of the values

ghost: Done

ghost:
@tweise  - Did not even attempt a manual verification of the console output :) . Ended up altering the logic of the application ( Logged the output to a file under the target directory ) instead of writing to the console. Subsequently added some asserts which should match the expected counts with the numbers by rereading the generated output file.

examples/enricher/src/test/java/org/apache/apex/examples/enricher/ApplicationTest.java
```diff
@@ -46,6 +56,36 @@ public void testApplication() throws Exception
     } catch (ConstraintViolationException e) {
```
tweise: The check is great, can you also remove the timing dependency? Instead of running for 10s, have it run until the expected result appears (check in loop) or timeout. See https://github.com/apache/apex-malhar/blob/master/examples/exactly-once/src/test/java/org/apache/apex/examples/exactlyonce/ExactlyOnceJdbcOutputTest.java#L118

Perhaps also use the new embedded mode API and get rid of the exception catch.

ghost:
@tweise  - Changed the to use the Embedded launcher. Used a output check logic to start asserts only after the expected output processing is complete.

examples/enricher/src/test/java/org/apache/apex/examples/enricher/ApplicationTest.java
```diff
@@ -36,16 +44,65 @@
...
+    EmbeddedAppLauncher<?> launcher = Launcher.getLauncher(Launcher.LaunchMode.EMBEDDED);
+    Attribute.AttributeMap launchAttributes = new Attribute.AttributeMap.DefaultAttributeMap();
+    launchAttributes.put(EmbeddedAppLauncher.RUN_ASYNC, true);
+    EnricherAppWithJSONFile enricherAppWithJSONFile = new EnricherAppWithJSONFile();
+    Launcher.AppHandle appHandle = launcher.launchApp(enricherAppWithJSONFile, conf, launchAttributes);
+    File folder = new File("target/enrichedData");
+    File[] listOfFiles = null;
+    // First ensure that the output file is generated and the expected lines match
+    // The test main thread sleeps until the expected lines are generated in the output file
+    while (true) {
```
tweise: should have a timeout

ghost: Added a timeout and refactored the code a bit as the test function was turning out to be too long.

tweise (approved):

tweise:
@ananthc I sent you a PR for replacing the file output with embedded result collector, with which the test completed in 3s vs > 30s (file output latency).

PR #683, opened 2017-12-26
APEXMALHAR-2260.PythonExecutionOperator-Execute python code as part of DAG

ghost:
@tweise / @patilvikram  - Could you please review or assign someone who could help me with the review of this PR ?

### DIFF
--- a/python/pom.xml
+++ b/python/pom.xml
@@ -1,0 +1,94 @@
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache.apex</groupId>
+    <artifactId>malhar</artifactId>
+    <version>4.0.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>malhar-python</artifactId>
+  <name>Apex library python support</name>
+  <packaging>jar</packaging>
+
+  <properties>
+    <jep-version>3.7.0</jep-version>
+    <jepInstallPath>/usr/local/lib/python3.5/site-packages/jep</jepInstallPath> <!-- Purposefully empty as it is ideally set from command line basing on env but
+     avoid errors in IDEs validating a pom.xml which requires a property to be set when referred in the configs-->
+    <disruptor-queue-conversant-media-version>1.2.10</disruptor-queue-conversant-media-version>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>malhar-library</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>malhar-library</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+      <classifier>tests</classifier>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>apex-common</artifactId>
+      <version>${apex.core.version}</version>
+      <type>jar</type>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.codehaus.janino</groupId>
+      <artifactId>janino</artifactId>
+      <version>2.7.8</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>black.ninia</groupId>
+      <artifactId>jep</artifactId>
+      <version>${jep-version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.conversantmedia</groupId>
+      <artifactId>disruptor</artifactId>
+      <version>${disruptor-queue-conversant-media-version}</version>
+      <classifier>jdk7</classifier><!-- Set classifier to jdk8 when malhar switches to JDK 8 -->
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <forkMode>once</forkMode>
+          <argLine>-Djava.library.path=${jepInstallPath}</argLine>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+</project>
--- a/python/src/main/java/org/apache/apex/malhar/python/base/ApexPythonEngine.java
+++ b/python/src/main/java/org/apache/apex/malhar/python/base/ApexPythonEngine.java
@@ -1,0 +1,185 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.python.base;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.BlockingQueue;
+
+import org.apache.apex.malhar.python.base.requestresponse.PythonInterpreterRequest;
+import org.apache.apex.malhar.python.base.requestresponse.PythonRequestResponse;
+
+/**
+ * Defines the methods that needs to be implemented by the Python Engine implementations. The first implementation
+ *  takes the approach of in-memory interpreter using JEP. Other possibilities are using Py4J which is an inter process
+ *   communication model. An Apex operator would use an instance of the Python engine implementations to run
+ * python code using the chosen engine.
+ */
+public interface ApexPythonEngine
+{
+  /***
+   * Used to perform any pre interpreter processing.
+   * @param preInitConfigs The configuration that is going to be used by the interpreter
+   * @throws ApexPythonInterpreterException if there is an issue in executing the pre interpreter logic
+   */
+  void preInitInterpreter(Map<PythonInterpreterConfig,Object> preInitConfigs) throws ApexPythonInterpreterException;
+
+  /***
+   * Starts the interpreter.
+   * @throws ApexPythonInterpreterException if library not locatable or any other issue starting the interpreter
+   */
+  void startInterpreter() throws ApexPythonInterpreterException;
+
+  /***
+   * Used to perform any logic that needs to be executed after the interpreter is started but before any tuples start
+   *  getting processed. Example, setting the starting state of the variables that are used in tuple processing.
+   * @throws ApexPythonInterpreterException
+   */
+  void postStartInterpreter() throws ApexPythonInterpreterException;
+
+  /***
+   * Runs a series of commands. The implementation engine could make use of a worker pool to execute the command.
+   * @param executionMode Whether these commands need to be run on all worker thread or any of the worker thread.Please see
+   *    *                       {@link WorkerExecutionMode} for choices available
+   * @param windowId used to select the worker from the worker thread pool. This parameter gains significance if
+   *                 we want to implement a sticky worker in the near future. This will allow for a basic approach to
+   *                 route the command/s to the same worker if the application logic needs it to be. In the case of ANY
+   *                 worker logic, the window ID along with the Request ID is used to implement a round robin approach
+   *                 to select the next worker. Note that Sticky worker might be required since python interpreter
+   *                 state is accumulated over as the commands run and a command can reference a variable created in a
+   *                 previous command etc. Such references might want to route all commands to a specific interpreter
+   *                 instance. If the Apex python engine is not being used by an operator implementation directly,
+   *                 the caller can pass in any number as it is not used in anything more than selecting a worker from a
+   *                 worker pool.
+   * @param requestId The parameter is used to select a worker from the
+   *                  worker pool along with the window Id. If the Apex python engine is not being used by an
+   *                  operator implementation directly, the caller can pass in any number as it is not used in anything
+   *                  more than selecting a worker from a worker pool.
+   * @param request Represents the request to be processed.
+   * @return A map with key as the command run and boolean as the value. True represents that the command successfully
+   *  run.
+   * @throws ApexPythonInterpreterException if interrupted or if the command cannot be executed
+   */
+  Map<String,PythonRequestResponse<Void>> runCommands(WorkerExecutionMode executionMode, long windowId, long requestId,
+      PythonInterpreterRequest<Void> request) throws ApexPythonInterpreterException;
+
+  /***
+   * Executes a method call
+   * @param executionMode If the method call needs to be invoked on all workers or any single worker. Please see
+   *    *                       {@link WorkerExecutionMode} for choices available
+   * @param windowId used to select the worker from the worker thread pool. This parameter gains significance if
+   *                 we want to implement a sticky worker in the near future. This will allow for a basic approach to
+   *                 route the command/s to the same worker if the application logic needs it to be. In the case of ANY
+   *                 worker logic, the window ID along with the Request ID is used to implement a round robin approach
+   *                 to select the next worker. Note that Sticky worker might be required since python interpreter
+   *                 state is accumulated over as the commands run and a command can reference a variable created in a
+   *                 previous command etc. Such references might want to route all commands to a specific interpreter
+   *                 instance. If the Apex python engine is not being used by an operator implementation directly,
+   *                 the caller can pass in any number as it is not used in anything more than selecting a worker from a
+   *                 worker pool.
+   * @param requestId The parameter is used to select a worker from the
+   *                  worker pool along with the window Id. If the Apex python engine is not being used by an
+   *                  operator implementation directly, the caller can pass in any number as it is not used in anything
+   *                  more than selecting a worker from a worker pool.
+   * @param req Represents the request to be processed.
+   * @param <T>
+   * @return A map containing the worker ID as key and boolean as successful or not
+   * @throws ApexPythonInterpreterException
+   */
+  <T> Map<String,PythonRequestResponse<T>> executeMethodCall(WorkerExecutionMode executionMode,long windowId,
+      long requestId, PythonInterpreterRequest<T> req) throws ApexPythonInterpreterException;
+
+  /***
+   * Executes a script that is locatable via a file path
+   * @param executionMode  If the method call needs to be invoked on all workers or any single worker. Please see
+   *                       {@link WorkerExecutionMode} for choices available
+   * @param windowId used to select the worker from the worker thread pool. This parameter gains significance if
+   *                 we want to implement a sticky worker in the near future. This will allow for a basic approach to
+   *                 route the command/s to the same worker if the application logic needs it to be. In the case of ANY
+   *                 worker logic, the window ID along with the Request ID is used to implement a round robin approach
+   *                 to select the next worker. Note that Sticky worker might be required since python interpreter
+   *                 state is accumulated over as the commands run and a command can reference a variable created in a
+   *                 previous command etc. Such references might want to route all commands to a specific interpreter
+   *                 instance. If the Apex python engine is not being used by an operator implementation directly,
+   *                 the caller can pass in any number as it is not used in anything more than selecting a worker from a
+   *                 worker pool.
+   * @param requestId The parameter is used to select a worker from the
+   *                  worker pool along with the window Id. If the Apex python engine is not being used by an
+   *                  operator implementation directly, the caller can pass in any number as it is not used in anything
+   *                  more than selecting a worker from a worker pool.
+   * @param request Represents the request to be processed.
+   * @return A map containing the worker ID as key and boolean as successful or not
+   * @throws ApexPythonInterpreterException
+   */
+  Map<String,PythonRequestResponse<Void>>  executeScript(WorkerExecutionMode executionMode,long windowId,long requestId,
+      PythonInterpreterRequest<Void> request) throws ApexPythonInterpreterException;
+
+  /***
+   * Evaluates a string as a python expression and also supports passing in variables from JVM to the python interpreter
+   * @param executionMode If the method call needs to be invoked on all workers or any single worker. Please see
+   *    *                       {@link WorkerExecutionMode} for choices available
+   * @param windowId used to select the worker from the worker thread pool. This parameter gains significance if
+   *                 we want to implement a sticky worker in the near future. This will allow for a basic approach to
+   *                 route the command/s to the same worker if the application logic needs it to be. In the case of ANY
+   *                 worker logic, the window ID along with the Request ID is used to implement a round robin approach
+   *                 to select the next worker. Note that Sticky worker might be required since python interpreter
+   *                 state is accumulated over as the commands run and a command can reference a variable created in a
+   *                 previous command etc. Such references might want to route all commands to a specific interpreter
+   *                 instance. If the Apex python engine is not being used by an operator implementation directly,
+   *                 the caller can pass in any number as it is not used in anything more than selecting a worker from a
+   *                 worker pool.
+   * @param requestId The parameter is used to select a worker from the
+   *                  worker pool along with the window Id. If the Apex python engine is not being used by an
+   *                  operator implementation directly, the caller can pass in any number as it is not used in anything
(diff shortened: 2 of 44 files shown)


tweise:
will look at it soon

python/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache.apex</groupId>
+    <artifactId>malhar</artifactId>
+    <version>3.9.0-SNAPSHOT</version>
```
tweise: rebase

python/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <forkMode>once</forkMode>
+          <argLine>-Djava.library.path=/usr/local/lib/python3.5/site-packages/jep</argLine>
```
tweise: shouldn't this be a property

python/src/main/java/org/apache/apex/malhar/python/base/ApexPythonEngine.java
```diff
@@ -0,0 +1,132 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.BlockingQueue;
+
+import org.apache.apex.malhar.python.base.requestresponse.PythonInterpreterRequest;
+import org.apache.apex.malhar.python.base.requestresponse.PythonRequestResponse;
+
+/**
+ * Defines the methods that needs to be implemented by the Python operator implementations. The first implementation
```
tweise: The interface needs to be implemented by? Looks like an operator would use it to interface with a Python engine?

python/src/main/java/org/apache/apex/malhar/python/base/ApexPythonInterpreterException.java
```diff
@@ -0,0 +1,35 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.apex.malhar.python.base;
+
+/**
+ * An exception used to denote issues when using the ApexPython Engine.
+ */
+public class ApexPythonInterpreterException extends Exception
```
tweise: make this an unchecked exception?

python/src/main/java/org/apache/apex/malhar/python/base/PythonInterpreterConfig.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.Map;
+
+import org.apache.apex.malhar.python.base.jep.JepPythonEngine;
+
+/***
+ * Used as key configs while passing the pre interpreter configuration. See
+ *  {@link JepPythonEngine#preInitInterpreter(Map)}
+ */
+public enum PythonInterpreterConfig
```
tweise, line 29: use Apex attributes instead, since they provide type information and defaults

python/src/main/java/org/apache/apex/malhar/python/base/ApexPythonEngine.java
```diff
@@ -0,0 +1,132 @@
...
+   * @throws ApexPythonInterpreterException if library not locatable or any other issue starting the interpreter
+   */
+  void startInterpreter() throws ApexPythonInterpreterException;
+
+  /***
+   * Used to perform any logic that needs to be executed after the interpreter is started but before any tuples start
+   *  getting processed. Example, setting the starting state of the variables that are used in tuple processing.
+   * @throws ApexPythonInterpreterException
+   */
+  void postStartInterpreter() throws ApexPythonInterpreterException;
```
tweise, line 54: why can such logic not be included in startInterpreter?

python/src/main/java/org/apache/apex/malhar/python/base/ApexPythonEngine.java
```diff
@@ -0,0 +1,132 @@
...
+
+  /***
+   * Runs a series of commands.
```
tweise: Here the concept of a worker appears suddenly. How are such workers setup and managed? I assumed that an interpreter is either embedded or a single process?

python/src/main/java/org/apache/apex/malhar/python/base/WorkerExecutionMode.java
```diff
@@ -0,0 +1,43 @@
...
+ *   deserializing a machine learning model can be used as a ALL_WORKERS model as the scoring can then be invoked across
+ *   all worker threads if required.
+ *  - {@link WorkerExecutionMode#ANY_WORKER} Represents a state wherein any worker can be used to execute the code.
+ *   Example would be scoring an incoming tuple on which the model has already been deserialized across all nodes
+ *  - {@link WorkerExecutionMode#STICKY_WORKER} Use STICKY_WORKER if the same worker needs to service the request.
+ *    The downside of this is that it may or may not complete on time and depends on the queue length.</p>
+ *
+ *   <p><b>Ensure the {@link PythonInterpreterRequest#hashCode()} is overridden if STICKY_WORKER is chosen </b></p>
+ */
+public enum WorkerExecutionMode
```
tweise: Perhaps the term "Worker" can be misleading here. Maybe just "ExecutionMode"? Values don't need to repeat "WORKER" also. ALL == BROADCAST?

python/src/main/java/org/apache/apex/malhar/python/base/ApexPythonEngine.java
```diff
@@ -0,0 +1,132 @@
...
+   * @param executionMode Whether these commands need to be run on all worker nodes or any of the worker node
+   * @param windowId used to select the worker from the worker pool.Can be any long if an operator is not using this.
```
tweise: windowId and requestId need more explanation. Not clear how they "select" a worker?

tweise (commented):
Please rebase the PR so that some imports get adjusted for 4.x

patilvikram:
@ananthc  Is it possible to share high-level document about this approach explaining all concepts.

tweise:
@patilvikram see fairly detailed email thread: https://lists.apache.org/thread.html/d01335f8aa7b24ddc79fffe022e7344f66b245084328e2ee43f7d178@%3Cdev.apex.apache.org%3E

ghost:
@tweise  - Apologies for the delayed response on this PR. Was not having connectivity to work on the review comments.

python/pom.xml
ghost: rebased

python/pom.xml
ghost: Changed it to be a java program argument based approach.  The original approach was to use a boolean to signify if Python JEP tests are to be executed. With this change, the value is now a string representing the location of the JEP dynamic library location. The -D switch uses the key jepInstallPath and value is the File system location.

Not specifying the above key will result in all JEP based unit tests automatically skipped as part of the build process. 

Sample invocation for JEP based tests is "mvn clean install -DjepInstallPath==/Users/userA/py-environments/jep/lib/python3.6/site-packages/jep"

The tests are executed in a forked JVM as the java.library.path cannot be manipulated within the primary maven JVM

python/src/main/java/org/apache/apex/malhar/python/base/ApexPythonEngine.java
ghost: Yes. You are correct. The interface needs to be implemented by the python implementation ( and not by the operator and hence is not a operator interface ). Since there can be multiple implementations possible for python engine implementations, this interface is just for the engine implementations to implement as a contract. 

Updated the javadocs to reflect this more clearly.

python/src/main/java/org/apache/apex/malhar/python/base/PythonInterpreterConfig.java
ghost: I have added more logic in the BasePythonExecutionOperator to cater the configuration elements as Operator properties. 

I am a bit confused about defining them as Apex attributes. I am assuming you are referring to the approach of setting these properties via configuration xml files. Changed the logic for the operator to "relay" the operator configured properties to the python engine implementations.

I have consciously separated python engine implementation from the aspects of Apex operator implementations as this might help us in the future use cases like Beam hybrid computation models.

python/src/main/java/org/apache/apex/malhar/python/base/ApexPythonEngine.java
ghost: The startInterpreter() deals with booting up the engine. For in memory implementations this means creating instances of the interpreter. For remote instances, it might mean creating IPC channels to an existing python process ( ex: Py4J implementations ) 

The interpreter state is preserved as long as the interpreter is not disposed off. ( i.e. variables created in various calls are available for reuse in subsequent calls etc). The postStartInterpreter  hook is provided for Apex operators to implement use cases like deserialise an ML model into memory etc. This is then used for every tuple call. The startInterpreter is more for the engine start logic whereas the postStartInterpreter is for the business logic implementation hooks like executing the python code which needs to be done once after the interpreter engine is started ( in the startInterpreter() method)

python/src/main/java/org/apache/apex/malhar/python/base/ApexPythonEngine.java
ghost: Added more javadocs to explain the concept of a worker. 

The workers are setup automatically by the python engine implementations. The number of workers is configurable by calling a setter that sets the number of threads in the worker thread pool.

Here is a brief details:

As part of the engine initialisation, a worker pool is created with each worker given one request queue and a response queue. The worker thread keeps listening on the request queues to perform work and submit the responses back. The Apex operator drains off the responses. The stragglers are emitted into a straggler port and the tuples which were processed within the SLAs are emitted into the normal port. The BasePythonExecution operator takes care of all of this logic. 

The javadocs inside InterpreterThread.java and InterpreterThreadWrapper.java gives a lot more details about the worker pool and related details. JepPythonEngine.java encapsulates a collection of InterterpreterWrapper instances ( which is a thin wrapper around a single worker a.k.a InterpreterThread).

python/src/main/java/org/apache/apex/malhar/python/base/WorkerExecutionMode.java
ghost: Done.

python/src/main/java/org/apache/apex/malhar/python/base/ApexPythonEngine.java
ghost: Added more javadocs to explain the usage of windowId and requestId. There are two uses of these two params. 1. A way to pick the matching response as submitted to a request queue. and 2. The implementation needed a way to determine the best fit worker basing on the worker execution mode. 1. BROADCAST/ANY/STICKY.

For picking up the corresponding response from a pool of responses ( as there might be stragglers that might have submitted their responses to the response queue ) we need a mechanism to match the response to a request. The best way was to use the windowId and requestId to achieve this. The operator implementation automatically provides these monotonically increasing values as requests are submitted into the request queue.


In case of choosing the worker from the worker pool:
If it is broadcast mode, then these values are not used at all.
If it is any worker mode, the windowid and requestid are used to roundrobin requests basing on the most free worker. In other modes, it is possible that the request is queued and the ANY mode chooses the least loaded worker to enqueue the request. Hashcode would have been an alternate choice but was not sure if dashcode would provide for an even distribution of work. Hence used this monotonically increasing sequence as a way to achieve round robin mechanisms.
if it is sticky, then the hashcode() of the request object is used to select the worker.

ghost:
@tweise  - Addressed the review comments and resubmitted the PR. Could you please provide additional thoughts if any ?

ghost:
Jenkins retest this please

ghost:
Jenkins retest this please

ghost:
Jenkins retest this please

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
```diff
@@ -0,0 +1,591 @@
...
+ *    This essentially is generated when JEP is not able to load a library that the python code is using. Ensure
+ *    that the PYTHONPATH environment variable is set pointing to the right location. This is especially
+ *     required when there are multiple python versions in the code.
+ *  </p>
+ *
+ *  <p>For very low time SLA requirements, it is encouraged to set the spin policy to be busy wait using the
+ *   following configuration snippet for the JEP engine
+ *   <code>
+ *       Map<PythonInterpreterConfig,Object> preInitConfigs = new HashMap<>();
+ *       preInitConfigs.put(PythonInterpreterConfig.IDLE_INTERPRETER_SPIN_POLICY, ""+ SpinPolicy.BUSY_SPIN);
```
tweise: should this be BUSY_SPIN.name()

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
```diff
@@ -0,0 +1,591 @@
...
+        if ( result != null) {
+          responseCounterForThisWindow += 1;
+          outputPort.emit(result);
+        } else {
+          numNullResponsesPerWindow += 1;
+        }
+      } catch (ApexPythonInterpreterException e) {
+        responseCounterForThisWindow += 1; // Error tuples need to be accounted for totals
+        errorPort.emit(tuple);
+        LOG.error("Error while processing tuple", e);
```
tweise: this should be debug or commented out since errors are emitted to a port?

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
```diff
@@ -0,0 +1,591 @@
...
+  }
+
+  @Override
+  public void deactivate()
+  {
+
+  }
+
+  /***
+   * Used to emit the stragglers into the Stragglers port. INvoked both when a new tuple arrives or when Idle time
```
tweise: type

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
```diff
@@ -0,0 +1,591 @@
...
+    numNullResponsesPerWindow = 0;
+    currentWindowId = windowId;
+  }
+
+  @Override
+  public void endWindow()
+  {
+    super.endWindow();
+    if (responseCounterForThisWindow < requestCounterForThisWindow) {
+      LOG.info("Detected stragglers and configured flag/state for blocking at end of window is " +
```
tweise: should be debug and possibly not logged repeatedly?

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
```diff
@@ -0,0 +1,591 @@
...
+          blockAtEndOfWindowForStragglers);
+      if (blockAtEndOfWindowForStragglers) {
+        while (responseCounterForThisWindow < requestCounterForThisWindow) {
+          LOG.debug("Trying to emit all stragglers before the next window can be processed");
```
tweise: move logging outside the loop? Is there a need to add a small sleep interval to avoid the busy loop? What happens when the expected responses don't arrive, I suppose the operator will be killed by the engine and then it may be good to document this here?

python/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+    <version>4.0.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>malhar-python</artifactId>
+  <name>Apex library python support</name>
+  <packaging>jar</packaging>
+
+  <properties>
+    <jep-version>3.7.0</jep-version>
+    <jepInstallPath></jepInstallPath> <!-- Purposefully empty as it is ideally set from command line basing on env but
```
tweise: perhaps it will be helpful to keep the default location here instead of empty?

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
```diff
@@ -0,0 +1,591 @@
...
+  {
+    emitStragglers();
+  }
+
+  @Override
+  public Response processStats(BatchedOperatorStats stats)
+  {
+    Response response = new Response();
+    if ( apexPythonEngine == null) {
+      response.repartitionRequired = true;
```
tweise: Since the engine is initialized in setup and the stats listener executes in the AM, how do we ever get here? In order to implement dynamic partitioning, the operator partitions would need to provide metrics that are then evaluated by the stats listener.

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
```diff
@@ -0,0 +1,591 @@
...
+    }
+    return response;
+  }
+
+  public ApexPythonEngine getApexPythonEngine()
+  {
+    return apexPythonEngine;
+  }
+
+  public void setApexPythonEngine(ApexPythonEngine apexPythonEngine)
```
tweise, line 415: how is this supposed to be used (it appears unused)?

tweise (changes requested):

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
```diff
@@ -0,0 +1,591 @@
...
+  @Override
+  public void setup(Context.OperatorContext context)
+  {
+    super.setup(context);
+    apexPythonEngine = initApexPythonEngineImpl(context);
+    Map<PythonInterpreterConfig,Object> preInitConfigurations = getPreInitConfigurations();
+    apexPythonEngine.preInitInterpreter(preInitConfigurations);
+    apexPythonEngine.startInterpreter();
+    apexPythonEngine.postStartInterpreter();
+    initPartitioner();
```
tweise: this shouldn't be here, since the partitioner is executed in the master and not in the worker container (setup is only called in the latter).

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
```diff
@@ -0,0 +1,591 @@
...
+              partitioner = new ThreadStarvationBasedPartitioner(this);
+              break;
+          }
+        }
+      }
+    }
+  }
+
+  /***
+   * Starts the python engine and also sleeps for sometime ( configurable) to ensure that the interpreter is
```
tweise, line 293: sleep intervals are always problematic, they result in non-deterministic behavior

python/src/main/java/org/apache/apex/malhar/python/base/ApexPythonEngine.java
tweise: Looking at the implementation of setup, I see the following:
```
     apexPythonEngine.preInitInterpreter(preInitConfigurations);
     apexPythonEngine.startInterpreter();
     apexPythonEngine.postStartInterpreter();
```
If the operator has these calls as a single logical unit, then why are they broken out in the interpreter interface? I understand that interpreter internal the differentiation makes sense, just wondering if that means it should also be visible in the interface. Is that motivated by usage outside of an Apex operator?

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
```diff
@@ -0,0 +1,591 @@
...
+
+/***
+ * <p>The operator provides a mechanism to execute arbitrary python code by using the {@link JepPythonEngine} as its
+ *  default engine.</p>
+ *
+ * <p>See {@link JepPythonEngine} and {@link InterpreterThread} for more detailed javadocs about the interpreter
+ *  itself and the API patterns possible through this engine</p>
+ *
+ * <p>Note that the JVM option of the library path needs to be set to a value which contains the JEP dynamic library.
+ *  For example, -Djava.library.path=/usr/local/lib/python3.5/site-packages/jep assuming that the JEP library is
```
tweise, line 57: That needs to be configured as part of JVM_OPTIONS attribute?

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
```diff
@@ -0,0 +1,591 @@
...
+  {
+    initPartitioner();
+    partitioner.partitioned(partitions);
+  }
+
+  @Override
+  public void beforeCheckpoint(long windowId)
+  {
+    accumulatedCommandHistory.clear();
+    accumulatedCommandHistory.addAll(getApexPythonEngine().getCommandHistory());
```
tweise, line 377: What are the guarantees for the interpreter state? Is that documented somewhere?

tweise (changes requested):

python/src/main/java/org/apache/apex/malhar/python/base/ApexPythonEngine.java
ghost: The primary motivation was for support beyond the methods of an Apex operator ( example as a utility usage etc). Bundling all of these in the single setup() method does not seem a good idea as you pointed out. 

Moved the postStartInterpreter() to another method which can be overridden if required that (by default includes the postStartInterpreter() call). Operator implementors can make use this decoupled approach to only "touch" the post interpreter logic without worrying about setup() which now only deals with the interpreter boot up.

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
ghost: Done.

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
ghost: Changed it to debug.

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
ghost: Done.

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
ghost: Changed to debug and moved the other repeated log statement out of the loop.

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
ghost: Moved logging out of the loop. Also added a small sleep to avoid the busy loop. 

This loop is triggered only if the operator user decides to wait ( by means of a configuration switch names "blockAtEndOfWindowForStragglers"). If the stragglers belonging to a window do not arrive and if the above config switch is true, the operator will eventually be killed as the endWindow() call would never return.

python/pom.xml
ghost: Set the default value to that of a linux environment and python 3.5 version installation of jep installed to the standard paths by the end user.

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
ghost: This was more of a convenient way to inject a custom interpreter implementation if JEP does not meet their use cases. The current default is JEP and hence the overridden approach need did not exist.

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
ghost: Removed the initPartitioner from this. Thanks for pointing this out.

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
ghost: Yes.

This is how I am configuring in the test application for this feature under META-INF/properties.xml:
`
<property>
    <name>dt.operator.pythonprocessor.attr.JVM_OPTIONS</name>
    <value> -Djava.library.path=${jepInstallPath}</value>
  </property>
`

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
ghost: The sleep can be configured to be 0 in which case the thread will continue ahead. It was only in the case of Tensorflow testing that this sleep was required because the complete loading of the tensor flow library seem to have been done on another thread. In this case, immediately issuing a python command resulted in errors. The only way seemed to be letting the main thread to sleep. A busy spin might not solve the problem either. The other plausible way is to use a test command that does not interfere with the users interpreter commands ( and the command makes use of the library tensorflow in this case ) and the main thread continuously issues this configured command and see if the return status is true to signify the completion of the boot up of the interpreter.

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
ghost: Thanks for pointing this out. Removed the stats listener interface completely. 

The partitioner uses the serialised operator metrics to decide the number of partitions at the time of "definePartitions()"

python/src/main/java/org/apache/apex/malhar/python/base/BasePythonExecutionOperator.java
ghost: The interpreter  state is maintained in the memory. The current implementation piggy backs on the JEP engine to access a variable that is available as part of that state ( or create new ones basing on its needs ) using the JEP apis. ( using the getXXX method())

Here are some documentation links that describe how the state values of an in-memory instance can be accessed/set 
- https://github.com/ninia/jep/wiki/Getting-Started
- https://github.com/ninia/jep/wiki/Numpy-Usage

ghost:
@tweise  - Could you please consider the revised version for further comments if any.

ghost:
Jenkins retest this please

python/src/main/java/org/apache/apex/malhar/python/base/PythonInterpreterConfig.java
tweise: You are right, so far operator level attributes have not been used to configure operators, although that might work. I was suggesting that instead of coming up with yet another way of map based configuration with defaults. I think the properties as implemented now are sufficient.

tweise:
I think this is sufficient for the initial implementation of the feature and improvements can be taken up separately. One thing I would like to see added is an example that uses the operator, which will make it much easier for users to get started. I suspect you have that example elsewhere already :)

Also it would be good to add some user documentation.

Thanks for the contribution!

tweise (approved):

ghost:
Thanks for the review Thomas. 

There is a sample apex app under the test sources that shows how this operator can be used. I will attempt a blog entry later which explains the various aspects that were discussed in this PR and other general features.

